PR #692, opened 2015-06-20
Country flags in scoreboard

joemmanuel:
This change appends a country flag in the contest scoreboard.


lhchavez:
:+1:. nada más dale squash y ya.

PR #1282, opened 2017-06-02
Agregando ACLs a los grupos

lhchavez:
Este es un pre-requisito para #1243, #1244 y schools.

### DIFF
--- a/frontend/database/20_groups_acls.sql
+++ b/frontend/database/20_groups_acls.sql
@@ -1,6 +1,6 @@
 -- Groups
 ALTER TABLE `Groups`
-	ADD COLUMN `acl_id` int(11) NULL AFTER `group_id`;
+  ADD COLUMN `acl_id` int(11) NULL AFTER `group_id`;
 
 DELIMITER $$
 CREATE PROCEDURE `Groups_ACLs_Migrate`()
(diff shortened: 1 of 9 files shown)


lhchavez:
@omegaup/reviewers PTAL

frontend/database/20_groups_acls.sql
```diff
@@ -0,0 +1,47 @@
...
+  OPEN cur;
+
+  read_loop: LOOP
+    FETCH cur INTO group_id, owner_id;
+    IF done THEN
+      LEAVE read_loop;
+    END IF;
+
+    INSERT INTO `ACLs` (`owner_id`) VALUES (owner_id);
+    SET acl_id = LAST_INSERT_ID();
```
joemmanuel: Nota que este `LAST_INSERT_ID()` puede ser propenso a race conditions..

lhchavez: https://dev.mysql.com/doc/refman/5.7/en/getting-unique-id.html disagrees :P

> For LAST_INSERT_ID(), the most recently generated ID is maintained in the server on a per-connection basis. It is not changed by another client. It is not even changed if you update another AUTO_INCREMENT column with a nonmagic value (that is, a value that is not NULL and not 0). Using LAST_INSERT_ID() and AUTO_INCREMENT columns simultaneously from multiple clients is perfectly valid. Each client will receive the last inserted ID for the last statement that client executed.

Este código se ejecuta mediante un script de Python single-threaded, así que aunque haya otras cosas corriendo concurrentemente, no debería tener ningún problema porque mantiene una conexión única. :)

joemmanuel: Nice

Issue #1383, opened 2017-08-13
As an English speaking/reading user, I would like to be able to see information in English only, so that my experience on omegaUp does not feel impersonal. 


tiqtok:
Tasks:

1. Hide Spanish links depending on a user's language - _var language = window.navigator.userLanguage || window.navigator.language;
alert(language);

When a user who's language is set to eng visits omegaUp, they should not see the links that take user's to pages in spanish.

pabloaguilar:
Can you list specific examples, please?
Do you mean the links to the blog, Q&A and book?

gramila:
Yeah

1. Blog
2. Q & A

What do you mean by book?

pabloaguilar:
The "Suggested book" section on the homepage

PR #1388, opened 2017-08-13
Development

tiqtok:
Removes links to pages with spanish text in menu, jumbotron info area, and arena page. #1383


pabloaguilar:
I'm not convinced that hiding these links is the right thing to do.
I do like the localization changes, where you fix messages currently hardcoded to Spanish.

Hiding content, on the other hand I'd rather not do.
Given that translation can be just a single click away (installing a 'translate' extension or built-in for Chrome) and that automatic translation is mostly understandable, I think having 'translatable' information available is better than not having it at all, e.g. which compiler version are we using? can I change my name? can I change my e-mail?
Furthermore, the correct fix for the blog posts and tutorial is to actually translate them and add multilingual support to the blog. This should be pretty easy to finish in a week's time and I don't want to submit this change and have it reverted a few days later.

@omegaup/reviewers Thoughts?

frontend/templates/en.lang
```diff
@@ -771,7 +771,9 @@ wordsTopics = "Topics"
 wordsTotal = "Total"
 wordsUnimplemented = "This feature has not been implemented yet :("
 wordsUser = "User"
+wordsUsername = "Username"
 wordsValidator = "Validator"
 wordsVerdict = "Verdict"
+wordsVerdictFor = "Verdict for "
```
pabloaguilar: Please use a template in combination with `omegaup.UI.formatString`

`wordsVerdictFor = 'Verdict for %(title)'`
and then use it in the JS code as follows:
`title: {text: omegaup.UI.formatString(omegaup.T.wordsVerdictFor, {'title':title})}`

This allows localization to languages where the sentence structure is different

alanboy:
My thoughts:

> Hiding content, on the other hand I'd rather not do. Given that translation can be just a single click away (installing a 'translate' extension or built-in for Chrome) 

I've seen users open a page in Spanish and they don't usually bother to find a translate service to see whats written, plus this is coming from Msft leap and for the most part they use Edge. If we officially support a language, i don't think we should rely on external plugins to do the live/runtime translation for us. (Using those tools to translate and then us provide the content, that would be fine IMO)

>I think having 'translatable' information available is better than not having it at all

I've actually seen the oposite in the LEAP classroom. It's a big turn-down to click a few links and end up in a Spanish site, after 3 pages in Spanish you just lose interest. Speaking one language really does make the difference.

>Furthermore, the correct fix for the blog posts and tutorial is to actually translate them and add multilingual support to the blog. This should be pretty easy to finish in a week's time and I don't want to submit this change and have it reverted a few days later.

I absolutely agree, translation our content should be our long term goal, but today we just don't have translated content and it's resulting in a bad experience. We can take the same approach as the experiments, where we `if` the feature, in this case English/Portuguese blog or wiki and just enable it for the general public once its ready. It's really not trivial to translate the content: take a look at this post https://blog.omegaup.com/2015/08/chocolates/ with images in Spanish, are we going to re-create the images?

This change is not for the sake of making the change, it's an actual issue that users struggle as David knows first hand since he has been sitting down with students and having the go trough the initial omegaUp experience. I know that there is a feeling that "omegaUp isnt ready" because of this.

lhchavez:
First of all, our mission is to increase the number of talented software engineers in Latin America, so this runs against that :( It was a conscious decision from day one to have all our platform in spanish because it is the language that the vast majority of to-be engineers in LA use, and we should embrace that. It's great that we can be useful to programs like LEAP, but we should not bend over backwards if it is not aligned with the mission.

hiding something because it's not translated is a bit hostile. how about just editing the english and portuguese translations so they say e.g. `Blog (Spanish)`? I wouldn't mind doing that because it still preserves the content and it sets the expectation upfront.

pabloaguilar:
To be clear: I agree with having full support for the languages we officially support. I'm just saying we should do it the right way and have everything translated, as opposed to hiding untranslated content.

@alanboy I understand there are some things might be hard to translate but I think it's fine if we have a 90% translated blog than to wait until it's 100% before being able to show it. Quote the saying: don't let perfection be the enemy of the good.

I like @lhchavez's suggestion to add a warning label that content is currently in Spanish.

So, lets get down to action items:
* What is the deadline for this change? We have volunteer translators ready to go and I claim we can have the blog translated in a ~2 weeks time. Is there a pressing need to have this earlier than that?
* Is the middle ground of adding a notice that content is in Spanish acceptable in the meantime?
* Find a different book to suggest when in English or Portuguese, and update the "suggested book" section to point to those instead of hiding it

alanboy:
I really appreciate the candor. Here are my conclusions on this, and then i'll leave the final decision to the technical committee:

>First of all, our mission is to increase the number of talented software engineers in Latin America, so this runs against that

English is a supported lang. Let's limit the discussion to how we should delight our English, Spanish and Portuguese speaking users in this bug. 

> It was a conscious decision from day one to have all our platform in spanish because it is the language that the vast majority of to-be engineers in LA use, and we should embrace that.

That's not what's being discussed here, site in Spanish is not being affected at all.

> It's great that we can be useful to programs like LEAP but we should not bend over backwards if it is not aligned with the mission.

They are not using the platform by coincidence, they are using it because its an agreement which both parties have been working on for some time now. This is a strategic partnership which is benefiting both of us. As you mention, our mission actually says Latin America, so that's the public we should focus on, to me translating the entire omegaUp content to English for _this_ partnership, now _that_ would be bending over backwards. Specially because the LEAP candidates are in a intensive 12-week program, they don't care about post about the new raking system from 3 years ago, they just don't have the time.

Translating the entire blog is a massive effort which to this day only 16% of visitors _could_ benefit from. How many of those visits actually try to read the blog? 1,2%?

image


>hiding something because it's not translated is a bit hostile

Hostile to who? Not the users. On some English site, clicking an English button and ending up in a Spanish page is worse IMO but I haven't raised it a as big deal because, you know, just my opinion. But with proof that this is causing confusion, poor adoption with our partnership and thus not really benefiting from it, now I deem it necessary to make the change. This is feedback directly from the Schools Technical Solutions team, those that literally sit next to the users and see them interact with the site, it's not _us_ sitting next to an English-only speaking person browsing the site, so when they say it's important I trust them. 

> I understand there are some things might be hard to translate but I think it's fine if we have a 90% translated blog than to wait until it's 100%

Oh, I am a fan of the 90% solution, but let me put it this way: when we get 90% of the content we are going to flip the switch and English user is never going to ever see Spanish content ever again, right? So why not hide it now as opposed to waiting? 

Don't take my word for it, [please open this study](http://itpro.nikkeibp.co.jp/article/COLUMN/20120717/409422/) which exactly mentions this problem.

Interesting? Useful? Was it the best use of our time for me to include the study here in this (english) conversation? Or did you prefer i simply skip it? OK, let's try again:

Don't take my word for it, [please open this study (JAPANESE)](http://itpro.nikkeibp.co.jp/article/COLUMN/20120717/409422/) which exactly mentions this problem.

Did that make it more useful? Did that make you have a better experience? 

I really don't understand what's wrong with hiding untranslated content, it's common practice and it's the right thing to do. If the feature isn't ready (omegaUp for work?) don't show it, if the content isn't useful/readable to the current audience don't show it:

image

image

I'm all up for translation to English, don't get me wrong, but right now it's not a matter of what we think it's best/want, we have a team for that. omegaUp TSE is there to tell us what's wrong when you actually put a user in front of the omegaUp.com/schools/ page. Right now it's this English/Spanish problem, tomorrow it'll be something in when Universidad Tecnológica del Norte de Aguascalientes starts to use it or the Peruvian ministry of education. Let's be receptive of their feedback.

>What is the deadline for this change? We have volunteer translators ready to go and I claim we can have the blog translated in a ~2 weeks time. 

I'll discuss the dates in private. If you seriously can have some external team translate the blog in 2 weeks (heck i'll do everything in my power to help you), then let's leave the blog link, but we still need to ask this same question with "Preguntas".

> Is there a pressing need to have this earlier than that?

This is hard, but I would say it's mostly relevant when a COHORT starts. The first week users start using the platform. 

> Is the middle ground of adding a notice that content is in Spanish acceptable in the meantime?

Personally, I *really* dislike the idea. If it weren't like the biggest menu on the site, perhaps, but having this in the main menu, hmmmm, i really really wish we didn't do this:
image


> Find a different book to suggest when in English or Portuguese, and update the "suggested book" section to point to those instead of hiding it

This should be easy.

tiqtok:
If we can get the Spanish pages translated in two weeks, then we don't need to hide the links, even if the pages are 90% translated. I think that's a suitable option. 

The only reason to hide the links in the first place is that until the pages were translated, it is my perspective (as a person who does not read well in Spanish) that a user with their browser set to english would not only prefer to read things in English but feel that links sending me to pages in Spanish is impersonal to the point that I would decide a site is not for me. User's make decisions like this subconsciously and quickly. Some of those reasons include: a site's community not representing oneself and having different goals, interests, culture, technological preferences, etc. These reasons don't have to be fact based but just based in one's perception, even if erroneously. We may have an impression that a person should be ok with a site that has content in Spanish, but we shouldn't allow the content of the site to inevitably alienate people who don't hold that opinion.  

There are however different user profiles that we want to accommodate of course, such as people who's browser is set to English but who also read Spanish well enough to be ok with such content and find it beneficial, people who's browser is set to Spanish, etc. 

I do not believe that this is a hostile move for people in my user profile but rather an agile development approach in line with "don't let perfection be the enemy of good." We could have simply removed the links and added them back for english speakers when the language was translated. Although keeping them around for people with browsers set to **english who also read in Spanish is fine too.** especially since at the moment, this possibly outweighs the former user profile. We just need to understand that at some point we can be alienating people, especially people outside of programs like LEAP, people of Latin American heritage, who don't speak Spanish and find omegaUp on their own. The type of people who would be drawn to a site like omegaUp coming from a non-traditional (bootcamp or otherwise) background. 

Which brings me to the point on the vision. Since I wasn't around for the inception, I'd like to ask what the original intent was of "[increasing] the number of talented software engineers in **Latin America**" Is that to mean people who live in latin america only (*not in the US)? Does that mean only Spanish speaking people? Or does that mean anyone of Latin decent whether or not they speak Spanish and whether or not they are in the Americas?

I once asked @lhchavez  about the purpose of getting into LEAP in the context of being used to get customer feedback or whether we were expanding into new markets. I believe the response was both. In that case, are we sticking to the vision or has a decision been made to move beyond what was originally agreed upon? 

In either case, and whatever is agreed upon, we need to define which user profiles we're willing to support and in what priority since we can't handle everything at the same time. This way we can know what users any given implementation will support, who we may be alienating, and determine whether we're ok with that for any given moment. 

As for the task at hand @pabloaguilar I think your solution is more than adequate. My perspective on your questions:

1. Roughly two weeks is more than enough.
2. I don't think that's necessary. 
3. That will have to be one of the things we'll have to change in support of the three supported languages.

lhchavez:
> "don't let perfection be the enemy of good." [...] Although keeping [the links] around for people with browsers set to english who also read in Spanish is fine too.

all right, let's go forwards with the `(Spanish)` option, especially since a good chunk of Portuguese speakers can read Spanish.

tiqtok:
ok - Just want to clarify that by Spanish link I meant a link going to a Spanish site, not a link with the text "Blog (Spanish)" next to it, and this in the context of getting translation of that section working in roughly two weeks time. 

In other words, we leave the links as is for now while we get the translation working.

PR #1760, opened 2018-02-19
Added initial support for user's profile privacy

AtharvaKhare:
Added a commit that addresses the profile privacy by hiding (https://github.com/omegaup/omegaup/issues/1688)
- Name, country, state, school, Graduation date,
- Statistics
if user has opted the profile to be private.
Note: By default, all profiles are NOT private.

TODO (Feel free to extend this list):
- [x] **Modify frontend/templates/user.edit.tpl to allow user to set privacy flag**
- [x] Fix a bug that prevents new user signup - ./stuff/update-dao.sh worked!
- [x] Write a unit test

Screenshots:


frontend/server/controllers/UserController.php
```diff
@@ -1219,6 +1219,34 @@ public static function apiProfile(Request $r) {
...
+            $response['userinfo']['recruitment_optin'] = null;
+            $response['userinfo']['scholar_degree'] = null;
+            $response['userinfo']['school'] = null;
+            $response['userinfo']['school_id'] = null;
+            $response['userinfo']['solved'] = null;
+            $response['userinfo']['state'] = null;
+            $response['userinfo']['state_id'] = null;
+            $response['userinfo']['submissions'] = null;
+            $response['status'] = 'ok';
+            return $response;
```
lhchavez: you can simplify this significantly by just returning the array literal:

```php
return [
  'problems' => [],
  'userinfo' => [
    'username' => $r['user']->username,
    'birth_date' => null,
    // ...
  ],
];
```

alternatively, you can "clean" `$response` once it's been constructed (after L1250) so you don't miss anything:

```php
if ((is_null($r['current_user']) || $r['current_user']->username != $r['user']->username)
      && $r['user']->is_private == 1) {
  $response['problems'] = [];
  foreach ($response['userinfo'] as $k => $v) {
    $response['userinfo'][$k] = null;
  }
  // Re-add the information we need:
  $response['userinfo']['username'] = $r['user']->username;
  $response['userinfo']['rankinfo'] = [
    'name' => null,
    'problems_solved' => null,
    'rank' => null,
  ];
}
```

frontend/database/44_user_privacy.sql
```diff
@@ -0,0 +1,3 @@
+-- Users
+ALTER TABLE `Users`
+  ADD COLUMN `is_private` tinyint(1) NOT NULL DEFAULT 0 AFTER `in_mailing_list`;
```
lhchavez: Can you add the following?

```sql
COMMENT 'Determina si el usuario eligió no compartir su información de manera pública'
```

sorry for the spanglish :(

frontend/server/controllers/UserController.php
AtharvaKhare: I totally forgot that PHP has foreach loop! Thanks!

AtharvaKhare:
Need some help with failing builds: https://travis-ci.org/omegaup/omegaup/jobs/343675996#L887

https://github.com/omegaup/omegaup/pull/1760/files#diff-7c2d43f5d9d193f767ccdb0b92b548eeR1655
Please suggest a better way to write this Validator. I did not include ```is_null``` because in the database, ```is_private``` is ```NOT NULL```.

frontend/server/controllers/UserController.php
```diff
@@ -1631,9 +1652,12 @@ public static function apiUpdate(Request $r) {
             $r['current_user']->language_id = $query[0]->language_id;
         }
 
+        Validators::isNumber($r['is_private'], 'is_private', true);
```
lhchavez: ah I see what you mean. What's going to happen is that since you specified that the field is `NOT NULL` in the database, the Data Access Object (dao) will not attempt to add `is_private` to the `INSERT` statement, which will then make the database insert the default value of 0.

What the

```php
if (!is_null($r['something'])) {
  Validators::isNumber($r['something'], 'something, true);
}
```

pattern is doing is that _if_ a user explicitly specifies a value for `something`, then it will make sure it is valid and throw an exception otherwise. if it's not explicitly set, then let the database do its thing.

If you follow the pattern, the error should go away!

AtharvaKhare:
Hey @lhchavez , I am a bit confused about adding a test, where should it be added? 
The test would be a PHP file in ```frontend/tests/ui```, and not a python file, right?
Should I create a new file for the test, or append to which file.

Thanks!

lhchavez:
Oh sorry about that, the PHP tests in that directory are _super_ bitrotten D: I sent out https://github.com/omegaup/omegaup/pull/1762 to delete them.

Normally for controllers we have all the controller unit tests in [`frontend/tests/controllers`](https://github.com/omegaup/omegaup/tree/master/frontend/tests/controllers). Concretely, you can add the test here: https://github.com/omegaup/omegaup/blob/master/frontend/tests/controllers/UserProfileTest.php

You can run _all_ the tests with `./stuff/runtests.sh` (which will take a while) or just that one file with `cd frontend/tests && phpunit controllers/UserProfileTest.php`.

AtharvaKhare:
@lhchavez Please review! :smile: 
The first commit's test passes on my local machine, so it must be the usual weird Selenium stuff.

lhchavez:
yeah, Selenium has been known to be a bit flaky :( will review later today :D

Issue #1711, opened 2018-01-28
Como usuario quiero saber que problemas intente resolver para intentarlos en un futuro.


MiguelRomero:
En la olimpiada veracruzana de informatica queríamos saber que problemas resolvió un usuario con finalidad de saber cual era el workflow de entrenamiento que siguió.

lhchavez:
Esto necesita hacer un query a la base de datos que regrese todos los problemas que hayas enviado pero que no tengas AC. como esto suena caro, no se podría poner por default en el perfil, así que sería una pesaña aparte.

también (como issue aparte): hay que definir un rol de maestro, de alguna manera otorgar este permiso de parte de los estudiantes para los maestros, y que el maestro pueda ver esta lista, aún si el perfil es privado.

lhchavez:
Este issue es únicamente para hacer la vista, una vez que #1716 se haya resuelto.

CarlosCordova18:
I would like to do the view also, could you please suggest me how to work on it as you did on:
https://github.com/omegaup/omegaup/issues/1716#issuecomment-366811775

lhchavez:
Sure!

Right now the list of problems you _have_ solved are displayed in your profile: https://omegaup.com/profile/ . What we would need to do is create a new table in that same view. A super quick tour of what happens when you open a page:

* First, [nginx](https://nginx.org/en/), our web server, goes through its internal configuration, which includes a set of rewrite rules stored here: https://github.com/omegaup/omegaup/blob/master/frontend/server/nginx.rewrites . This translates nice, REST-looking URLs into ugly `.php` filenames. In this example, this is the rule that is being invoked: https://github.com/omegaup/omegaup/blob/8c21617376773f9d2b9412eb04c26559233d4bb9/frontend/server/nginx.rewrites#L97
* Next, the appropriate `.php` file is invoked. In this case, it is https://github.com/omegaup/omegaup/blob/master/frontend/www/profile.php, which does very little except set up PHP and invoke a [smarty](https://www.smarty.net/) template.
* The smarty [template](https://github.com/omegaup/omegaup/blob/master/frontend/templates/profile.tpl) is a kind of outdated way of creating HTML injecting some values in PHP. We're trying to get away from it and use [Vue](https://vuejs.org) everywhere, but that takes time :( On the flipside, it's relatively simple.
[...]

PR #1780, opened 2018-02-25
Creando las vistas de los problemas no resueltos de cada usuario

CarlosCordova18:
En respuesta al Issue #1711


frontend/server/libs/dao/Problems.dao.php
```diff
@@ -341,7 +341,7 @@ class ProblemsDAO extends ProblemsDAOBase {
         $user_id
     ) {
         $sql = "
-            SELECT
+            SELECT DISTINCT
```
lhchavez: a-ha! nice catch, we missed that in the previous iteration.

can you add a test for this? just make any of the users that sent an AC send another AC.

frontend/templates/profile.tpl
```diff
@@ -60,6 +60,25 @@
...
 		</div>
 
+		<div class="panel panel-default">
+			<div class="panel-heading">
+				<h2 class="panel-title">{#profileUnsolvedProblems#} <span class="badge" id="problems-unsolved-total">0</span></h2>
+			</div>
+			<table class="table table-striped" id="problems-unsolved">
+				<thead>
+					<tr>
+						<th>{#profileUnsolvedProblemsTableTitle#}</th>
```
lhchavez: I realize the other table has exactly the same problem and it's definitely not your fault, but do you mind making this

```html
<th colspan="3">{#profileUnsolvedProblemsTableTitle#}</th>
```

? also, bonus points if you fix the other table :D

frontend/server/libs/dao/Problems.dao.php
CarlosCordova18: I think the test should be sending another to not accepted run to a unsolved problem, but I will create the test you are asking for also.

CarlosCordova18:
I've added also the little modification for #1767

PR #1784, opened 2018-02-26
Issue # 1517

HorsemanJnr:
Added a condition on line 69 to make it so that banned problems are not showing to administrators.
The new code throws an exception so that both privately and publicly banned problems are not visible to the administrator of the platform

Fixes: #1517


frontend/server/controllers/RunController.php
```diff
@@ -66,7 +66,7 @@ private static function validateCreateRequest(Request $r) {
             // Check that problem exists
             $r['problem'] = ProblemsDAO::getByAlias($r['problem_alias']);
 
-            if ($r['problem']->deprecated) {
+            if ($r['problem']->visibility == ProblemController::VISIBILITY_PUBLIC_BANNED || $r['problem']->visibility == ProblemController::VISIBILITY_PRIVATE_BANNED) {
```
lhchavez: we still need the original condition. Can you preserve the original condition add the new condition on a new block and make it `throw new NotFoundException('problemNotFound');`?

Also, can you add a test? https://github.com/omegaup/omegaup/blob/master/frontend/tests/controllers/RunCreateTest.php holds all the run creation tests.

HorsemanJnr:
hi,
I am having some troubles with writing the unit test. I'm trying to manually set the visibility of the problem to ProblemController::PUBLIC_VISIBILITY_BANNED, but the function does not throw the necessary exception. This makes the test fail. 
Any pointers on the best way forward from here? Am I thinking through this the right way?

lhchavez:
It's currently throwing `NptFoundException`, where it should throw a `NotFoundException`: the test works!

HorsemanJnr:
I fixed the error in my local dev env. I'm committing and pushing now.

frontend/tests/controllers/RunCreateTest.php
```diff
@@ -742,4 +742,19 @@ public function testRunInAssignmentFromStudentAfterDeadline() {
         // Call API
         $response = RunController::apiCreate($r);
     }
+
+    /**
+     * Should not allow sending banned problems to administrators
+     * @expectedException
```
lhchavez: nit: `@expectedException NotFoundException`

frontend/tests/controllers/RunCreateTest.php
```diff
@@ -742,4 +742,19 @@ public function testRunInAssignmentFromStudentAfterDeadline() {
...
+     */
+    public function testShouldNotAllowToSendBannedProblems() {
+        $r = $this->setValidRequest();
+
+        $r['visibility'] = ProblemController::VISIBILITY_PUBLIC_BANNED;
```
lhchavez: This unfortunately does not change the visibility of the problem. That can be done using https://github.com/omegaup/omegaup/blob/88db08f12c34904e5f904dc820edc11735d8b14f/frontend/tests/controllers/ProblemUpdateTest.php#L498-L503

In order to create the problem (to get the `$problem->alias`) you can do https://github.com/omegaup/omegaup/blob/88db08f12c34904e5f904dc820edc11735d8b14f/frontend/tests/controllers/ProblemUpdateTest.php#L453-L456

and finally to create the run you can just call `RunController::apiCreate(new Request(...));`. You can create the `new Request` like this: https://github.com/omegaup/omegaup/blob/88db08f12c34904e5f904dc820edc11735d8b14f/frontend/tests/controllers/RunCreateTest.php#L45-L51 (without the `contest_alias` and setting the `problem_alias` to `$problem->alias`).

frontend/tests/controllers/RunCreateTest.php
```diff
@@ -742,4 +742,19 @@ public function testRunInAssignmentFromStudentAfterDeadline() {
...
+        echo $r['visiblity'];
```
lhchavez: Before you submit any pull request, remember to remove any debugging statements and commented-out code (like in line 758).

frontend/tests/controllers/RunCreateTest.php
HorsemanJnr: hi, so there is already a check in RunController on line 64 that checks if a problem is banned, and throws an InvalidParameterException, with the message "parameterEmpty". This means if a problem is banned, that exception is thrown, instead of the NotFoundException.
Is the new check redundant? and show my test case expect the InvalidParameterException, instead of the NotFoundException?

lhchavez: huh, that's odd. the exception at line 64 only checks whether `problem_alias` was set and is non-empty: https://github.com/omegaup/omegaup/blob/a62d6663d7cf8bada2ce4a659a146efd8dd71999/frontend/server/controllers/RunController.php#L64

so something must be wrong with setting that parameter in the `Request`.

HorsemanJnr: thanks @lhchavez I had a typo in my test, so the problem_alias wasn't set.
I'm adding different cases for both publicly and privately banned questions just to be sure.

Issue #1771, opened 2018-02-23
[Color Hierarchy] make apiProfile() return user's color


heduenas:
The algorithm to compute the user's color in described in the design doc:
https://docs.google.com/document/d/1R14hS-6TEhE34Aw3Pu_85jqIZjCyxVx6fiCHL8cvLg0/edit#

heduenas:
Cost estimate: 4 hours including tests.

randika96:
Can you tell me where I can find apiProfile()?  Then I can start work on this.

heduenas:
It's right here:
https://github.com/omegaup/omegaup/blob/004156793d50b97c9fc92af19077d784cd857b89/frontend/server/controllers/UserController.php#L1222

randika96:
i will start working on this issue

lhchavez:
assigned! let us know if you need any help.

randika96:
As a solution for this I am thinking of declaring a function to calculate the appropriate color which takes the marks as an argument and return the color. In case of that can you tell what is the best place for the place this method.  And call the method inside the apiProfile().
Can you tell me where I can find a class for user?

lhchavez:
It's probably better to have this be a CSS classname instead of a hardcoded color. I've already sent https://github.com/omegaup/omegaup/pull/1781 out so that the cron job that calculates the user ranks also calculates the cutoff points and the classnames. You'll need to re-fetch master to get that code, and run `./stuff/db-migrate.py migrate` to install the new table.

Since it's a cronjob, it's run on a schedule, so you'll want to run it manually for testing:

```shell
ubuntu@ubuntu-xenial:/opt/omegaup$ sudo su -c '/opt/omegaup/stuff/cron/update_user_rank.py'
2018-02-25 16:00:44,318:update_user_rank.py:Started
2018-02-25 16:00:44,335:update_user_rank.py:Updating accepted stats for problems...
2018-02-25 16:00:44,336:update_user_rank.py:Updating user rank...
2018-02-25 16:00:44,338:update_user_rank.py:Updating ranking cutoffs...
2018-02-25 16:00:44,340:update_user_rank.py:Done
```

The method to calculate the classname, given that it will perform some interaction with the database, should go in the Data Access Objects class [`Users.dao.php`](https://github.com/omegaup/omegaup/blob/master/frontend/server/libs/dao/Users.dao.php).

randika96:
How is the marks are stored along with user?

lhchavez:
[...]

PR #1791, opened 2018-03-01

randika96:
added function ``FindClassName()`` to the class `Users.dao.php` which returns the css class name.
Call the function from `apiProfile()` and added the class name to userinfo


randika96:
There are some unwanted files came up with my IDE. Ignore them

frontend/server/controllers/UserController.php
```diff
@@ -1260,6 +1260,8 @@ public static function apiProfile(Request $r) {
             ];
             $response['userinfo']['is_private'] = true;
         }
+        $classname=UsersDAO::FindClassName($response['userinfo']['username']);
```
lhchavez: you are not following the project's style guidelines :( your code should look exactly like the surrounding code, which means there should be a space around the `=`.

If you do `git push` from within the VM, the pre-upload hook will offer to fix these kinds of problems for you :D alternatively, you can run `./stuff/git-hooks/pre-push` manually if you absolutely want to push from outside the VM.

also, tiny request: we don't like to create temporary variables unless they increase readability and break very complicated expressions. here it's a simple assignment, so

```php
$response['userinfo']['classname'] =
    UsersDAO::FindClassName($response['userinfo']['username']);
```

is enough.

frontend/server/libs/dao/Users.dao.php
```diff
@@ -148,4 +148,28 @@ public static function getHideTags($user_id) {
         global $conn;
         return $conn->GetOne($sql, $params);
     }
+    public static function FindClassName($username){
+        global $conn;
+        $sql = 'select User_Rank.score
```
lhchavez: same here, please make the format of the SQL string the same as the other strings in this file: keywords go in all caps, table/column names go between backticks, we try to add newlines between the parts of the SQL query to highlight its structure, etc.

frontend/server/libs/dao/Users.dao.php
```diff
@@ -148,4 +148,28 @@ public static function getHideTags($user_id) {
         global $conn;
         return $conn->GetOne($sql, $params);
     }
+    public static function FindClassName($username){
```
lhchavez: If you pass in the `$user_id` instead of the `$username`, you don't need to make a `JOIN` with `Users`!

frontend/server/libs/dao/Users.dao.php
```diff
@@ -148,4 +148,28 @@ public static function getHideTags($user_id) {
...
+                from Users
+                inner join User_Rank on User_Rank.user_id=Users.user_id
+                where username = ?';
+        $params = [$username];
+        $rs = $conn->GetRow($sql, $params);
+        if (count($rs)==0){
+            return null;
+        }else{
+            $score = $rs[0];
+            $sql = 'select score, classname from User_Rank_Cutoffs order by score desc';
```
lhchavez: You can avoid this second roundtrip to the database by using a subquery:

```sql
SELECT
  `urc`.`classname`,
  `urc`.`score`,
  `urc`.`percentile`
FROM
  `User_Rank_Cutoffs` `urc`
WHERE
  `urc`.cutoff` > (
    SELECT
      `ur`.`score`
    FROM
      `User_Rank` `ur`
    WHERE
      `ur`.user_id` = ?
  )
ORDER BY
  `urc`.`percentile` ASC
LIMIT
  1;
```

Then since [`getOne()`](http://adodb.org/dokuwiki/doku.php?id=v5:reference:connection:getone) only returns the first column, you can just do

```php
return $conn->GetOne($sql, $params) ?? 'user-rank-unranked';
```

PR #1792, opened 2018-03-01

randika96:
Modified `apiProfile()` to return the CSS class name with userinfo. 
In oder to do that I declared a new function in `Users.dao.php` to compute appropriate class name when user name is given. And call it from `aprProfile()` and update userinfo.

### DIFF
--- a/frontend/server/controllers/UserController.php
+++ b/frontend/server/controllers/UserController.php
@@ -1240,6 +1240,8 @@ class UserController extends Controller {
             ];
             $response['userinfo']['is_private'] = true;
         }
+        $classname=UsersDAO::cd($response['userinfo']['username']);
+        $response['userinfo']['classname']=$classname; //update the userinfo
         $response['status'] = 'ok';
         return $response;
     }
--- a/frontend/server/controllers/UserController.php
+++ b/frontend/server/controllers/UserController.php
@@ -1240,7 +1240,7 @@ class UserController extends Controller {
             ];
             $response['userinfo']['is_private'] = true;
         }
-        $response['userinfo']['classname'] = UsersDAO::cd($response['userinfo']['user_id']);
+        $response['userinfo']['classname'] = UsersDAO::getRankingClassName($r['user']->user_id);
         $response['status'] = 'ok';
         return $response;
     }
--- a/frontend/server/controllers/UserController.php
+++ b/frontend/server/controllers/UserController.php
@@ -1240,8 +1240,7 @@ class UserController extends Controller {
             ];
             $response['userinfo']['is_private'] = true;
         }
-        $classname=UsersDAO::cd($response['userinfo']['username']);
-        $response['userinfo']['classname']=$classname; //update the userinfo
+        $response['userinfo']['classname'] = UsersDAO::cd($response['userinfo']['user_id']);
         $response['status'] = 'ok';
         return $response;
     }
--- a/frontend/server/libs/dao/Users.dao.php
+++ b/frontend/server/libs/dao/Users.dao.php
@@ -148,4 +148,28 @@ class UsersDAO extends UsersDAOBase {
         global $conn;
         return $conn->GetOne($sql, $params);
     }
+    public static function FindClassName($username){
+        global $conn;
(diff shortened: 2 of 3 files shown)


frontend/server/controllers/UserController.php
```diff
@@ -1240,6 +1240,7 @@ public static function apiProfile(Request $r) {
             ];
             $response['userinfo']['is_private'] = true;
         }
+        $response['userinfo']['classname'] = UsersDAO::cd($response['userinfo']['user_id']);
```
lhchavez: Shouldn't this be `UsersDAO::getRankingClassName`?

Also, `$response['userinfo']` does not have a `user_id` member (see line 1235 to see what the object contains). What you want is `$r['user']->user_id`.

frontend/server/libs/dao/Users.dao.php
```diff
@@ -148,4 +148,29 @@ public static function getHideTags($user_id) {
         global $conn;
         return $conn->GetOne($sql, $params);
     }
+    public static function FindClassName($user_id){
```
lhchavez: nit: try to follow the naming conventions and make this function start with a lowercase. Also, can you rename it to `getRankingClassName` to better reflect what it does?

Also, add a newline before this function and remove the newline in Line 174.

frontend/server/libs/dao/Users.dao.php
```diff
@@ -167,7 +167,7 @@ public static function FindClassName($user_id){
                 ORDER BY
                     `urc`.`percentile` ASC
                 LIMIT
-                    1;';
+                    1';
```
lhchavez: This semicolon was good!

frontend/server/controllers/UserController.php
randika96: Okay I will make the change.

lhchavez:
All right, this is now a good time to add the test!

You can start by copying https://github.com/omegaup/omegaup/blob/a62d6663d7cf8bada2ce4a659a146efd8dd71999/frontend/tests/controllers/UserRankTest.php#L91-L111 and modifying the `Call API` section to invoke the `apiProfile` instead:

```php
$response = UserController::apiProfile(new Request([
    'username' => $contestant->username
]);
```

Now you should assert that the `$response['userinfo']['classname']` is not `'user-rank-unranked'` and you're done!

Please refer to other tests in that same directory to get a feel for how tests should flow. The documentation for PHPUnit can be found here: https://phpunit.de/documentation.html

frontend/tests/controllers/UserRankTest.php
```diff
@@ -110,6 +110,23 @@ public function testUserRankByProblemsSolved() {
         $this->assertEquals($response['problems_solved'], 1);
     }
 
+    public function testUserRankingClassName() {
+        // Create a user and sumbit a run with him
```
lhchavez: Going forward we should try to avoid using genered pronouns as much as possible. Please change to `them`.

frontend/server/libs/dao/Users.dao.php
```diff
@@ -148,4 +148,29 @@ public static function getHideTags($user_id) {
         global $conn;
         return $conn->GetOne($sql, $params);
     }
+    
+    public static function getRankingClassName($user_id){
+        $sql = 'SELECT
+                    `urc`.`classname`,
+                    `urc`.`score`,
```
lhchavez: Given that you need to change the query, do you mind dropping `score` and `percentile`? We're not using them in PHP, so there's no use of shipping them across the wire.

frontend/server/libs/dao/Users.dao.php
```diff
@@ -148,4 +148,29 @@ public static function getHideTags($user_id) {
...
+                            `ur`.`score`
+                        FROM
+                            `User_Rank` `ur`
+                        WHERE
+                            `ur`.user_id` = ?
+                    )
+                ORDER BY
+                    `urc`.`percentile` ASC
+                LIMIT
+                    1';
```
lhchavez: Can you re-add the semicolon at the end of the query?

PR #1799, opened 2018-03-03

randika96:
Modified apiProfile() to return CSS class name according to the ranking

### DIFF
--- a/frontend/server/libs/dao/Users.dao.php
+++ b/frontend/server/libs/dao/Users.dao.php
@@ -148,4 +148,27 @@ class UsersDAO extends UsersDAOBase {
         global $conn;
         return $conn->GetOne($sql, $params);
     }
+
+       public static function getRankingClassName($user_id) {
+        $sql = 'SELECT
+                    `urc`.`classname`
+                FROM
+                    `User_Rank_Cutoffs` `urc`
+                WHERE
+                    `urc`.score >= (
+                        SELECT
+                            `ur`.`score`
+                        FROM
+                            `User_Rank` `ur`
+                        WHERE
+                            `ur`.user_id = ?
+                    )
+                ORDER BY
+                    `urc`.percentile ASC
+                LIMIT
+                    1;';
+        $params = [$user_id];
+        global $conn;
+        return $conn->GetOne($sql, $params) ?? 'user-rank-unranked';
+    }
 }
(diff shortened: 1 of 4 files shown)


frontend/server/libs/dao/Users.dao.php
```diff
@@ -148,4 +148,27 @@ public static function getHideTags($user_id) {
...
         return $conn->GetOne($sql, $params);
     }
+    
+   public static function getRankingClassName($user_id) {
+        $sql = 'SELECT
+                    `urc`.`classname`
+                FROM
+                    `User_Rank_Cutoffs` `urc`
+                WHERE
+                    `urc`.score < (
```
lhchavez: This needs to be

```mysql
`urc`.`score` >= (
```

otherwise you'll be getting the wrong cutoffs

lhchavez:
hmm are you still running `git push` outside the VM? Travis has some suggestions to fix the linting issue: https://travis-ci.org/omegaup/omegaup/jobs/348549443#L5689

And there are still a couple of failing unit tests :( Can you run the tests before pushing the next time?

Also, it seems like you need to add `classname` to https://github.com/omegaup/omegaup/blob/e51780ad71a96af28200cb295d8db7885b954f93/frontend/tests/controllers/UserProfileTest.php#L59

And one final thjing, you don't need to close the PR every time you push something. Just push a commit and the PR will update!

frontend/server/libs/dao/Users.dao.php
```diff
@@ -151,22 +151,18 @@ public static function getHideTags($user_id) {
     
    public static function getRankingClassName($user_id) {
         $sql = 'SELECT
-                    `urc`.`classname`
+                    urc . classname
```
lhchavez: oh no, the backticks were good. can you add them back?

Also, the condition is still wrong, it should be

```mysql
`urc`.`score` >= (
```

You just flipped the order but kept the condion the same, since `a < b` <=> `b > a`. Adding the = and the `DESC` will make it pass the unit test, but if you have more users, it will start to behave incorrectly :(

frontend/server/libs/dao/Users.dao.php
randika96: I can not push the code from inside the VM that is why I am pushing from outside :(

lhchavez: all right, let's fix that because otherwise you're just fighting against the tools and they will win since they don't get tired :( let's continue on slack.

PR #1805, opened 2018-03-04
Vueificar scoreboardmerge.php

CarlosCordova18:
Moving scoreboardmerge view to VueJS.
Tests and more information about this change could be found here: https://docs.google.com/document/d/1uzvJk1W0tStr0YsPzIts-7x4ZhQpetmQqtpbncJNPKg/edit?usp=sharing


frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
```diff
@@ -0,0 +1,98 @@
+<template>
+  <div>
+    <div class="post">
+      <div class="copy">
+        <legend>Concurso: <select class="contests"
```
lhchavez: Can you use `v-model="contests"`? https://vuejs.org/v2/guide/forms.html (see the "Multiple select (bound to Array)" section).

You'll also need to add `contests: []` to the `data` section.

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
```diff
@@ -0,0 +1,98 @@
...
+                id='contests'
+                multiple="multiple"
+                name='contests'
```
lhchavez: the `name` is also unneded here.

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
```diff
@@ -0,0 +1,98 @@
...
+    </div>
+  </div>
+</template>
+
+<script>
+import {T} from '../../omegaup.js';
+
+export default {
+  props: {
+    contests: Array,
```
lhchavez: can this be called `availableContests`?

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
```diff
@@ -0,0 +1,98 @@
...
+           id="get-merged-scoreboard"
+           v-on:click.prevent="displayTable">
+        Ver scoreboard total
+      </div>
+    </div>
+    <div class="post">
+      <div class="copy"
+           id="ranking"
+           v-if="showTable">
+        <table class="merged-scoreboard">
```
lhchavez: How about adding `v-if="scoreboard.length > 0"`? Then you can remove the `<div class="post">` and `<div class="copy">` AND get rid of `showTable` altogether :D

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
```diff
@@ -0,0 +1,98 @@
...
+          <tr>
+            <td></td>
+            <td><strong>Username</strong></td>
```
lhchavez: Can this be `{{ T.User }}`?

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
```diff
@@ -0,0 +1,98 @@
...
+          </tr>
+          <tr v-for="rank in scoreboard"
+              v-if="scoreboard">
+            <td><strong>{{ rank.place }}</strong></td>
+            <td>
+              <div class="username">
+                {{ rank.username }}
+              </div>
+              <div class="name">
+                {{ rank.username != rank.name ? rank.name : '&nbsp;' }}
```
lhchavez: you don't need the `&nbsp;` here. Just `''` is fine.

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
```diff
@@ -0,0 +1,98 @@
...
+              </div>
+            </td>
+            <td class="numeric"
+                colspan="2"
+                v-for="alias in aliases">({{ rank.contests[alias].points }}<span class=
+                "scoreboard-penalty"
+                  v-if="showPenalty">{{ ' ' + rank.contests[alias].penalty }}</span>)</td>
```
lhchavez: nit:

```
v-if="showPenalty"> {{ rank.contests[alias].penalty }}</span>)</td>
```

(so the space is before the curlies). same below.

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
```diff
@@ -0,0 +1,98 @@
...
+      <div class="copy"
+           id="ranking"
+           v-if="showTable">
+        <table class="merged-scoreboard">
+          <tr>
+            <td></td>
+            <td><strong>Username</strong></td>
+            <td colspan="2"
+                v-for="alias in aliases"
+                v-if="isMoreThanZero"><strong>{{ alias }}</strong></td>
```
lhchavez: Do you need the `v-if` here? It should do the right thing if `aliases` is empty. If you absolutely must use the `v-if` here, use `v-if="aliases.length > 0"`.

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
```diff
@@ -0,0 +1,98 @@
...
+            <td colspan="2"><strong>{{ total }}</strong></td>
+          </tr>
+          <tr v-for="rank in scoreboard"
+              v-if="scoreboard">
```
lhchavez: Same here, `v-if` should be unneeded.

frontend/www/js/omegaup/contest/scoreboardmerge.js
```diff
@@ -0,0 +1,93 @@
...
+                               })
+                .then(function(scoreboard) {
+                  const ranking = scoreboard.ranking;
+                  let sc = [], aliases = [];
+                  let data, place;
+                  let showPenalty = false;
+                  // Get values to pass through props
+                  if (ranking.length > 0) {
+                    // Show penalty or not
+                    for (var entry in ranking) {
```
lhchavez: here you can use

```javascript
for (const entry of ranking) {
  showPenalty |= !!entry.total.penalty;
}
```

frontend/www/js/omegaup/contest/scoreboardmerge.js
```diff
@@ -0,0 +1,93 @@
...
+            scoreboardMerge.showTable = false;
+            omegaup.API.Contest.scoreboardMerge({
+                                 contest_aliases:
+                                     contestAliases.map(encodeURIComponent)
+                                         .join(','),
+                               })
+                .then(function(scoreboard) {
+                  const ranking = scoreboard.ranking;
+                  let sc = [], aliases = [];
+                  let data, place;
```
lhchavez: You don't need to forward declare stuff in JavaScript. In fact, they're trying to remove hoisting altogether!

frontend/www/js/omegaup/contest/scoreboardmerge.js
```diff
@@ -0,0 +1,93 @@
...
+                        'place': place,
+                        'username': data.username,
+                        'name': data.name,
+                        'contests': data.contests,
+                        'totalPoints': data.total.points,
+                        'totalPenalty': data.total.penalty
+                      });
+                    }
+                  } else {
+                    sc = null;
```
lhchavez: You don't need this. You can leave the array empty.

frontend/www/js/omegaup/contest/scoreboardmerge.js
```diff
@@ -0,0 +1,93 @@
...
+                      data = ranking[entry];
+                      showPenalty |= !!data.total.penalty;
+                    }
+                    // Get aliases for indexing in the same order all rows
+                    for (var entry in ranking[0].contests) {
+                      aliases.push(entry);
+                    }
+
+                    // Create the scoreboard object
+                    for (var entry in ranking) {
```
lhchavez: same here:

```javascript
for (const index in ranking) {
   if (!ranking.hasOwnProperty(index)) continue;
   const entry = ranking[index];
   const place = parseInt(index) + 1;
   // ...
}
```

frontend/www/js/omegaup/contest/scoreboardmerge.js
```diff
@@ -0,0 +1,93 @@
...
+          'get-scoreboard': function(contestAliases) {
+            scoreboardMerge.showTable = false;
+            omegaup.API.Contest.scoreboardMerge({
+                                 contest_aliases:
+                                     contestAliases.map(encodeURIComponent)
+                                         .join(','),
+                               })
+                .then(function(scoreboard) {
+                  const ranking = scoreboard.ranking;
+                  let sc = [], aliases = [];
```
lhchavez: can you use `scoreboard` instead of `sc`?

frontend/www/js/omegaup/contest/scoreboardmerge.js
```diff
@@ -0,0 +1,93 @@
...
+                        'totalPoints': data.total.points,
+                        'totalPenalty': data.total.penalty
+                      });
+                    }
+                  } else {
+                    sc = null;
+                  }
+                  // Update the props values
+                  scoreboardMerge.aliases = aliases;
+                  scoreboardMerge.showPenalty = showPenalty ? true : false;
```
lhchavez: here

```javascript
scoreboardMerge.showPenalty = showPenalty;
```

is sufficient. in general every time you use `foo ? true : false` you can replace it with just `foo`.

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
```diff
@@ -0,0 +1,98 @@
+<template>
+  <div>
+    <div class="post">
+      <div class="copy">
+        <legend>Concurso: <select class="contests"
+                id='contests'
```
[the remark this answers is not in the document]
CarlosCordova18: I didn't think on that :open_mouth:

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
```diff
@@ -0,0 +1,98 @@
...
+    displayTable: function() {
+      let self = this;
+      let contestAliases = $('select.contests option:selected')
+                               .map(function() { return this.value })
+                               .get();
+      this.$emit('get-scoreboard', contestAliases);
+    }
+  },
+  data: function() {
+    return { total: T.wordsTotal, }
```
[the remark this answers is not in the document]
CarlosCordova18: So T is an object used to display strings in different languages?

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
```diff
@@ -0,0 +1,98 @@
...
+    showTable: Boolean,
+    isMoreThanZero: Boolean,
+    scoreboard: Array,
+    showPenalty: Boolean,
+    aliases: Array,
+  },
+  methods: {
+    displayTable: function() {
+      let self = this;
+      let contestAliases = $('select.contests option:selected')
```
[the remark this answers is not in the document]
CarlosCordova18: I think that `map` function will not be necessary anymore, as the `contest.alias` is in the value property of each option, and, as you suggested below, I'm using `v-model` to bind the selectedContests data property with the options selected. :smile:

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
CarlosCordova18: Definitely I need to read deeper that guide

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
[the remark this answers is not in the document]
lhchavez: correct!

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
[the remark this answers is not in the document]
lhchavez: oh right, i wrote this comment before the others.

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
CarlosCordova18: I'll need to set a background white to the table in order to display it similar to the old one.

frontend/www/js/omegaup/contest/scoreboardmerge.js
CarlosCordova18: As I'm reading, **hoisting** consists on the JS capability of using variables before they are declared, it doesn't work when a variable is declared and iniatilized at the same time.

What I'm doing here is just that, declaring and initializing the variables in the same expression, so I removing the hoisting. Why would it be bad?

CarlosCordova18:
I don't know why Webpack is giving me too much problems when recompiling the `.vue` and `.js` files.
Every time I make a change, I need to re-run the `yarn run dev` because webpack isn't watching my changes and automatically recompiling the file. In the compiled file, sometimes there are missing lines of code or missing semicolons.
Does this have anything to do with my code or is a webpack error or an error on the virtual machine? What should I do?

lhchavez:
hmm i've only seen that happen when you add new files. if you're just modifying stuff it _should_ work :S

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
CarlosCordova18: I'm making this change, but `T.User` is not defined yet, maybe you will define it later or `User` is not the name of the property?

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
```diff
@@ -39,13 +38,16 @@
...
             colspan="2"
-            v-for="alias in aliases">({{ rank.contests[alias].points }}<span class=
-            "scoreboard-penalty" v-if="showPenalty"> {{ rank.contests[alias].penalty }}</span>)</td>
+            v-for="alias in aliases">({{ rank.contests[alias].points }} <span class=
+            "scoreboard-penalty"
+              v-if="showPenalty">{{ rank.contests[alias].penalty }}</span>)</td>
         <td class="numeric"
-            colspan="2">({{ rank.totalPoints }}<span class="scoreboard-penalty" v-if="showPenalty"> {{ rank.totalPenalty }}</span>)</td>
+            colspan="2">({{ rank.totalPoints }} <span class="scoreboard-penalty"
+              v-if="showPenalty">{{ rank.totalPenalty }}</span>)</td>
```
CarlosCordova18: In this part and the previous one, the linter is automatically deleting the blank space before ` {{rank.totalPenalty}}`. How could I avoid that?

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
pabo99: You can define define the property in all language templates:

`en.lang` you need add `User = "Username"`
`es.lang` you need add `User = "Nombre de usuario"`
`pt.lang` you need add `User = "Nome de usuário"`

Alternatively you can look up this word in the templates to use it in your vue  file  (you can use `T.wordsUser` or `T.profileUsername`)

CarlosCordova18: Thanks, I'm pushing the commit for this change also.

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
```diff
@@ -0,0 +1,95 @@
...
+            v-for="alias in aliases">({{ rank.contests[alias].points }} <span class=
+            "scoreboard-penalty"
+              v-if="showPenalty">{{ rank.contests[alias].penalty }}</span>)</td>
+        <td class="numeric"
+            colspan="2">({{ rank.totalPoints }} <span class="scoreboard-penalty"
+              v-if="showPenalty">{{ rank.totalPenalty }}</span>)</td>
+      </tr>
+    </table>
+    <table class="merged-scoreboard"
+           v-else="">
```
CarlosCordova18: Linter is also adding that `v-else=""`, don't know what to do to avoid that corrections.

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
lhchavez: just ignore it u_u. The linter sometimes does suboptimal arrangements, but it's better to have something enforce some amount of consistency.

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
lhchavez: oh this is a bit more worrisome. let me take a look.

CarlosCordova18: So, should I push a new commit ignoring the linter for the cases that I mentioned, or you would fix it directly?

lhchavez: just saw that we have a lot of other `v-else=""` in the codebase, so Vue is able to cope with this :D let's not worry about it then, both this and the other linter issues.

CarlosCordova18: Nice! Is there any other modification that I need to do to finish this PR?
I think this is the first time I see this amount of comments in a single pull request 😞

lhchavez: nope, all's good! :D I'm just going through it manually and it looks like it's doing the right thing.

Issue #1797, opened 2018-03-02
Agregar links administrativos en la vista de editar concurso


lhchavez:
Como administrador de un concurso, me gustaría poder acceder a los enlaces que aparecen en `/contest/mine/` (como los links al scoreboard y el reporte de actividad):

![image](https://user-images.githubusercontent.com/168028/36914071-d7094252-1e00-11e8-9094-59f37488a875.png)

Dentro de la vista de edición del concurso.

## Posible Solución

Agregar una pestaña nueva en la [vista de edición de concurso](https://github.com/omegaup/omegaup/blob/master/frontend/templates/contest.edit.tpl) donde se agreguen los siguientes links:  https://github.com/omegaup/omegaup/blob/75da57a398cc7794c435a58bb00d21e890e2021a/frontend/www/js/omegaup/components/contest/ContestList.vue#L75-L99

alirizwi:
I want to take up this issue.

lhchavez:
assigned to @alirizwi!

alirizwi:
@lhchavez What should be the name of the new tab containing the links?

lhchavez:
`Links`, `Enlaces`, `Vínculos` in spanish, english, portuguese respectively.

alirizwi:
@lhchavez I want to use scorecard url and admin scorecard url in contest.edit.js file, but it seems that ```contest.scoreboard_url``` and ```contest.scoreboard_url_admin``` do not work. How can I use the scorecard urls?

lhchavez:
Oh, right. You also need to add those two columns to https://github.com/omegaup/omegaup/blob/8855dcb75086722877a887d146402a88ba94b9d8/frontend/server/controllers/ContestController.php#L551 and remove them from [`apiDetails()`](https://github.com/omegaup/omegaup/blob/8855dcb75086722877a887d146402a88ba94b9d8/frontend/server/controllers/ContestController.php#L656), probably just with

```php
unset $result['scoreboard_url'];
unset $result['scoreboard_url_admin'];
```

alirizwi:
@lhchavez Its still not working. On accessing through contest.edit.js file,  ```scoreboard_url``` and ```scoreboard_admin_url``` are returned as empty values.

lhchavez:
Can you create a new branch (`git commit checkout -b add-administrative-links && git push origin HEAD:add-administrative-links`) and send a pull request with all the changed files? I can probably help debug that.

alirizwi:
[...]

PR #1806, opened 2018-03-05
Added administrative links in the edit view contest

alirizwi:
Added administrative links in the edit view contest

Fixes: #1797


frontend/www/js/contest.edit.js
```diff
@@ -466,8 +470,44 @@ omegaup.OmegaUp.on('ready', function() {
...
+                                              .attr('href',
+                                                    '/arena/' +
+                                                        contestAlias + '/print/')
+                                              .text( 'Printable Version' )))
+                      
+                      .append($('<td></td>')
+                                  .append($('<a></a>')
+                                              .attr('href',
+                                                  '/arena/' +
+                                                  contestAlias + '/scoreboard/')
```
alirizwi: @lhchavez I want to use `scoreboard_url` and `scoreboard_admin_url` here, but `contest.scoreboard_url` and `contest.scoreboard_admin_url` does not work.

lhchavez: oh i see what'happening: `contest` is out of scope, so you cannot access it. you need to move this to after line 69, where the variable is still in scope.

frontend/www/js/contest.edit.js
```diff
@@ -118,6 +120,8 @@ omegaup.OmegaUp.on('ready', function() {
           feedback: $('#feedback').val(),
           penalty: $('#penalty').val(), public: public,
           scoreboard: $('#scoreboard').val(),
+          scoreboard_url: $('#scoreboard_url').val(),
```
lhchavez: you don't need this: this is only when you want to update the contest, and you never want to update the `scoreboard_url`.

frontend/www/js/contest.edit.js
```diff
@@ -48,6 +48,8 @@ omegaup.OmegaUp.on('ready', function() {
         $('.new_contest_form #public').val(contest.public);
         $('.new_contest_form #register').val(contest.contestant_must_register);
         $('.new_contest_form #scoreboard').val(contest.scoreboard);
+        $('#scoreboard-url').val(contest.scoreboard_url);
```
lhchavez: This is also unneeded: There are no elements that match `$('#scoreboard-url')`.

frontend/templates/contest.edit.tpl
```diff
@@ -183,6 +184,19 @@
 			</table>
 		</div>
 	</div>
+
+	<div class='tab-pane' id='links'>
```
lhchavez: Can you use double quotes `"` instead of single quotes `'`? for consistency.

frontend/templates/en.lang
```diff
@@ -99,6 +99,7 @@ contestDetailsWindowLength = "Window Length"
 contestDetailsYes = "Yes"
 contestEdit = "Edit contest"
 contestEditAddProblems = "Add problems"
+contestEditAdministrativeLinks = "Administrative Links"
```
lhchavez: https://travis-ci.org/omegaup/omegaup/jobs/349255615#L5691 is complaining that you didn't add this entry to all languages. Add it to es.lang and pt.lang and run `./stuff/i18n.py`. If it fails again, try merging the `upstream/master` branch (with `git pull upstream master`), which should bring a couple of improvements to that script.

frontend/www/js/contest.edit.js
alirizwi: Got your point. Will be updating PR soon.

frontend/templates/es.lang
```diff
@@ -99,6 +99,7 @@ contestDetailsWindowLength = "Window Length"
 contestDetailsYes = "Si"
 contestEdit = "Editar concurso"
 contestEditAddProblems = "Agrégale problemas!"
+contestEditAdministrativeLinks = "Administrativo Enlaces"
```
lhchavez: Can you use `Enlaces administrativos`?

frontend/templates/pt.lang
```diff
@@ -99,6 +99,7 @@ contestDetailsWindowLength = "Tamanho da janela"
 contestDetailsYes = "Sim"
 contestEdit = "Editar concurso"
 contestEditAddProblems = "Adicionar problemas"
+contestEditAdministrativeLinks = "Administrativo Vínculos"
```
lhchavez: Can you use `Vínculos administrativos`?

frontend/www/js/contest.edit.js
```diff
@@ -466,8 +471,52 @@ omegaup.OmegaUp.on('ready', function() {
...
+          $('#contest-admin-links')
+              .append(
+                  $('<tr></tr>')
+
+                      .append($('<td></td>')
+                                  .append($('<a></a>')
+                                              .attr('href', '/arena/' +
+                                                                contestAlias +
+                                                                '/admin/')
+                                              .text(' Submissions')))
```
lhchavez: Could it be possible to move all this to the `.tpl` file? Building stuff in jQuery is kind of hard :(

Also, can you use the translation strings for all the texts? so things would be like

```html
<tr>
  <td><a class="submissions">{#wordsSubmissions#}</a></td>
  ...
</tr>
```

And then on the jQuery side you can do

```javascript
$('#contest-admin-links .submissions').attr('href', ...);
...
```

which is much simpler :D It also avoids declaring `scoreboard_url` and `scoreboard_url_admin` at all!

frontend/templates/es.lang
alirizwi: Yeah sure!

frontend/templates/pt.lang
alirizwi: Yeah sure!

frontend/www/js/contest.edit.js
alirizwi: Yeah, this can be done! Will update the PR shortly.

PR #1811, opened 2018-03-07
Sanitizar datos de Grado academico

randika96:
Modified code to validate the Users academic degree


frontend/server/controllers/UserController.php
```diff
@@ -27,6 +27,16 @@ public static function apiCreate(Request $r) {
 
         Validators::isEmail($r['email'], 'email');
 
+        $allowedScholar_degrees = array_intersect(
+                $allowedScholar_degrees,
```
lhchavez: this is not defined anywhere. Can you do

```php
public static $kAllowedScholarDegrees = [
    'none', 'early_childhood', 'pre_primary', 'primary', 'lower_secondary',
    'upper_secondary', 'post_secondary', 'tertiary', 'bachelors', 'master',
    'doctorate'
];
```

at the top of the file? See https://github.com/omegaup/omegaup/blob/master/frontend/server/controllers/RunController.php for an example.

frontend/server/controllers/UserController.php
```diff
@@ -1260,6 +1270,7 @@ public static function apiProfile(Request $r) {
             ];
             $response['userinfo']['is_private'] = true;
         }
+        $response['userinfo']['classname'] = UsersDAO::getRankingClassName($r['user']->user_id);
```
lhchavez: this seems like leftovers from the previous change. can you remove it? Maybe also follow the guide shared by @heduenas to synchronize your code to `upstream/master`.

randika96: When make a new pr why still showing my previous changes? As those are already merged?

randika96: How to remove those from PR?

frontend/templates/user.edit.tpl
```diff
@@ -76,13 +76,17 @@
...
 						<div class="col-md-7">
 							<select name="scholar_degree" id="scholar_degree" class="form-control">
-								<option value="Elementary">{#userEditElementary#}</option>
-								<option value="Middle school">{#userEditMiddleSchool#}</option>
-								<option value="High school">{#userEditHighSchool#}</option>
-								<option value="Bachelor's">{#userEditBachelors#}</option>
-								<option value="Master's">{#userEditMasters#}</option>
-								<option value="Doctorate">{#userEditDoctorate#}</option>
-								<option value="Post-doc">Post-doc</option>
+								<option value="none">{#userEditNone#}</option>
```
lhchavez: Not all of these translation strings exist :( There are two options:

* You can remove any `option` that has a `userEditXxx` label that does not exist in `frontend/templates/en.lang`.
* Keep all the options and add all the missing `userEditXxx` to `frontend/templates/en.lang`, `es.lang`, and `pt.lang` (using Google Translate for the languages you don't know). Run `./stuff/i18n.py` once you're done to regenerate all the files and avoid Travis' complaints.

frontend/server/controllers/UserController.php
```diff
@@ -27,6 +27,14 @@ public static function apiCreate(Request $r) {
 
         Validators::isEmail($r['email'], 'email');
 
+        $kAllowedScholarDegrees = ['none', 'early_childhood', 'pre_primary', 'primary', 'lower_secondary','upper_secondary', 'post_secondary', 'tertiary', 'bachelors', 'master', 'doctorate'];
```
lhchavez: mind moving this with the rest of the `public static` things? Line 13 should be okay. If possible, also keep the format I suggested in the comment.

frontend/tests/factories/UserFactory.php
```diff
@@ -16,7 +16,7 @@ class UserFactory {
     * @param string $email optional
     * @return user (DAO)
     */
-    public static function createUser($username = null, $password = null, $email = null, $verify = true, $is_private = false) {
+    public static function createUser($username = null, $password = null, $email = null, $scholar_degree = 'none', $verify = true, $is_private = false) {
```
lhchavez: oh this'll probably break a bunch of tests (thanks, PHP, for not allowing named parameters :angry:).

the easiest fix is to move this parameter to the very end so that any callsite that is explicitly passing all other parameters will still work. i'll file another bug to clean this up later similar to https://github.com/omegaup/omegaup/issues/1440 so you don't have to do it (you've already helped plenty! someone else on our team will handle it)

lhchavez: Yeah, Travis confirmed that a bunch of stuff broke: https://travis-ci.org/omegaup/omegaup/jobs/350670991

randika96: Will it fix by moving to the end of the parameter list? Or have to follow any other approach? I would like to finish what I start

lhchavez: just move it to the end

frontend/tests/factories/UserFactory.php
```diff
@@ -16,7 +16,7 @@ class UserFactory {
     * @param string $email optional
     * @return user (DAO)
     */
-    public static function createUser($username = null, $password = null, $email = null, $scholar_degree = 'none', $verify = true, $is_private = false) {
+    public static function createUser($username = null, $password = null, $email = null, $scholar_degree = null, $verify = true, $is_private = false) {
```
lhchavez: A bit of explanation why it's necessary to move `$scholar_degree` to the very end of the parameter list:

There are some callsites that take advantage of the fact that all parameters are optional and take a default value, like https://github.com/omegaup/omegaup/blob/f56927b6fd7bc251d2c6f1c2db998725ac4cfcf0/frontend/tests/controllers/UserCreateTest.php#L298

since this relies on the fact that the order of the parameters is fixed, that callsite will send what it was intending to send to `$verify` into `$scholar_degree` , confusing both this code and the callsite!

So the correct fix is to declare this as

```php
public static function createUser($username = null, $password = null, $email = null, $verify = true, $is_private = false, $scholar_degree = 'none') {
```

(the `if` you have at line 33 is unneeded)

randika96: Oh.. I understood why it must be end of the list

PR #1817, opened 2018-03-08
Current coder of the month features

### LINKED ISSUE #1790: Visualizar top users del mes
*Como un usuario, quiero visualizar los top coders del mes para poder motivar la competición y se aprecie claramente al ganador*

## Comportamiento Esperado

*Así como podemos ver los top coders y las esculeas más activas, debería mostrarse una tabla con los top coders del mes*

## Comportamiento Actual

*No se muestra una tabla con los top coders del mes*

## Posible Solución

*Tendríamos que calcular el puntaje de los usuarios en el mes, seguro un query que agrupe por users y sume el puntaje en el úlitmo mes, y crear una tabla para la visualización*

SpaceWhite:
Fix #1790


frontend/server/controllers/UserController.php
```diff
@@ -1332,6 +1332,61 @@ public static function apiCoderOfTheMonth(Request $r) {
...
+    /**
+     * Get coder of the month by trying to find it in the table using
+     * today date. If there's no coder of the month for the given
+     * date, calculate it and save it.
+     *
+     * @param Request $r
+     * @return array
+     * @throws InvalidDatabaseOperationException
+     */
+    public static function apiCurrentCoderOfTheMonth() {
```
lhchavez: How is this different from `apiCoderOfTheMonth()`? Make sure you clearly document what the function does in the comment above.

frontend/server/controllers/UserController.php
```diff
@@ -1358,6 +1413,32 @@ public static function apiCoderOfTheMonthList(Request $r) {
...
+    /**
+     * Returns the Current top of coders of the month
+     *
+     * @param Request $r
+     */
+    public static function apiCurrentCoderOfTheMonthList() {
+        $response = [];
+        $response['coders'] = [];
+        try {
+            $coders = CoderOfTheMonthDAO::getCurrentCodersOfTheMonth();
```
lhchavez: can you change the name of these functions to `UserController::apiMonthlyRankList` and `CoderOfTheMonthDAO::getMonthlyRankList`?

Also, can you pass in the date range to calculate into `CoderOfTheMonthDAO::getMonthlyRankList`?

frontend/server/libs/dao/Coder_Of_The_Month.dao.php
```diff
@@ -94,6 +92,37 @@ public static function calculateCoderOfTheMonth($firstDay) {
...
+        return $allData;
+    }
+
+    /**
+     * Get all Current top coders of the month
+     *
+     * @static
+     * @return Array
+     */
+    final public static function getCurrentCodersOfTheMonth() {
```
lhchavez: This function does not take into account the formula in L48, so you're not actually calculating who would be winning for this month so far :(

Probably the better solution is to go with the suggestion I made earlier:

* Create a helper function that receives three parameters: start date, end date, and whether or not it should not include people who have been coders of the month in the past year.
* If it shouldn't include past coders of the month, include lines 65-67 into the SQL query, otherwise don't include them.
* Both `calculateCoderOfTheMonth` and `getMonthlyRankList` can call this helper function with the right parameters, avoiding duplicating any code!

frontend/server/nginx.rewrites
```diff
@@ -57,6 +57,7 @@ rewrite ^/arena/[a-zA-Z0-9_+-]+/admin/?$ /arena/admin.php last;
 rewrite ^/arena/[a-zA-Z0-9_+-]+/scoreboard/[a-zA-Z0-9_+-]+/?$ /arena/scoreboard.php last;
 rewrite ^/arena/([a-zA-Z0-9_+-]+)/?$ /arena/contest.php?contest_alias=$1 last;
 rewrite ^/coderofthemonth/?$ /coderofthemonth.php last;
+rewrite ^/currentcoderofthemonth/? /currentcoderofthemonth.php last;
```
lhchavez: same here: `monthlyranklist`

frontend/templates/index.tpl
```diff
@@ -87,8 +87,30 @@
 		</div>
 		{/if}
 
+		{if isset($currentCoderOfTheMonthData)}
```
lhchavez: @npave is currently trying to change the frontpage, so can you leave this file untouched? You can definitely change http://localhost:8080/coderofthemonth/ add a new tab to display the people that are on the top of this month's ranking.

frontend/server/controllers/UserController.php
SpaceWhite: @lhchavez 
Basically `apiCoderOfTheMonth()` and `apiCurrentCoderOfTheMonth()` do the same thing, only despite using the first day of the first month `apiCurrentCoderOfTheMonth` use the today's date `+1 month`.

By using today's date `+1 month`, `CoderOfTheMonthDAO::calculateCoderOfTheMonth` is calculating the coder ranking of this month

frontend/server/controllers/UserController.php
SpaceWhite: @lhchavez I'll change it

frontend/server/libs/dao/Coder_Of_The_Month.dao.php
SpaceWhite: Can you elaborate more about why it doesn't calculate the formula in line 48?

frontend/server/controllers/UserController.php
lhchavez: oh i see. but the issue asks for a list of top users for the month, not just one, so I think you can drop `apiCurrentCoderOfTheMonth` altogether and just keep `UserController::apiMonthlyRankList`.

frontend/server/libs/dao/Coder_Of_The_Month.dao.php
lhchavez: `CoderOfTheMonthDAO::getMonthlyRankList` should return exactly the same thing that `calculateCoderOfTheMonth` does with a few differences:

* It should return a list of the top ... let's say 50 coders given the current month so far.
* It should not exclude coders who have been Coder of the Month in the last year.
* It should perform the calculations based on different dates.

Your function merely returns the list of previous Coders of the Month :(

SpaceWhite:
@lhchavez It seems like I have misunderstood the problem, I think that the issue is about showing what candidate will be the Coder the month next month every day.

I'll make a new pull request for this issue

Issue #1750, opened 2018-02-15
Hacer opcional el mostrar usuario en el rank general


tvanessa:
*Como usuario, quiero tener la posibilidad de no aparecer en el rank general*

## Comportamiento Esperado

Que sea opcional aparecer en el rank

lhchavez:
This query needs to take into account the user's `is_private` setting:

https://github.com/omegaup/omegaup/blob/8855dcb75086722877a887d146402a88ba94b9d8/stuff/cron/update_user_rank.py#L37-L64

CarlosCordova18:
@lhchavez  I want this one!!!

lhchavez:
Assigned to @CarlosCordova18 !

CarlosCordova18:
One question, where do I need to create the test for this modification?

lhchavez:
Here's the best place to do so: https://github.com/omegaup/omegaup/blob/master/frontend/tests/controllers/UserRankTest.php there's already a bunch of tests, so copy one and modify it to fit your needs.

CarlosCordova18:
I've created the PR, if you can, please assign me another issue to fix :D

PR #1818, opened 2018-03-08
Haciendo opcional el mostrar usuario en el rank general.

CarlosCordova18:
This fixes the issue #1750


frontend/tests/controllers/UserRankTest.php
```diff
@@ -49,6 +49,34 @@ public function testFullRankByProblemSolved() {
...
+        $runDataPrivate = RunsFactory::createRunToProblem($problemData, $contestantPrivate);
+        RunsFactory::gradeRun($runDataPrivate);
+
+        // Refresh Rank
+        $this->refreshUserRank();
+
+        // Call API
+        $response = UserController::apiRankByProblemsSolved(new Request());
+
+        // Contestant should not appear in the rank as he's private.
```
lhchavez: nit: I'm pretty sure you just copied this from somewhere else in the codebase, but can you change the `he's` into `they're`? We're trying to move away from gendered language to be more inclusive :)

CarlosCordova18:
Actually, I created that comment from myself, good to know that is better to not set genders for inclusion +1

PR #1820, opened 2018-03-08
fixes #1772

### LINKED ISSUE #1772: [Color Hierarchy] Display "Rank" field below "Graduation date" in user's profile page
This depends on back end change #1771 

We should add the field “Rank” below “Graduation date” in https://omegaup.com/profile/xyz. Its value will be Unrated/Beginner/Specialist/Expert/Master/International Master and should be displayed with the color associated with it as described in the desing doc: https://docs.google.com/document/d/1R14hS-6TEhE34Aw3Pu_85jqIZjCyxVx6fiCHL8cvLg0/edit#

pj1527:
Adding Rank field below Graduation Date in User's Profile Page


frontend/templates/pseudo.lang
```diff
@@ -583,6 +583,7 @@ profileEdit = "(Edi7 pr0fi13)"
 profileEmail = "(Emai1)"
 profileGraduationDate = "(Gradua7i0n da73)"
 profileInfo = "(Inf0rma7i0n)"
+profileRank = "(Rank)"
```
heduenas: Can you run `./stuff/i18n.py`, this line should be generated automatically by that script rather than manually.

.gitignore
```diff
@@ -95,3 +95,5 @@ frontend/www/js/dist/
 # Ignore C++ / Pascal docs
 frontend/www/docs/cpp
 frontend/www/docs/pas
+frontend/server/libs/third_party/facebook-php-graph-sdk
```
lhchavez: welp, this shouldn't be in .gitignore. we use both of those O_O

can you remove them?

frontend/templates/profile.basicinfo.tpl
```diff
@@ -33,6 +33,11 @@
 						<div class="content">
 						</div>
 						</li>
+
+						<li class="fbSettingsListItem clearfix uiListItem"><a class="pvm phs fbSettingsListLink clearfix" ><span class="pls fbSettingsListItemLabel"><strong>{#profileRank#}</strong></span><span class="fbSettingsListItemContent fcg"><strong id="user-rank">{$profile.userinfo.rankinfo.rank}</strong></span></a>
```
lhchavez: can you use `{#wordsRanking#}` instead of `{#profileRank#}`?

.gitignore
```diff
@@ -95,3 +95,5 @@ frontend/www/js/dist/
...
+frontend/www/karel.js
```
lhchavez: instead can you add `ubuntu-xenial-16.04-cloudimg-console.log`?

frontend/templates/en.lang
```diff
@@ -583,6 +583,7 @@ profileEdit = "Edit profile"
 profileEmail = "Email"
 profileGraduationDate = "Graduation date"
 profileInfo = "Information"
+profileRank = "Rank"
```
lhchavez: oh right, can you also remove `profileRank` from all the `.lang` files?

lhchavez:
oh but that prints the ranking position, not the string like the issue description mentioned :( what needs to happen is that you need to convert `$profile.userinfo.classname` (which currently is a string like `user-rank-international-master`) into a translation string (which should be something like `userRankInternationalMaster`) and then use the translation string to get the actual translated content in `frontend/www/profile.php`:

```php
$smarty->assign('rankName', $smarty->getTemplateVars(preg_replace_callback(
    '/-[a-z]/',
    function ($matches) {
        return strtoupper($matches[0][1]);
    },
    $smarty->getTemplateVars('profile')['userinfo']['classname'])
));
```

Then you can display `{$rankName}` in the `.tpl` file. You'll also need to add all the necessary translation strings for all the ranks in the document and finally add the `$profile.userinfo.classname` as classname for the span. So in total, that should display

```html
<span class="user-rank-international-master">Maestro Internacional</span>
```

if it were displayed in Spanish.

lhchavez:
do you need any help with this? maybe @heduenas can chime in.

tvanessa:
El issue #1772 ya está cerrado, este PR se puede revertir.

PR #1837, opened 2018-03-14
[issue: 1831] modify logout redirect url

Sugirjan:
According to[ issue: 1831](https://github.com/omegaup/omegaup/issues/1831), When a user logs out omegaUp is redirected to '/logout?logout' and for the second time, if user logs in omegaUp tries to log out the user.([Code](https://github.com/omegaup/omegaup/blob/9fc24707fef3f7a4f7de52572cadbe5b26344ea5/frontend/www/js/login.js#L50)) So I removed the '/?logout' part from redirect URL.


Sugirjan:
Honestly, First I thought it was a bug related to google authentication process and dug into that. But after I came up with this easy solution. Thank you for merging PR

PR #1840, opened 2018-03-14
QualityNominationController checks for duplicate tags

dgarvit:
QualityNominationController checks for duplicate tags and returns an error if a tag is repeated.

Fixes #1838.

PR #1845, opened 2018-03-15
Language improvements on Contest Edit Page

### LINKED ISSUE #1844: Contest edit page has grammatical errors
The input label says "Problems" but actually we can add only 1 problem at a time. Also, it does not mention anywhere that one has to only enter the problem alias.

dgarvit:
Also added input placeholders to enhance user experience.
Fixes #1844.


frontend/templates/contest.edit.tpl
```diff
@@ -24,9 +24,9 @@
 			<div class="panel-body">
 				<form class="form" id="add-problem-form">
 					<div class="form-group">
-						<label for="problems-dropdown">{#wordsProblems#}</label>
+						<label for="problems-dropdown">{#wordsProblem#}</label>
```
lhchavez: Ah! good catch!

frontend/templates/contest.edit.tpl
```diff
@@ -24,9 +24,9 @@
...
 						<input class="typeahead form-control" name="problems" id="problems-dropdown"
-							autocomplete="off" />
+							autocomplete="off" placeholder="Enter problem alias" />
```
lhchavez: Think you can use a translated string like you did for https://github.com/omegaup/omegaup/pull/1840 ? otherwise people who browse the site in spanish or portuguese won't understand this :(

dgarvit: Sure I can do that! :)

So I need to create another entry in the lang files right?

lhchavez: yes, and here in this file you just need to use the `{#curlyWithPoundSigns#}` thing.

dgarvit: @lhchavez i made the requested changes. Please review :)

PR #1867, opened 2018-03-30
moved recaptcha script to head

princepurohit153:
moved recaptcha script to head

### DIFF
--- a/frontend/templates/head.tpl
+++ b/frontend/templates/head.tpl
@@ -76,7 +76,7 @@
 		<script type="text/plain" id="omegaup-enabled-experiments">{','|implode:$ENABLED_EXPERIMENTS}</script>
 {/if}
 {if isset($recaptchaFile)}
-	<script type="text/javascript" src="{$recaptchaFile}"></script>
+		<script type="text/javascript" src="{$recaptchaFile}"></script>
 {/if}
 	</head>
 	<body{if isset($bodyid) and $bodyid} id="{$bodyid|escape}"{/if}{if $smarty.const.OMEGAUP_LOCKDOWN} class="lockdown"{/if}>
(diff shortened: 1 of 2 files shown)


frontend/templates/login.tpl
```diff
@@ -121,5 +121,5 @@
 
 <script type="text/javascript" src="{version_hash src="/js/login.js"}"></script>
 <script src="https://apis.google.com/js/platform.js?onload=renderButton" defer></script>
-<script src='https://www.google.com/recaptcha/api.js'></script>
+<!-- <script src='https://www.google.com/recaptcha/api.js'></script> -->
```
lhchavez: Since we use git for everything, there's no point in committing commented-out code, since we can always view all the files' history and they only look ugly.

frontend/templates/head.tpl
```diff
@@ -75,6 +75,12 @@
 {if !empty($ENABLED_EXPERIMENTS)}
 		<script type="text/plain" id="omegaup-enabled-experiments">{','|implode:$ENABLED_EXPERIMENTS}</script>
 {/if}
+
+	{if isset($recaptchaFile)}
```
lhchavez: Do you mind following the style in the rest of the file?

```
{if isset($recaptchaFile)}
	<script type="text/javascript" src="{$recaptchaFile}"></script>
{/if}
```

frontend/templates/head.tpl
```diff
@@ -75,6 +75,12 @@
...
+	<script type="text/javascript" src="{$recaptchaFile}"></script>
+	{/if}
+
```
lhchavez: Oh, also avoid adding consecutive newlines: https://travis-ci.org/omegaup/omegaup/jobs/360380948

PR #1880, opened 2018-04-04
Vuefication of footer template

nvzard:


frontend/www/js/omegaup/components/Footer.vue
```diff
@@ -0,0 +1,54 @@
+<template>
```
lhchavez: Based on https://travis-ci.org/omegaup/omegaup/jobs/362129861, `<template>...</template>` should have exactly one child, so you need to wrap all the contents with a `<div>...</div>`.

Do that, and run the linter (see the invocation in https://travis-ci.org/omegaup/omegaup/jobs/362129860 ) and we should be good!

Issue #1222, opened 2017-05-26
el sistema esta en español pero los paises estan en ingles


alanboy:
Ademas de que no están en orden alfabético

![image](https://cloud.githubusercontent.com/assets/527987/26479512/aef875d0-4188-11e7-88d6-5216ce33239a.png)

heduenas:
Countries show up in sorted order as per https://github.com/omegaup/omegaup/pull/1861 but they still show up in English even when user language is set to Spanish, we need to fix that.

Divyansh00:
In the database, only the English name of countries are stored, so to show spanish names for the country we have to add spanish name for each country

NdubuisiO:
This link seems to have name in Spanish can you confirm  @heduenas http://www.101languages.net/spanish/country-names-spanish/

heduenas:
Yes, those are Spanish names. I think we will also need the 2 letter codes and Portuguese names.

Divyansh00:
@heduenas if we are using the same country codes, we can store the country name in all languages(English, Spanish and Portuguese)  separated by some delimiter(like comma) in the database and then extract the required country language by converting them to string by using split function.
This may not be optimal solution but will do the work.

heduenas:
I think a cleaner solution would be to store the translations in the .lang files as wordsMX, wordsJP, wordsUS, etc. and let the existing localization tools do the localization for us. @lhchavez, do you think that's a good approach?

SpaceWhite:
@heduenas Do you think #1861 is working. In the docs it states that the return of `sort` function is either true or false. So I think it is still not working http://php.net/manual/en/function.sort.php#refsect1-function.sort-returnvalues. And it seems like this issue has not been assigned to anyone else. Can someone assign me this issue?

heduenas:
@SpaceWhite What I understand from the link that you shared is that sort() does sort the array and additionally "returns TRUE on success or FALSE on failure". So, from that perspective, I don't see issues with #1861. However, we still need to fix the localization part of this issue.

SpaceWhite:
@heduenas But the for looping is using that sort return value (true/false) as looping array. I think this would be help https://ideone.com/AZZsJz.

heduenas:
[...]

PR #1889, opened 2018-04-07
Countries in spanish and portuguese

SpaceWhite:
Fix issue #1222


frontend/templates/user.edit.tpl
```diff
@@ -45,8 +45,16 @@
...
+                                {foreach $COUNTRIES as $key=>$country}
+                                    {$COUNTRIES.$key=[
+                                        'country_id' => $country->country_id,
+                                        'name' => $smarty.config["words`$country->country_id`"]
+                                    ]}
+                                {/foreach}
+                                {usort($COUNTRIES, create_function('$val1, $val2', 'return strcmp($val1[\'name\'], $val2[\'name\']);'))}
+
+								{foreach $COUNTRIES as $country}
+								<option value="{$country['country_id']}">{$country['name']}</option>
```
heduenas: Is this working correctly? I think it should be printed out as "#words*#" to work. Can you please attach a screenshot of your what local web app is showing?

SpaceWhite: @heduenas It is hard to get a screenshot so I'll give the page source code https://pastebin.com/T1akRFqW (You can see the spanish language in the options). https://www.smarty.net/docs/en/language.syntax.variables.tpl Actually if you see the documentation, it is mentioned that `#hashmark#` can also be called via `$smarty.config` variable.

heduenas: Makes sense.
Seems like you need to run ./stuff/i18n.py to please travis. Other than that, your PR looks good to me.

frontend/templates/es.lang
```diff
@@ -862,23 +957,87 @@ wordsGood = "Buena"
...
+wordsHR = "Croacia"
+wordsHT = "Haiti"
+wordsHU = "Hungary"
 wordsHard = "Difícil"
 wordsHighQualityProblem = "Este problema ha sido reconocido como un Problema de Calidad por un comité de jueces de omegaUp"
 wordsHomework = "Tarea"
 wordsHomeworks = "Tareas"
 wordsHours = "horas"
-wordsID = "ID"
+wordsID = "Indonesia"
```
heduenas: Seems like there is a collision between "Indonesia" and "ID". Maybe you can rename countries to "wordsCountryID", "wordsCountryIE", etc? It is possible that some other collisions like this one could happen if we leave it like "wordsID"/"wordsIE".

SpaceWhite: I didn't notice that, thanks for letting me know :D

frontend/templates/user.edit.tpl
lhchavez: welp, can we instead store these things in the database? something like extending the schema of https://github.com/omegaup/omegaup/blob/master/frontend/database/24_iso_3166_2.sql to include translations.

lhchavez: if you can find a spanish/portuguese version of the ISO-3166-2 dataset (which should also contain the administrative divisions in addition to country names), it would be ideal!

SpaceWhite: @lhchavez I can only found spanish version for now. I'll create a new pull request for this

lhchavez: all right, let's wait for the ISO dataset to avoid touching the template files at all :D

PR #1898, opened 2018-04-10
Unvalidated open redirect

### LINKED ISSUE #1883: Open redirect vulnerability in login page
Sample malicious url `https://omegaup.com/login/?redirect=https%3A//www.example.com/`
## Current behaviour
Redirect URL is not filtered, so that users are vulnerable to phising attack

## Expected behaviour
Redirect URL should be only available with only `omegaup.com` domain

## Possible solution
Add redirect validation in https://github.com/omegaup/omegaup/blob/0f63ff8b4c477155b2231ab4f79e1eb66e69253d/frontend/www/login.php#L48

SpaceWhite:
Fix #1883  Open redirect vuln

### DIFF
--- a/frontend/www/login.php
+++ b/frontend/www/login.php
@@ -46,10 +46,16 @@ if (isset($_GET['shva'])) {
 
 if ($c_Session->CurrentSessionAvailable()) {
     if (isset($_GET['redirect'])) {
-        die(header('Location: ' . $_GET['redirect']));
-    } else {
-        die(header('Location: /profile/'));
-    }
+        $redirect_hostname = parse_url($_GET['redirect'], PHP_URL_HOST);
+        //In case malformed URL is given redirect
+        if ($redirect_hostname !== false) {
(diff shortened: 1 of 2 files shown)


frontend/www/login.php
```diff
@@ -46,10 +46,16 @@
...
 if ($c_Session->CurrentSessionAvailable()) {
     if (isset($_GET['redirect'])) {
-        die(header('Location: ' . $_GET['redirect']));
-    } else {
-        die(header('Location: /profile/'));
-    }
+        $redirect_hostname = parse_url($_GET['redirect'], PHP_URL_HOST);
+        //In case malformed URL is given redirect
+        if ($redirect_hostname !== false) {
+            $allowed_hosts = ['blog.omegaup.com', $_SERVER['SERVER_NAME']];
```
pabloaguilar: We can tighten this further by checking `redirect.starts_with(OMEGAUP_URL)`
There is no login for the blog, so that's not needed as an allowed host, and the SERVER_NAME doesn't include the port number, which is important in the general case.

PR #1901, opened 2018-04-11
Fix open redirection vulnerability

### LINKED ISSUE #1883: Open redirect vulnerability in login page

SpaceWhite:
Fix #1883

### DIFF
--- a/frontend/www/login.php
+++ b/frontend/www/login.php
@@ -46,10 +46,16 @@ if (isset($_GET['shva'])) {
 
 if ($c_Session->CurrentSessionAvailable()) {
     if (isset($_GET['redirect'])) {
-        die(header('Location: ' . $_GET['redirect']));
-    } else {
-        die(header('Location: /profile/'));
-    }
+        $redirect_hostname = parse_url($_GET['redirect'], PHP_URL_HOST);
+        //In case malformed URL is given redirect
+        if ($redirect_hostname !== false) {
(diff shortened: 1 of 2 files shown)


frontend/www/login.php
```diff
@@ -46,10 +46,16 @@
 
 if ($c_Session->CurrentSessionAvailable()) {
     if (isset($_GET['redirect'])) {
-        die(header('Location: ' . $_GET['redirect']));
-    } else {
-        die(header('Location: /profile/'));
-    }
+        $redirect_hostname = parse_url($_GET['redirect'], PHP_URL_HOST);
+        //In case malformed URL is given redirect
+        if ($redirect_hostname !== false) {
```
lhchavez: do you mind combining the two conditions?

```php
if (in_array($redirect_hostname, $allowed_hosts, true))
```

given that you're asking for strict compariton, the `false` check is not needed :D

Issue #1902, opened 2018-04-11
System admin potential CSRF target on change password API


SpaceWhite:
## Current behaviour
Currently admin can directly change user password without any additional authentication (such as old_password, captcha, etc). This can be used as a target of CSRF.

Example: omegaup.com/api/user/changepassword?user=user1&password=maliciouspass

## Expected behaviour
There should be some additional authentication (maybe captcha) to mitigate CSRF when admin want to attempt password changes

lhchavez:
welp

lhchavez:
Turns out this particular API is only used in one place, and we already have a better alternative when there is a previous interaction with the support team, so don't need it anymore \o/ but our APIs still don't have any CSRF protection, which is kind of scary. I'll fix this quickly (by simply not using this API at all) and open a new one to enable CSRF protection across the board.

PR #1906, opened 2018-04-11
Primer intento de arreglar CSRF

lhchavez:
Este cambio agrega una verificación del Referer como mitigación de CSRF.
Este es el primer paso para #1905


lhchavez:
Thanks to @SpaceWhite for reporting!

SpaceWhite:
@lhchavez But this will not mitigate CSRF like #1883, because the referrer is the the same as `$_SERVER['SERVER_NAME']`.

lhchavez:
This will mitigate only a subset. Say you are on the open web and click a random link to an arbitrary API like you reported on https://github.com/omegaup/omegaup/issues/1902 . CSRF that comes from the omegaup.com domain are still affected (like the one you report in #1883), but I'd still like to have several layers of defenses.

frontend/www/api/ApiCaller.php
```diff
@@ -52,6 +52,10 @@ public static function httpEntryPoint() {
         $r = null;
         try {
             $r = self::init();
+            if (!empty($_SERVER['HTTP_REFERER']) &&
+                parse_url($_SERVER['HTTP_REFERER'], PHP_URL_HOST) != $_SERVER['SERVER_NAME']) {
```
pabloaguilar: How about using `OMEGAUP_URL` instead of trusting `_SERVER` vars? Something along the lines of `referrer.starts_with(OMEGAUP_URL)`

`OMEGAUP_URL` is what we use in `SessionController` for generating OAuth callback URLs.

lhchavez: oh! using `OMEGAUP_URL` might be what's needed to make Travis happy. Having said that, we might need to still do some parsing because `OMEGAUP_URL` doesn't have the trailing slash, so someone could try to do CSRF with `https://omegaup.com.evildomain.com`. All that's left to do is to figure out what hostname the Selenium magic is using.

lhchavez: that did it :D thanks!

PR #1918, opened 2018-04-13
Fix redirect glitch when anonymous user visit course page.

### LINKED ISSUE #1888: Glitch when anonymous user visit course
## Current behaviour
When anonymous user (not logger in user) visit `omegaup.com/course` page, user will be redirected to login page, but there is some UI glitch on redirection

## Expected behaviour
There should be no glitch on redirection

dgarvit:
Fixes #1888.


dgarvit:
It was a quick fix. :)

PR #1922, opened 2018-04-14
Pseudo language option should only be shown to user in developer environments.

### LINKED ISSUE #1891: Pseudo language option should only be shown to user in developer/sandbox environments
## Current behaviour
In `omegaup/profile/edit` pseudo language is available

## Expected behaviour
This option should be disable because it is just an i18n identifier

dgarvit:
Fixes #1891.


frontend/templates/user.edit.tpl
```diff
@@ -74,7 +74,9 @@
 							<option value="es">{#wordsSpanish#}</option>
 							<option value="en">{#wordsEnglish#}</option>
 							<option value="pt">{#wordsPortuguese#}</option>
+							{if $DEVELOPMENT_MODE}
```
lhchavez: How about 

```
{if "OMEGAUP_DEVELOPMENT_MODE"|defined && $smarty.const.OMEGAUP_DEVELOPMENT_MODE}
```

instead?

dgarvit:
I made the changes. @lhchavez please review! :)

PR #1923, opened 2018-04-14
Preferred programming language should have full language names on profile edit page

### LINKED ISSUE #1917: Preferred programming language should have full language names on profile edit page
## Current behaviour

Preferred programming language on `profile/edit` page has a list of languages with only language extensions (eg: py for Python).

## Expected behaviour

Preferred programming language on `profile/edit` page has a list of languages with complete names of languages.

dgarvit:
Fixes #1917.


frontend/server/controllers/RunController.php
```diff
@@ -10,6 +10,22 @@ class RunController extends Controller {
...
+    public static $languageFullNames = [
+        'kp' => 'Karel (Pascal)',
+        'kj' => 'Karel (Java)',
+        'c' => 'C',
+        'cpp' => 'C++',
+        'cpp11' => 'C++ 11',
+        'java' => 'Java',
+        'py' => 'Python',
+        'rb' => 'Ruby',
+        'pl' => 'PL/I',
```
lhchavez: This is supposed to be `Perl`.

frontend/server/controllers/RunController.php
```diff
@@ -10,6 +10,22 @@ class RunController extends Controller {
         'kp', 'kj', 'c', 'cpp', 'cpp11', 'java', 'py', 'rb', 'pl', 'cs', 'pas',
         'cat', 'hs', 'lua'
     ];
+    public static $languageFullNames = [
```
lhchavez: Would it be possible to merge this into `$kSupportedLanguages`? All the places where [`$kSupportedLanguages` is used](https://github.com/omegaup/omegaup/search?utf8=%E2%9C%93&q=kSupportedLanguages&type=) would need to be updated to support using a hash instead of a normal array. e.g. where `Validators::isInEnum` is used, a new function `isInArrayKeys` should be used instead, and use [`array_keys`](http://php.net/array_keys) in the other places.

frontend/templates/user.edit.tpl
```diff
@@ -103,8 +103,8 @@
 						<div class="col-md-7">
 						<select id="programming_language" name="programming_language" class="form-control" >
 							<option value=""></option>
-							{foreach from=$PROGRAMMING_LANGUAGES item=programming_language}
-							<option value="{$programming_language}">{$programming_language}</option>
+							{foreach from=$PROGRAMMING_LANGUAGES key=keyword item=programming_language}
```
lhchavez: nit: maybe `key=extension item=name`?

frontend/server/controllers/RunController.php
```diff
@@ -10,6 +10,22 @@ class RunController extends Controller {
...
+        'c' => 'C',
+        'cpp' => 'C++',
+        'cpp11' => 'C++ 11',
+        'java' => 'Java',
+        'py' => 'Python',
+        'rb' => 'Ruby',
+        'pl' => 'PL/I',
+        'cs' => 'C#',
+        'pas' => 'Pascal',
+        'cat' => 'Cat',
```
lhchavez: Ah forgot about this :( this needs some translation string, because this is supposed to be `Output only`. The problem is that it should render as `Sólo salida` in spanish and `Somente saída` in portuguese.

frontend/server/controllers/RunController.php
dgarvit: I think we should go by this procedure. Your opinion?

lhchavez: no real opinon. just take into account that translation will be required for both, because of the `cat` / `Output Only` entry. That one must be translated regardless.

dgarvit:
@lhchavez I made the changes! Please review!

PR #1924, opened 2018-04-14
User without authorisation cannot clone a contest

### LINKED ISSUE #1920: User without authorization can clone contest
## Current behaviour
User without authorization to access other person private contest, can create clone contest by only providing `contest_alias`

## Expected behaviour
Before clone a contest, check if the contest is available to the user

## Possible solution
Add `self::canAccessContest` function to here
https://github.com/omegaup/omegaup/blob/3696e0adeafd710cde61bed1bffb7303a8bb97fb/frontend/server/controllers/ContestController.php#L835

dgarvit:
Fixes #1920.


frontend/server/controllers/ContestController.php
```diff
@@ -833,6 +833,8 @@ public static function apiClone(Request $r) {
         // Authenticate user
         self::authenticateRequest($r);
 
+        self::canAccessContest($r); // Make sure user has access to the contest to be able to clone it
```
lhchavez: this should be

```php
if (!Authorization::isContestAdmin($r['current_user_id'], $r['contest'])) {
    throw new ForbiddenAccessException();
}
```

otherwise contestants without (administrative) access to the contest can still clone it.

SpaceWhite: That makes sense :D

dgarvit:
Fixed the authentication, and added the test! @lhchavez please review! :)

PR #1928, opened 2018-04-15
Environment in default config should be 'development'

pabloaguilar:
#1922 accidentally set the environment to 'production' by default


lhchavez:
all the defaults should be production-safe. it's preferable to accidentally break development environments than production D:

pabloaguilar:
Ack.
Closing this in favor of https://github.com/omegaup/puppet/pull/5

PR #1938, opened 2018-04-18
Issue1790

### LINKED ISSUE #1790: Visualizar top users del mes

SpaceWhite:
Fix #1790


frontend/server/controllers/UserController.php
```diff
@@ -1388,41 +1388,42 @@ public static function apiCoderOfTheMonth(Request $r) {
             // Get first day of the current month
             $firstDay = date('Y-m-01');
         }
+
         try {
-            $coderOfTheMonth = null;
+            $codersOfTheMonth = [];
```
lhchavez: nit: this is unnecessary

frontend/database/57_coder_of_the_month_rank.sql
```diff
@@ -0,0 +1,2 @@
+-- Add rank column to coder of the month
+ALTER TABLE Coder_Of_The_Month ADD COLUMN rank int(11);
```
lhchavez: Can you also add an `UPDATE Coder_Of_The_Month SET rank=1;`? Otherwise all previous CotM will have rank = 0 D:

Also, once you're done with that, can you `ALTER` the column so that it's `NOT NULL`?

frontend/server/libs/dao/Coder_Of_The_Month.dao.php
```diff
@@ -64,31 +64,27 @@ public static function calculateCoderOfTheMonth($firstDay) {
...
+                score DESC
+			LIMIT 100
 		";
 
         $val = [$startTime, $endTime, $endTime];
 
         global $conn;
-        $rs = $conn->GetRow($sql, $val);
-        if (count($rs) == 0) {
+        $results = $conn->getAll($sql, $val);
```
lhchavez: For this query, you don't need the `username` or `name` anymore since you'll only need the `user_id` in the caller. that'll make this way faster :)

frontend/server/libs/dao/Coder_Of_The_Month.dao.php
```diff
@@ -64,31 +64,27 @@ public static function calculateCoderOfTheMonth($firstDay) {
...
 				Coder_Of_The_Month cm on u.user_id = cm.user_id
 			WHERE
-				cm.user_id IS NULL OR DATE_ADD(cm.time, INTERVAL 1 YEAR) < ?
+				cm.user_id IS NULL OR (DATE_ADD(cm.time, INTERVAL 1 YEAR) < ? AND cm.rank = 1)
 			GROUP BY
 				user_id
 			ORDER BY
-				score DESC
-			LIMIT 1
+                score DESC
```
lhchavez: nit: can you leave this using tabs instead of spaces? I guess we'll have to fix that later on a separate change to avoid this looking inconsistent.

frontend/www/coderofthemonth.php
```diff
@@ -1,6 +1,7 @@
 <?php
 
 require_once('../server/bootstrap.php');
+$smarty->clearAllCache();
```
lhchavez: hmm this is going to be way too expensive. Why do you need it? We'll need to figure out a way to avoid adding it.

frontend/www/js/omegaup/components/coderofthemonth/CoderOfTheMonth.vue
```diff
@@ -0,0 +1,78 @@
+<template>
+  <div class="wait_for_ajax panel panel-default">
+    <div class="panel-heading">
+      <ul class="nav nav-tabs">
+        <li class="active"
+            v-on:click="show = true">
+          <a data-toggle="tab">{{UI.formatString(T.codersOfTheMonth)}}</a>
```
lhchavez: All of these translated strings don't have any format arguments, so you can just do

    {{ T.codersOfTheMonth }}

Same throughout the file.

frontend/www/js/omegaup/components/coderofthemonth/CoderOfTheMonth.vue
```diff
@@ -0,0 +1,78 @@
...
+    var mm = today.getMonth() + 1;
+    var dd = today.getDate();
+
+    var todayString = [
+      today.getFullYear(),
+      (mm > 9 ? '' : '0') + mm,
+      (dd > 9 ? '' : '0') + dd
+    ].join('-');
+    API.User.coderOfTheMonthList({date: todayString})
+        .then(function(response) { self.coders_monthly = response.coders; })
```
lhchavez: I'm a bit confused about which array represents what. Can you change `coders` and `coders_monthly` to something more appropriate? Like `pastMonthCoders` and `currentMonthCoders`?

frontend/www/js/omegaup/components/coderofthemonth/CoderOfTheMonth.vue
```diff
@@ -0,0 +1,78 @@
...
+    API.User.coderOfTheMonthList()
+        .then(function(response) { self.coders = response.coders; })
+        .fail(UI.apiError);
+
+    // coder of the month list
+    var today = new Date();
+    var mm = today.getMonth() + 1;
+    var dd = today.getDate();
+
+    var todayString = [
```
lhchavez: How about

```javascript
var nextMonth = new Date();
nextMonth.setMonth(nextMonth.getMonth()+1);
...
date: nextMonth.format('{yyyy}-{MM}-{dd}')
```

? JavaScript's [`Date.setMonth`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setMonth) takes care of overflow :)

frontend/www/js/omegaup/components/coderofthemonth/CoderOfTheMonth.vue
```diff
@@ -0,0 +1,78 @@
...
+
+<script>
+import {API, T} from '../../omegaup.js';
+import UI from '../../ui.js';
+
+export default {
+  props: {},
+  computed: {},
+  data: function() {
+    return {T: T, UI: UI, coders: [], coders_monthly: [], show: true};
```
lhchavez: Can you change `show` into `showCurrentMonth` or something more descriptive?

frontend/www/js/omegaup/components/coderofthemonth/CoderOfTheMonth.vue
```diff
@@ -0,0 +1,78 @@
...
+          <th></th>
+          <th>{{UI.formatString(T.codersOfTheMonthCountry)}}</th>
+          <th>{{UI.formatString(T.codersOfTheMonthUser)}}</th>
+          <th>{{UI.formatString(T.codersOfTheMonthDate)}}</th>
+        </tr>
+      </thead>
+      <tbody v-if="show">
+        <tr v-for="coder in coders">
+          <td><img v-bind:src="coder.gravatar_32"></td>
+          <td><img v-bind:src="'/media/flags/' + coder.country_id.toLowerCase() + '.png'"
```
lhchavez: @CarlosCordova18 taught me that we can use [`template literals`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) in Vue due to Babel :D, like in https://github.com/omegaup/omegaup/blob/master/frontend/www/js/omegaup/components/user/Profile.vue#L43

```html
<img v-bind:src="`/media/flags/${coder.country_id.toLowerCase()}.png`" ...>
```

lhchavez:
oh right, can you expand a bit on what the code is doing write a one or two paragraphs about it in the Pull Request description? See all the other Pull Requests for examples.

It took me a bit of time to figure it out, and we'd like for someone who's not super familiar with the codebase to be able to figure it out quicker.

SpaceWhite:
@lhchavez Thanks for the review, I've read all your review and I'll change my code. And I'll also thanks for reminding me, I'll provide PR description from now on.

frontend/www/js/omegaup/components/coderofthemonth/CoderOfTheMonth.vue
```diff
@@ -0,0 +1,76 @@
...
+import UI from '../../ui.js';
+
+export default {
+  props: {},
+  computed: {},
+  data: function() {
+    return {
+      T: T,
+      UI: UI,
+      coders: [],
```
lhchavez: OK, I think I now get what all this is doing. Can you rename

* `coders` => `codersOfCurrentMonth`
* `coders_monthly` => `codersOfPreviousMonth`
* `showCoderOfTheMonth` => `showCurrentMonth`

Update the comments in L63, L69, and I *think* that should be it!

frontend/server/libs/dao/Coder_Of_The_Month.dao.php
```diff
@@ -64,31 +64,27 @@ public static function calculateCoderOfTheMonth($firstDay) {
 			LEFT JOIN
 				Coder_Of_The_Month cm on u.user_id = cm.user_id
 			WHERE
-				cm.user_id IS NULL OR DATE_ADD(cm.time, INTERVAL 1 YEAR) < ?
+				cm.user_id IS NULL OR (DATE_ADD(cm.time, INTERVAL 1 YEAR) < ? AND cm.rank = 1)
```
pabloaguilar: This doesn't look correct to me. It reads to me as:
"A candidate to coder of the month is someone who hasn't been considered for coder of the month, or was the top ranked coder over a year ago"
But I think we want 
"A candidate to coder of the month is someone who hasn't been considered for coder of the month, or hasn't been the the top ranked in at least a year"
That is, I think you need to do another sub-select to get the "latest time that a user was number one":
```
LEFT JOIN
  (SELECT user_id, MAX(time) latest_time FROM Coder_Of_The_Month GROUP BY user_id HAVING rank = 1) AS cm
WHERE
 cm.user_id IS NULL OR DATE_ADD(cm.latest_time, INTERVAL 1 YEAR) < ?
```

frontend/www/js/omegaup/components/coderofthemonth/CoderOfTheMonth.vue
```diff
@@ -0,0 +1,76 @@
...
+      <thead>
+        <tr>
+          <th></th>
+          <th>{{T.codersOfTheMonthCountry}}</th>
+          <th>{{T.codersOfTheMonthUser}}</th>
+          <th>{{T.codersOfTheMonthDate}}</th>
+        </tr>
+      </thead>
+      <tbody v-if="showCoderOfTheMonth">
+        <tr v-for="coder in coders">
```
pabloaguilar: This is better done by using a computed property, `visibleCoders` whose value depends on `showCurrentMonth` and then you only have one table definition here.

frontend/www/js/omegaup/components/coderofthemonth/CoderOfTheMonth.vue
```diff
@@ -0,0 +1,76 @@
...
+  </div>
+</template>
+
+<script>
+import {API, T} from '../../omegaup.js';
+import UI from '../../ui.js';
+
+export default {
+  props: {},
+  computed: {},
```
pabloaguilar: Add a computed property here:
`visibleCoders: function() { return showCurrentMonth ? codersOfTheCurrentMonth : codersOfPreviousMonths; }`

frontend/www/js/omegaup/components/coderofthemonth/CoderOfTheMonth.vue
```diff
@@ -0,0 +1,76 @@
...
+          <th>{{T.codersOfTheMonthDate}}</th>
+        </tr>
+      </thead>
+      <tbody v-if="showCoderOfTheMonth">
+        <tr v-for="coder in coders">
+          <td><img v-bind:src="coder.gravatar_32"></td>
+          <td><img v-bind:src="`/media/flags/${coder.country_id.toLowerCase()}.png`"
+               v-if="coder.country_id != null"></td>
+          <td>{{coder.username}}</td>
+          <td>{{coder.date}}</td>
```
pabloaguilar: Maybe hide the date when showing the users of the current month?
I mean, I don't think it's useful to have 100 rows showing the exact same month

frontend/server/libs/dao/Coder_Of_The_Month.dao.php
SpaceWhite: @pabloaguilar In order to do this we need to disable `ONLY_FULL_GROUP_BY` in mysql global. Should I set the global value in migration?

pabloaguilar: Hmm... I think you can just rewrite the query, either by changing `HAVING` for `WHERE` or by naming `rank` in the `GROUP BY` clause. It shouldn't be necessary to change the `ONLY_FULL_GROUP_BY` configuration

SpaceWhite:
PTAL @lhchavez @pabloaguilar

frontend/database/58_coder_of_the_month_rank.sql
```diff
@@ -1,4 +1,5 @@
 -- Add rank column to coder of the month
+-- Rank here implies rank of user for a particular time perod, not the same thing as user's global rank
```
lhchavez: i think @pabloaguilar meant something like

```sql
ALTER TABLE
    Coder_Of_The_Month
ADD COLUMN
    rank int(11) NOT NULL COMMENT 'El lugar en el que el usuario estuvo durante ese mes';
```

Issue #1865, opened 2018-03-30
When editing statements in multiple languages only one is saved.


frcepeda:
## Expected Behavior

When I edit a problem statement in multiple languages in one go, all of my changes are saved.

## Current Behavior

When I edit multiple languages of a statement, only the one currently selected when I hit submit gets saved, and the edits for other languages discarded.

## Steps to Reproduce

1. Create a problem
2. Go to the statement edit view
3. Edit one language (e.g. Spanish)
4. Edit another language (e.g. English)
5. Select submit
6. Only the currently selected language in the dropdown gets updated.

SpaceWhite:
This bug probably caused by failing Promise
![screenshot from 2018-04-28 23-58-32](https://user-images.githubusercontent.com/15065085/39403130-80751412-4b40-11e8-80b0-31c4155cce7b.png)
This is because fail prototype isn't available on js promise
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#Promise_prototype

PR #1953, opened 2018-04-29
Hotfix Promise prototype

SpaceWhite:
Rename Promise fail method to catch #1865 

We need to also add catch function to hook_tools


SpaceWhite:
@lhchavez How about despite using promise in javascript, we use `async`? By doing this maybe we can get rid of linting error.

PR #1954, opened 2018-04-29
Fix delay social button on load

### LINKED ISSUE #1872: [Performance bug] Botones sociales retrasan el on-load de la página
*Como un usuario normal, quiero que la página cargue lo más pronto posible. (para disminuir bounce rates!)

## Comportamiento Esperado

La página carga en no más de un segundo.

## Comportamiento Actual

Hay muchas cosas que la hacen lenta.. pero una de ellas es que el on-load no sucede sino hasta que se cargaron los botones sociales (Twitter y FB que aparecen en el footer).

## Posible Solución

Cargar los botones sociales en el evento `OmegaUp.on('ready')` para evitar bloquear el tiempo de carga.

Screenshot de Chrome dev console:
image

SpaceWhite:
Render facebook and twitter buttons after omegaup ready was triggered.  Fix #1872


frontend/www/ux/social-button.js
```diff
@@ -0,0 +1,33 @@
+omegaup.OmegaUp.on('ready', function() {
+  // facebook button
+  (function(d, s, id) {
```
lhchavez: is there any benefit to introducing these [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE)? Otherwise it might be better to just inline them to reduce the cognitive burden while reading the code.

frontend/www/ux/social-button.js
```diff
@@ -0,0 +1,33 @@
...
+    var js, fjs = d.getElementsByTagName(s)[0];
+    if (d.getElementById(id)) return;
+    js = d.createElement(s);
+    js.id = id;
+    js.src = 'https://connect.facebook.net/en_US/sdk.js#xfbml=1&version=v2.12';
+    fjs.parentNode.insertBefore(js, fjs);
```
lhchavez: Is there any benefit to inserting the script before any other scripts? By the time this is evaluated, every single other script should have been loaded, so you should be able to append it to the very end.

Same below.

frontend/www/ux/social-button.js
```diff
@@ -0,0 +1,33 @@
+omegaup.OmegaUp.on('ready', function() {
+  // facebook button
+  (function(d, s, id) {
+    var js, fjs = d.getElementsByTagName(s)[0];
```
lhchavez: nit: Try to avoid declarations without an initialization. I know that JavaScript does this kind of hoisting for you (ugh), so let's leave that magic to the compiler and try to use more modern style and [Declare Variables At First Use](http://wiki.c2.com/?DeclareVariablesAtFirstUse). This will be especially important when we start using Babel and `let` more and more, which has different semantics for this.

frontend/www/ux/social-button.js
```diff
@@ -0,0 +1,33 @@
...
+    if (d.getElementById(id)) return;
+    js = d.createElement(s);
```
lhchavez: nit: can you use a bit more descriptive variable name? Something like

```javascript
var facebookScriptElement = document.createElement('facebook-jssdk');
```

frontend/www/ux/social-button.js
```diff
@@ -0,0 +1,33 @@
+omegaup.OmegaUp.on('ready', function() {
+  // facebook button
+  (function(d, s, id) {
+    var js, fjs = d.getElementsByTagName(s)[0];
+    if (d.getElementById(id)) return;
```
lhchavez: IIUC the `omegaup.OmegaUp.on('ready', ...)` should guarantee that this is only run once. Is it not what you are observing? If so, we have a bug and we should fix it D:

frontend/www/ux/social-button.js
```diff
@@ -0,0 +1,33 @@
...
+    js = d.createElement(s);
+    js.id = id;
+    js.src = 'https://connect.facebook.net/en_US/sdk.js#xfbml=1&version=v2.12';
+    fjs.parentNode.insertBefore(js, fjs);
+  }(document, 'script', 'facebook-jssdk'));
+
+  // twitter button
+  twttr = (function(d, s, id) {
+    var js, fjs = d.getElementsByTagName(s)[0], t = window.twttr || {};
```
lhchavez: I'm afraid that this is depending a lot on the implementation details :( Whenever they change the implementation, this will break.

SpaceWhite:
@lhchavez PTAL

PR #2004, opened 2018-05-15
[Ghost mode] Server side development

SpaceWhite:
## Description
This PR enable OmegaUp server side to create ghost mode (aka virtual contest) for contestant.
NB : client side ghost mode will be in another PR

### API create
To create ghost mode trigger the `apiCreate` by the following:
`domain.com/api/contest/create?auth_token=(put your token here)&contest_alias=cont1&virtual=1`

This API will trigger `validateCreateVirtual` if `virtual` query exists. Then it will recreate a contest from the real contest and also changing the `ACL`, `Problemset`, and etc

### Ghost mode restriction
Ghost mode will utilize the API like the real contest (eg. `apiRuns`, `apiDetails`, etc). But there are some API that need to be restricted in ghost mode.

### Scoreboard event
Ghost mode will simulate scoreboard, hence scoreboard event is needed in order to do so. Scoreboard event will find the real contest scoreboard and merge with contestant runs.

### Future works
Currently `apiRuns` doesn't show the list of runs from real contest, should we simulate the @lhchavez  apiRuns too?


frontend/database/1_initial_schema.sql
```diff
@@ -108,7 +108,7 @@ CREATE TABLE IF NOT EXISTS `Contests` (
   `finish_time` timestamp NOT NULL DEFAULT '2000-01-01 06:00:00' COMMENT 'Hora de finalizacion de este concurso',
   `window_length` int(11) DEFAULT NULL COMMENT 'Indica el tiempo que tiene el usuario para envíar solución, si es NULL entonces será durante todo el tiempo del concurso',
   `director_id` int(11) NOT NULL COMMENT 'el userID del usuario que creo este concurso',
-  `rerun_id` int(11) NOT NULL COMMENT 'Este campo es para las repeticiones de algún concurso',
+  `rerun_id` int(11) NOT NULL COMMENT 'Este campo es para las repeticiones de algún concurso, Contiene el id del concurso original.',
```
lhchavez: can you do the `ALTER TABLE` statement for this in `64_ghost_mode_contest.sql` and leave this file untouched? I should add a lint validator that older scripts are not modified to prevent this.

frontend/server/controllers/ContestController.php
```diff
@@ -2321,6 +2389,16 @@ public static function apiRuns(Request $r) {
         return $response;
     }
 
+    public static function forbiddenInVirtual(Request $r) {
+        if (isset($r['virtual']) and $r['virtual'] == true) {
+            throw new ForbiddenAccessException();
+        }
+    }
+
+    public static function isVirtual(Request $r) {
```
lhchavez: instead of repurposing `/api/contest/create` with an additional parameter to distinguish between virtual and real contest, how about creating a brand new API? `/api/contest/createVirtual`? that way you don't need this.

frontend/server/controllers/ContestController.php
```diff
@@ -136,6 +136,8 @@ function (Request $r) use ($page, $page_size, $active_contests, $recommended, $q
     public static function apiAdminList(Request $r) {
         self::authenticateRequest($r);
 
+        self::forbiddenInVirtual($r);
```
lhchavez: Why do you need all these restrictions? I think it's better if you remove all of them if possible to make maintenance easier. Given that you are creating a brand new ACL for both the (cloned) problemset and the contest, there should be no messing around with the original contest.

Note that you can currently bypass all of them by adding `?virtual=false` to the request.

frontend/server/controllers/ContestController.php
```diff
@@ -936,6 +947,41 @@ public static function apiCreate(Request $r) {
...
+
+        $r['public'] = 0;
+        $r['title'] = $real_contest->title;
+        $r['description'] = $real_contest->description;
+        $r['problems'] = ProblemsetProblemsDAO::getProblems($real_contest->problemset_id);
+        $r['start_time'] = time();
+        $r['finish_time'] = time() + strtotime($real_contest->finish_time) - strtotime($real_contest->start_time);
+        $r['window_length'] = $real_contest->window_length;
+        $r['rerun_id'] = $real_contest->contest_id;
+        $r['alias'] = null;
```
lhchavez: can you generate an alias? otherwise how can the user invite their friends to the new contest?

frontend/server/controllers/ContestController.php
```diff
@@ -906,8 +916,9 @@ public static function apiCreate(Request $r) {
                 foreach ($r['problems'] as $problem) {
                     $problemset_problem = new ProblemsetProblems([
                                 'problemset_id' => $problemset->problemset_id,
-                                'problem_id' => $problem['id'],
-                                'points' => $problem['points']
+                                'problem_id' => $problem['id'] ?? $problem['problem_id'],
```
lhchavez: welp, I just realized that this is a security problem. if you create a contest and add problems one by one, we check if the user has permission to do so https://github.com/omegaup/omegaup/blob/6ce61ccea1ad58e8a76e0ac3c695dd73b003f3ab/frontend/server/controllers/ContestController.php#L1158-L1205

but if you somehow guess the problems' ids, this just lets you add them, even if they are banned!

so i think what we should do is remove L915-L926 altogether, and then in the new API you are going to create, make a function (say, `ProblemsetProblemsDAO::copyProblemset($new_problemset, $old_problemset)`) that does an [`INSERT ... SELECT`](https://dev.mysql.com/doc/refman/8.0/en/insert-select.html).

frontend/server/libs/Scoreboard.php
```diff
@@ -23,6 +23,7 @@ public function __construct(array $params) {
         ScoreboardParams::validateParameter('show_all_runs', $params, false /*is_required*/, false);
         ScoreboardParams::validateParameter('auth_token', $params, false /*is_required*/, null);
         ScoreboardParams::validateParameter('only_ac', $params, false /*is_required*/, false);
+        ScoreboardParams::validateParameter('virtual', $params, false, false);
```
lhchavez: why do you need to receive this as a parameter? Can't you infer if it's virtual from the fact that a contest has a non-`null` `rerun_id`?

and also, I thought the whole idea was to make no modifications to the scoreboard, given that you can merge them in the client side.

frontend/server/libs/dao/Contests.dao.php
```diff
@@ -156,6 +156,29 @@ class ContestsDAO extends ContestsDAOBase {
...
 
+    final public static function getVirtualByContestAndUser(Contests $real_contest, $user) {
+        $sql = 'SELECT
+                    *
+                FROM
+                    Contests
+                INNER JOIN
+                    ACLs ON ACLs.acl_id = Contests.acl_id
+                WHERE
+                    ACLs.owner_id = ? AND Contests.alias IS NULL AND Contests.rerun_id = ?
```
lhchavez: this is also related to the comment that the virtual contests should also have aliases: this way you cannot replay a contest more than once. if you have an alias, you can.

frontend/server/libs/dao/Runs.dao.php
```diff
@@ -587,6 +587,19 @@ class RunsDAO extends RunsDAOBase {
         return (Time::get() + $submission_gap);
     }
 
+    /**
+     *  This function is used to change the time in runs param relative
+     *  to new start time
+     *  param $runs array of Runs
+     */
+    public static function calibrateRuns($runs, $contest_start, $contest_new_start) {
```
lhchavez: You should be able to do this in the client side.

frontend/server/controllers/ContestController.php
SpaceWhite: @lhchavez  I thought the whole idea of virtual contest is to let user to only simulate the real contest. Not to edit the virtual contest he/she has created (like adding problem, adding admin, adding contestant, etc should be disabled). That is the reason why I think we should add restriction to the contest.

And about the bypass, if the user provide `virtual=false` in the query than it will visit the real contest, not the virtual contest.

frontend/server/controllers/ContestController.php
SpaceWhite: @lhchavez  So there will be multiple same aliases in the table?

frontend/server/controllers/ContestController.php
lhchavez: hmm you got a point with modifying the problemset, but adding admins/contestants should be allowed. otherwise you can't invite your friends to your virtual contest.

frontend/server/controllers/ContestController.php
lhchavez: aliases must be unique (guaranteed by the `UNIQUE` restriction in mysql). you can either add a counter or a random suffix.

SpaceWhite: In that case I'll close this PR. I'll make a new PR today @lhchavez

PR #2008, opened 2018-05-16
[GHOST MODE] Server side development

SpaceWhite:
# Description
Ghost mode for server side.

Changes:
1. Create virtual contest API
When virtual contest is created `rerun_id` value will be set to original contest contest id
2. API details will return original contest alias
This will be used to determine which scoreboard to merge in the client side. So in the later development, omegaup client side will call 2 scoreboardevents API and merge it
3. API access restriction
API like add problem and remove problem will be restricted

Additional development:
[✔️] Add `rerun_id` in `apimylist` JSON (So the JS can distinguish ghost mode contest and real contest)
[✔️] Restrict owner to change  virtual contest public contest (Restrict `apiUpdate`?)
[✔️] User can input contest alias with the `alias-zz`. Fix `generateAlias`
[✔️] Add `virtual` flag in `apiIntroDetails` aka `apiPublicDetails` so that the intro page know it is ghost mode.
[✔️] Add unittest
[✔️] Ghost mode in `apiList`
[] Users are not able to create virtual contest if he/she has not joined original contest

@lhchavez Why should we let user to add admin? And should we restrict user to `apiUpdate` (so that user can't change title, time, contest alias etc)?


frontend/server/controllers/ContestController.php
```diff
@@ -1066,6 +1126,19 @@ private static function validateCreateOrUpdate(Request $r, $is_update = false) {
...
+    /**
+     * This function is used to restrict API in virtual contest
+     *
+     * @param Request $r
+     * @return void
+     * @throws ForbiddenAccessException
+     */
+    private static function forbiddenInVirtual(Contests $contest) {
+        if (ContestsDAO::isVirtual($contest)) {
+            throw new ForbiddenAccessException('forbiddenInGhostMode');
```
lhchavez: I feel a slight cognitive dissonance of this being referred to as "ghost mode" and "virtual" in the same method :S would it be possible to choose one name for it and use it consistently?

frontend/server/controllers/ContestController.php
```diff
@@ -828,6 +832,74 @@ public static function apiClone(Request $r) {
...
+        $r['penalty_calc_policy'] = $original_contest->penalty_calc_policy;
+        $r['show_scoreboard_after'] = $original_contest->show_scoreboard_after;
+        $r['languages'] = $original_contest->languages;
+        $r['auth_token'] = $auth_token;
+
+        $response = self::apiCreate($r);
+
+        try {
+            $virtual_contest = ContestsDAO::getByAlias($virtual_contest_alias);
+            ContestsDAO::updateContestToVirtual($virtual_contest, $original_contest);
```
lhchavez: I think it's going to be easier if you refactor `apiCreate` in a way in which both methods can call the same validation methods (as required) and insert the record into the database. that way you can inject the `rerun_id` from the beginning instead of having to update it later. that way you _also_ avoid having to perform the lookup in L889, thus reducing the number of database roundtrips!

frontend/server/controllers/ContestController.php
```diff
@@ -828,6 +832,74 @@ public static function apiClone(Request $r) {
...
+        }
+
+        $start_time = strtotime($original_contest->start_time);
+        $finish_time = strtotime($original_contest->finish_time);
+
+        if ($finish_time < Time::get()) {
+            throw new ForbiddenAccessException('originalContestHasNotEnded');
+        }
+
+        $virtual_contest_alias = ContestsDAO::generateAlias($r['alias']);
```
lhchavez: here you already have the `Contests` object, so you shouldn't need to pass in the alias to `generateAlias`, which in turn gets another copy of the `Contests` object. Instead, pass `$originalContest`.

frontend/server/controllers/ContestController.php
```diff
@@ -828,6 +832,74 @@ public static function apiClone(Request $r) {
...
 
+    public static function apiCreateVirtual(Request $r) {
+        if (OMEGAUP_LOCKDOWN) {
+            throw new ForbiddenAccessException('lockdown');
+        }
+        // Authenticate user
+        self::authenticateRequest($r);
+
+        try {
+            $original_contest = ContestsDAO::getByAlias($r['alias']);
```
lhchavez: nit: use `$camelCase`.

frontend/server/libs/dao/Contests.dao.php
```diff
@@ -762,6 +762,84 @@ public static function getNeedsInformation($problemset_id) {
...
+    public static function getLastVirtualContest($contest, $user = null) {
+        $sql = '
+            SELECT
+                *
+            FROM
+                Contests
+        ';
+        $params = [];
+        if (!is_null($user)) {
+            $sql .= '
```
lhchavez: given that aliases must be globally unique, you can drop the `owner_id` `INNER JOIN` since two users might want to create a virtual contest and the second one will fail.

Alternatively, you can use the current time as a way of avoiding collisions without needing to query the database. fewer roundtrips! :D

frontend/server/controllers/ContestController.php
```diff
@@ -828,6 +832,74 @@ public static function apiClone(Request $r) {
...
+
+        $virtual_contest_alias = ContestsDAO::generateAlias($r['alias']);
+
+        $contest_length = strtotime($original_contest->finish_time) - strtotime($original_contest->start_time);
+
+        $auth_token = isset($r['auth_token']) ? $r['auth_token'] : null;
+
+        //Initialiaze request
+        $r = new Request();
+        $r['title'] = $original_contest->title;
```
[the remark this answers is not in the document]
SpaceWhite: I'm thinking about appending the word `(virtual)` or emoji in the client side. Javascript can distinguish the original and the virtual one by checking `rerun_id` value

frontend/server/controllers/ContestController.php
SpaceWhite: If I'm not mistaken you want to change the code from
```
$contest->rerun_id=0
``` 
to 
```
$contest->rerun_id = is_null($r['rerun_id']) ? 0 : $r['rerun_id']
``` 

By doing this we can reduce the number of database access, but it may cause some issues. User can call `apiCreate` with `rerun_id` parameter inside. What should I do to mitigate this? @lhchavez

SpaceWhite:
PTAL @lhchavez

frontend/server/controllers/ContestController.php
```diff
@@ -902,18 +973,6 @@ public static function apiCreate(Request $r) {
             // Save the contest object with data sent by user to the database
             ContestsDAO::save($contest);
 
-            if (!is_null($r['problems'])) {
```
heduenas: For my own understanding, why are you deleting this?

frontend/server/libs/dao/Contests.dao.php
```diff
@@ -762,6 +762,72 @@ public static function getNeedsInformation($problemset_id) {
...
+     * @return string of unique virtual contest alias
+     */
+    public static function generateAlias(Contests $contest) {
+        $last_virtual_contest = self::getLastVirtualContest($contest);
+        $alias = $contest->alias;
+        if (is_null($last_virtual_contest)) {
+            $alias = $alias . '-virtual-1';
+        } else {
+            $str = explode('-', $last_virtual_contest['alias']);
+            $alias = $alias .'-virtual-' . strval(intval(end($str)) + 1);
```
heduenas: Can you add a comment about this and the previous line?

frontend/server/libs/dao/Contests.dao.php
```diff
@@ -762,6 +762,72 @@ public static function getNeedsInformation($problemset_id) {
...
+        return $alias;
+    }
+
+    /**
+     * Update contest table to virtual contest
+     *
+     * @param String alias, Number original contest
+     * @return void
+     */
+    public static function updateContestToVirtual($virtual_contest, $original_contest) {
```
heduenas: What's the use case for this method?

frontend/tests/factories/ContestsFactory.php
```diff
@@ -301,6 +301,12 @@ public static function forcePublic($contestData, $last_updated = null) {
         ContestsDAO::save($contest);
     }
 
+    public static function forceFinish($contestData) {
```
heduenas: What's the use case for this function?

frontend/server/libs/dao/Contests.dao.php
SpaceWhite: Welp, This code is from the previous PR. I forgot to delete it.

frontend/server/controllers/ContestController.php
SpaceWhite: Because in `apiCreate` we only do create the contest. Adding problem process will be after we created the contest. Hence this line is not needed.

frontend/server/libs/dao/Contests.dao.php
SpaceWhite: But I think we should update the `rerun_id` after `apiCreate` because user can forge `rerun_id`. I need @lhchavez opinion about this.

lhchavez: no need. as long as you don't populate the contest's `rerun_id` in `apiCreate` and always overwrite it in `apiCreateVirtual` (which you are already doing), the user cannot forge it.

frontend/server/libs/dao/Contests.dao.php
```diff
@@ -762,6 +763,79 @@ public static function getNeedsInformation($problemset_id) {
...
 
+    /**
+     * Generate alias of virtual contest / ghost mode
+     *
+     * @param Contests $contest
+     * @return string of unique virtual contest alias
+     */
+    public static function generateAlias(Contests $contest) {
+        // Get the last virtual contest to get virtual contest number
+        $last_virtual_contest = self::getLastVirtualContest($contest);
```
lhchavez: oh i just realized this is racy :( if two users try to call this at roughly the same time, they will both see there is no previous virtual contest and they will both try to create `-virtual-1` :(

frontend/server/libs/dao/Contests.dao.php
```diff
@@ -762,6 +763,79 @@ public static function getNeedsInformation($problemset_id) {
...
+        // Virtual contest alias format (alias-virtual-n)
+        $alias = $contest->alias;
+
+        if (is_null($last_virtual_contest)) {
+            // No virtual contest? Assign alias as (alias-virtual-1)
+            $alias = $alias . '-virtual-1';
+        } else {
+            // ANother virtual contest exist. Take the last n of virtual contest
+            $str = explode('-', $last_virtual_contest['alias']);
```
lhchavez: what if a malicious user creates `mycoolcontest-virtual-notanint` and you try to create a virtual contest for `mycoolcontest`?

frontend/server/controllers/ContestController.php
lhchavez: what i'd suggest is to avoid calling `self::apiCreate`. instead, create the `Contests` object here:

```php
new Contests([
    // ...
    'rerun_id' => $originalContest->contest_id,
    // ...
]);
```

that way it's unforgeable.

frontend/server/controllers/ContestController.php
lhchavez: this was added for testing purposes, but we realized a user can bypass the restrictions to adding problems in a problemset (e.g. disallowing to add banned problems) by guessing the ids for the problems and using this feature D:

frontend/server/controllers/ContestController.php
```diff
@@ -1066,6 +1140,19 @@ private static function validateCreateOrUpdate(Request $r, $is_update = false) {
...
+    /**
+     * This function is used to restrict API in virtual contest
+     *
+     * @param Request $r
+     * @return void
+     * @throws ForbiddenAccessException
+     */
+    private static function forbiddenInVirtual(Contests $contest) {
+        if (ContestsDAO::isVirtual($contest)) {
+            throw new ForbiddenAccessException('forbiddenInVirtualContest');
```
lhchavez: oh one more thing, any string that appears in an exception should have its own translation entry.

frontend/server/libs/dao/Contests.dao.php
SpaceWhite: How about we prevent user from using `virtual` keyword as alias when user try to create a new contest? @lhchavez

frontend/server/libs/dao/Contests.dao.php
SpaceWhite: How about if the user directly trigger the `apiCreate` with `rerun_id`? @lhchavez

frontend/server/libs/dao/Contests.dao.php
SpaceWhite: Hmm.. If that is the case, is this solvable by using transaction?

frontend/server/libs/dao/Contests.dao.php
lhchavez: I'd prefer to avoid calling `apiCreate` at all costs. you _can_ create a common function that both in turn call.

frontend/server/libs/dao/Contests.dao.php
lhchavez: i think a transaction won't help you because you need to do it in a single database roundtrip.

frontend/server/libs/dao/Contests.dao.php
lhchavez: that makes things a bit brittle :(

frontend/server/controllers/ContestController.php
```diff
@@ -828,6 +831,122 @@ public static function apiClone(Request $r) {
...
+        $contest->penalty_calc_policy = $original_contest->penalty_calc_policy;
+        $contest->show_scoreboard_after = true;
+        $contest->languages = $original_contest->languages;
+        $contest->rerun_id = $original_contest->contest_id;
+
+        $acl = new ACLs();
+        $acl->owner_id = $r['current_user_id'];
+
+        $problemset = new Problemsets([
+            'needs_basic_information' => $original_contest_problemset->needs_basic_information,
```
lhchavez: can you make both of these `false`?

frontend/server/controllers/ContestController.php
```diff
@@ -828,6 +831,122 @@ public static function apiClone(Request $r) {
...
+            'requests_user_information' => $original_contest_problemset->requests_user_information
+        ]);
+
+        self::createContest($acl, $problemset, $contest);
+
+        try {
+            // Copy problemset problems from original contest to virtual contest
+            ProblemsetProblemsDAO::copyProblemset($contest->problemset_id, $original_contest->problemset_id);
```
lhchavez: This still needs to be run within the transaction, so can you pass in the `$originalContest->problemset_id` to `self::createContest()` (where the last parameter is a `$original_problemset_id = null`). So you can do everything as atomically as possible.

frontend/server/controllers/ContestController.php
```diff
@@ -828,6 +831,122 @@ public static function apiClone(Request $r) {
...
+        } catch (DuplicatedEntryInDatabaseException $e) {
+            throw $e;
+        } catch (exception $e) {
+            throw new InvalidDatabaseOperationException($e);
+        }
+
+        return ['status' => 'ok', 'alias' => $contest->alias];
+    }
+
+    public static function createContest(ACLs $acl, Problemsets $problemset, Contests $contest) {
```
lhchavez: nit: this can be private.

frontend/server/controllers/ContestController.php
```diff
@@ -855,7 +974,7 @@ public static function apiCreate(Request $r) {
         $contest->start_time = gmdate('Y-m-d H:i:s', $r['start_time']);
         $contest->finish_time = gmdate('Y-m-d H:i:s', $r['finish_time']);
         $contest->window_length = $r['window_length'] === 'NULL' ? null : $r['window_length'];
-        $contest->rerun_id = 0; // NYI
+        $contest->rerun_id = is_null($r['rerun_id']) ? 0 : $r['rerun_id'];
```
lhchavez: nit: this should be 0.

frontend/server/controllers/ContestController.php
```diff
@@ -1637,6 +1730,7 @@ public static function apiScoreboardEvents(Request $r) {
         // Push scoreboard data in response
         $response = [];
         $response['events'] = $scoreboard->events();
+        $response['original_alias'] = $r['contest']->contest_id;
```
lhchavez: nit:

```php
return [
  'events' => $scoreboard->events(),
  'original_alias' => $r['contest']->alias,
];
```

frontend/server/libs/dao/Contests.dao.php
```diff
@@ -762,6 +763,31 @@ public static function getNeedsInformation($problemset_id) {
...
+    /**
+     * Generate alias of virtual contest / ghost mode
+     *
+     * @param Contests $contest
+     * @param Users $user
+     * @return string of unique virtual contest alias
+     */
+    public static function generateAlias(Contests $contest) {
+        // Virtual contest alias format (alias-virtual-random)
+        $alias = $contest->alias;
```
lhchavez: nit:

```php
return $alias . '-virtual-' . SecurityTools::randomString(3);
```

note that you might need to truncate the `$alias` if it is longer than 32 - 12 = 20 characters.

frontend/server/controllers/ContestController.php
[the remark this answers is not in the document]
heduenas: I'd prefer to use a`is_virtual` flag and append the word `(virtual)` in javascript.

frontend/server/controllers/ContestController.php
heduenas: +1

frontend/server/libs/dao/Contests.dao.php
```diff
@@ -762,6 +763,34 @@ public static function getNeedsInformation($problemset_id) {
...
+
+        // Truncate alias
+        if (strlen($alias) > 20) {
```
lhchavez: [`substr`](http://php.net/substr) already handles the case where the length is greater than the string's length, so you can just do

```php
        return substr($alias, 0, 20) . '-virtual-' .
            SecurityTools::randomString(3);
```

frontend/server/controllers/ContestController.php
```diff
@@ -828,6 +831,113 @@ public static function apiClone(Request $r) {
...
+    public static function apiCreateVirtual(Request $r) {
+        if (OMEGAUP_LOCKDOWN) {
+            throw new ForbiddenAccessException('lockdown');
+        }
+        // Authenticate user
+        self::authenticateRequest($r);
+
+        try {
+            $original_contest = ContestsDAO::getByAlias($r['alias']);
+            $original_contest_problemset = ProblemsetsDAO::getByPK($original_contest->problemset_id);
```
lhchavez: nit: unneeded

frontend/server/controllers/ContestController.php
```diff
@@ -828,6 +831,113 @@ public static function apiClone(Request $r) {
...
+
+        $acl = new ACLs();
+        $acl->owner_id = $r['current_user_id'];
+
+        $problemset = new Problemsets([
+            'needs_basic_information' => false,
+            'requests_user_information' => 'no',
+        ]);
+
+        self::createContest($acl, $problemset, $contest, $original_contest, $original_contest_problemset);
```
lhchavez: here you just need `$original_contest->problemset_id`, since there's no need to get the whole `Problemsets` object if you just need the ID. That way you avoid one database roundtrip :D

frontend/server/controllers/ContestController.php
```diff
@@ -828,6 +831,113 @@ public static function apiClone(Request $r) {
...
+        $contest->partial_score = $original_contest->partial_score;
+        $contest->feedback = $original_contest->feedback;
+        $contest->penalty = $original_contest->penalty;
+        $contest->penalty_type = $original_contest->penalty_type;
+        $contest->penalty_calc_policy = $original_contest->penalty_calc_policy;
+        $contest->show_scoreboard_after = true;
+        $contest->languages = $original_contest->languages;
+        $contest->rerun_id = $original_contest->contest_id;
+
+        $acl = new ACLs();
```
lhchavez: nit: you can also move the creation of the `ACLs` object into `self::createContest`

frontend/server/controllers/ContestController.php
lhchavez: this is still pending.

btw, going forward, you can add a reaction or reply to a comment to track what things are still missing.

frontend/server/controllers/ContestController.php
```diff
@@ -828,6 +831,111 @@ public static function apiClone(Request $r) {
...
+            $originalContest = ContestsDAO::getByAlias($r['alias']);
+        } catch (Exception $e) {
+            throw new InvalidDatabaseOperationException($e);
+        }
+
+        if (is_null($originalContest)) {
+            throw new NotFoundException('contestNotFound');
+        }
+
+        $start_time = strtotime($originalContest->start_time);
```
lhchavez: nit: camelCase all across the board.

frontend/server/controllers/ContestController.php
```diff
@@ -828,6 +831,111 @@ public static function apiClone(Request $r) {
...
+            'needs_basic_information' => false,
+            'requests_user_information' => 'no',
+        ]);
+
+        self::createContest($r, $problemset, $contest, $originalContest->problemset_id);
+
+        return ['status' => 'ok', 'alias' => $contest->alias];
+    }
+
+    private static function createContest(Request $r, Problemsets $problemset, Contests $contest, $original_problemset = null) {
```
lhchavez: nit: `$original_problemset_id`

frontend/server/controllers/ContestController.php
lhchavez: Done

frontend/server/libs/dao/Contests.dao.php
lhchavez: Done

frontend/tests/controllers/VirtualContestTest.php
```diff
@@ -0,0 +1,188 @@
...
+        $login = self::login($contestant);
+        $r = new Request([
+            'alias' => $contestData['request']['alias'],
+            'auth_token' => $login->auth_token
+        ]);
+
+        $response = ContestController::apiCreateVirtual($r);
+
+        // Get generated virtual contest alias
+        $virtual_contest_alias = $response['alias'];
```
lhchavez: same here, camelCase across the board.

frontend/server/controllers/ContestController.php
SpaceWhite: Why should we change this to camel case? I think we use snake case in [Code 1](https://github.com/omegaup/omegaup/blob/43b34399ce8f40a7864fe089dfddcb13c9f12aca/frontend/server/controllers/ContestController.php#L37), [Code 2](https://github.com/omegaup/omegaup/blob/43b34399ce8f40a7864fe089dfddcb13c9f12aca/frontend/server/controllers/ContestController.php#L63), etc @lhchavez

frontend/tests/controllers/VirtualContestTest.php
SpaceWhite: Same as here why should we change to camel case? @lhchavez

frontend/server/controllers/ContestController.php
lhchavez: pretty please?

in all seriousness, the consistency of PHP is a mess (both in our codebase and elsewhere). We're going to try to follow _a_ naming standard, and the [Zend](https://framework.zend.com/manual/1.12/en/coding-standard.naming-conventions.html#coding-standard.naming-conventions.variables) one seems to be the most popular, and they use camelCase.

lhchavez: I'm also going to add this to the linter in the near future to enforce a bit of consistency at least in our codebase.

frontend/server/controllers/ContestController.php
```diff
@@ -664,6 +666,10 @@ public static function apiDetails(Request $r) {
                     $r['current_identity_id'],
                     $r['contest']->problemset_id
                 );
+            } catch (ForbiddenAccessException $e) {
+                if (!ContestsDAO::hasFinished($r['contest'])) {
```
lhchavez: Why is this needed? With this, IIUC anybody can view the details of a contest that has already finished, and we don't want that D:

SpaceWhite: Oops I used the wrong API for the client side. I should have called `apiPublicDetails` rather than `apiDetails`.

SpaceWhite:
You’re welcome. 

@lhchavez Actually “Users are not able to create virtual contest if he/she has not joined original contest” issue was because I’m using apiDetails as the data for the intro page.

Issue #2009, opened 2018-05-19
Smarty template vuefication


SpaceWhite:
This issue is used as master issue to track all vue migration progress
List of smarty template that needs to be converted to `.vue`

### Task 1 [✔️]
Vuefication problem list. These files will take effects : `problems.tpl`, `problem.list.tpl`, `pager_bar.tpl`
[Issue link](https://github.com/omegaup/omegaup/issues/1999) - Assign to @CarlosCordova18 (Completed)

### Task 2 [✔️]
Vuefication arena index. These files will take effects :  `arena.contest_list.tpl` [PR merged](https://github.com/omegaup/omegaup/pull/2127), `arena.index.tpl` [PR merged](https://github.com/omegaup/omegaup/pull/3523)

### Task 3
Vuefication arena competition. These files will take effects : `arena.contest.tpl`, `arena.contest.course.tpl`, `arena.contest.admin.tpl`, `arena.contest.practice.tpl`, `arena.contest.practice.tpl`, `arena.contest.interview.tpl`,
`arena.clarification_list.tpl`, `arena.problem.tpl`, 
`arena.runs.tpl`  [PR merged](https://github.com/omegaup/omegaup/pull/3725)

### Task 4 [✔️]
Vuefication contest intro. Only this fill will take effects : `arena.contest.intro.tpl`  [PR merged](https://github.com/omegaup/omegaup/pull/4000)

### Task 5
Vuefication html navbar and header. These files will take effects : `head.tpl`, `common.navbar.tpl` [PR Merged](https://github.com/omegaup/omegaup/pull/2905) , `common.navbar.notifications.tpl` [PR Merged](https://github.com/omegaup/omegaup/pull/2909)

### Task 7 [✔️]
Vuefication contest related tpl. These files will take effects

- [X] `contest.edit.tpl` [PR merged](https://github.com/omegaup/omegaup/pull/2127)
- [X] `contest.new.form.tpl` [PR merged](https://github.com/omegaup/omegaup/issues/3925)
- [X] `contest.report.tpl` [PR merged](https://github.com/omegaup/omegaup/pull/2295)

### Task 8 [✔️]
Vuefication small graph on omegaup index page. This file will take effects : `contest.stats.tpl`
[PR link](https://github.com/omegaup/omegaup/pull/2087) - Assigned to @sarvesh211999 

### Task 9 [✔️]
[...]

PR #2062, opened 2018-05-27
[GHOST MODE] Client side development

SpaceWhite:
## Description
Ghost mode client side development
Changes : 
1. Create a new page `http://domain.com/arena/contest_alias/virtual/` for user to start virtual contest
screenshot-35 227 25 77-2018-05-23-06-41-12
2. Add links to `/arena` page so that users are able to create virtual contest
screenshot-35 227 25 77-2018-05-23-06-42-05
3. Restict some edit contest feature in virtual contests
screenshot-2018-5-27 contestedit omegaup
4. Simulate scoreboard
screenshot-2018-6-2 omegaup

## Additional development
[✔️] Clean virtual contest `/admin`, `/stats`, `/print`, `/activity` column in `contest/mylist` table
[] Virtual contest start date limit ? (New PR)
[] Append virtual contest keyword, so users can distinguish between the original and the virtual contest
[] Convert `contest.edit.tpl` to Vue? (New PR)


frontend/www/js/omegaup/components/arena/Virtual.vue
```diff
@@ -0,0 +1,94 @@
...
+    }
+  },
+  created: function() {
+    var self = this;
+
+    API.Contest.publicDetails({contest_alias: this.contest_alias})
+        .then(function(response) {
+          var detail = response;
+
+          // convert time to H:i:s
```
lhchavez: This looks a lot like https://github.com/omegaup/omegaup/blob/fb251f8d3b8d1514ac345cf4ebfa58b129c05cff/frontend/www/js/omegaup/arena/arena.js#L13-L35 . Do you mind moving that function over to [`frontend/www/js/omegaup/ui.js`](https://github.com/omegaup/omegaup/blob/master/frontend/www/js/omegaup/ui.js) so that you can use it from both places?

frontend/www/js/omegaup/components/arena/Virtual.vue
```diff
@@ -0,0 +1,94 @@
...
+<script>
+import {API, T, UI, OmegaUp} from '../../omegaup.js';
+import {Arena} from '../../arena/arena.js';
+import DateTimePicker from '../DateTimePicker.vue';
+
+export default {
+  props: {},
+  data: function() {
+    return {
+      T: T, UI: UI, detail: {}, start_time: new Date(),
```
lhchavez: nit: camelCase across the board.

frontend/www/js/omegaup/components/arena/Virtual.vue
```diff
@@ -0,0 +1,94 @@
...
+          contest_alias: /\/arena\/([^\/]+)\/virtual/.exec(
+              window.location.pathname)[1]
+    }
+  },
+  created: function() {
+    var self = this;
+
+    API.Contest.publicDetails({contest_alias: this.contest_alias})
```
lhchavez: The convention is that .vue files should not invoke API calls directly, but let the JS handle that for them since .vue files are the presentation layer only. Can you move this to `frontend/www/js/omegaup/arena/virtual.js`? And pass in the details into a `prop` called `contest`.

frontend/www/js/omegaup/components/arena/Virtual.vue
```diff
@@ -0,0 +1,94 @@
...
+          var detail = response;
+
+          // convert time to H:i:s
+          let delta = detail.finish_time - detail.start_time;
+          let deltaHour = Math.floor(delta / 3600000);
+          let deltaMinute = Math.floor(delta % 3600000 / 60000);
+          let deltaSecond = delta % 3600000 % 60000;
+
+          // String Format
+          detail.contestDurationString =
```
lhchavez: these three properties should be [Vue computed properties](https://vuejs.org/v2/guide/computed.html).

frontend/www/js/omegaup/components/arena/Virtual.vue
```diff
@@ -0,0 +1,94 @@
...
+              {window_length: Math.floor(detail.submissions_gap / 60)});
+
+          self.detail = detail;
+          self.start_time = new Date();
+        })
+        .fail(UI.apiError);
+  },
+  methods: {
+    onSubmit: function() {
+      API.Contest.createVirtual({
```
lhchavez: Same here, this should just emit an event, so that the JS file can call the API.

frontend/www/js/omegaup/components/arena/Virtual.vue
```diff
@@ -0,0 +1,94 @@
+<template>
+  <div class="contest panel">
+    <div class="panel-body">
+      <div class="text-center">
+        <h2>{{detail.title + ' - ' +
```
lhchavez: nit:

```
{{ UI.formatString(T.virtualTitle, {title: contest.title }) }}
```

Where `virtualTitle` should be `%(title) — Virtual contest`. We should try to avoid concatenating strings that are supposed to be translated in the presentation layer, since it makes it harder to correctly i18n-ize some things. For instance, if we ever want to support RTL languages (like hebrew or arabic), it would make it harder if we hardcode the order in which things are concatenated. Also note that the character that separates the "Virtual contest" part should be an [em dash](http://www.thepunctuationguide.com/em-dash.html) instead of a hyphen.

frontend/www/js/contest.edit.js
```diff
@@ -87,6 +88,15 @@ omegaup.OmegaUp.on('ready', function() {
             contest.contestant_must_register == '0') {
           $('#requests').hide();
         }
+
```
lhchavez: Is there a way in which you can avoid adding the columns in the first place instead of removing them after-the-fact?

frontend/templates/arena.contest_list.tpl
```diff
@@ -10,6 +10,7 @@
 				<th class="time col-md-2" data-bind="visible: showTimes">{#wordsEndTime#}</th>
 				<th class="col-md-2" data-bind="visible: showTimes">{#wordsDuration#}</th>
 				<th class="col-md-2" data-bind="visible: showPractice" ></th>
+                <th class="col-md-2" data-bind="visible: showVirtual" ></th>
```
lhchavez: nit: use tabs here to match the surrounding indentation.

frontend/www/js/omegaup/components/arena/Virtual.vue
```diff
@@ -0,0 +1,72 @@
...
+
+export default {
+  props: {detail: Object},
+  data: function() {
+    return { T: T, UI: UI, startTime: new Date(), }
+  },
+  computed: {
+    contestDurationString: function() {
+      let detail = this.detail;
+      let deltaTime = UI.formatDelta(detail.finish_time - detail.start_time);
```
lhchavez: why not just `return UI.formatDelta(detail.finish_time - detail.start_time)`? it's better if we use the same format everywhere.

frontend/templates/en.lang
```diff
@@ -964,6 +968,7 @@ wordsRuntime = "Runtime"
 wordsSaveChanges = "Save changes"
 wordsScoreboard = "Scoreboard"
 wordsSearch = "Search"
+wordsSecond = "second"
```
lhchavez: You should be able to remove this.

frontend/templates/en.lang
```diff
@@ -990,5 +995,6 @@ wordsUser = "User"
 wordsValidator = "Validator"
 wordsVerdict = "Verdict"
 wordsVeryEasy = "VeryEasy"
+wordsVirtual = "Virtual contest"
```
lhchavez: This is redundant with `virtualContest`. You should be able to remove this and only use the other one.

(sorry, all the `wordsXxx` are terrible names :cry: )

frontend/www/js/contest.edit.js
SpaceWhite: We can do this by calling `publicdetail` API in the `www/contestedit.php` and check if the contest is virtual contest or not. And by using that value we hide the UI in `tpl` file.
 But in the future we need to remove this code and convert it to vue. Shouldn't this be okay as temporary fix? I know there will be UI glitch when javascript tries to hide the elements. LMK what do you think @lhchavez :)

lhchavez:
looks good! are you planning to tackle the additional development on a future change? If so, we might need to hide some of the functionality in production. you can achieve that with a change similar to https://github.com/omegaup/omegaup/commit/240300fccaf54a5b49611613a9537c131edf30d7 (but in reverse).

You can do that in a separate change, and we'll land that one first to avoid messing with this one too much (except to maybe just hide the virtual contest link under the experiment flag).

SpaceWhite:
@lhchavez I will make another commit for scoreboard simulation and clean up UI around today or tomorrow. But yes I'll make add experiment constant. PTAL

frontend/www/js/contest.edit.js
lhchavez: if there are plans to fix it, this works for me as-is.

SpaceWhite: I'll do this in the future :) #2009

frontend/templates/arena.contest_list.tpl
```diff
@@ -10,6 +10,9 @@
 				<th class="time col-md-2" data-bind="visible: showTimes">{#wordsEndTime#}</th>
 				<th class="col-md-2" data-bind="visible: showTimes">{#wordsDuration#}</th>
 				<th class="col-md-2" data-bind="visible: showPractice" ></th>
+			{if !empty($ENABLED_EXPERIMENTS) && in_array('virtual', $ENABLED_EXPERIMENTS)}
```
lhchavez: nit: align the `{if` to the beginning of the line.

frontend/www/js/omegaup/arena/arena.js
```diff
@@ -584,7 +564,28 @@ export class Arena {
 
   refreshRanking() {
     let self = this;
-    if (self.options.contestAlias) {
+    if (self.options.originalContestAlias != null) {
+      let events = [];
```
lhchavez: nit: in JavaScript you should declare stuff only in the function in which you need it to prevent scope creep. And with `let`, you should declare it in the _block_ you need it. So it's considered a best practice in most languages (except C99) to declare stuff as late as possible.

So move this to L571:

```javascript
let events = response.events;
```

frontend/www/js/omegaup/arena/arena.js
```diff
@@ -598,7 +599,71 @@ export class Arena {
...
+    let delta = (new Date()).getTime() - self.startTime.getTime();
+    let rank = [];
+
+    let problemOrder = [];
+    let problems = [];
+    let initialScores = [];
+
+    let count = 0;
+    for (var key in self.problems) {
+      problemOrder[self.problems[key].alias] = count;
```
lhchavez: can't you use `problems.length` instead of `count`?

frontend/www/js/omegaup/arena/arena.js
```diff
@@ -598,7 +599,71 @@ export class Arena {
...
+      problems.push({order: count, alias: self.problems[key].alias});
+      count++;
+    }
+
+    data.forEach(function(env) {
+      if (env.delta <= delta) {
```
lhchavez: nit: use the [guard clause pattern](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html):

```javascript
if (env.delta > delta) return;

// rest
```javascript

File: frontend/www/js/omegaup/arena/arena.js [hunk 16]
```diff
@@ -598,7 +599,71 @@ export class Arena {
...
+        if (!rank.hasOwnProperty(env.username)) {
+          rank[env.username] = [];
```
lhchavez: it looks super weird to declara a JavaScript array and then add arbitrary properties to it. Can't you use an object literal instead?

```javascript
rank[env.username] = {
  country: env.country,
  // ...
};
```

frontend/www/js/omegaup/arena/virtual.js
```diff
@@ -0,0 +1,37 @@
...
+import virtual from '../components/arena/Virtual.vue';
+
+OmegaUp.on('ready', function() {
+  let contestAlias =
+      /\/arena\/([^\/]+)\/virtual/.exec(window.location.pathname)[1];
+  let detail;
+  API.Contest.publicDetails({contest_alias: contestAlias})
+      .then(function(response) {
+        let detail = response;
+        let virtual_ = new Vue({
```
lhchavez: Maybe `let arenaVirtual = new Vue({` to follow the namespacing convention.

frontend/www/js/omegaup/arena/virtual.js
```diff
@@ -0,0 +1,37 @@
...
+import Vue from 'vue';
+import virtual from '../components/arena/Virtual.vue';
+
+OmegaUp.on('ready', function() {
+  let contestAlias =
+      /\/arena\/([^\/]+)\/virtual/.exec(window.location.pathname)[1];
+  let detail;
+  API.Contest.publicDetails({contest_alias: contestAlias})
+      .then(function(response) {
+        let detail = response;
```
lhchavez: Why not make this `function(detail)` to avoid an unnecessary alias?

frontend/www/js/omegaup/arena/virtual.js
```diff
@@ -0,0 +1,37 @@
...
+            return createElement('virtual', {
+              props: {detail: detail},
+              on: {
+                submit: function(ev) {
+                  API.Contest.createVirtual({
+                               alias: contestAlias,
+                               start_time: ev.startTime.getTime() / 1000
+                             })
+                      .then(function(response) {
+                        let virtual_contest_alias = response.alias;
```
lhchavez: nit: camelCase: `virtualContestAlias`.

frontend/www/js/omegaup/components/arena/Virtual.vue
```diff
@@ -0,0 +1,67 @@
...
+  </div>
+</template>
+
+<script>
+import {API, T, UI, OmegaUp} from '../../omegaup.js';
+import {Arena} from '../../arena/arena.js';
+import DateTimePicker from '../DateTimePicker.vue';
+
+export default {
+  props: {detail: Object},
```
lhchavez: Can you split this `detail` into `finishTime`, `startTime`, `scoreboard`, and `submissionGap`? Each with better typing information than `Object`. That way you can replace the

```javascript
let detail = this.detail;
let foo = detail.foo;
... foo ...
```

with just

```javascript
... this.foo ...
```

frontend/www/js/omegaup/components/contest/ContestList.vue
```diff
@@ -61,16 +61,18 @@
           <td v-if="!isAdmin"></td>
           <td v-else-if="contest.public == '1'">{{ T.wordsYes }}</td>
           <td v-else="">{{ T.wordsNo }}</td>
-          <td v-if="contest.scoreboard_url &amp;&amp; isAdmin">
+          <td v-if="isAdmin">
```
lhchavez: how are you guaranteeing that `contest.scoreboard_url` is always set when `isAdmin` is true? If the guarantee is bidirectional, i'd prefer to check whether `contest.scoreboard_url` is non-empty so that the check matches what is being used.

frontend/www/js/omegaup/components/contest/ContestList.vue
```diff
@@ -80,7 +82,8 @@
           <td v-if="isAdmin">
             <a class="glyphicon glyphicon-dashboard"
                 v-bind:href="'/arena/' + contest.alias + '/admin/'"
-                v-bind:title="T.contestListSubmissions"></a>
+                v-bind:title="T.contestListSubmissions"
+                v-if="contest.rerun_id == 0"></a>
```
lhchavez: You should still be able to admin a virtual contest if you're an admin. For instance, I _need_ to be able to admin other virtual contests when they request assistance.

webpack.config.js
```diff
@@ -28,6 +28,7 @@ module.exports = {
     qualitynomination_details: './frontend/www/js/omegaup/qualitynomination/details.js',
     user_charts: './frontend/www/js/omegaup/user/charts.js',
     user_profile: './frontend/www/js/omegaup/user/profile.js',
+    virtual: './frontend/www/js/omegaup/arena/virtual.js',
```
lhchavez: nit: call this `arena_virtual` to keep the namespacing convention.

frontend/www/js/omegaup/arena/virtual.js
```diff
@@ -0,0 +1,37 @@
+import {API, UI, OmegaUp, T} from '../omegaup.js';
+import Vue from 'vue';
+import virtual from '../components/arena/Virtual.vue';
```
lhchavez: `arena_Virtual`

frontend/www/js/omegaup/arena/virtual.js
```diff
@@ -0,0 +1,37 @@
...
+
+OmegaUp.on('ready', function() {
+  let contestAlias =
+      /\/arena\/([^\/]+)\/virtual/.exec(window.location.pathname)[1];
+  let detail;
+  API.Contest.publicDetails({contest_alias: contestAlias})
+      .then(function(response) {
+        let detail = response;
+        let virtual_ = new Vue({
+          el: '#virtual',
```
lhchavez: `#arena-virtual`

frontend/www/js/omegaup/arena/virtual.js
```diff
@@ -0,0 +1,37 @@
...
+          render: function(createElement) {
+            return createElement('virtual', {
```
lhchavez: can you call this `omegaup-arena-virtual` to follow convention?

frontend/www/js/omegaup/components/contest/ContestList.vue
SpaceWhite: It is not guaranteed that `contest.scoreboard_url` is always set when `isAdmin` is true. And number of headers `<td>` is not the same as number of `<td>`.
So it will create UI like this:
screenshot-2018-6-2 contest omegaup 1
NB: See the `edit` button is below the `scoreboard` header

These changes create UI like this:
screenshot-2018-6-2 contest omegaup

frontend/server/libs/Scoreboard.php
```diff
@@ -714,11 +715,11 @@ private static function calculateEvents(
 
             $problem_data = &$identity_problems_score[$identity_id][$problem_id];
 
-            if ($problem_data['points'] >= $contest_score) {
+            if ($problem_data['points'] >= $contest_score and $params['unique']) {
```
lhchavez: Why is this `unique` needed? Can't we keep this code as-is?

frontend/www/js/omegaup/arena/arena.js
```diff
@@ -598,7 +598,69 @@ export class Arena {
...
+      };
+      rank[env.username].total = env.total;
+    });
+
+    let ranking = [];
+
+    for (var username in rank) ranking.push(rank[username]);
+
+    ranking.sort(function(rank1, rank2) {
+      return rank1.total.points < rank2.total.points;
```
lhchavez: note that in JavaScript, [`Array.sort`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) expects an integer <, =, or > 0 instead of a boolean (like in C++).

frontend/www/js/omegaup/components/contest/ContestList.vue
lhchavez: oh i see, you're splitting the condition between the `<td>` and the `<a>`. Got it, what you're doing is correct.

frontend/server/controllers/ContestController.php
```diff
@@ -1710,13 +1713,13 @@ public static function apiScoreboardEvents(Request $r) {
         self::validateDetails($r);
 
         $params = ScoreboardParams::fromContest($r['contest']);
-        $params['show_all_runs'] = Authorization::isContestAdmin($r['current_identity_id'], $r['contest']);
+        $params['show_all_runs'] = (Authorization::isContestAdmin($r['current_identity_id'], $r['contest']) and !ContestsDAO::isVirtual($r['contest']));
```
lhchavez: style nit: use `&&` instead of `and` here and everywhere else in the change. Also, if possible, break the line to make it more readable:

```php
$params['show_all_runs'] = (
    Authorization::isContestAdmin($r['current_identity_id'], $r['contest']) &&
    !ContestsDAO::isVirtual($r['contest'])
);
```

could we also have a way for site-admins to always see all runs? So if it's a site admin, `$params['show_all_runs']` should be true as well.

frontend/server/libs/Experiments.php
```diff
@@ -38,7 +38,9 @@ class Experiments {
     /**
      * An array with all the known experiments.
      */
+    const VIRTUAL = 'virtual';
```
lhchavez: nit:

```php
/**
 * Support for virtual contest (a.k.a. Ghost mode).
 */
const VIRTUAL = 'virtual';

/**
 * An array with all the known experiments.
 */
private static $kKnownExperiments = [
    self::VIRTUAL
];
```

frontend/www/js/omegaup/arena/arena.js
```diff
@@ -584,7 +564,27 @@ export class Arena {
 
   refreshRanking() {
     let self = this;
-    if (self.options.contestAlias) {
+    if (self.options.originalContestAlias != null) {
+      API.Contest.scoreboardEvents({contest_alias: self.options.contestAlias})
+          .then(function(response) {
+            let events = response.events;
+            for (var i in events) {
```
lhchavez: nit:

```javascript
for (let event of events) {
  event.username ...
  ...
}
```

frontend/www/js/omegaup/arena/arena.js
```diff
@@ -584,7 +564,27 @@ export class Arena {
...
+              events[i].username += '-(virtual)';
+              events[i].name += '-(virtual)';
+            }
+            API.Contest.scoreboardEvents(
+                           {contest_alias: self.options.originalContestAlias})
+                .then(function(response) {
+                  var original_events = response.events;
+                  events = events.concat(original_events);
```
lhchavez: nit: `events.push.apply(events, response.events);`

frontend/www/js/omegaup/arena/arena.js
```diff
@@ -584,7 +564,27 @@ export class Arena {
...
+                  var data = {events: events};
+                  self.virtualRankingChange(events);
+                  self.onRankingEvents(data);
```
lhchavez: nit: `self.onRankingEvents({events: events});` (or use `let` for `data`).

frontend/www/js/omegaup/arena/arena.js
```diff
@@ -598,7 +598,69 @@ export class Arena {
...
+  virtualRankingChange(data) {
+    let self = this;
+    let delta = (new Date()).getTime() - self.startTime.getTime();
+    let rank = [];
+
+    let problemOrder = [];
+    let problems = [];
+    let initialScores = [];
+
+    for (var key in self.problems) {
```
lhchavez: nit:

```javascript
for (let problem of self.problems) {
  problemOrder[problem.alias] ...
}
```

frontend/www/js/omegaup/arena/arena.js
```diff
@@ -598,7 +598,69 @@ export class Arena {
...
+        penalty: env.problem.penalty,
+        points: env.problem.points,
+        runs: problem.runs + 1
+      };
+      rank[env.username].total = env.total;
+    });
+
+    let ranking = [];
+
+    for (var username in rank) ranking.push(rank[username]);
```
lhchavez: nit:

```javascript
for (let user in rank) ranking.push(user);
```

frontend/www/js/omegaup/arena/arena.js
SpaceWhite: Welp, after some researches I just realized that `event` is not reserved word in javascript.:)

frontend/server/libs/Scoreboard.php
SpaceWhite: For example if user A made submissions like this
```
1. delta: 12, problem_alias: prob1, score: 50, penalty: 0
2. delta: 22, problem_alias: prob1, score: 50, penalty: 0
3. delta: 32, problem_alias: prob1, score: 80, penalty: 20
```
`scoreboardevents` API will only returns data like this
```
1. delta: 22, problem_alias: prob1, score: 50, penalty: 0
2. delta: 32, problem_alias: prob1, score: 80, penalty: 20
```

`scoreboardevents` API will ignore the first submission because the first verdict is the same as the second one. That's why I create a new `unique` parameter in the scoreboard.

lhchavez: i see. `unique` is probably not the best name since it does not convey all of this information. maybe `show_all_runs` would be better, but that's already being used to be sort of an "admin" flag. if we can't think of a better name, how about renaming the current `show_all_runs` for `admin` and using `show_all_runs` for this feature?

frontend/www/js/omegaup/arena/arena.js
SpaceWhite: `self.problems` type is `dictionary` so it can't be iterated with `for (let of )` command. But in ES2017 we can use this command https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries

lhchavez: derp, I missed [`Object.values`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values):

```javascript
for (let problem of Object.values(self.problems)) {
  problemOrder[problem.alias] ...
}
```

SpaceWhite:
PTAL @lhchavez

SpaceWhite:
:D

PR #2087, opened 2018-06-13
Contest stats

sarvesh211999:
Convert contest.stats.tpl to vue.

#2009


sarvesh211999:
I am unable to understand the error in selenium test Please help !!

frontend/www/js/omegaup/components/contest/Stats.vue
```diff
@@ -0,0 +1,13 @@
+<template>
+  <div class="post">
+    <div class="copy">
+      <h1>Estadísticas en vivo</h1>
+      <div class="total-runs">
+        <h2></h2>
+      </div>
+      <div class="verdict-chart"></div>
```
SpaceWhite: Instead of changing `id` to `class`. Can you change these `verdict-chart`, `distribution-chart`, `pending-runs-chart` into vue?

sarvesh211999: okay I will change

frontend/www/js/omegaup-graph.js
```diff
@@ -68,7 +68,7 @@ OmegaupGraph.prototype.pendingRuns = function(refreshRate, updateStatsFn) {
       type: 'spline',
       animation: Highcharts.svg,  // don't animate in old IE
       marginRight: 10,
-      renderTo: 'pending-runs-chart',
+      renderTo: $('.pending-runs-chart')[0],
```
lhchavez: nit: `document.querySelector('.pending-runs-chart')`. Same everywhere else you are using jQuery.

frontend/templates/contest.stats.tpl
```diff
@@ -1,16 +1,8 @@
...
-		<div id="verdict-chart"></div>
-		<div id="distribution-chart"></div>
-		<div id="pending-runs-chart"></div>
-	</div>
-</div>
-
-<script type="text/javascript" src="{version_hash src="/js/contest.stats.js"}"></script>
+<div id="contest-stats"></div>
+<script type="text/javascript" src="{version_hash src="/js/dist/contest_stats.js"}"></script>
+<!-- <script type="text/javascript" src="{version_hash src="/js/contest.stats.js"}"></script> -->
```
lhchavez: nit: Try to avoid including commented-out code.

frontend/www/js/omegaup/components/contest/Stats.vue
```diff
@@ -0,0 +1,13 @@
+<template>
+  <div class="post">
+    <div class="copy">
+      <h1>Estadísticas en vivo</h1>
```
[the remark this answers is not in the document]
sarvesh211999: I am not able to understand ??

frontend/www/js/omegaup-graph.js
sarvesh211999: I will try convertong into vue.

sarvesh211999:
Can I install another module i.e. vue-highcharts for generating??

frontend/www/js/omegaup/components/contest/Stats.vue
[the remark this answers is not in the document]
SpaceWhite: https://en.wikipedia.org/wiki/Internationalization_and_localization#Naming. Can you add this word translation into `en.lang`, `pt.lang` and etc?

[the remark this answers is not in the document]
sarvesh211999: oh okay

frontend/www/js/omegaup/components/contest/Stats.vue
sarvesh211999: I will have to install another module . vue-highcharts
So should I install it ??

SpaceWhite:
You can try to use `$(self.$el)` https://vuejs.org/v2/api/#Options-DOM. And make change similar to `DateTimePicker` https://github.com/omegaup/omegaup/blob/9c82418d00b6aa4b5908d4402648880f9cdb1911/frontend/www/js/omegaup/components/DateTimePicker.vue#L25 @sarvesh211999

sarvesh211999:
I was referring Charts.vue from 
https://github.com/omegaup/omegaup/blob/9c82418d00b6aa4b5908d4402648880f9cdb1911/frontend/www/js/omegaup/components/user/Charts.vue#L26-L27

Is it advisable or it was due to mistake . Because lint does not show error in this.

sarvesh211999:
As @SpaceWhite suggested using `this.$el` and @lhchavez suggested about using **Query Selector**
I made changes .
This is changed only for verdict-charts i.e. pie charts .
If it is ok I can change them for rest.

sarvesh211999:
Please help about the errors in TRAVIS

SpaceWhite:
IMHO the idea of separating charts into their components is better because it is more elegant than generating component with class name. But we will see what does @lhchavez think?

sarvesh211999:
So you are suggesting to create a different component for every chart ??
Right @SpaceWhite

frontend/www/js/omegaup/contest/stats.js
```diff
@@ -1,4 +1,19 @@
...
+import Vue from 'vue';
+import {OmegaUp} from '../omegaup.js';
+
+OmegaUp.on('ready', function() {
+  var verdict = null;
+  let Stats = new Vue({
+    el: '#contest-stats',
+    render: function(createElement) { return createElement('contestStats'); },
+    mounted: function() {
+      verdict = this.$el.querySelector('.copy').querySelectorAll('div')[0];
```
lhchavez: `querySelectorAll(...)[0]` is equivalent to `querySelector(...)` and is harder to reason about.

frontend/www/js/omegaup/contest/stats.js
```diff
@@ -30,7 +44,7 @@ omegaup.OmegaUp.on('ready', function() {
   }
 
   function drawCharts() {
-    $('#total-runs')
+    $($('.total-runs')[0])
```
lhchavez: try to avoid using jQuery as much as possible:

```javascript
document.querySelector('.total-runs').textContent = ...
```

frontend/www/js/omegaup/contest/stats.js
```diff
@@ -42,11 +56,11 @@ omegaup.OmegaUp.on('ready', function() {
...
     // Draw verdict counts pie chart
     window.run_counts_chart =
-        oGraph.verdictCounts('verdict-chart', contestAlias, stats);
+        oGraph.verdictCounts(verdict, contestAlias, stats);
 
     // Draw distribution of scores chart
-    window.distribution_chart =
-        oGraph.distributionChart('distribution-chart', contestAlias, stats);
+    window.distribution_chart = oGraph.distributionChart(
+        $('.distribution-chart')[0], contestAlias, stats);
```
lhchavez: `$(...)[0]` is equivalent to `document.querySelector(...)` and uses jQuery (which we want to avoid doing going forward).

lhchavez:
yes, we should try to componentize things as much as possible :D the thing with the class names vs. the IDs is just that we were previously using IDs for everything (yuck!) and we're slowly moving away from that (and jQuery altogether).

making actual components for things takes things a step further!

sarvesh211999:
OKay I will do it

sarvesh211999:
Added component for verdict-chart.
For now please ignore the rem. comments I will remove it once all components are added.
Please review is this ok. So I can proceed further .

frontend/www/js/omegaup/components/VerdictChart.vue
```diff
@@ -0,0 +1,27 @@
...
+export default {
+  props: {
+    stats: Object,
+    contestAlias: String,
+  },
+  mounted: function() { this.draw_pie_chart()},
+  methods: {
+    draw_pie_chart: function() {
+      if (this.stats != null) {
+        if (window.run_counts_chart != null) {
```
lhchavez: let's try to avoid setting global state. declare something in `data` and use that instead of `window`.

frontend/www/js/omegaup/components/VerdictChart.vue
```diff
@@ -0,0 +1,27 @@
...
+          return;
+        }
+
+        // Draw verdict counts pie chart
+        window.run_counts_chart =
+            oGraph.verdictCounts(this.$el, this.contestAlias, this.stats);
+      }
+      setTimeout(this.draw_pie_chart, 1000);
```
lhchavez: it's probably better for this to not drive the updating. it's going to be better if you have a [`watch`](https://vuejs.org/v2/guide/computed.html#Watchers) for every time `stats` changes.

frontend/www/js/omegaup/components/contest/Stats.vue
```diff
@@ -0,0 +1,27 @@
...
+      <div class="total-runs">
+        <h2></h2>
+      </div><!-- <div class="verdict-chart"></div> -->
+      <verdict-chart></verdict-chart>
```
SpaceWhite: You need to specify the `verdict-chart` prop value https://vuejs.org/v2/guide/components.html#Passing-Data-to-Child-Components-with-Props

sarvesh211999: Sorry It was some mistake in some commit I have done it earlier

frontend/www/js/omegaup/components/VerdictChart.vue
sarvesh211999: I am not able to understand . Can you clarify more??

sarvesh211999: I have changed somethings. Please review it.

frontend/www/js/omegaup/components/VerdictChart.vue
sarvesh211999: I am not able to understand.

frontend/www/js/omegaup/components/contest/Stats.vue
sarvesh211999: Done

frontend/www/js/omegaup/components/VerdictChart.vue
SpaceWhite: The usage of `setTimeout` is to call `draw_pie_chart` function every T milliseconds. See the [docs](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout). We used this because `stats` value may changed in the middle of execution. In vue we have [watcher](https://vuejs.org/v2/guide/computed.html#Watchers) which allows you to run specific commands when a variable changed its value.

lhchavez: so what we're trying to say is there is a way to avoid the `setTimeout` call, since it's very annoying to see everything move every second even if the data did not change.

sarvesh211999: Okay I got .
I am little busy now
I will change by tommorow and commit.

sarvesh211999: I have added watch for stats.

sarvesh211999:
Sorry I just have to call the function

sarvesh211999:
Please review it. And please help about selenium test.

frontend/www/js/omegaup/components/VerdictChart.vue
```diff
@@ -0,0 +1,33 @@
...
+      if (this.stats != null) {
+        if (this.$el.run_counts_chart != null) {
+          return;
+        }
+
+        // Draw verdict counts pie chart
+        this.$el.run_counts_chart =
+            oGraph.verdictCounts(this.$el, this.contestAlias, this.stats);
+      }
+      setTimeout(this.updateRunCountsData, 10000);
```
SpaceWhite: The purpose of using `watch` is to remove `setTimeout` command

frontend/www/js/omegaup/components/VerdictChart.vue
```diff
@@ -0,0 +1,33 @@
...
+    },
+    updateRunCountsData: function() {
+      this.$el.run_counts_chart.series[0].setData(
+          oGraph.normalizeRunCounts(this.stats));
+      setTimeout(this.updateRunCountsData, 10000);
```
SpaceWhite: Same here

frontend/www/js/omegaup/contest/stats.js
```diff
@@ -24,29 +48,29 @@ omegaup.OmegaUp.on('ready', function() {
   }
 
   function updateRunCountsData() {
-    window.run_counts_chart.series[0].setData(oGraph.normalizeRunCounts(stats));
+    // window.run_counts_chart.series[0].setData(oGraph.normalizeRunCounts(stats));
     window.distribution_chart.series[0].setData(oGraph.getDistribution(stats));
     setTimeout(updateRunCountsData, updateRunCountsChartTimeout);
```
SpaceWhite: Same here. It would be nice to first fix issues lhchavez and I mentioned previously and clean up the commits.

sarvesh211999:
Okhay I will first then componentize all the charts

sarvesh211999:
I have added verdictChartRunCountsChart as a flag so that distribution chart does not get redrawn everytime stats change. Used event bus to pass value from verdict chart component to distribution chart component

frontend/www/js/omegaup/components/DistributionChart.vue
```diff
@@ -0,0 +1,45 @@
...
+  mounted: function() { this.draw_distribution_chart();},
+  watch: {
+    stats: function() {
+      this.draw_distribution_chart();
+      this.updateRunCountsData();
+      this.distributionChart();
+    }
+  },
+  methods: {
+    draw_distribution_chart: function() {
```
lhchavez: nit: use camelCase consistently for method names.

frontend/www/js/omegaup/components/DistributionChart.vue
```diff
@@ -0,0 +1,45 @@
...
+      if (this.stats != null) {
```
lhchavez: nit: use the [guard clause pattern](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html) consistently:

```javascript
if (!this.stats) return;
// rest
```

frontend/www/js/omegaup/components/DistributionChart.vue
```diff
@@ -0,0 +1,45 @@
...
+    stats: Object,
+    contestAlias: String,
+  },
+  data: function() {
+    return { verdictChartRunCountsChart: false, }
+  },
+  mounted: function() { this.draw_distribution_chart();},
+  watch: {
+    stats: function() {
+      this.draw_distribution_chart();
```
lhchavez: do you need to call this every time `stats` is updated?

frontend/www/js/omegaup/components/DistributionChart.vue
```diff
@@ -0,0 +1,45 @@
...
+      this.updateRunCountsData();
+      this.distributionChart();
```
lhchavez: why do you need to invoke this? The caller (`frontend/www/js/omegaup/contest/stats.js
`, who sets `stats`) already knows when this is updated.

you should be able to just remove it altogether. same with `runCountsChart`.

frontend/www/js/omegaup/contest/stats.js
```diff
@@ -1,18 +1,43 @@
-omegaup.OmegaUp.on('ready', function() {
+import contest_Stats from '../components/contest/Stats.vue';
+import Vue from 'vue';
+import {OmegaUp} from '../omegaup.js';
+export const bus = new Vue();
+
+OmegaUp.on('ready', function() {
+  var stats = null;
```
lhchavez: seems like you don't need this anymore. how about you rename `StatsVue` to just `stats` and use `stats.stats` instead of `stats`? That way you only have one variable.

frontend/www/js/omegaup/contest/stats.js
```diff
@@ -1,18 +1,43 @@
-omegaup.OmegaUp.on('ready', function() {
+import contest_Stats from '../components/contest/Stats.vue';
+import Vue from 'vue';
+import {OmegaUp} from '../omegaup.js';
+export const bus = new Vue();
```
lhchavez: This is not used anywhere, so you can just remove it from all files.

frontend/www/js/omegaup/contest/stats.js
```diff
@@ -1,18 +1,43 @@
...
+
+OmegaUp.on('ready', function() {
+  var stats = null;
   var contestAlias =
       /\/contest\/([^\/]+)\/stats\/?.*/.exec(window.location.pathname)[1];
+  var StatsVue = new Vue({
+    el: '#contest-stats',
+    props: {
+      abc: String,
```
lhchavez: this seems to be unused.

frontend/www/js/omegaup/components/contest/Stats.vue
```diff
@@ -0,0 +1,30 @@
+<template>
+  <div class="post">
+    <div class="copy">
+      <h1>Estadísticas en vivo</h1>
+      <div class="total-runs">
+        <h2></h2>
+      </div><verdict-chart v-bind:contestalias="contestAlias"
```
lhchavez: nit: `v-bind:contestAlias="contestAlias"`. Same below.

frontend/www/js/omegaup/contest/stats.js
```diff
@@ -23,36 +48,14 @@ omegaup.OmegaUp.on('ready', function() {
...
 
-  function updateRunCountsData() {
-    window.run_counts_chart.series[0].setData(oGraph.normalizeRunCounts(stats));
-    window.distribution_chart.series[0].setData(oGraph.getDistribution(stats));
-    setTimeout(updateRunCountsData, updateRunCountsChartTimeout);
-  }
-
   function drawCharts() {
-    $('#total-runs')
+    $($('.total-runs')[0])
```
lhchavez: same as in the other places where you touched a line that uses jQuery: let's try to avoid using jQuery.

```javascript
document.querySelector('.total-runs').textContent = omegaup....;
```

frontend/www/js/omegaup/components/DistributionChart.vue
lhchavez: i am talking about `distributionChart` and `runCountsChart` in particular.

frontend/www/js/omegaup/components/DistributionChart.vue
sarvesh211999: The problem that is coming is that iniially we can call the function as mounted but intially stats is coming undefined it is updated later. So by keeping watch we can call the function again.

sarvesh211999: Without that it is not possible to render charts. Because mounted function is called only once.
Any stats are getting update after a particular time.

sarvesh211999: There is only 1 second delay in receving data in vuejs . After 1 second stats become defined. 
But it is not advisable to use this 1 second.

sarvesh211999: Yess I am trying to fix that. I am trying to render the component after we get stats. This way it will be better.

frontend/www/js/omegaup/components/contest/Stats.vue
sarvesh211999: on lint fix it again become contestalias

sarvesh211999:
Please review @lhchavez @SpaceWhite

frontend/www/js/omegaup/components/contest/Stats.vue
```diff
@@ -0,0 +1,39 @@
...
+<script>
+import verdict_chart from '../VerdictChart.vue';
+import distribution_chart from '../DistributionChart.vue';
+export default {
+  props: {
+    stats: Object,
+    contestAlias: String,
+  },
+  components: {
+    'verdict-chart': verdict_chart,
```
lhchavez: nit: all component names should start with `omegaup-`, for consistency: `omegaup-verdict-chart`, `omegaup-distribution-chart`.

frontend/www/js/omegaup/contest/stats.js
```diff
@@ -0,0 +1,53 @@
...
+                updatePendingRunsChartTimeout,
+                function() { return stats.pending_runs.length; });
+          }
+        })
+        .fail(omegaup.UI.apiError);
+    updateStats();
+  };
+
+  function updateStats() {
+    setTimeout(function() { getStats(); }, callStatsApiTimeout);
```
lhchavez: I know this was in the original code, but can you change this to use [`setInterval`](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setInterval) instead? Then you don't need to explicitly call (or even declare!) `updateStats()`. Also, use a lambda:

```javascript
const callStatsApiTimeout = 10 * 1000;  // 10s
setInterval(() => getStats(), callStatsApiInterval);
getStats();

File: frontend/www/js/omegaup/contest/stats.js [hunk 25]
```diff
@@ -0,0 +1,53 @@
...
+  var callStatsApiTimeout = 10 * 1000;
+  var updateRunCountsChartTimeout = callStatsApiTimeout;
+  var updatePendingRunsChartTimeout = callStatsApiTimeout / 2;
+  function getStats() {
+    omegaup.API.Contest.stats({contest_alias: contestAlias})
+        .then(function(s) {
+          stats = s;
+          if (ChartsDrawn != true) {
+            ChartsDrawn = true;
+            var StatsVue = new Vue({
```
lhchavez: What happens if you render this unconditionally before getting the stats? that way you wouldn't need to have this logic to check whether the charts have been drawn.

you might need a null-check in the components, but that's totally fine since they need one already.

frontend/www/js/omegaup/contest/stats.js
```diff
@@ -0,0 +1,53 @@
...
+
+OmegaUp.on('ready', function() {
+  var stats = null;
+  var ChartsDrawn = false;
+  var contestAlias =
+      /\/contest\/([^\/]+)\/stats\/?.*/.exec(window.location.pathname)[1];
+
+  Highcharts.setOptions({global: {useUTC: false}});
+  var callStatsApiTimeout = 10 * 1000;
+  var updateRunCountsChartTimeout = callStatsApiTimeout;
```
lhchavez: This seems to be unneeded now.

frontend/www/js/omegaup/contest/stats.js
```diff
@@ -0,0 +1,53 @@
...
+  var updatePendingRunsChartTimeout = callStatsApiTimeout / 2;
+  function getStats() {
+    omegaup.API.Contest.stats({contest_alias: contestAlias})
+        .then(function(s) {
+          stats = s;
```
lhchavez: this might be unneeded. instead, you might need to do `StatsVue.stats = s;`

frontend/www/js/omegaup/contest/stats.js
```diff
@@ -0,0 +1,53 @@
...
+              data: {
+                stats: stats,
+                contestAlias: contestAlias,
+              },
+              components: {
+                'contestStats': contest_Stats,
+              },
+            });
+            // Pending runs chart
+            window.pending_chart = oGraph.pendingRuns(
```
lhchavez: Can you store this in a local variable instead of in `window`?

```javascript
let pendingChart = null;
...
pendingChart = oGraph.pendingRuns(...);
```

frontend/www/js/omegaup/contest/stats.js
```diff
@@ -0,0 +1,53 @@
...
+                updatePendingRunsChartTimeout,
+                function() { return stats.pending_runs.length; });
```
lhchavez: nit: use a lambda: `() => stats.pending_runs.length`

frontend/www/js/omegaup-graph.js
sarvesh211999: Using document querySelector is fine right ??

frontend/www/js/omegaup/components/contest/Stats.vue
```diff
@@ -0,0 +1,39 @@
...
+    'distribution-chart': distribution_chart,
+  },
+  mounted: function() { this.updateTotalRuns();},
+  methods: {
+    updateTotalRuns: function() {
+      this.$el.querySelector('.total-runs').textContent =
```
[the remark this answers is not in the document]
sarvesh211999: yeah sure

frontend/www/js/omegaup/components/contest/Stats.vue
sarvesh211999: okay

frontend/www/js/omegaup/contest/stats.js
sarvesh211999: if we render before getting stats. Then we might have to again call the function in the vue files to draw graph after getting stats. In earlier commits i have done this but this way was much better so i have done this. And in that also we might need event bus and many flags. So It is better in this way.

frontend/www/js/omegaup/components/contest/Stats.vue
sarvesh211999: Why it is used in veridict count function to give title. It is necessary to pass contestAlias. It is argument of function

frontend/www/js/omegaup/components/DistributionChart.vue
```diff
@@ -0,0 +1,26 @@
...
+  props: {
+    stats: Object,
+    contestAlias: String,
+  },
+  mounted: function() { this.draw_distribution_chart();},
+  watch: {stats: function() { this.updateRunCountsData();}},
+  methods: {
+    draw_distribution_chart: function() {
+      // Draw distribution of scores chart
+      this.$el.distribution_chart =
```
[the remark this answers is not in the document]
sarvesh211999: no it does not work . I think it should be in DOM

frontend/www/js/omegaup-graph.js
lhchavez: totally! it's not jQuery, so it's okay to use.

frontend/www/js/omegaup/components/contest/Stats.vue
lhchavez: oh i might have missed that one. if it's still used, then keep it.

frontend/www/js/omegaup/contest/stats.js
sarvesh211999: In previous commits I have done that but problem in that is that we have to use event bus to check distribution chart whether they had been drawn.
So to prevent that one check is only added.

sarvesh211999:
Please review @lhchavez

frontend/www/js/omegaup/contest/stats.js
```diff
@@ -3,22 +3,20 @@ import Vue from 'vue';
...
   var updatePendingRunsChartTimeout = callStatsApiTimeout / 2;
+  var pendingChart = null;
   function getStats() {
     omegaup.API.Contest.stats({contest_alias: contestAlias})
         .then(function(s) {
-          stats = s;
           if (ChartsDrawn != true) {
             ChartsDrawn = true;
-            var StatsVue = new Vue({
+            var stats = new Vue({
```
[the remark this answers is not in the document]
sarvesh211999: yeah in that case first stats become undefined so for that we have to call draw char again and again.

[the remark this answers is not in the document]
sarvesh211999: What to do ??
Changes done according to your other comments.

frontend/www/js/omegaup/contest/stats.js
```diff
@@ -0,0 +1,50 @@
...
+            // Pending runs chart
+            pendingChart = oGraph.pendingRuns(updatePendingRunsChartTimeout,
+                                              () => s.pending_runs.length);
+            stats.stats = s;
+          }
+        })
+        .fail(omegaup.UI.apiError);
+  };
+
+  setInterval(() => getStats(), callStatsApiTimeout);
```
SpaceWhite: Instead of redrawing Vue component every time `t`, can you just update the variable inside Vue?
Something like
```
var stats = new Vue({
(existing vue code)....
methods: {
    loadData: function() {
        API.Contest.stats... {
              (codes).....
              this.stats = s;
        }
    },
},
ready: function() {
    this.loadData();
    setInterval(() => this.loadData(), callStatsApiTimeout);
}
})
```

sarvesh211999: In this also there will be same problem that stats will come undefined.

sarvesh211999:
@lhchavez what to do ??

frontend/www/js/omegaup/contest/stats.js
lhchavez: Maybe this is fine. API methods shouldn't be called from Vue components anyways.

PR #2095, opened 2018-06-20
[Plagiarism detector] Disqualify submission feature

SpaceWhite:
## Description
Allow admin to disqualify a submission

## Changes
1. Change `Runs.test` table to (normal, test, disqualify)
2. Add apiDisqualify
3. Fix cache bug (Scoreboard caches aren't renewed when `UserController::deleteCache` is triggered)
4. Add disqualify UI to admin arena

Demo:
untitled


frontend/database/68_run_test.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE `Runs` MODIFY COLUMN `test` ENUM("normal", "test", "disqualify");
```
lhchavez: you might need to do this in several steps:

```sql
CREATE COLUMN `type` ENUM('normal', 'test', 'disqualified') DEFAULT 'normal';
UPDATE ... SET type = 'test' WHERE test = 1;
DROP COLUMN test;
```

frontend/server/controllers/RunController.php
```diff
@@ -558,6 +558,33 @@ public static function apiRejudge(Request $r) {
...
+     * @param Request $r
+     * @throws InvalidDatabaseOperationException
+     */
+    public static function apiDisqualify(Request $r) {
+        // Get the user who is calling this API
+        self::authenticateRequest($r);
+
+        self::validateDetailsRequest($r);
+
+        if (!(Authorization::canEditRun($r['current_identity_id'], $r['run']))) {
```
lhchavez: nit: there's a superfluous paren:

```php
if (!Authorization::...) {
```

frontend/server/controllers/RunController.php
```diff
@@ -558,6 +558,33 @@ public static function apiRejudge(Request $r) {
...
+            throw new ForbiddenAccessException('userNotAllowed');
+        }
+
+        $r['run']->test = 'disqualify';
+        RunsDAO::save($r['run']);
+
+        $response = [];
```
lhchavez: nit:

```php
return [
    'status' => 'ok',
];
```

heduenas:
As discussed offline, please add some unit testing for disqualification.

frontend/database/68_run_test.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE `Runs` ADD COLUMN `type` ENUM('normal', 'test', 'disqualify') DEFAULT 'normal';
+UPDATE `Runs` SET type='test' WHERE test=1;
+ALTER TABLE `Runs` DROP COLUMN test;
+ALTER TABLE `Runs` CHANGE COLUMN `type` `test` ENUM('normal', 'test', 'disqualify');
```
lhchavez: wait, the goal was to drop the `test` column altogether D: can you remove this line and the one below?

frontend/tests/controllers/RunDisqualifyTest.php
```diff
@@ -0,0 +1,76 @@
+<?php
+
+/**
+ * Unittest for disqualifying run
+ *
+ * @author SpaceWhite
+ */
+class RunDisqualifyTest extends OmegaupTestCase {
```
heduenas: These two tests share a good chunk of code, can you make a common function for them?

SpaceWhite: I've changed the code a little bit. I think we don't need to do that :)

frontend/server/controllers/RunController.php
```diff
@@ -559,6 +559,32 @@ public static function apiRejudge(Request $r) {
...
+     * @param Request $r
+     * @throws InvalidDatabaseOperationException
+     */
+    public static function apiDisqualify(Request $r) {
+        // Get the user who is calling this API
+        self::authenticateRequest($r);
+
+        self::validateDetailsRequest($r);
+
+        if (!Authorization::canEditRun($r['current_identity_id'], $r['run'])) {
```
lhchavez: We recently made it possible for contest/course admins to be able to view the details of a run (even if they themselves are not admins of the problem) *if* they have solved the problem already:

https://github.com/omegaup/omegaup/blob/4a04d79657cbfb8887eb43660010d8bbc6218dbc/frontend/server/controllers/RunController.php#L623-L624

maybe we should refactor this check so that `apiDisqualify` also uses it.

frontend/database/69_run_test.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE `Runs` ADD COLUMN `type` ENUM('normal', 'test', 'disqualify') DEFAULT 'normal';
```
lhchavez: argh, missed this earlier D:

nit: `disqualified`

frontend/server/controllers/RunController.php
SpaceWhite: I don't see the point why we need to refactor the code.
In that commit, we allow those who have solved the problem to view others submission. But it doesn't mean that user who have solved the problem can disqualify others submission. And apiDisqualify doesn't use anything in APIDetails on the first place.
Let me know what you think @lhchavez :)

lhchavez:
all right, Travis has been fixed, now the failures are legit. Can you fix them please?

SpaceWhite:
@lhchavez welp, I missed something in RunController. PTAL :)

Issue #2116, opened 2018-06-29
Add dependencies between Selenium subtasks


lhchavez:
We would love to have some Selenium subtasks depend on other subtests. So for example, one subtest can create a course and add some students, and then several other subtests can just assume the course has been created.

This can be solved by something like https://pytest-dependency.readthedocs.io/en/latest/

pabo99:
Selenium is no longer used in omegaUp. This issue can be closed

PR #2122, opened 2018-07-01
Selenium dependency

AtharvaKhare:
Added dependencies between selenium subtasks
Fixes: #2116 

Checklist:

- [x] Complete test_contest.py
- [ ] Modify test_course.py
- [ ] Modify test_smoke.py
- [ ] Make flaky compatible with the tests
- [x] Update pytest-dependency by using custom pip wheel

### DIFF
--- a/frontend/tests/ui/conftest.py
+++ b/frontend/tests/ui/conftest.py
@@ -62,9 +62,10 @@ class JavaScriptLogCollector:
             yield entry
 
 
-class Driver:
+class Driver:  # pylint: disable=too-many-instance-attributes
     '''Wraps the state needed to run a test.'''
 
+    # pylint: disable=too-many-arguments
     def __init__(self, browser, wait, url, options):
         self.browser = browser
         self.wait = wait
--- a/frontend/tests/ui/conftest.py
+++ b/frontend/tests/ui/conftest.py
@@ -62,15 +62,14 @@ class JavaScriptLogCollector:
             yield entry
 
 
-class Driver:  # pylint: disable=too-many-instance-attributes
+class Driver:
     '''Wraps the state needed to run a test.'''
 
-    # pylint: disable=too-many-arguments
-    def __init__(self, browser, wait, url, worker_id, options):
+    def __init__(self, browser, wait, url, options):
         self.browser = browser
         self.wait = wait
-        self._worker_id = worker_id
         self._next_id = 0
+        self._session_id = None
         self._url = url
         self.options = options
         self.user_username = self.create_user()
@@ -81,7 +80,18 @@ class Driver:  # pylint: disable=too-many-instance-attributes
         '''Generates a relatively unique id.'''
 
         self._next_id += 1
-        return '%s_%d_%d' % (self._worker_id, int(time.time()), self._next_id)
+        return '%d_%d' % (int(time.time()), self._next_id)
+
+    def set_session_id(self):
+        '''Sets an id for particular test session.'''
+        self._session_id = self.generate_id()
+        return self._session_id
+
+    def get_session_id(self):
+        '''Returns current id, assuming set_session_id() is called earlier.'''
+
+        assert self._session_id is not None, "Please use set_session_id()."
+        return self._session_id
 
     def url(self, path):
         '''Gets the full url for :path.'''
@@ -535,7 +545,6 @@ def driver(request, browser_name):
 
         try:
             yield Driver(browser, wait, request.config.option.url,
-                         os.environ.get('PYTEST_XDIST_WORKER', 'w0'),
                          request.config.option)
         finally:
             if util.CI:
--- a/frontend/tests/ui/test_contest.py
+++ b/frontend/tests/ui/test_contest.py
@@ -32,7 +32,7 @@ def test_create_users(driver):
                                  "test_create_users[chrome]"])
 @util.no_javascript_errors()
 @util.annotate
-def test_create_contest_admin(driver, user):
+def test_create_contest_admin(driver):
     '''Creates a contest as an admin.'''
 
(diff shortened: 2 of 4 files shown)


AtharvaKhare:
@lhchavez review please :)

Should there be any modification in test_course.py ? I feel that it should be kept as is.
Feedback appreciated.

.pylintrc
```diff
@@ -11,7 +11,7 @@
 #
 # TODOs are useful to locally indicate that something is not (yet) working.
 #  fixme = (warning notes in code comments; message varies).
-disable=I0011,R0903,E1101,fixme
+disable=I0011,R0903,E1101,fixme,W0621
```
lhchavez: Let's try to avoid touching this file. Instead, fix whatever pylint was complaining about.

frontend/tests/ui/conftest.py
```diff
@@ -549,3 +549,12 @@ def driver(request, browser_name):
     except:
         logging.exception('Failed to initialize')
         raise
+
+
+@pytest.fixture(scope='session')
+def strings(driver):
```
lhchavez: given that we want to just generate the ID just once, wouldn't it be better to just make `driver` calculate the `run_id` once and store it as a member? that way you don't need `strings` at all!

stuff/travis/selenium.sh
```diff
@@ -14,6 +14,7 @@ stage_install() {
 	pip3 install --user selenium
 	pip3 install --user pytest
 	pip3 install --user pytest-xdist
+	pip3 install --user pytest-dependency
```
lhchavez: Can you remove `pytest-xdist`?

frontend/tests/ui/conftest.py
AtharvaKhare: Actually `strings` stores the items already in the database.

If I remove `strings`, and `test_create_contest_admin` fails after adding one user, flaky will rerun the test and will attempt to re-add the user with same username (since `run_id` is same). This leads to test failing.

Another advantage of storing already present items in `strings` is that we can skip some tests, such as `test_smoke.py::test_create_user()`.

stuff/travis/selenium.sh
AtharvaKhare: Yes will do.

I was thinking that `test_contest.py` will run first in travis, and then use xdist for other two tests. How does that sound to yoU?

frontend/tests/ui/conftest.py
AtharvaKhare: Alright, I will remove use of strings from test_smoke.py

I cannot remove use of strings from test_contest.py due to flaky. Commit without use of strings is this one: https://github.com/omegaup/omegaup/pull/2122/commits/138534484ec3586416b864d71e74e0320dbfbebc and would lead to errors like this: https://travis-ci.org/omegaup/omegaup/jobs/399013701
driver.register_user fails on second run since the same user (due to constant run_id) is already present in the database.

(The test didnt have any error in itself - see commit https://github.com/omegaup/omegaup/pull/2122/commits/3f6d92307c12973c7d146aa4b836224fd0981ee9 passes the selenium tests.)

lhchavez: what's stopping you from creating a test that just creates that user and letting the dependency management to take care of ensuring that the tests that depend on that user being in the database have passed?

stuff/travis/selenium.sh
AtharvaKhare: Yes. So we run tests in two parts:
1 - Run tests sequentially which use pytest-dependency - `test_contest.py`
2- Run tests with xdist which dont use the package - `test_smoke.py` and `test_course.py`, since these two tests do not contain any dependency (yet).

lhchavez: given that you'll get rid of `strings` and will most likely add dependencies on `test_smoke.py` and/or `test_course.py` due to that (and also, we'll probably start adding more dependencies soon now that we have that capability), let's remove the temptation to use `xdist` sooner rather than later.

frontend/tests/ui/conftest.py
AtharvaKhare: I honestly didn't think of doing that. I will break down test_contest_admin into subtests and add dependency, thanks!
I will store run_id in driver and remove strings all together. :)

stuff/travis/selenium.sh
AtharvaKhare: Alright, thanks! :)

AtharvaKhare:
Hi @lhchavez 
Please check if the changes in `test_contest.py` and `class Driver` are appropriate. :)

I could not come up with dependencies in test_course and test_smoke. Please tell me if there are any.
Also, should I remove the newly created stub in `test_smoke.py` or keep it to demonstrate inter-file dependency? (assert statement is there to prevent lint errors)

Additionally, I am using a custom pytest-dependency package, which is hosted on my profile. The original one uses deprecated function, creating about 39 warnings, and does not support inter-class dependency, so I decided to compile a new one.

frontend/tests/ui/test_contest.py
```diff
@@ -5,29 +5,88 @@
...
-    run_id = driver.generate_id()
-    contest_alias = 'ut_contest_%s' % run_id
-    problem = 'sumas'
+    run_id = driver.set_session_id()
     user1 = 'ut_user_1_%s' % run_id
     user2 = 'ut_user_2_%s' % run_id
     password = 'P@55w0rd'
 
     driver.register_user(user1, password)
     driver.register_user(user2, password)
```
AtharvaKhare: How should I handle selenium failing here(on a rare case)? Flaky restarts the function, and tries to register user1 again (which has the same id) and fails.

Issue #2076, opened 2018-06-07
Virtual contest related bugs and improvement


SpaceWhite:
### Current implementation
In order to distinguish virtual contest contestants and original contest contestants, omegaup append the username with `-(virtual)` word [Link](https://github.com/omegaup/omegaup/blob/d90f10ac471ecc4647151e38e7a390f5f4edb4ad/frontend/www/js/omegaup/arena/arena.js#L574-L575). The results will be like this:
![screenshot-2018-6-6 omegaup](https://user-images.githubusercontent.com/15065085/41074987-984d3518-69d8-11e8-8343-980e5137962c.png)
### Expected result
It would be nice to change to scoreboard contestant name from `username - (virtual) (name - (virtual))` to `username - (name) - (virtual)` or `username - (name) 👻`

### Current implementation
[Link](https://github.com/omegaup/omegaup/blob/d90f10ac471ecc4647151e38e7a390f5f4edb4ad/frontend/www/js/contest.edit.js#L92-L99). #2009 This code is used to hide feature in the edit contest in virtual contest. Adding problem, delete problem, change description and etc are not available in virtual contest. But this code hide the features after the page has loaded, which may results in UI glitch. 
### Expected solution
Convert `contest.edit.tpl` to vue. And hide those UI in vue

## 3. Virtual contest bug
- Virtual contest starting time can't be set
- Add `contest name - (virtual)` to `/arena` past contest, so that contestant can distinguish between virtual contest or original contest

SpaceWhite:
@lhchavez  Regarding the third issue, it seems like there is no `start_time` restriction in the current implementation. Should we restrict start_time in normal contest?

PR #2127, opened 2018-07-06
[Ghost mode] Convert contest edit to vue and bug fixes

SpaceWhite:
## List of changes
1. Move contest.edit.js to edit.js 
2. Convert to tab panes into vues (`ContestNewForm.vue`, `ContestClone.vue`, etc)
3. Convert auto complete to vue (`AutocompleteUser.vue`, etc)
4. Fixes bugs in simulating virtual contest scoreboard

Reference #2076  no 2


SpaceWhite:
@lhchavez PTAL. Any idea how to update child component after API call in edit.js? I tried to use something like `ev.problems = problems` after API call, and it doesn't work properly. Do you have any suggestion?

frontend/www/js/omegaup/components/AutocompleteGroupAdmins.vue
```diff
@@ -0,0 +1,15 @@
+<template>
+  <input class="typeahead form-control"
+        ref="input">
+</template>
+
+<script>
+import {UI, API} from '../omegaup.js';
+export default {
+  data: function() {
```
lhchavez: How about you create a generic `Autocomplete.vue` and you pass in a callback that takes one parameter (the element where the typeahead will be mounted). I haven't tried this myself, but does this work?

```html
<omegaup-autocomplete v-bind:init="el => UI.typeahead(el, API.Group.list)"
                      v-model="admin"></omegaup-autocomplete>
```

(only suggesting it because https://stackoverflow.com/a/46101329/688337 suggests it _should_ work).

frontend/www/js/omegaup/components/contest/ContestAddProblem.vue
```diff
@@ -0,0 +1,60 @@
...
+        </tr>
+      </thead>
+      <tbody>
+        <tr v-for="problem in problems">
+          <td>{{problem.order}}</td>
+          <td>
+            <a v-bind:href="`/arena/problem/${problem.alias}/`">{{problem.alias}}</a>
+          </td>
+          <td>{{problem.points}}</td>
+          <td><button class="close">x</button></td>
```
lhchavez: Can you use [`×`](https://www.fileformat.info/info/unicode/char/00d7/index.htm) instead of `x`? Bootstrap uses this in their docs: https://getbootstrap.com/docs/4.0/utilities/close-icon/ (`&times;` is the HTML entity for `×`).

Same in all the other places where `x` is used.

frontend/www/js/omegaup/components/contest/ContestAddProblem.vue
```diff
@@ -0,0 +1,60 @@
...
+  </div>
+</template>
+
+<script>
+import {T} from '../../omegaup.js';
+import AutocompleteProblems from '../AutocompleteProblems.vue';
+export default {
+  props: {problems: Array},
+  data: function() {
+    return { T: T, alias: "", point: 100, order: 1, }
```
lhchavez: nit: `points`

frontend/www/js/omegaup/components/contest/ContestAddProblem.vue
```diff
@@ -0,0 +1,60 @@
...
+  },
+  methods: {onSubmit: function() { this.$parent.$emit('addProblem', this);}},
```
lhchavez: nit throughout the change: most of the other event names use `kebab-case`, so let's follow that convention for consistency.

frontend/www/js/omegaup/components/contest/ContestAdmins.vue
```diff
@@ -0,0 +1,58 @@
...
+          <th>{{T.contestEditRegisteredAdminUsername}}</th>
+          <th>{{T.contestEditRegisteredAdminRole}}</th>
+          <th>{{T.contestEditRegisteredAdminDelete}}</th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr v-for="admin in admins"
+            v-if="(admin.role != 'site-admin') || showSiteAdmin">
+          <td>
+            <a v-bind:href="`/profile/${admin.username}/`">{{admin.username}}</a>
```
lhchavez: Would it be too complicated to use https://github.com/omegaup/omegaup/blob/master/frontend/www/js/omegaup/components/user/Username.vue instead?

frontend/www/js/omegaup/components/contest/ContestEdit.vue
```diff
@@ -0,0 +1,112 @@
+<template>
+  <div>
+    <div class="page-header">
+      <h1>{{T.contestEdit + ' ' + contest.title}} <small><a v-bind:href=
```
lhchavez: nit: Can you add a new translation string that can be used with `UI.formatString`? We should avoid concatenating translation strings because that makes it difficult to support languages where the order of the words changes in the future.

frontend/www/js/omegaup/components/contest/ContestGroupAdmins.vue
```diff
@@ -0,0 +1,48 @@
...
+        <tr>
+          <th>{{T.contestEditRegisteredGroupAdminName}}</th>
+          <th>{{T.contestEditRegisteredAdminRole}}</th>
+          <th>{{T.contestEditRegisteredAdminDelete}}</th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr v-for="group in groupAdmins">
+          <td>
+            <a v-bind:href="`/group/${group.alias}/edit/`"></a>{{group.name}}
```
lhchavez: Shouldn't `{{group.name}}` be inside the `<a></a>`?

frontend/www/js/omegaup/components/contest/ContestPublish.vue
```diff
@@ -0,0 +1,33 @@
+<template>
+  <div class="panel panel-primary">
+    <div class='panel-body'>
+      <form class='contest-publish-form'
+            v-on:submit.prevent="onSubmit">
+        <div class="form-group">
+          <label>{{T.contestNewFormPublic}}</label> <select class="form-control"
+               v-model="public">
+            <option value='0'>
```
lhchavez: nit: can you change all the apostrophes to double quotes to be consistent in HTML code?

frontend/www/js/omegaup/contest/edit.js
```diff
@@ -0,0 +1,169 @@
+import {OmegaUp, UI, API, T} from '../omegaup.js';
+import Vue from 'vue';
+import ContestEdit from '../components/contest/ContestEdit.vue';
+
+OmegaUp.on('ready', function() {
+  var contestAlias =
+      /\/contest\/([^\/]+)\/edit\/?.*/.exec(window.location.pathname)[1];
+
+  API.Contest.adminDetails({contest_alias: contestAlias})
```
lhchavez: welp, [arrow code](https://blog.codinghorror.com/flattening-arrow-code/) / [callback hell](http://callbackhell.com/) D:

How about you submit all these requests in parallel?

```javascript
$.when(
  API.Contest.adminDetails({contest_alias: contestAlias}),
  API.Contest.problems({contest_alias: contestAlias}),
  API.Contest.users({contest_alias: contestAlias}),
  API.Contest.admins({contest_alias: contestAlias})
).done(contest, problems, users, admins) {
  problems = problems.problems;
  // ....
}).fail(UI.apiError);

File: frontend/www/js/omegaup/contest/edit.js [hunk 10]
```diff
@@ -0,0 +1,169 @@
...
+                                        .fail(UI.apiError);
+                                  },
+                                  addProblem: function(ev) {
+                                    API.Contest.addProblem({
+                                                 contest_alias: contestAlias,
+                                                 order_in_contest: ev.order,
+                                                 problem_alias: ev.alias,
+                                                 points: ev.point,
+                                               })
+                                        .then(function(response) {
```
lhchavez: does converting these inline functions to lambdas make the code look cleaner?

frontend/www/js/omegaup/ui.js
```diff
@@ -198,7 +208,8 @@ let UI = {
               },
             })
         .on('typeahead:select', cb)
-        .on('typeahead:autocomplete', cb);
+        .on('typeahead:autocomplete', cb)
+        .trigger("change");
```
lhchavez: nit: use single quotes consistently in JavaScript code.

frontend/www/js/omegaup/ui.js
```diff
@@ -42,6 +42,16 @@ let UI = {
     return template;
   },
 
+  contestUpdated: function(data) {
+    if (data.status == 'ok') {
+      UI.success(omegaup.T.contestEditContestEdited + ' <a href="/arena/' +
+                 $('.new_contest_form #alias').val() + '">' +
+                 T.contestEditGoToContest + '</a>');
+    } else {
```
lhchavez: nit: use the guard clause pattern:

```javascript
if (data.status != 'ok') {
  UI.error(data.error || 'error');
  return;
}
UI.success(UI.formatString(..., document.querySelector(...).value));
```

also, can you move this to the file where you use it? We shouldn't be declaring component-specific code here.

frontend/www/js/omegaup/components/contest/ContestAddProblem.vue
```diff
@@ -0,0 +1,61 @@
...
+  </div>
+</template>
+
+<script>
+import {T, UI} from '../../omegaup.js';
+import Autocomplete from '../Autocomplete.vue';
+export default {
+  props: {problems: Array},
+  data: function() {
+    return { T: T, UI: UI, alias: "", points: 100, order: 1, }
```
lhchavez: nit: use single quotes in all JavaScript and PHP code. Double quotes is only used in HTML.

frontend/www/js/omegaup/components/contest/ContestNewForm.vue
```diff
@@ -0,0 +1,262 @@
...
+                {{T.wordsNo}}
+              </option>
+            </select>
+            <p class="help-block">{{T.contestNewFormScoreboardAtEndDesc}}</p>
+          </div>
+          <div class="form-group col-md-6">
+            <label>{{T.wordsLanguage}}s</label><br>
+            <select class="form-control"
+                 multiple="multiple">
+              <option value="{$language}">
```
lhchavez: This doesn't work :(

frontend/www/js/omegaup/components/contest/ContestNewForm.vue
```diff
@@ -0,0 +1,262 @@
...
+                {{T.wordsYes}}
+              </option>
+              <option value="0">
+                {{T.wordsNo}}
+              </option>
+            </select>
+            <p class="help-block">{{T.contestNewFormScoreboardAtEndDesc}}</p>
+          </div>
+          <div class="form-group col-md-6">
+            <label>{{T.wordsLanguage}}s</label><br>
```
lhchavez: Can you add a translation string for the plural version? Otherwise languages that don't make the plural the same way as english/spanish/portuguese will show this weirdly.

frontend/www/js/omegaup/contest/edit.js
```diff
@@ -0,0 +1,220 @@
+import {OmegaUp, UI, API, T} from '../omegaup.js';
+import Vue from 'vue';
+import ContestEdit from '../components/contest/ContestEdit.vue';
+
+OmegaUp.on('ready', function() {
+  const contestAlias =
+      /\/contest\/([^\/]+)\/edit\/?.*/.exec(window.location.pathname)[1];
+
+  const trigger =
```
lhchavez: Can you inline all the methods where they should be? It hurts readability a bit that you need to scroll to another place if you want to see the implementation of a callback.

IIUC you should be able to remove `trigger` altogether. If you stil have any methods that are used more than once, just declare a plan function.

frontend/www/js/omegaup/components/contest/ContestEdit.vue
```diff
@@ -0,0 +1,119 @@
+<template>
+  <div>
+    <div class="page-header">
+      <h1>{{UI.formatString(T.contestEditWithTitle, {title: contest.title})}}
+      <small><a v-bind:href="`/arena/${contest.alias}/`">{{T.contestDetailsGoToContest}}</a></small></h1>
+    </div>
+    <ul class="nav nav-tabs nav-justified">
+      <li class="active"
```
lhchavez: Can you conditionally add the `active` class to all the `li` and `div`s only when they are actually active? Note that you'll also need to fix the Selenium tests to wait for this to happen instead of waiting on `#problems` being loaded (the old school way): https://travis-ci.org/omegaup/omegaup/jobs/403921064#L5860

frontend/tests/ui/test_contest.py
```diff
@@ -5,11 +5,11 @@
 
 import urllib
 
+from ui import util
```
lhchavez: Can you revert this change? It causes lint failures locally :(

frontend/tests/ui/test_contest.py
```diff
@@ -224,8 +224,10 @@ def add_students_contest(driver, users):
     '''Add students to a recently contest.'''
 
     util.add_students(
-        driver, users, container_id='contestants',
-        parent_xpath='*[@id="contestants"]',
+        driver, users,
+        container_selector='.nav-tabs > li:nth-child(4) > a:nth-child(1)',
```
lhchavez: would it be possible to add a classname or somehow expose the refs? using `nth-child` is kind of brittle and things will break if we ever move tabs around.

Same everywhere you use that selector.

frontend/tests/ui/test_contest.py
```diff
@@ -235,23 +237,25 @@ def add_students_bulk(driver, users):
 
     driver.wait.until(
         EC.element_to_be_clickable(
-            (By.XPATH, ('//a[contains(@href, "#contestants")]')))).click()
+            (By.CSS_SELECTOR,
+             ('.nav-tabs > li:nth-child(4) > a:nth-child(1)')))).click()
     driver.wait.until(
         EC.visibility_of_element_located(
-            (By.CSS_SELECTOR, '#contestants')))
+            (By.CSS_SELECTOR, '#content')))
```
lhchavez: Same here, can you add something that's relevant to the contestants? Waiting for `#content` does nothing since that element should have been visible already.

frontend/tests/ui/test_contest.py
```diff
@@ -235,23 +237,25 @@ def add_students_bulk(driver, users):
...
 
     driver.wait.until(
         EC.visibility_of_element_located(
             (By.XPATH, (
-                '//textarea[contains(@name, "usernames")]')))).send_keys(
+                '//textarea[contains(@class, "form-control")]')))).send_keys(
```
lhchavez: Same here, `form-control` is too generic and we might have more of these in the page, resulting in the input going to wrong places.

frontend/www/js/omegaup/ui.js
```diff
@@ -50,14 +50,13 @@ let UI = {
...
-  contestUpdated: function(data) {
+  contestUpdated: function(data, contestAlias) {
     if (data.status != 'ok') {
       UI.error(data.error || 'error');
       return;
     }
     UI.success(omegaup.T.contestEditContestEdited + ' <a href="/arena/' +
-               $('.new_contest_form #alias').val() + '">' +
-               T.contestEditGoToContest + '</a>');
+               contestAlias + '">' + T.contestEditGoToContest + '</a>');
```
lhchavez: Can you make this a templated string to avoid doing concatenation?

frontend/www/js/omegaup/components/contest/ContestNewForm.vue
```diff
@@ -223,8 +228,13 @@ export default {
       submissionsGap: this.data.submissions_gap,
       title: this.data.title,
       titlePlaceHolder: '',
-      windowLength: this.data.window_length == 0 ? '' : this.data.window_length,
-      windowLengthEnabled: this.data.window_length != 0,
+      windowLength:
+          (this.data.window_length == 0 || this.data.window_length == null) ?
```
lhchavez: nit: `this.data.window_length || ''`

frontend/tests/ui/util.py
```diff
@@ -30,15 +30,17 @@
...
     '''Add students to a recently :instance.'''
 
     driver.wait.until(
         EC.element_to_be_clickable(
-            (By.XPATH, ('//a[contains(@href, "%s")]' % container_id)))).click()
+            (By.CSS_SELECTOR, container_selector))).click()
     driver.wait.until(
         EC.visibility_of_element_located(
-            (By.CSS_SELECTOR, '#%s' % container_id)))
+            (By.CSS_SELECTOR, container)))
```
lhchavez: seems like what you want to do is pass in two things: the _tab_ selector (what you currently call `container_selector`) and the _container_ selector (what you currently call `container`).

Do you mind renaming the arguments so they reflect that?

frontend/tests/ui/util.py
```diff
@@ -48,7 +50,7 @@ def add_students(driver, users, container_id, parent_xpath, submit_locator):
         driver.wait.until(
             EC.visibility_of_element_located(
                 (By.XPATH,
-                 '//*[@id="%s"]//a[text()="%s"]' % (container_id, user))))
+                 '//a[text()="%s"]' % (user))))
```
lhchavez: This can match all sort of random stuff. Can you somehow use the container selector to restrict this? You might need to change the container selector to be a `container_xpath` instead, because xpath and css selectors are kind of hard to mix.

frontend/tests/ui/test_contest.py
lhchavez: Oh. Can you run `python3.5 -m pip install --user pylint --upgrade` to avoid that? And then revert it.

frontend/tests/ui/test_contest.py
```diff
@@ -235,23 +237,25 @@ def add_students_bulk(driver, users):
...
                     ', '.join(users))
     driver.wait.until(
         EC.element_to_be_clickable(
             (By.CLASS_NAME, ('user-add-bulk')))).click()
     for user in users:
         driver.wait.until(
             EC.visibility_of_element_located(
-                (By.XPATH, '//*[@id="contest-users"]//a[text()="%s"]' % user)))
+                (By.XPATH,
+                 '//*[contains(@class, "table")]//a[text()="%s"]' % user)))
```
lhchavez: How about using `table` instead of `*[contains(@class, "table")]`? Should be equivalent.

But having said that, is there a way in which we can reference this one table instead of any table?

frontend/tests/ui/test_contest.py
```diff
@@ -260,16 +264,19 @@ def add_problem_to_contest(driver, problem):
 
     driver.wait.until(
         EC.element_to_be_clickable(
-            (By.XPATH, '//a[@href = "#problems"]'))).click()
+            (By.CSS_SELECTOR,
+             'li.problems > a'))).click()
 
-    driver.typeahead_helper('*[@id="problems"]', problem)
+    driver.typeahead_helper('*[contains(@class, "twitter-typeahead")]',
```
lhchavez: Same here, can we be more specific to choose one specific typeahead instead of any typeahead that just happens to match?

frontend/tests/ui/test_contest.py
```diff
@@ -260,16 +265,19 @@ def add_problem_to_contest(driver, problem):
...
+             'li.problems > a'))).click()
 
-    driver.typeahead_helper('*[@id="problems"]', problem)
+    driver.typeahead_helper('*[contains(@class, "problems-container")]',
+                            problem)
     driver.wait.until(
         EC.element_to_be_clickable(
             (By.CSS_SELECTOR,
-             '#add-problem-form button[type="submit"]'))).click()
+             '.btn'))).click()
```
lhchavez: this and the one in L280 are still missing to be explicit-ified.

frontend/www/js/omegaup/ui.js
lhchavez: this is still pending.

SpaceWhite:
@lhchavez  I've added some commits to this PR, could you please take a look?

frontend/server/libs/dao/Runs.dao.php
```diff
@@ -354,15 +355,19 @@ class RunsDAO extends RunsDAOBase {
...
+                        i.user_id NOT IN (SELECT ur.user_id FROM User_Roles ur WHERE ur.acl_id IN (?, ?) AND ur.role_id = ?)';
                 $val = [
                     $problemset_id,
                     $acl_id,
-                    $acl_id,
                     Authorization::SYSTEM_ACL,
                     Authorization::ADMIN_ROLE,
                 ];
+                if ($exclude_admin) {
+                    $log->info('EXCLUDE ADMIN');
```
lhchavez: this doesn't seem too useful. remove?

frontend/server/controllers/ContestController.php
```diff
@@ -915,13 +917,13 @@ public static function apiCreateVirtual(Request $r) {
 
         // Initialize contest
         $contest = new Contests();
-        $contest->title = $originalContest->title;
+        $contest->title = $originalContest->title . ' [virtual]';
```
lhchavez: can you make this a translation string instead? in general concatenation of things that get displayed in the UI is discouraged.

frontend/server/libs/dao/Runs.dao.php
```diff
@@ -354,15 +355,19 @@ class RunsDAO extends RunsDAOBase {
                         Problemset_Identities pi ON i.identity_id = pi.identity_id
                     WHERE
                         pi.problemset_id = ? AND
-                        i.user_id != (SELECT a.owner_id FROM ACLs a WHERE a.acl_id = ?) AND
-                        i.user_id NOT IN (SELECT ur.user_id FROM User_Roles ur WHERE ur.acl_id IN (?, ?) AND ur.role_id = ?);';
+                        i.user_id NOT IN (SELECT ur.user_id FROM User_Roles ur WHERE ur.acl_id IN (?, ?) AND ur.role_id = ?)';
```
lhchavez: does this work?  you are adding a `;` at the end and then in L367 concatenating more stuff to it.

frontend/server/libs/dao/Runs.dao.php
```diff
@@ -341,8 +341,9 @@ class RunsDAO extends RunsDAOBase {
 	 *
 	 */
 
-    final public static function getAllRelevantIdentities($problemset_id, $acl_id, $showAllRuns = false, $filterUsersBy = null, $group_id = null) {
+    final public static function getAllRelevantIdentities($problemset_id, $acl_id, $showAllRuns = false, $filterUsersBy = null, $group_id = null, $exclude_admin = false) {
```
lhchavez: the previous behavior is to exclude admins, so can you make this

```php
$excludeAdmin = true
```

? Also, we're trying to get camelCase everywhere, so every variable that you introduce should be in camelCase.

frontend/www/js/omegaup/arena/arena.js
```diff
@@ -571,88 +571,116 @@ export class Arena {
 
   refreshRanking() {
     let self = this;
-    if (self.options.originalContestAlias != null) {
-      API.Problemset.scoreboardEvents(
-                        {problemset_id: self.options.problemsetId})
+    if (self.options.contestAlias != null) {
```
lhchavez: Can this go in a separate change? Itś already really large as is :(

SpaceWhite: Welp I'd like to separate these javascript changes, but it seems like I mixed these javascript changes with other changes. Do you have any idea how to revert git with only specified lines?

frontend/server/libs/dao/Runs.dao.php
SpaceWhite: I think it is working. It concatenates exclude admin condition to the query if `exclude_admin` flag value is true. And the `;` at the end will be added in the end of `if`

frontend/www/js/omegaup/arena/arena.js
lhchavez: yes! you can do the following:

```shell
git checkout $LAST_MASTER_MERGE -- $FILES_THAT_HAVE_CHANGES
# your last merge was 11f4d4042c05948481bea9f2949e2fc2a02c5689,
# and the corresponding master commit was then 6a6bbb4, so
# $LAST_MASTER_MERGE=6a6bbb4 . this effectively does a revert of a subset
# of files.
git reset HEAD
# This unstages all the reverted files. Now you want to choose a subset of the
# lines of those files to actually commit:
git add -p
# This allows you to choose what changes to add, chunk by chunk!
git commit -m 'lines that i want'
git commit -a -m 'lines that i don't want'
```

you can then cherry-pick the last change to another branch :D

frontend/server/libs/dao/Runs.dao.php
lhchavez: oh nevermind, for some reason i thought there was a `;` added _inside_ the query. please disregard this.

frontend/www/js/omegaup/arena/arena.js
SpaceWhite: The `add -p` command is really cool. Thanks :)

frontend/www/api/ApiCaller.php
```diff
@@ -65,6 +65,7 @@ private static function isCSRFAttempt() {
         $allowed_hosts = [
             parse_url(OMEGAUP_URL, PHP_URL_HOST),
             OMEGAUP_LOCKDOWN_DOMAIN,
+            '35.231.177.254'
```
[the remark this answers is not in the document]
SpaceWhite: Welp, we don't need this. I comitted it by mistake. Actually I develop omegaUp platform on cloud, so I did this to whitelist my server ip address.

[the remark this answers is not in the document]
SpaceWhite: omg, that is very cool trick. Thanks for letting me know :)

PR #2135, opened 2018-07-14
Change virtual contest contestant naming

SpaceWhite:
## Description
Change virtual contestant naming from `username - (virtual) (name - (virtual))` to `username - (name) - (virtual)`
screenshot_2018-07-14 omegaup

#2076 No 1 issue

### DIFF
--- a/frontend/www/js/omegaup/arena/arena.js
+++ b/frontend/www/js/omegaup/arena/arena.js
@@ -576,10 +576,7 @@ export class Arena {
                         {problemset_id: self.options.problemsetId})
           .then(function(response) {
             let events = response.events;
-            for (let event of events) {
-              event.username += '-(virtual)';
-              event.name += '-(virtual)';
-            }
+            for (let event of events) event.virtual = true;
             API.Problemset.scoreboardEvents({
                             problemset_id: self.options.originalProblemsetId
                           })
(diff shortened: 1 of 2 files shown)


frontend/www/js/omegaup/arena/arena.js
```diff
@@ -715,10 +714,11 @@ export class Arena {
...
       newRanking[rank.username] = i;
 
-      let username = rank.username + ((rank.name == rank.username) ?
+      let username = rank.username +
+                     ((rank.name == rank.username) ?
                                           '' :
-                                          (' (' + UI.escape(rank.name) + ')'));
-
+                                          (' (' + UI.escape(rank.name) + ')')) +
+                     (rank.virtual ? ' - virtual' : '');
```
lhchavez: how about using `[...]` instead of parentheses? That way it's unambiguous because users' names already use parentheses. And maybe drop the hyphen.

Also, how about moving this to `UI`? That way you can use it here and in `Scoreboard.vue`.

frontend/www/js/omegaup/arena/arena.js
```diff
@@ -604,39 +601,41 @@ export class Arena {
...
+          rank[key].problems.push(
               {penalty: 0, percent: 0, points: 0, runs: 0});
         }
-        rank[env.username].total = {points: 0, penalty: 0};
+        rank[key].total = {points: 0, penalty: 0};
       }
-      let problem =
-          rank[env.username].problems[problemOrder[env.problem.alias]];
-      rank[env.username].problems[problemOrder[env.problem.alias]] = {
+      var problem =
```
lhchavez: let's keep the `let` here.

frontend/www/js/omegaup/arena/arena.js
```diff
@@ -604,39 +601,41 @@ export class Arena {
...
-      problemOrder[problem.alias] = problems.length + 1;
+      problemOrder[problem.alias] = problems.length;
       problems.push({order: problems.length + 1, alias: problem.alias});
     }
 
     data.forEach(function(env) {
       if (env.delta > delta) return;
-      if (!rank.hasOwnProperty(env.username)) {
-        rank[env.username] = {
+      var key = env.username + (env.virtual ? '-virtual' : '');
```
lhchavez: nit: `let`

frontend/templates/en.lang
```diff
@@ -1006,5 +1006,6 @@ wordsUsersFeedback = "Users Feedback"
 wordsValidator = "Validator"
 wordsVerdict = "Verdict"
 wordsVeryEasy = "VeryEasy"
+wordsVirtualSuffix = " [virtual]"
```
heduenas: Wouldn't it be better to make this string "%s [virtual]" in case we later introduce some language where it's better to put the "[virtual]" before the %s?

PR #2174, opened 2018-08-17
Fix simulate scoreboard and add sync scoreboard feature

SpaceWhite:
## Description
Virtual contest scoreboard related bug fix
## Changes
- [x] Fix how scoreboard was simulated in virtual contest (Thanks @heduenas)
- [x] Enable sync scoreboard when contestant create submission
- [x] invoke sync scoreboard only when things change
- [x] Disable sync scoreboard after contest ended


frontend/www/js/omegaup/arena/arena.js
```diff
@@ -110,7 +110,10 @@ class EventsSocket {
         self.arena.updateClarification(data.clarification);
       }
     } else if (data.message == '/scoreboard/update/') {
-      if (self.arena.contestAdmin && data.scoreboard_type != 'admin') return;
+      if (self.arena.contestAdmin && data.scoreboard_type != 'admin') {
+        if (self.arena.options.originalContestAlias == null) return;
+        self.arena.virtualRankingChange(data.scoreboard);
```
lhchavez: do you want to add another `return;` here to preserve the original behavior?

frontend/www/js/omegaup/arena/arena.js
```diff
@@ -571,88 +574,114 @@ export class Arena {
...
-        penalty: env.problem.penalty,
-        points: env.problem.points,
-        runs: problem.runs + 1
-      };
-      rank[key].total = env.total;
-    });
+    // Merge original contest scoreboard and virtual contest
+    let self = this;
+
+    API.Problemset.scoreboardEvents(
```
lhchavez: is there a way to avoid requesting this every time the ranking changes?

frontend/www/js/omegaup/arena/arena.js
SpaceWhite: Ah sure! Thanks for the fast response :)

frontend/www/js/omegaup/arena/arena.js
SpaceWhite: Yes, create a new function and trigger it with `setInterval`. Calling API every time is expensive

lhchavez: hmm but the old contest is finished by this point in time. so you should be able to call it exactly once.

frontend/www/js/omegaup/arena/arena.js
```diff
@@ -571,88 +581,127 @@ export class Arena {
...
+    let events = self.originalContestScoreboardEvent;
+    let currentDelta =
+        ((new Date()).getTime() - self.startTime.getTime()) / (1000 * 60);
+
+    for (let rank of data.ranking) rank.virtual = true;
 
     let problemOrder = {};
     let problems = [];
-    let initialScores = [];
+    let initProblems = [];
```
lhchavez: nit: `initialProblems`. We try to not abbreviate things too much.

frontend/www/js/omegaup/arena/arena.js
```diff
@@ -571,88 +581,127 @@ export class Arena {
...
 
     for (let problem of Object.values(self.problems)) {
       problemOrder[problem.alias] = problems.length;
+      initProblems[problems.length] = {
```
lhchavez: can you instead do `initialProblems.push({...})`?

frontend/www/js/omegaup/arena/arena.js
```diff
@@ -571,88 +581,127 @@ export class Arena {
...
-    data.forEach(function(env) {
-      if (env.delta > delta) return;
-      let key = env.username + (env.virtual ? '-virtual' : '');
-      if (!rank.hasOwnProperty(key)) {
-        rank[key] = {
+    // Calculate original contest scoreboard with current delta time
+    let originalContestRanking = {};
+    let originalContestEvents = [];
+
+    events.forEach(function(env) {
```
lhchavez: nit: `evt`. `env` normally refers to the environment variables. sorry i missed this in a previous review.

frontend/www/js/omegaup/arena/arena.js
```diff
@@ -571,88 +581,127 @@ export class Arena {
...
-      rank[key].problems[problemOrder[env.problem.alias]] = {
+      if (env.delta > currentDelta) return;
+      originalContestEvents.push(env);
+      let problem =
+          originalContestRanking[key].problems[problemOrder[env.problem.alias]];
+      originalContestRanking[key].problems[problemOrder[env.problem.alias]] = {
         penalty: env.problem.penalty,
         points: env.problem.points,
-        runs: problem.runs + 1
+        runs: problem ? problem.runs + 1 : 1
```
lhchavez: can you add a comment about this `+1` thing?

frontend/www/js/omegaup/arena/arena.js
```diff
@@ -571,88 +581,127 @@ export class Arena {
...
+            self.originalContestScoreboardEvent = response.events;
+            self.virtualContestRefreshInterval = setInterval(function() {
+              self.onVirtualRankingChange(data);
+              data = JSON.parse(JSON.stringify(defaultData));
+            }, 5000);
+          })
+          .fail(UI.apiError);
+    } else {
+      self.onVirtualRankingChange(data);
+      data = JSON.parse(JSON.stringify(defaultData));
```
lhchavez: this statement has no effect. can you drop it? same with L697

frontend/www/js/omegaup/arena/arena.js
```diff
@@ -571,88 +581,127 @@ export class Arena {
...
+    let self = this;
+
+    var defaultData = JSON.parse(JSON.stringify(data));
+    if (self.originalContestScoreboardEvent == null) {
+      API.Problemset.scoreboardEvents(
+                        {problemset_id: self.options.originalProblemsetId})
+          .then(function(response) {
+            self.originalContestScoreboardEvent = response.events;
+            self.virtualContestRefreshInterval = setInterval(function() {
+              self.onVirtualRankingChange(data);
```
lhchavez: is it really needed to invoke this on a timer? can you find a way to instead invoke it only when things change?

frontend/www/js/omegaup/arena/arena.js
```diff
@@ -571,88 +581,143 @@ export class Arena {
...
+        .then(function(response) {
+          // Change username to username-virtual
+          for (let evt of response.events) {
+            evt.username =
+                UI.formatString(T.virtualSuffix, {username: evt.username});
+            evt.name = UI.formatString(T.virtualSuffix, {username: evt.name});
+          }
+
+          // Merge original contest and virtual contest scoreboard events
+          response.events = response.events.concat(originalContestEvents);
```
lhchavez: do these need to be re-sorted?

SpaceWhite: Actually it is not necessary to be sorted. Because these [lines](https://github.com/omegaup/omegaup/blob/b822fee041b1bc885fae869288fbed53df7f86f6/frontend/www/js/omegaup/arena/arena.js#L783-L798) could handle semi sorted data (Sorted relative to corresponding username). But for precautions, I'll add the sorting part

SpaceWhite:
Is the virtual scoreboard ready to be deployed to omegaUp platform? I'd like to delete the experiment variable if it is ready

PR #2184, opened 2018-08-24
upgrade webpack 4

SpaceWhite:
### Description

1. Upgrade webpack from 3 to 4 and its dependencies
2. Modify optimization code in `webpack.config.js`
3. Merge all languages into the main bundle
4. Some fixes for the new split chunk model.

Fixes: #1783


SpaceWhite:
It suppose to 60% faster in webpack4 but`yarn run dev --progress` takes a lot of times when building modules. Does anyone know the reason?

webpack.config.js
```diff
@@ -62,16 +62,18 @@ module.exports = [{
...
-      name: 'omegaup',
-    }),
     new VueLoaderPlugin(),
     new ExtractTextPlugin({
       filename: 'css/dist/[name].css',
       allChunks: true,
     }),
     new RemoveSourceWebpackPlugin([omegaupStylesRegExp]),
   ],
+  optimization: {
```
pabloaguilar: Travis complains about the `UglifyJsPlugin` call later in this file and says that you should change it to `config.optimization.minimize`

PR #2273, opened 2018-12-04
#2009 Rank.Table.tpl

sarvesh211999:
Rank.table.tpl is converted into vue

Part of #2009


frontend/www/js/omegaup/components/RankTable.vue
```diff
@@ -0,0 +1,177 @@
...
+    },
+    prev_page_filter: function() {
+      if (this.filter != null)
+        return "/rank/?page=" + (this.page - 1).toString() + "&filter=" +
+               this.filter;
+      else
+        return "/rank/?page=" + (this.page - 1).toString();
+    },
+    data_filter: function() {
+      if (this.filter != null) return this.filter
```
lhchavez: What's the purpose of this? If `this.filter` is null, this method will return `undefined`, which is not very different from `null`.

frontend/www/js/omegaup/components/RankTable.vue
```diff
@@ -0,0 +1,177 @@
...
+<script>
+import {T} from '../omegaup.js';
+import UI from '../ui.js';
+import {OmegaUp} from '../omegaup.js';
+
+export default {
+  props: {
+    page: Number,
+    length: Number,
+    is_index: Boolean,
```
lhchavez: nit: `isIndex`

frontend/www/js/omegaup/components/RankTable.vue
```diff
@@ -0,0 +1,177 @@
...
+        </template>
+      </template>
+      <template v-else="">
+        <h3 class="panel-title">{{ UI.formatString(T.rankHeader,{count:length}) }}</h3>
+      </template>
+    </div>
+    <div class="panel-body no-padding">
+      <div class="table-responsive">
+        <table class="table table-striped table-hover no-margin"
+               v-bind:data-filter="data_filter"
```
lhchavez: I think all these `data-...` are no longer needed. You can consult those values from the payload instead of embedding them here.

frontend/templates/rank.table.tpl
```diff
@@ -10,60 +10,15 @@
...
+{/if}
+
+{if !isset($filter)}
+	{$filter = ''}
+{/if}
+
+<script type="text/json" id="payload">{['page' => $page, 'length' => $length, 'is_index' => $is_index, 'availableFilters' => $availableFilters, 'filter' => $filter]|json_encode}</script>
+<div id="rank-table"></div>
+<script type="text/javascript" src="{version_hash src="/js/dist/rank_table.js"}"></script>
+<!-- <script language="javascript" src="{version_hash src="/js/rank.table.js"}"></script> -->
```
lhchavez: Can you remove this?

frontend/www/js/omegaup/components/RankTable.vue
```diff
@@ -0,0 +1,177 @@
...
+          }
+        })
+        .fail(omegaup.UI.apiError);
+  },
+  methods: {
+    filterSelected: function(key) {
+      if (this.filter == key) return "selected";
+    },
+    filterChange: function() {
+      var queryParameters = {}, queryString = location.search.substring(1),
```
lhchavez: Can you preserve the comment?

frontend/www/js/omegaup/ranktable.js
```diff
@@ -0,0 +1,31 @@
...
+OmegaUp.on('ready', function() {
+  const payload = JSON.parse(document.getElementById('payload').innerText);
+  let rankTable = new Vue({
+    el: '#rank-table',
+    render: function(createElement) {
+      return createElement('rankTable', {
+        props: {
+          page: this.page,
+          length: this.length,
+          is_index: this.is_index,
```
lhchavez: nit: `isIndex`.

frontend/www/js/omegaup/components/RankTable.vue
sarvesh211999: What do you mean by preserving the comment ?

sarvesh211999:
@lhchavez Changes suggested by you are done .

frontend/www/js/omegaup/components/RankTable.vue
```diff
@@ -0,0 +1,170 @@
...
+  },
+  mounted: function() {
+    var self = this;
+    var problemsSolved = self.$el.querySelector("table");
+    var length = self.length;
+    var page = self.page;
+    var filter = self.filter;
+    var isIndex = (self.isIndex === true);
+
+    omegaup.API.User.rankByProblemsSolved(
```
lhchavez: Is it possible to invoke this API from frontend/www/js/omegaup/ranktable.js ? You can have a structure similar to

```javascript
OmegaUp.on('ready', function() {
  const payload = JSON.parse(document.getElementById('payload').innerText);
  omegaup.API.User.rankByProblemsSolved(...)
    .then(function(result) {
       // process result...
       let rankTable = new Vue({
         // pass in the processed rank as a prop
       });
    });
});
```

normally we don't allow APIs to be called from within Vue files so that they are a bit more stateless.

frontend/www/js/omegaup/components/RankTable.vue
```diff
@@ -0,0 +1,170 @@
...
+        {lowCount:(page-1)*length+1,highCount:page*length}) }}</h3>
+        <template v-if="page &gt; 1">
+          <a class="prev"
+                    v-bind:href="`/rank/?page=${page-1}`">{{ T.wordsPrevPage }}</a> <span class=
+                    "delimiter">|</span>
+        </template><a class="next"
+                  v-bind:href="`/rank/?page=${page+1}`">{{ T.wordsNextPage }}</a>
+        <template v-if="Object.keys(availableFilters).length &gt; 0">
+          <select class="filter"
+                    v-model="filter_key"
```
lhchavez: can this be `v-model="filter"`? That way you don't need to use `v-bind:selected="filterSelected(key)"` at all!

sarvesh211999:
Okay I will make the changes .

frontend/www/js/omegaup/components/RankTable.vue
sarvesh211999: I can do this but it is gives Vue Warning 
Avoid mutating a prop directly since the value will be overwritten whenever the parent component re-renders. Instead, use a data or computed property based on the prop's value.

sarvesh211999:
@lhchavez Review

frontend/www/js/omegaup/components/RankTable.vue
lhchavez: oh right, you need to call the prop `initialFilter` and then do:

```javascript:
data: function() {
  return {
    T: T,
    filter: this.initialFilter,
   ...
  };
},
```

similar to how it's done in https://github.com/omegaup/omegaup/blob/4bc8eaa517b8a141379bbc8d320a0f8a69a15151/frontend/www/js/omegaup/components/user/EmailEdit.vue#L36-L42

sarvesh211999:
@lhchavez Changes Done

Issue #2344, opened 2019-01-27
Search bar in rankings to search for a specific User.


akki111singh:
![screenshot_2019-01-28-01-41-33-808_com android chrome](https://user-images.githubusercontent.com/45921807/51806309-8569ee80-229e-11e9-8379-765ff5d42a0a.png)
https://omegaup.com/rank/
We can add a search bar on the top right corner so that it will be easy for the user to search his/her name in the list.

ss62171:
can you please elaborate on this??

akki111singh:
@ss62171 currently there is no search bar for a user to search his name in the rank list in page https://omegaup.com/rank/ (can't able to attach image,don't know why ?) so we can implement this.

lhchavez:
adding a way to search for a user in the ranking list sounds reasonable. but yes, please amend the original comment so that it better explains the problem at hand. a screenshot might not be enough since it's not easy to notice something that's missing.

lhchavez:
sure thing! oh and also, it's possible to view your own rank by going to your profile. granted it might not be the most discoverable way, but it's a workaround.

PR #2361, opened 2019-02-03
Added the search user feature in user ranking

akki111singh:
Added the search user feature in the ranking page.

Fixes: #2344


akki111singh:
@lhchavez can u please review it.

frontend/www/js/omegaup/components/RankTable.vue
```diff
@@ -2,6 +2,21 @@
   <div class="panel panel-default">
     <div class="panel-heading">
       <template v-if="!isIndex">
+        <div class="text-right">
```
lhchavez: How about you re-use the same control that's used in http://localhost:8080/contest/prueba/edit/ ? that way you can also use an AJAX call instead of a POST, which should avoid all of the PHP changes. That will also add support for autocomplete out of the box!

akki111singh: Can u please explain in detail?

lhchavez: here's the source code for that: https://github.com/omegaup/omegaup/blob/d2180611dad2a5c755b6ad5826d10d8ca9ba9f3e/frontend/www/js/omegaup/components/contest/Contestant.vue#L8

frontend/server/libs/dao/Users.dao.php
```diff
@@ -34,6 +34,16 @@ public static function FindByUsername($username) {
...
     }
 
+    public static function CheckUsername($username) {
+        global  $conn;
+        $sql = 'SELECT u.* FROM Users u WHERE username = ? LIMIT 1';
+        $rs = $conn->GetRow($sql, [$username]);
+        if (count($rs)==0) {
+            return null;
+        }
+        return ($username);
```
SpaceWhite: nit: remove bracket

frontend/www/rank.php
```diff
@@ -20,10 +22,20 @@
         $availableFilters['school'] = $smarty->getConfigVars('wordsFilterBySchool');
     }
 }
+if (strlen($to_search)>0) {
+    $Searched=UsersDAO::CheckUsername($to_search);
```
SpaceWhite: This operation could be very expensive since it searches for the entire database every time user provide to query string. It would be better to create a new api called `apiFindUsername` instead doing this because we migeht use this API in the futur. And also use Cache like in this code https://github.com/omegaup/omegaup/blob/dcc994ff1549851498b48ab205af95678cdb847c/frontend/server/controllers/ContestController.php#L69. By using cache we could reduce sql query cost

lhchavez: no need for a cache: we already check for usernames in lots of places, since it's indexed and pretty inexpensive.

using the autocomplete component referenced in another comment of mine above is the way to go since it would avoid a lot of duplicate functionality.

akki111singh: @lhchavez i would work on this after 4-5 days as i have my exams now

lhchavez: sure thing! your exams have priority :)

thanks for letting us know so that the expectations are set accordingly.

akki111singh: @lhchavez can u please review it.

frontend/templates/searched_user.tpl
```diff
@@ -0,0 +1,4 @@
+<div id="content">
```
lhchavez: Can you remove this file?

frontend/www/js/omegaup/components/RankTable.vue
```diff
@@ -97,9 +111,14 @@ export default {
     resultTotal: Number,
   },
   data: function() {
-    return { T: T, UI: UI, }
+    return { T: T, UI: UI, Searched_user: '', }
```
lhchavez: nit: all JavaScript variables should use camelCase, so this should be called `searchedUsername`

frontend/www/js/omegaup/components/RankTable.vue
```diff
@@ -2,6 +2,19 @@
...
+        <div class="text-right">
+          <form v-on:submit.prevent="onSubmit">
+            <div class="form-inline">
+              <div class="form-group">
+                <omegaup-autocomplete class="form-control"
+                     v-bind:init="el =&gt; UI.userTypeahead(el)"
+                     v-model="Searched_user"></omegaup-autocomplete>
+              </div><input class="btn btn-primary btn-lg active"
+                   type="submit"
+                   value="Search User">
```
lhchavez: Can you add a translation string for this? (in `frontend/templates/{en,es,pt}.lang`) once you modify those files, run

```bash
./stuff/hook_tools/lint.py --linters=i18n fix --all
```

to regenerate all the needed files. you can use Google Translate for the spanish and portuguese strings.

frontend/www/js/omegaup/components/RankTable.vue
```diff
@@ -97,9 +111,14 @@ export default {
     resultTotal: Number,
   },
   data: function() {
-    return { T: T, UI: UI, }
+    return { T: T, UI: UI, Searched_user: '', }
   },
   methods: {
+    onSubmit: function() {
+      var Searched_user_url = '/profile/' + this.Searched_user;
```
lhchavez: it's 2019, so we should avoid using `var`. [`const`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const) should always be preferred whenever it makes sense, and [`let`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let) when you need to reassign.

frontend/www/js/omegaup/components/RankTable.vue
```diff
@@ -97,9 +111,14 @@ export default {
...
+      window.location = Searched_user_url;
```
lhchavez: note that just blindly appending user-controlled input is dangerous: https://excess-xss.com/

you need to escape the input before adding it to a URL:

```suggestion
      window.location = `/profile/${encodeURIComponent(this.searchedUsername)}`;
```

You can read a lot more about why data validation is important and other ways it can go wrong at https://www.owasp.org/index.php/Data_Validation

frontend/templates/searched_user.tpl
akki111singh: Sure :))

frontend/www/js/omegaup/components/RankTable.vue
```diff
@@ -2,6 +2,19 @@
   <div class="panel panel-default">
     <div class="panel-heading">
       <template v-if="!isIndex">
+        <div class="text-right">
```
lhchavez: Can you make sure this looks as much as possible to http://localhost:8080/contest/mine/ ?

Here's the source code for that: https://github.com/omegaup/omegaup/blob/master/frontend/www/js/omegaup/components/contest/ContestList.vue

frontend/www/js/omegaup/components/RankTable.vue
```diff
@@ -2,6 +2,21 @@
...
+            <div class="form-inline">
+              <div class="form-group">
+                <div class="text-left">
+                  <omegaup-autocomplete class="form-control"
+                       v-bind:init="el =&gt; UI.userTypeahead(el)"
+                       v-model="searchedUsername"></omegaup-autocomplete>
+                </div>
+              </div><input class="btn btn-primary btn-lg active"
+                   type="submit"
+                   value="Search User">
```
lhchavez: You're not using the translation string. You need to use `<button...>{{ T.searchUser }}</button>`.

frontend/templates/en.lang
```diff
@@ -723,6 +723,7 @@ schoolsLearnProgBySolvProblems = "Learn programming by solving problems"
 schoolsManageProgressOfStudents = "Manage the progress of your students"
 schoolsStartUsingSchools = "Start using Schools"
 scoreboardmerge = "Merge scorboard"
+searchedUsername  = "Searched user name."
```
lhchavez: can this be

```suggestion
searchUser  = "Search user"
```

same for the other languages.

lhchavez:
you don't need to close and reopen this issue. every time you push to this branch, it will update here.

frontend/www/js/omegaup/components/RankTable.vue
lhchavez: this is still pending.

akki111singh:
> you don't need to close and reopen this issue. every time you push to this branch, it will update here.

Yes i know that it will be update here but don't know due to which reason the build is failing i.e why i have done this.

frontend/www/js/omegaup/components/RankTable.vue
akki111singh: Can u explain clearly what need to be done here!!

lhchavez:
> > you don't need to close and reopen this issue. every time you push to this branch, it will update here.
> 
> Yes i know that it will be update here but don't know due to which reason the build is failing i.e why i have done this.

If the Travis link (e.g. https://travis-ci.org/omegaup/omegaup/builds/491355455) says "job errored", it's not your fault. on the other hand, if it says failed, it _might_ be your fault. you can see the logs on Travis as well: https://travis-ci.org/omegaup/omegaup/jobs/491355458 . please avoid closing the issue unnecessarily.

akki111singh:
> > > you don't need to close and reopen this issue. every time you push to this branch, it will update here.
> > 
> > 
> > Yes i know that it will be update here but don't know due to which reason the build is failing i.e why i have done this.
> 
> If the Travis link (e.g. https://travis-ci.org/omegaup/omegaup/builds/491355455) says "job errored", it's not your fault. on the other hand, if it says failed, it _might_ be your fault. you can see the logs on Travis as well: https://travis-ci.org/omegaup/omegaup/jobs/491355458 . please avoid closing the issue unnecessarily.

Ok got it! Thanks :)

frontend/www/js/omegaup/components/RankTable.vue
lhchavez: make sure both the code and the rendered contents look as similar as possible as the two links provided. no right-alignment, no form elements on the header, etc.

akki111singh: @lhchavez solved

akki111singh:
@lhchavez review please! :3

frontend/www/js/omegaup/components/RankTable.vue
```diff
@@ -3,7 +3,14 @@
     <div class="panel-heading">
       <template v-if="!isIndex">
         <h3 class="panel-title">{{ UI.formatString(T.rankRangeHeader,
-        {lowCount:(page-1)*length+1,highCount:page*length}) }}</h3>
+        {lowCount:(page-1)*length+1,highCount:page*length})
```
yoniihdc: Can we include the result of this function into a [computed property](https://vuejs.org/v2/guide/computed.html) it would make it easier to maintain if we want to modify the value

yoniihdc:
Hi @akki111singh for UI changes, can you include screenshots of how the results look like, please.

akki111singh:
> Hi @akki111singh for UI changes, can you include screenshots of how the results look like, please.

/home/akhil/Videos/Peek 2019-02-12 12-05.gif

frontend/www/js/omegaup/components/RankTable.vue
akki111singh: @yoniihdc  are we supposed to change other part of code,as this is the PR for search bar only ?

yoniihdc: you're right, maybe we can create another one, since this is the rank table component 😄

yoniihdc:
> > Hi @akki111singh for UI changes, can you include screenshots of how the results look like, please.
> > peek 2019-02-12 12-05
> > Above is the attached gif of current look. ;3

Can we use the button with the same color as we have in the arena, just to keep the same pattern design in the app.
imagen

akki111singh:
screenshot from 2019-02-13 00-14-19
Is it ok.

frontend/www/js/omegaup/components/RankTable.vue
lhchavez: whatever happened to making it look and feel similar to http://localhost:8080/contest/mine/ and https://github.com/omegaup/omegaup/blob/master/frontend/www/js/omegaup/components/contest/ContestList.vue ?

it doesn't matter if it doesn't look like the other link in arena that @yoniihdc shared, that one's using the wrong layout :( (maybe we need to update that one too)

frontend/www/js/omegaup/components/RankTable.vue
```diff
@@ -3,7 +3,14 @@
...
+        }}</h3><label><omegaup-autocomplete class="form-control"
+                              v-bind:init="el =&gt; UI.userTypeahead(el)"
+                              v-model="searchedUsername"></omegaup-autocomplete></label>
+        <div class="btn-group">
+          <button type="button"
```
lhchavez: can you add `class="btn btn-primary"`?

frontend/www/js/omegaup/components/RankTable.vue
lhchavez: this is still missing :(

lhchavez: concretely i'm interested in following the same code structure of splitting the header/body/table of the panel.

akki111singh: @lhchavez are we supposed to change other part of code in this pr as this  pr is for search bar only?

lhchavez: you just need to add a couple of `div` and shuffle things around a bit.

normally code changes should be limited in scope to just one thing, but in this case, touching the immediately surrounding code to make it better is okay and probably not worth doing another PR for just 2-3 extra lines of code.

akki111singh: Right!!   :)   Can u explain what exactly need to be  done ?

lhchavez: make the code look like this: https://github.com/omegaup/omegaup/blob/4d9a6e777c3d68a2a5582ff4a3196c92d011da12/frontend/www/js/omegaup/components/contest/ContestList.vue#L3-L33

akki111singh: {lowCount:(page-1)*length+1,highCount:page*length}) }}</h3>
and for this part ?

lhchavez: leave it as is, it's part of the real header.

akki111singh: @lhchavez the filter part should be in body of header ?

akki111singh:
Final view after splitting the header/body/table of the panel.
final

frontend/www/js/omegaup/components/RankTable.vue
akki111singh: @lhchavez updated accordingly. :3

frontend/www/js/omegaup/components/RankTable.vue
```diff
@@ -29,53 +37,51 @@
...
-              <a class="prev"
-                        v-bind:href="prevPageFilter">{{ T.wordsPrevPage }}</a> <span class=
-                        "delimiter"
-                        v-show="shouldShowNextPage">|</span>
-            </template><a class="next"
-                      v-bind:href="nextPageFilter"
-                      v-show="shouldShowNextPage">{{ T.wordsNextPage }}</a>
-          </template><br>
-        </div>
+    <div class="table-responsive">
```
lhchavez: note that the code is still not similar in structure to the other snippet: this div does not exist there, and the table doesn't have as many classes.

can you make sure this is as close as possible to the other code?

frontend/www/js/omegaup/components/RankTable.vue
```diff
@@ -1,9 +1,17 @@
 <template>
   <div class="panel panel-default">
     <div class="panel-heading">
+      <h3 class="panel-title">{{ UI.formatString(T.rankRangeHeader,
```
lhchavez: can you add the template with the `v-if="isIndex"` to keep the original functionality (copy L14 to here)? also move the block in L36-39 to here.

frontend/www/js/omegaup/components/RankTable.vue
```diff
@@ -1,9 +1,17 @@
...
+      {lowCount:(page-1)*length+1,highCount:page*length}) }}</h3>
+    </div>
+    <div class="panel-body no-padding">
```
lhchavez: since we don't want to show this in the homepage at all, can you add the `v-if="isIndex"` here and remove the template tag in L14?

frontend/www/js/omegaup/components/RankTable.vue
akki111singh: Line 14 has vif="!isIndex" so i have to add v-if="isIndex" of v-if="!isIndex"

frontend/www/js/omegaup/components/RankTable.vue
akki111singh: v-if="isIndex" or v-if="!isIndex" and am is supposed to put it in div tag ?

lhchavez: `v-if="!isIndex"`, and yes, in the `div` tag.

frontend/www/js/omegaup/components/RankTable.vue
lhchavez: since you need to add the `v-else` too, structure it so that you put the `v-if="isIndex"` first and then the `v-else`.

akki111singh:
@lhchavez structured :)

Issue #2372, opened 2019-02-10
When i click on edit profile,error message is displayed at the top.


akki111singh:
When i click on the edit profile in the profile section an unnecessary error message is displayed at the top.
![image](https://user-images.githubusercontent.com/45921807/52534266-82becd00-2d65-11e9-8750-15813a30a36c.png)

lhchavez:
JavaScript console says

```
error_handler.js?ver=5dd116:61 Unhandled exception  TypeError: Cannot read property 'trim' of null
    at Object.country (iso3166.min.js?ver=2c7e40:9)
    at HTMLSelectElement.<anonymous> (user.edit.js?ver=46ca98:20)
    at HTMLSelectElement.dispatch (jquery-3.2.1.js:8770)
    at HTMLSelectElement.q.handle (jquery-3.2.1.js:8484)
    at Object.trigger (jquery-3.2.1.js:10252)
    at HTMLSelectElement.<anonymous> (jquery-3.2.1.js:10252)
    at Function.each (jquery-3.2.1.js:4343)
    at e.fn.init.each (jquery-3.2.1.js:4090)
    at e.fn.init.trigger (jquery-3.2.1.js:10252)
    at user.edit.js?ver=46ca98:59
```

so we just need to check if the country is not null and avoid this call: https://github.com/omegaup/omegaup/blob/78ec494e522993d543cf931d598190079c45f173/frontend/www/js/user.edit.js#L20

akki111singh:
@lhchavez i would like to do it.

lhchavez:
assigned!

PR #2374, opened 2019-02-10
Edit profile error when country is not mentioned

akki111singh:
Fixes: #2372

### DIFF
--- a/frontend/www/js/user.edit.js
+++ b/frontend/www/js/user.edit.js
@@ -17,11 +17,7 @@ omegaup.OmegaUp.on('ready', function() {
         $('#state_id option').remove();
         $('#state_id').val('');
 
-        try {
-          var country = iso3166.country($('#country_id').val());
-        } catch (err) {
-          country = '';
-        }
+        var country = iso3166.country($('#country_id').val() || '');
 
         if (!country.sub) {
           $('#state_id').attr('disabled', 'disabled');
--- a/frontend/www/js/user.edit.js
+++ b/frontend/www/js/user.edit.js
@@ -17,7 +17,12 @@ omegaup.OmegaUp.on('ready', function() {
         $('#state_id option').remove();
         $('#state_id').val('');
 
-        var country = iso3166.country($('#country_id').val());
+        try {
+          var country = iso3166.country($('#country_id').val());
+        } catch (err) {
+          country = '';
+        }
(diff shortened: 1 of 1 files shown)


frontend/www/js/user.edit.js
```diff
@@ -17,7 +17,12 @@ omegaup.OmegaUp.on('ready', function() {
         $('#state_id option').remove();
         $('#state_id').val('');
 
-        var country = iso3166.country($('#country_id').val());
+        try {
```
yoniihdc: you can use [short-circuit evaliation](https://en.wikipedia.org/wiki/Short-circuit_evaluation) to avoid try/catch.

```
var country = iso3166.country($('#country_id').val()) || '';
```

akki111singh: var country = iso3166.country($('#country_id').val()) || ''; 
Not working :(

akki111singh:
Please review :))

frontend/www/js/user.edit.js
lhchavez: please elaborate. "not working" does not allow you yo help resolve the issue.

frontend/www/js/user.edit.js
```diff
@@ -17,7 +17,12 @@ omegaup.OmegaUp.on('ready', function() {
...
+          var country = iso3166.country($('#country_id').val());
+        } catch (err) {
+          country = '';
```
lhchavez: note that using exceptions as control flow (i.e. as replacement for `if`/`else`) is an anti-pattern: http://wiki.c2.com/?DontUseExceptionsForFlowControl

you already know what the cause of the exception is (the country being `null`) so make an explicit check for that:

```
var country = '';
var countryId = $('country_id').val();
if (countryId) {
  ...
}
```

frontend/www/js/user.edit.js
akki111singh: Still giving the same error as before  TypeError: Cannot read property 'trim' of null
    at Object.country (iso3166.min.js?ver=2c7e40:9)

lhchavez: oh that might be because the short-circuit evaluation should be within the call to `.country()`:

```suggestion
        var country = iso3166.country($('#country_id').val() || '');
```

PR #4379, opened 2020-07-14
Renaming 'Problema subido por' to 'Subido por'

FishyFeeshy:
# Descripción

Fixed the readings source to be consistent.

Changed:
`Problema subido por` to `Subido por`
`Problema proposto por` to `Proposto por`

Fixes: omegaup#4310


frontend/tests/ui/test_group.py
```diff
@@ -66,21 +66,6 @@ def test_create_group_with_identities_and_restrictions(driver):
...
-                    'li.nav-contests .dropdown-menu')))
-        for present_href in ['/arena/']:
-            assert contests_dropdown.find_elements_by_css_selector(
-                'a[href="%s"]' % present_href), (
-                    '%s item is not present!' % present_href)
-        for absent_href in ['/contest/new/', '/scoreboardmerge/']:
-            assert not contests_dropdown.find_elements_by_css_selector(
-                'a[href="%s"]' % absent_href), (
-                    '%s item is visible!' % absent_href)
-
```
lhchavez, line 83: why is this change here? in order to revert it, try running

```shell
git restore --source=upstream/master frontend/tests/ui/test_group.py
```

this translates to "restore the version of the file `frontend/tests/ui/test_group.py` as it was in the `upstream/master` branch".

frontend/server/src/Controllers/README.md
```diff
@@ -230,7 +230,8 @@
 
 ### Description
 
-Get stats for an overall platform report.
+- Get stats for an overall platform report.
```
lhchavez, line 233: if you run `composer install` inside the container / VM, and re-run the linter, it should get rid of all these changes.

frontend/tests/ui/test_group.py
FishyFeeshy: That command worked. It should be good now.

frontend/server/src/Controllers/README.md
FishyFeeshy: I ran the linter, but I was prompted to re-run it when I tried to push my changes, and that fixed it.

lhchavez:
all right, the only thing missing is removing the `stuff/hook_tools` change from this PR. the folks in slack can help with that if the command i shared above did not work.

lhchavez:
The [`stuff/hook_tools`](https://github.com/omegaup/omegaup/pull/4379/files#diff-68ee8676648dc5d9a64cab16f5b49582) file is still there :( ask around in slack for help with this.

lhchavez:
yay! thanks :D

PR #4444, opened 2020-07-31
Student Progress Bar

FishyFeeshy:
# Descripción

Adds a feature on the students' progress page that creates a student progress bar. The progress bar consists of several boxes for each problem. The color of the box is based on the score the student received, and the score can be viewed by hovering over the box. 

image

# Comentarios

CC: @carlosabcs 

# Checklist:

      omegaUp.


frontend/www/js/omegaup/components/course/StudentProgress.test.ts
```diff
@@ -0,0 +1,46 @@
...
+
+describe('StudentProgress.vue', () => {
+  it('Should handle scores', async () => {
+    const expectedDate = new Date('1/1/2020, 12:00:00 AM');
+    const wrapper = shallowMount(course_StudentProgress, {
+      propsData: {
+        course: {
+          alias: 'hello',
+        },
+        assignment: [
```
lhchavez, line 17: would it be possible to put the type annotation to all entries directly under `propsData`?

```suggestion
        assignment: <omegaup.Assignment[]>[
```

the types should match the ones declared in `StudentProgress.vue`. same for `course` and `student`.

frontend/www/js/omegaup/components/course/StudentProgress.vue
```diff
@@ -0,0 +1,103 @@
...
+<style>
+.box {
+  width: 20px;
+  height: 20px;
+  border: 1px solid black;
+  margin: -0.5px;
+}
+
+.bg-green {
+  background: rgb(53, 184, 53);
```
lhchavez, line 38: Can these colors be declared in [`frontend/www/sass/base/_colors.scss`](https://github.com/omegaup/omegaup/blob/master/frontend/www/sass/base/_colors.scss)? 
https://github.com/omegaup/omegaup/pull/4424/files#diff-9565d87813be34c611d8564cf7e6e539R67 has an example of how to do that.

frontend/www/js/omegaup/components/course/StudentProgress.vue
```diff
@@ -0,0 +1,103 @@
...
+@Component
+export default class StudentProgress extends Vue {
+  @Prop() course!: types.CourseDetails;
+  @Prop() student!: types.StudentProgress;
+  @Prop() assignments!: omegaup.Assignment[];
+
+  T = T;
+
+  problemScore(problem: number): string {
+    return Math.round(problem) + '%';
```
lhchavez, line 69: nit: the use of [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) is preferred over concatenation with `+`:

```suggestion
    return `{Math.round(problem)}%`;
```

(although this was suggested to be inlined above, so this method can be removed).

frontend/www/js/omegaup/components/course/AssignmentDetails.vue
```diff
@@ -222,7 +222,6 @@ import T from '../../lang';
 import course_ProblemList from './ProblemList.vue';
 import course_ScheduledProblemList from './ScheduledProblemList.vue';
 import DateTimePicker from '../DateTimePicker.vue';
-
```
carlosabcs, line 225: Try to restore this, and all the lines in all the files, that have been removed. Probably was your editor that automatically did that deletion.

frontend/www/js/omegaup/components/course/StudentProgress.vue
```diff
@@ -0,0 +1,103 @@
...
+      <div class="d-flex justify-content-center">
+        <div v-if="student.progress.hasOwnProperty(assignment.alias) == false">
+          {{ T.wordsProblemsUnsolved }}
+        </div>
+        <div
+          v-for="problem in student.progress[assignment.alias]"
+          v-bind:class="getProblemColor(Math.round(problem))"
+          data-toggle="tooltip"
+          data-placement="bottom"
+          v-bind:title="problemScore(problem)"
```
lhchavez, line 22:
```suggestion
          v-bind:title="`{Math.round(problemScore)}%`"
```

this expression is simple enough that it's probably better to inline it.

frontend/www/js/omegaup/components/course/StudentProgress.vue
```diff
@@ -0,0 +1,103 @@
...
+      class="score flex-column justify-content-center align-items-center"
+      v-for="assignment in assignments"
+    >
+      <p class="mb-1 text-center">{{ Math.round(score(assignment)) }}%</p>
+      <div class="d-flex justify-content-center">
+        <div v-if="student.progress.hasOwnProperty(assignment.alias) == false">
+          {{ T.wordsProblemsUnsolved }}
+        </div>
+        <div
+          v-for="problem in student.progress[assignment.alias]"
```
lhchavez, line 18: nit:

```suggestion
          v-for="problemScore in student.progress[assignment.alias]"
```

given that this is not a problem, but the problem's score, it might be better to call this `problemScore`.

frontend/www/js/omegaup/components/course/StudentProgress.test.ts
```diff
@@ -0,0 +1,46 @@
...
+import expect from 'expect';
+import Vue from 'vue';
+import T from '../../lang';
+import { omegaup } from '../../omegaup';
+
+import course_StudentProgress from './StudentProgress.vue';
+
+describe('StudentProgress.vue', () => {
+  it('Should handle scores', async () => {
+    const expectedDate = new Date('1/1/2020, 12:00:00 AM');
```
carlosabcs, line 11: This variable must be deleted as you're not using it.

frontend/www/js/omegaup/components/course/StudentProgress.test.ts
```diff
@@ -0,0 +1,46 @@
...
+                problem: 44,
+                problem2: 55,
+              },
+            },
+          ],
+          username: 'student',
+        },
+      },
+    });
+    expect(wrapper.find('div.bg-red'));
```
carlosabcs, line 44: Try adding another test to check if the value `44` or `55` appears in the component as you expect.

frontend/www/js/omegaup/components/course/StudentProgress.vue
```diff
@@ -0,0 +1,103 @@
...
+    </td>
+  </tr>
+</template>
+
+<style>
+.box {
+  width: 20px;
+  height: 20px;
+  border: 1px solid black;
+  margin: -0.5px;
```
carlosabcs, line 34: Is this  `margin` really necessary?

frontend/www/js/omegaup/components/course/StudentProgress.vue
```diff
@@ -0,0 +1,103 @@
...
+    } else if (problemScore >= 50) {
+      return 'box bg-yellow';
+    } else if (problemScore > 0) {
+      return 'box bg-red';
+    } else {
+      return 'box bg-black';
+    }
+  }
+
+  studentProgressUrl(): string {
```
carlosabcs, line 99: This should be a computed property, instead of a method.
```typescript
get studentProgressUrl(): string {
...
}
```

frontend/www/js/omegaup/components/problem/List.vue
```diff
@@ -260,6 +260,7 @@ import {
   faExclamationTriangle,
   faBan,
 } from '@fortawesome/free-solid-svg-icons';
+import StudentProgress from '../course/StudentProgress.vue';
```
lhchavez, line 263:
```suggestion
```

this is not being used here, so it can be safely removed.

frontend/www/js/omegaup/components/course/StudentProgress.vue
```diff
@@ -0,0 +1,103 @@
...
+      ) /
+        (Object.values(this.student.progress[assignment.alias]).length * 100)) *
+      100
+    );
+  }
+
+  getProblemColor(problemScore: number): String {
+    if (problemScore > 70) {
+      return 'box bg-green';
+    } else if (problemScore >= 50) {
```
carlosabcs, line 90: What is the best option here to avoid those `else if` @lhchavez ?

frontend/www/js/omegaup/course/students.ts
```diff
@@ -1,14 +1,17 @@
 import course_ViewProgress from '../components/course/ViewProgress.vue';
+import course_StudentProgress from '../components/course/StudentProgress.vue';
 import { OmegaUp } from '../omegaup';
 import { types } from '../api_types';
 import * as api from '../api';
 import * as ui from '../ui';
 import T from '../lang';
 import Vue from 'vue';
+import { data } from 'jquery';
```
lhchavez, line 9:
```suggestion
```

would it be possible to avoid using jquery? We (and a large part of the rest of the internet) have been slowly trying to remove it everywhere and wouldn't want to see it be reintroduced.

frontend/www/js/omegaup/course/students.ts
```diff
@@ -1,14 +1,17 @@
 import course_ViewProgress from '../components/course/ViewProgress.vue';
+import course_StudentProgress from '../components/course/StudentProgress.vue';
```
carlosabcs, line 2: Here you don't need to import the component, so you could delete this line.

frontend/www/js/omegaup/course/students.ts
carlosabcs, line 9: Why are you importing this?

frontend/www/js/omegaup/course/students.ts
```diff
@@ -1,14 +1,17 @@
...
 import * as api from '../api';
 import * as ui from '../ui';
 import T from '../lang';
 import Vue from 'vue';
+import { data } from 'jquery';
 
 OmegaUp.on('ready', function () {
   const payload = types.payloadParsers.StudentsProgressPayload();
 
+  //what to do for student progress component?
```
carlosabcs, line 14: You don't need to do anything here with that component :)

frontend/www/js/omegaup/components/course/AssignmentDetails.vue
lhchavez, line 225: would it be possible to avoid these whitespace changes?

These can be safely reverted with

```shell
git restore --version=upstream/master frontend/www/js/omegaup/components/course/AssignmentDetails.vue
```

same with all the other files with only whitespace changes.

frontend/www/js/omegaup/components/course/Form.test.ts
```diff
@@ -12,12 +12,12 @@ const baseCourseFormProps = {
     admission_mode: 'registration',
     alias: 'Newx',
     assignments: [],
-    needs_basic_information: false,
     description: 'New',
     finish_time: new Date(),
     is_admin: true,
     is_curator: true,
     name: 'Nuevo',
+    needs_basic_information: false,
```
lhchavez, line 20: is this change really needed? if not, can this be reverted too?

frontend/www/js/omegaup/components/course/StudentProgress.test.ts
lhchavez, line 44: what is the expected list of boxes? would it be possible to make a stronger assertion, like checking that all of the boxes exist, are in the right order, and have the expected percentage?

frontend/www/js/omegaup/course/students.ts
lhchavez, line 14: @tvanessa @heduenas can answer this probably.

frontend/www/js/omegaup/components/course/StudentProgress.vue
lhchavez: something like this:

```typescript
if (problemScore >= 70) return 'box bg-green';
if (problemScore >= 50) return 'box bg-yellow';
...
return 'box bg-black';
```

frontend/www/js/omegaup/components/course/StudentProgress.vue
FishyFeeshy: If I don't include it, the borders are next to each other instead of shared. Is that ok?
image

lhchavez: if the borders are made a bit less dark, AND the container also has a 1px border, it would make all the boxes have a consistent 2px border. 

This is how the demo in the Bootstrap docs for flex does it, and looks alright: https://getbootstrap.com/docs/4.5/utilities/flex/#direction

lhchavez:
oh forgot to ask one more thing! is there a way in which this can be displayed in a way that is friendly for people with [color vision deficiency](https://en.wikipedia.org/wiki/Color_blindness)?

maybe putting the number in the box with enough contrast suffices. (@carlosabcs @tvanessa for second opinion)

frontend/www/js/omegaup/components/course/StudentProgress.vue
FishyFeeshy: This should be better. 
image

frontend/www/js/omegaup/components/course/StudentProgress.vue
FishyFeeshy: Should this be changed to `return /course/${this.course.alias}/student/${this.student.name || this.student.username}/;`, in case the student doesn't have a name?

lhchavez: oh! we all missed this. it should be

```typescript
    return `/course/${this.course.alias}/student/${this.student.username}/`;
```

URLs should never use names: instead they should use usernames or aliases (or IDs in very rare cases, like for Schools).

frontend/www/js/omegaup/components/course/StudentProgress.vue
```diff
@@ -0,0 +1,97 @@
...
+        <div v-if="student.progress.hasOwnProperty(assignment.alias) == false">
+          {{ T.wordsProblemsUnsolved }}
+        </div>
+        <div v-else class="d-flex border border-dark">
+          <div
+            v-for="problemScore in student.progress[assignment.alias]"
+            v-bind:class="getProblemColor(Math.round(problemScore))"
+            data-toggle="tooltip"
+            data-placement="bottom"
+            v-bind:title="Math.round(problemScore) + '%'"
```
lhchavez, line 23: nit: the use of [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) is preferred over concatenation with `+`:

```suggestion
            v-bind:title="`{Math.round(problemScore)}%`"
```

FishyFeeshy: When I do that, I get this:
image
Am I doing that wrong? I thought you couldn't do [template literals inside attributes](https://vuejs.org/v2/guide/syntax.html#Attributes)

carlosabcs: Change that to:
```js
v-bind:title="`${Math.round(problemScore)}%`"
```
and it'll work

lhchavez: errr. yes, sorry, i missed a `$` ^^;;

lhchavez:
@carlosabcs any final comments?

@FishyFeeshy there's one conflicting file. you might need to do `git pull upstream master` and resolve the conflicts before reuploading.

frontend/www/js/omegaup/components/course/StudentProgress.vue
```diff
@@ -0,0 +1,97 @@
...
+        {{ student.name || student.username }}
+      </a>
+    </td>
+    <td
+      class="score flex-column justify-content-center align-items-center"
+      v-for="assignment in assignments"
+    >
+      <p class="mb-1 text-center">{{ Math.round(score(assignment)) }}%</p>
+      <div class="d-flex justify-content-center">
+        <div v-if="student.progress.hasOwnProperty(assignment.alias) == false">
```
carlosabcs, line 14: This could be modified to: `v-if="!student.progress.hasOwnProperty(assignment.alias)"`

PR #4548, opened 2020-08-24
Course statistics page

FishyFeeshy:
# Descripción
-Frontend portion of Course Statistics page
Creates a statistics page (accessible under the statistics button on the course details page) that displays a graph of general statistics on progress for all students in the course. Each problem is represented by a bar on the graph. There currently are 6 stats to choose from. 
image

# Comentarios

CC: @carlosabcs

# Checklist:

      omegaUp.


frontend/server/src/DAO/Assignments.php
```diff
@@ -44,73 +44,83 @@ public static function getProblemset(
...
         int $courseId,
         int $groupId
     ): array {
         $sql = '
+        SELECT
+            bpr.assignment_alias,
+            bpr.problem_alias,
+            VARIANCE(bpr.max_user_score_for_problem) AS variance,
+            AVG(bpr.max_user_score_for_problem) AS average,
+            COUNT(
```
heduenas, line 59: Can you change this and the next metric to percentages? You can do it with:
`AVG(
                CASE WHEN bpr.max_user_percent_for_problem > 0.6 THEN 1 ELSE 0 END
            )`

frontend/templates/en.lang
```diff
@@ -1331,6 +1331,8 @@ wordsAssignment = "Assignment"
 wordsAssignments = "Assignments"
 wordsAuthor = "Author"
 wordsAutoAccepted = "User has been explicitly invited to the contest"
+wordsAverage = "Average"
+wordsAverageScore = "Average score"
```
lhchavez, line 1335: given that all of these translation strings are related to course statistics, would it be possible to have them all start with `courseStatistics` instead of `words`?

The proliferation of `words` as a prefix is something we kind of regret u_u

frontend/www/js/omegaup/components/course/Statistics.vue
```diff
@@ -0,0 +1,64 @@
+<template>
+  <div class="container-fluid">
+    <h2 class="text-center">
+      <a v-bind:href="courseUrl">{{ course.name }}</a>
```
lhchavez, line 4: nit:

```suggestion
      <a v-bind:href="`/course/${course.alias}/`">{{ course.name }}</a>
```

it might be better to inline this, since it's only used once.

frontend/www/js/omegaup/components/course/Statistics.vue
```diff
@@ -0,0 +1,64 @@
...
+          text: `${T.wordsStudentsAbove} 60%`,
+        },
+        {
+          value: this.lowScoreChartOptions,
+          text: `${T.wordsStudentsScored} 0%`,
+        },
+        { value: this.minimumChartOptions, text: T.wordsMinimumScore },
+        { value: this.maximumChartOptions, text: T.wordsMaximumScore },
+      ],
+    };
```
lhchavez, line 61: these are normally implemented as members of the class: https://class-component.vuejs.org/guide/class-component.html#data

so these can be moved together with the `T = T;` declaration. Here's an example: https://github.com/omegaup/omegaup/blob/b1660fe166a2083332e88bfe036ba0ad6da68458/frontend/www/js/omegaup/components/course/Clone.vue#L77-L79

frontend/www/js/omegaup/components/course/Statistics.vue
```diff
@@ -0,0 +1,64 @@
...
+  }
+  data() {
+    return {
+      selected: this.varianceChartOptions,
+      options: [
+        { value: this.varianceChartOptions, text: T.wordsVariance },
+        { value: this.averageChartOptions, text: T.wordsAverageScore },
+        {
+          value: this.highScoreChartOptions,
+          text: `${T.wordsStudentsAbove} 60%`,
```
lhchavez, line 52: We try to avoid concatenating translation strings whenever possible, since not all languages have the same word order (e.g. Arabic and Hebrew are written [right to left](https://en.wikipedia.org/wiki/Right-to-left)).

Would it be possible to move the "60%" to the translation string? Same with the "0%" below.

frontend/www/js/omegaup/course/statistics.ts
```diff
@@ -0,0 +1,310 @@
...
+    return problems;
+  };
+
+  const getVariance = (stats: types.CourseStatisticsPayload) => {
+    let variance = [];
+    for (const problem in stats.problemStats) {
+      variance.push(stats.problemStats[problem].variance || 0);
+    }
+    return variance;
+  };
```
lhchavez, line 29: there's a lot of repetition in these functions :S another way to express them would be:

```typescript
  getStatistic(name: 'variance'|'average'|...) {
    return this.problemStats.map((problem) => problem[name] || 0);
  };
```

and then the other getters can be written as

```typescript
  get variance() { return this.getStatistic('variance'); }
```

now, the type of `this.problemStats` should be `types.CourseProblemStatistics`, which needs to be created from PHP.

frontend/www/js/omegaup/course/statistics.ts
```diff
@@ -0,0 +1,310 @@
...
+            },
+          },
+        },
+        series: [
+          {
+            name: T.wordsVariance,
+            data: getVariance(payload),
+          },
+        ],
+      },
```
lhchavez, line 139: can all of these options be computed in the .vue file? Vue is a [Model-View-ViewModel](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93viewmodel) implementation, so the general idea is that the .vue file should be fully responsible for making all the conversions needed to display stuff.

Another way to think of this is that the .ts file is _only_ responsible for moving things around between PHP and the .vue files, and coordination among multiple .vue files if needed (although that might be better served by [Vuex](https://vuex.vuejs.org/)).

frontend/www/js/omegaup/course/statistics.ts
```diff
@@ -0,0 +1,310 @@
...
+import * as api from '../api';
+import * as ui from '../ui';
+import T from '../lang';
+import Vue from 'vue';
+import { numberFormat } from 'highcharts';
+
+OmegaUp.on('ready', function () {
+  const payload = types.payloadParsers.CourseStatisticsPayload();
+
+  const getProblems = (stats: types.CourseStatisticsPayload) => {
```
lhchavez, line 13: (read comment in L139 first)

when all these functions are moved to the .vue file, they should all be rewritten as [computed properties](https://vuejs.org/v2/guide/computed.html) (implemented in TypeScript as [`get`ters](https://class-component.vuejs.org/guide/class-component.html#computed-properties)).

frontend/www/js/omegaup/course/statistics.ts
```diff
@@ -0,0 +1,310 @@
...
+          },
+        },
+        series: [
+          {
+            name: T.wordsVariance,
+            data: getVariance(payload),
+          },
+        ],
+      },
+      averageChartOptions: {
```
lhchavez, line 140: most of these options look very similar to each other. Can there be a function that generates them all?

frontend/www/js/omegaup/components/course/Statistics.vue
```diff
@@ -0,0 +1,197 @@
...
+      this.maxPoints,
+      this.problems,
+    );
+  }
+  //helper functions
+  get problems() {
+    let problems = [];
+    for (const problem in this.problemStats) {
+      problems.push(
+        `${this.problemStats[problem].assignment_alias} - ${this.problemStats[problem].problem_alias}`,
```
lhchavez, line 118: the use of [`for...of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) is _almost_ always preferred over [`for...in`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in), since the latter has some [well-documented downsides](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#Why_Use_for...in) (and produces smaller code to boot!):

```suggestion
    for (const problem of this.problemStats) {
      problems.push(
        `${problem.assignment_alias} - ${problem.problem_alias}`,
```

same in all the other `for` loops in this change (even the one in L134).

since the body of the loop is simple enough, it could also be rewritten as:

```typescript
    return this.problemStats.map((problem) =>
        `${problem.assignment_alias} - ${problem.problem_alias}`);
```

frontend/www/js/omegaup/components/course/Statistics.vue
```diff
@@ -0,0 +1,197 @@
...
+      '{y}',
+      T.wordsScore,
+      this.getStatistic('maximum'),
+      this.maxPoints,
+      this.problems,
+    );
+  }
+  //helper functions
+  get problems() {
+    let problems = [];
```
lhchavez, line 115: minor nit: whenever a variable won't be reassigned, the use of [`const`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const) is preferred over `let`, since that more clearly marks the variable as holding the same value throughout the function:

```suggestion
    const problems = [];
```

frontend/www/js/omegaup/components/course/Statistics.vue
```diff
@@ -0,0 +1,192 @@
...
+      if (problem.maxPoints > maxPoints) maxPoints = problem.maxPoints;
+    }
+    return maxPoints;
+  }
+  get maxVariance() {
+    let maxVariance = 0;
+    const variance = this.getStatistic('variance');
+    for (let i = 0; i < variance.length; i++) {
+      if (variance[i] > maxVariance) maxVariance = variance[i];
+    }
```
lhchavez, line 131:
```suggestion
    for (const variance of this.getStatistic('variance')) {
      if (variance > maxVariance) maxVariance = variance;
    }
```

this can also use `for...of`!

PR #4556, opened 2020-08-25
Course Statistics Backend

FishyFeeshy:
# Descripción

Creates the backend for a course statistics page. Gets 6 different metrics on students' progress for each problem within the course to display:
- Variance
- Average score
- Percent of students scored above 60%
- Percent of students scored at 0%
- Minimum score
- Maximum score 

Frontend: #4548

# Comentarios

CC: @heduenas 

# Checklist:

      omegaUp.


frontend/server/src/DAO/Assignments.php
```diff
@@ -44,73 +44,83 @@ public static function getProblemset(
...
     ): array {
         $sql = '
+        SELECT
+            bpr.assignment_alias,
+            bpr.problem_alias,
+            VARIANCE(bpr.max_user_score_for_problem) AS variance,
+            AVG(bpr.max_user_score_for_problem) AS average,
+            AVG(
+                CASE WHEN bpr.max_user_percent_for_problem > 0.6 THEN 1 ELSE 0 END
+            ) AS highScoreCount,
```
lhchavez, line 61: nit:

```suggestion
            AVG(
                CASE WHEN bpr.max_user_percent_for_problem > 0.6 THEN 100 ELSE 0 END
            ) AS high_score_percentage,
```

all names in MySQL are using `snake_case` naming convention instead of `camelCase`. same in all other places where there is a `camelCase` name being assigned.

also, the word `count` confused me a bit since i thought it was going to be how many problems had a high score, but it is used as a percentage of problems that have a high score. therefore, can this be called `percentage` and be in the range [0, 100] for consistency with its intended use? Same in L64.

PR #4609, opened 2020-09-03
Course verdict distribution query 

FishyFeeshy:
# Descripción

Gets verdicts per problem within a course. Modified assignment statistics query to get average submissions.

# Comentarios

# Checklist:

      omegaUp.


frontend/server/src/DAO/Assignments.php
```diff
@@ -120,7 +122,83 @@ public static function getAssignmentsProblemsStatistics(
...
+     * Returns each problem with the count of each verdict
+     *
+     * @return list<array{assignment_alias: string, problem_alias: string, runs: int, verdict: string}>
+     */
+    public static function getAssignmentVerdictDistribution(
+        int $courseId,
+        int $groupId
+    ): array {
+        $sql = '
+        SELECT
```
lhchavez, line 143: since the outer SELECT does not require any grouping/ordering that the inner SELECT cannot do (like the one in the query above), would it be possible to remove the outer SELECT and just leave the inner one?

FishyFeeshy: I will need to make some adjustments to the frontend to account for the change, but it should create the same outcome.

lhchavez: huh, what kind of adjustments? I wouldn't have expected any if both queries produce the same results.

And now that I notice, this code is not being tested! Can there be a unit test? `frontend/tests/controllers/CourseUsersTest.php` seems like a good place to do it. The `\OmegaUp\Controllers\Course::getCourseStatisticsForSmarty()` function is the one that can be invoked from that test.

frontend/server/src/DAO/Assignments.php
```diff
@@ -120,7 +122,74 @@ public static function getAssignmentsProblemsStatistics(
...
+            pr.assignment_alias,
+            pr.problem_alias,
+            pr.problem_id,
+            pr.order,
```
lhchavez, line 147:
```suggestion
```

what happens if this is removed (the `problem_id` is fine to keep)? If MySQL does not like that, can it be removed from the array afterwards so that it never gets shipped to the end user? https://github.com/omegaup/omegaup/blob/e6904d86e63f70dee4be7ad96424458322ab39db/frontend/server/src/DAO/Courses.php#L82-L92 shows one possible way to achieve that.

frontend/tests/controllers/AssignmentProblemsTest.php
```diff
@@ -460,5 +467,124 @@ public function testAssignmentProblemsStatistics() {
...
+        \OmegaUp\Test\Factories\Run::gradeRun($runData, 0, 'WA');
+
+        $results = \OmegaUp\DAO\Assignments::getAssignmentVerdictDistribution(
+            $courseData['course']->course_id,
+            $courseData['course']->group_id
+        );
+
+        $this->assertEquals($assignmentAlias, $results[0]['assignment_alias']);
+
+        // Variance of the third problem should be 0, no user did anything
```
lhchavez, line 579: nit:

```suggestion
```

this comment no longer corresponds to the code.

frontend/tests/controllers/AssignmentProblemsTest.php
```diff
@@ -460,5 +467,124 @@ public function testAssignmentProblemsStatistics() {
...
+        $this->assertEquals('AC', $results[0]['verdict']);
```
lhchavez, line 580: can all of these assertions become one? sort of

```php
$this->assertEqualsCanonicalizing(
  $results,
  [
    [
      'verdict' => 'AC',
      'runs' => ...,
      'problem_id' => $problemsData[0]['problem']->problem_id,
      'problem_alias' => $problemsData[0]['problem']->alias,
      'assignment_alias' => $assignmentAlias,
    ],
    ...
  ]
),
```

The docs for [`assertEqualsCanonicalizing()`](https://phpunit.readthedocs.io/en/9.2/assertions.html#assertequalscanonicalizing).

lhchavez (commented):
just two minor things and we're good to go!

lhchavez (approved):
let's commit this as-is and revisit if there are problems later.

thanks for the changes!

PR #5597, opened 2021-08-10
Fix typos in Spanish translations in the same string that were in the…

erwaen:
Fix typos in Spanish translations in the same string that were in the files:
- es.lang
- lang.es.json
- lang.es.ts

cambie la palabra "apirantes" a "candidatos".


erwaen:
Ahora lo cambio

lhchavez (approved):
🎉

PR #6431, opened 2022-03-17
fix: redirect to homepage after login

denllx:
# Description

Redirect the login page to homepage after user is logged in.

Fixes: #6293 

# Comments

When a login page is loaded, read the user's login status first. The login status is stored in `CommonPayload`.

# Check list:

- [x] The code follows the [guidance style](https://github.com/omegaup/omegaup/wiki/Coding-guidelines) from
      omegaUp.
- [x] If the change is large (> 200 lines), try to split it into several pull requests. Preferably one for controllers + php unit

### DIFF
--- a/frontend/server/src/Controllers/User.php
+++ b/frontend/server/src/Controllers/User.php
@@ -4328,9 +4328,11 @@ class User extends \OmegaUp\Controllers\Controller {
      * @omegaup-request-param string $third_party_login
      */
     public static function getLoginDetailsForTypeScript(\OmegaUp\Request $r) {
-        if (!is_null(
-            \OmegaUp\Controllers\Session::getCurrentSession()['identity']
-        )) {
+        if (
+            !is_null(
+                \OmegaUp\Controllers\Session::getCurrentSession()['identity']
+            )
+        ) {
             header('Location: /');
             die();
         }
--- a/frontend/server/src/Controllers/User.php
+++ b/frontend/server/src/Controllers/User.php
@@ -4328,6 +4328,12 @@ class User extends \OmegaUp\Controllers\Controller {
      * @omegaup-request-param string $third_party_login
      */
     public static function getLoginDetailsForTypeScript(\OmegaUp\Request $r) {
+        if (!is_null(
+            \OmegaUp\Controllers\Session::getCurrentSession()['identity']
+        )) {
+            header('Location: /');
+            die();
+        }
         $thirdPartyLogin = $r->ensureOptionalString('third_party_login');
         if ($r->offsetExists('fb')) {
             $thirdPartyLogin = 'facebook';
--- a/frontend/server/src/Controllers/User.php
+++ b/frontend/server/src/Controllers/User.php
@@ -4334,7 +4334,7 @@ class User extends \OmegaUp\Controllers\Controller {
             )
         ) {
             header('Location: /');
-            die();
+            throw new \OmegaUp\Exceptions\ExitException();
         }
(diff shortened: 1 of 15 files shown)


frontend/www/js/omegaup/components/login/Signin.vue
```diff
@@ -41,7 +41,14 @@ import omegaup_Signup from './Signup.vue';
...
   @Prop() facebookUrl!: string;
+  @Prop() isLoggedIn!: boolean;
 
   T = T;
+
+  mounted() {
+    if (this.isLoggedIn) {
+      window.location.href = '/';
+    }
+  }
```
lhchavez, line 52: i still need to translate the coding guidelines to English, but we avoid any and all lifecycle hooks (https://github.com/omegaup/omegaup/wiki/Coding-guidelines#:~:text=Se%20debe%20evitar%20el,directa%20con%20el%20DOM).

in this case, signin.ts can do this directly instead of needing to involve the Vue component. or even better: if possible fix it from PHP: https://github.com/omegaup/omegaup/blob/657f6e19fa59e6e876f733e4c18d2ac9bc097b80/frontend/server/src/Controllers/User.php#L4330

denllx: Ok, I'll fix it. Btw, could I ask why lifecyle hooks are not allowed?

lhchavez: sure! we've found that they are often abused to do things that are better suited to do elsewhere:
* initialization of variables (better done through computed properties)
* changing the document location (better done by the script instead of letting multiple Vue components fight each other about what the canonical location should be)
* etc.

the only exception to this rule is to interact directly with DOM elements, and even that is something that Vue is built to prevent in the first place! so in practice only one component is allowed to use them: the one that renders Markdown.

frontend/server/src/Controllers/User.php
```diff
@@ -4328,6 +4328,14 @@ public static function isNonUserIdentity(
      * @omegaup-request-param string $third_party_login
      */
     public static function getLoginDetailsForTypeScript(\OmegaUp\Request $r) {
+        if (
+            !is_null(
+                \OmegaUp\Controllers\Session::getCurrentSession()['identity']
+            )
+        ) {
+            header('Location: /');
+            die();
```
lhchavez, line 4337:
```suggestion
            throw new \OmegaUp\Exceptions\ExitException();
```

we don't use `die()` because that causes problems for tests.

frontend/server/src/Controllers/User.php
```diff
@@ -4328,6 +4328,14 @@ public static function isNonUserIdentity(
...
      */
     public static function getLoginDetailsForTypeScript(\OmegaUp\Request $r) {
+        if (
+            !is_null(
+                \OmegaUp\Controllers\Session::getCurrentSession()['identity']
+            )
+        ) {
+            header('Location: /');
+            throw new \OmegaUp\Exceptions\ExitException();
+        }
```
lhchavez, line 4338:
```suggestion
        try {
            $r->ensureIdentity();
            // If the user has already logged in, redirect them to the home page.
            header('Location: /');
            throw new \OmegaUp\Exceptions\ExitException();
        } catch (\OmegaUp\Exceptions\UnauthorizedException $e) {
            // Do nothing.
        }
```

we don't use `getCurrentSession()` in a lot of places. This pattern is a lot more used.

lhchavez (approved):
thanks for the fix!

PR #6489, opened 2022-04-02
Moved the navbar items to right for a better mobile experience.Fix(#5934)

MohitOhlyan:
# Descripción

Move the Navbar Items to the right when using mobile. 
Screenshot from 2022-04-02 17-54-45

Fixes: #5934

# Comentarios

One thing I discovered is that the notification button is not show in the mobile. For now I couldn't get it to work. 
Maybe add a new issue to work on it later.

We can add the User profile above the Navbar items. It might look even better. Maybe in the next issue where the notification button appear too.   

Any suggestion, please let me know in comments. 

# Checklist:

      omegaUp.


frontend/www/js/omegaup/components/common/NavbarItems.vue
```diff
@@ -203,3 +203,11 @@ export default class NavbarItems extends Vue {
   T = T;
 }
 </script>
+
+<style lang="scss" scoped>
+@import '../../../../sass/main.scss';
+
+.mr-auto {
+  float: right;
```
lhchavez, line 211: using `float` might mess some layouts up a bit. is it possible to use text alignment instead? https://getbootstrap.com/docs/4.0/utilities/text/#text-alignment

MohitOhlyan: I think that is having some UI issues, I tried making the `<ul>` right sided but it was making the user profile act weird because the user profile icon is  separate `<div>` and not inside the `<ul>`. 
Screenshot from 2022-03-30 07-08-29

Also there was the problem of Notification Icon not appearing, for which I have made a new Issue and in that issue I am trying to solve the notification icon and possibly make it much better. 
 
 [Issue](https://github.com/omegaup/omegaup/issues/6499)

MohitOhlyan:
Made some better changes. Closing this Pull Request. Please refer to a new Pull Request.

lhchavez:
try not to close PRs if they are going to be replaced by another one.

frontend/www/js/omegaup/components/common/NavbarItems.vue
MohitOhlyan: I think that is having some UI issues, I tried making the `<ul>` right sided but it was making the user profile act weird because the user profile icon is  separate `<div>` and not inside the `<ul>`. 
Screenshot from 2022-03-30 07-08-29

Also there was the problem of Notification Icon not appearing, for which I have made a new Issue and in that issue I am trying to solve the notification icon and possibly make it much better. 
 
 Issue: 

> try not to close PRs if they are going to be replaced by another one. you can keep working on that branch!



> try not to close PRs if they are going to be replaced by another one. you can keep working on that branch!
I am sorry about that. I thought of solving this and another issue in the same PR. I will do as stated from next time.

Issue #6507, opened 2022-04-04
[FEATURE] In create problem page UI.


anmode:
## Some UI improvement in Create problem page. Backend working nicely!

`Suggestion first`
![createproblem1](https://user-images.githubusercontent.com/88332977/161523877-168dff90-c45c-4337-9ce1-463d06baafda.png)

`Improvemnt 1`
The alignment of tag name and delete

`Improvement 2`
The ?logo in Delete in public tag should have hover css so user can know that it shows an important information. So a hover css should be there.

`Suggestion second`
![createproblem2](https://user-images.githubusercontent.com/88332977/161525114-b5e0dcd5-30d9-48d8-8b70-bb727a0ae1b9.png)

`Improvement`
[Here is how to write the file] used two time in this I think upper one is okay but near attach the zip file there is not needed, if there is then UI not looks good there should be space between. So it looks good

## Contexto
UI improvements. So website looks fantastic!

## Tu ambiente

* Nombre del navegador (e.g. Chrome):
* Sistema Operativo y versión (ubuntu 21.10):
* Link al sitio en el que este problema pasa:https://omegaup.com/

anmode:
Could i work on this? Just want advics in 2nd suggestion.

denllx:
I want to solve this issue. Please assign it to me.

anmode:
> Hello, @tvanessa I want to solve this issue. Please assign it to me.

Will u apply for Gsoc?

PR #6520, opened 2022-04-07
Feat/improve create problem UI

denllx:
# Description

I improved the create problem UI.
1. Add hover css for the question mark. After you hover on it, it will become larger to call your attention.
2. Align the 'Tag name' and 'Delete'
3. Remove the duplicate 'how to write file' link

If your change touches the interface, include a screenshot.

Fixes: #6507 

Screenshot from 2022-04-07 18-03-54

Screenshot from 2022-04-07 18-08-55

# Check list:

- [x] The code follows the [guidance
      style](https://github.com/omegaup/omegaup/wiki/Coding-guidelines) from
      omegaUp.
- [x] If the change is large (> 200 lines), try to split it into
      several pull requests. Preferably one for controllers + phpunit


frontend/www/js/omegaup/components/problem/Tags.vue
```diff
@@ -19,17 +19,21 @@
       <table class="table table-striped">
         <thead>
           <tr>
-            <th class="text-center" scope="col">
+            <th class="text-center" width="50%" scope="col">
```
pabo99, line 22: Could it be possible to use a bootstrap-4 class instead? eg. [w-50](https://getbootstrap.com/docs/4.0/utilities/sizing/):
```suggestion
            <th class="text-center w-50" scope="col">
```
?
I think both have the same effect.

It applies in Line 86 too.

frontend/www/js/omegaup/components/problem/Tags.vue
```diff
@@ -19,17 +19,21 @@
...
               {{ T.contestEditTagName }}
             </th>
-            <th class="text-center" scope="col">
+            <th
+              style="text-align: left; width: 50%; padding-left: 15%"
```
pabo99, line 26: Same comment as previous one. 

And one more thing: We discourage the use of `style` attribute because of 
 possible xss attacks. Instead can you use `bootstrap-4` classes:
```suggestion
              class="text-left w-50 pl-1"
```
?
It applies in Line 90 too.

pabo99 (commented):
It looks good, only I have a couple of suggestions

frontend/www/js/omegaup/components/problem/Tags.vue
```diff
@@ -19,17 +19,18 @@
       <table class="table table-striped">
         <thead>
           <tr>
-            <th class="text-center" scope="col">
+            <th class="text-center w-50" scope="col">
               {{ T.contestEditTagName }}
             </th>
-            <th class="text-center" scope="col">
+            <th class="text-left w-50 pl-5" scope="col">
```
pabo99, line 25: Is the alignment change intentional?

frontend/www/js/omegaup/components/problem/Tags.vue
```diff
@@ -79,10 +80,10 @@
       <table class="table table-striped">
         <thead>
           <tr>
-            <th class="text-center" scope="col">
+            <th class="text-center w-5" scope="col">
```
pabo99, line 83: nit:
```suggestion
            <th class="text-center w-50" scope="col">
```

frontend/www/js/omegaup/components/problem/Tags.vue
denllx: yes, if use text-center, then the "delete" title will align with "delete+icon", so the the first "delete" will appear slightly left. thus i use text-left

frontend/www/js/omegaup/components/problem/Tags.vue
denllx: same as above

carlosabcs: `w-5` is not a class from Bootstrap. So it is equal to not writing `w-50`. If you don't need that class, please remove it.

frontend/www/js/omegaup/components/problem/Tags.vue
carlosabcs: What happens when you remove `text-center` or `text-left`? IIRC, `text-left` is the default so it might not be necessary.

frontend/www/js/omegaup/components/problem/Tags.vue
denllx: I fixed.

frontend/www/js/omegaup/components/problem/Tags.vue
denllx: I fixed.

pabo99 (approved):
It looks good to me!

frontend/www/js/omegaup/components/problem/Tags.vue
```diff
@@ -259,3 +260,22 @@ export default class ProblemTags extends Vue {
   }
 }
 </script>
+
+<style>
```
heduenas, line 264: Can you move this css to a common file (such as  `frontend/www/css/style.css`) so we can re-use it in the future for other question buttons?

Issue #6521, opened 2022-04-07
[BUG] User birthday could be later than today


denllx:
## Expected Behavior

In the Edit profile page, users could modify their birthday, but no later than current date.

## Current Behavior
![Screenshot from 2022-04-07 18-26-06](https://user-images.githubusercontent.com/42374096/162179932-0d4cd395-ba04-4473-811f-7a9d7bc8d0b8.png)

## Possible solution

Limit the data picker to only allow dates before today

## Steps to reproduce (for bugs)

1. open https://omegaup.com/profile/#edit-basic-information
2. find 'Birthday Date'
3. select a date later than today

## your environment

* Browser name (e.g. Chrome 39, node.js 5.4): Chrome 89.0.4389.82
* Operating System and version (desktop or mobile): Linux
* Link to the site where this problem happens: See above

denllx:
Meanwhile, I want to solve this issue. Please assign it to me.

MohitOhlyan:
Just curious, how can someone's birth date be later than today?

anmode:
> Just curious, how can someone's birth date be later than today?

+1

denllx:
> Just curious, how can someone's birth date be later than today?

That's also I'm curious about :)

PR #6525, opened 2022-04-07
fix: disable birthday later than today

denllx:
# Description

In the _edit profile_ page, I added limitation to the birth date pick so that it won't display dates later than today.

Screenshot from 2022-04-07 22-00-24

Fixes: #6521 

# Check list:

- [x] The code follows the [guidance
      style](https://github.com/omegaup/omegaup/wiki/Coding-guidelines) from
      omegaUp.
- [x] If the change is large (> 200 lines), try to split it into
      several pull requests. Preferably one for controllers + phpunit


frontend/www/js/omegaup/components/user/BasicInformationEdit.vue
```diff
@@ -56,6 +56,7 @@
       <omegaup-datepicker
         v-model="birthDate"
         :required="false"
+        :max="new Date().toISOString().split('T')[0]"
```
lhchavez, line 59:
```suggestion
        :max="new Date()"
```

the parameter must be a `Date` object according to the declaration in `DatePicker.vue` L32. the provided expression is not one, so the only possible thing that is a `Date` object is just `new Date()`.

frontend/www/js/omegaup/components/DatePicker.vue
```diff
@@ -27,6 +28,8 @@ export default class DatePicker extends Vue {
   @Prop({ default: true }) enabled!: boolean;
   @Prop({ default: T.datePickerFormat }) format!: string;
   @Prop({ default: false }) isInvalid!: boolean;
+  @Prop({ default: new Date(8640000000000000).toISOString().split('T')[0] })
+  max!: Date;
```
lhchavez, line 32:
```suggestion
  @Prop({ default: null }) max!: Date | null;
```

it's probably better to disable the maximum by explicitly using `null` instead of a magic constant if nothing was passed.

frontend/www/js/omegaup/components/DatePicker.vue
```diff
@@ -2,6 +2,7 @@
   <input
     v-model="stringValue"
     :name="name"
+    :max="max"
```
lhchavez, line 5: can this be omitted if the max is `null`? since the value of this needs to convert the date to an ISO date and the `max` property is declared to be a `Date` object, the `.toISOString().split('T')[0]` part can exist in a [computed property](https://vuejs.org/guide/essentials/computed.html) instead of in L31.

lhchavez (changes requested):

frontend/www/js/omegaup/components/DatePicker.vue
denllx: In fact, I used `new Date()` at first, but it doesn't work. It turns out that the <input type="date"> element only accepts dates in **ISO format**. That's why I intentionally convert it to ISOString.
A blog on [stackoverflow.com](https://stackoverflow.com/questions/38638424/html5-input-type-date-disable-dates-before-today) also explains this.

frontend/www/js/omegaup/components/user/BasicInformationEdit.vue
denllx: In fact, I used new Date() at first, but it doesn't work. It turns out that the element only accepts dates in ISO format. That's why I intentionally convert it to ISOString.
A blog on [stackoverflow.com](https://stackoverflow.com/questions/38638424/html5-input-type-date-disable-dates-before-today) also explains this.

frontend/www/js/omegaup/components/DatePicker.vue
denllx: ok. I changed the data type of _max_ to string, and use max='' to disable maximum.

frontend/www/js/omegaup/components/user/BasicInformationEdit.vue
lhchavez: this still needs to happen.

lhchavez (changes requested):

frontend/www/js/omegaup/components/DatePicker.vue
lhchavez: can this please be kept as `Date | null`? I left another comment with a bit more explanation.

frontend/www/js/omegaup/components/DatePicker.vue
```diff
@@ -27,6 +28,7 @@ export default class DatePicker extends Vue {
   @Prop({ default: true }) enabled!: boolean;
   @Prop({ default: T.datePickerFormat }) format!: string;
   @Prop({ default: false }) isInvalid!: boolean;
+  @Prop({ default: '' }) max!: string;
```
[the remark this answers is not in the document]
denllx: I understand. I'll modify it.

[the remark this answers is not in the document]
denllx: I fixed.

lhchavez (approved):
thanks for the fix!

PR #6527, opened 2022-04-08
Aligned the navigation bar buttons on the right side for mobile users…

MohitOhlyan:
# Descripción
For mobile users, the navigation bar buttons appeared on the left side while the navbar button was on the right which causes problem to many. 

In this pull request, I moved the navigation bar buttons on the right side for mobile users. It doesn't affect the  web view. 

I have used `align-items` tag instead of `float` as stated in the previous Pull Request. 

In the video below you can see the the PR doesn't affect the web view and make the mobile view much better. 
Screenshot from 2022-04-14 23-09-24

https://user-images.githubusercontent.com/63846699/162439035-35c1022d-0e38-4741-a9e3-997e08fab14e.mp4

Fixes:
#5934 

# Comentarios

If any improvements are needed please let me know. Thank You. 

# Checklist:

      omegaUp.
- [x] Si el cambio es grande (> 200 líneas), hay que intentar partirlo en

### DIFF
--- a/frontend/www/js/omegaup/components/common/Navbar.vue
+++ b/frontend/www/js/omegaup/components/common/Navbar.vue
@@ -1,6 +1,6 @@
 <template>
   <header>
-    <nav class="navbar navbar-expand-lg navbar-dark fixed-top p-0">
+    <nav class="navbar navbar-expand-lg navbar-dark fixed-top p-0 text-right">
       <div class="container-xl pl-0 pl-xl-3">
         <a class="navbar-brand p-3" href="/">
           <img
@@ -390,7 +390,6 @@ export default class Navbar extends Vue {
 @import '../../../../sass/main.scss';
 
 nav.navbar {
-  text-align: end;
   background-color: var(--header-primary-color);
   .navbar-brand {
     background-color: var(--header-navbar-brand-background-color);
--- a/frontend/www/js/omegaup/components/common/Navbar.vue
+++ b/frontend/www/js/omegaup/components/common/Navbar.vue
@@ -392,9 +392,6 @@ export default class Navbar extends Vue {
 nav.navbar {
   text-align: end;
   background-color: var(--header-primary-color);
-  .navbar-nav .dropdown-menu {
-    position: absolute;
-  }
   .navbar-brand {
     background-color: var(--header-navbar-brand-background-color);
   }
--- a/frontend/www/js/omegaup/components/common/Navbar.vue
+++ b/frontend/www/js/omegaup/components/common/Navbar.vue
@@ -105,7 +105,7 @@
               >
             </li>
           </ul>
-          <ul v-else class="navbar-nav navbar-right">
+          <ul v-else class="navbar-nav navbar-right align-items-end">
             <omegaup-notifications-clarifications
               v-if="inContest"
               :clarifications="clarifications"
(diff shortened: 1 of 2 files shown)


frontend/www/js/omegaup/components/common/Navbar.vue
```diff
@@ -390,8 +390,11 @@ export default class Navbar extends Vue {
 @import '../../../../sass/main.scss';
 
 nav.navbar {
+  text-align: end;
   background-color: var(--header-primary-color);
-
+  .navbar-nav .dropdown-menu {
+    position: absolute;
+  }
```
lhchavez, line 397: what is the purpose of this `position:absolute` here? is there any way to avoid it?

MohitOhlyan: It is so the the dropdown-menu items don't go out of screen on the right side. I tried a lot with other properties like making the dropdown-menu left aligned but this is the only way it worked.

MohitOhlyan: > what is the purpose of this `position:absolute` here? is there any way to avoid it?

If we don't do it, then when we click on the user profile, the entire navbar expands with it. 

I am sorry for my previous comment, I was confused about it.

MohitOhlyan: Screenshot from 2022-04-17 03-23-14
This is how it looks without the absolute Property. 
With absolute:-
Screenshot from 2022-04-17 03-25-14

lhchavez: is this only in mobile, or does it also happen on desktop? if it's only in mobile, it's probably better to have it be consistent with the other dropdowns (i.e. no `position: absolute`)

MohitOhlyan: Yes, this happens only in mobile and when aligned to the right(as in the PR). Okay, I will remove this property and update the PR. 
I had added this because if we are aligned to the right as of right now, then it doesn't happen. Also, having this property does not effect any changes on the Desktop Website.

frontend/www/js/omegaup/components/common/Navbar.vue
```diff
@@ -390,8 +390,8 @@ export default class Navbar extends Vue {
 @import '../../../../sass/main.scss';
 
 nav.navbar {
+  text-align: end;
```
lhchavez, line 393: last request! it's preferable to use Bootstrap classes over changing things in CSS. for this the `nav.navbar` element should have the `text-right` classname: https://getbootstrap.com/docs/4.0/utilities/text/#text-alignment

MohitOhlyan: No worries. I appreciate every suggestions you provide. I do as suggested and check if if does what is required without causing any problems. 
Correct me if I wrong, I should add the `text-right` property to the class and not to modify the `nabar-nav' class.

lhchavez: it's rather adding `text-right` to L3 (the `nav` element that already has the `navbar` class, which in css selector syntax is `nav.navbar`) so that line becomes

```html
<nav class="navbar navbar-expand-lg navbar-dark fixed-top p-0 text-right">
```

MohitOhlyan: Got it. This is what I meant but I think my phrasing was not correct. Thanks Again.

lhchavez (approved):
thank you!

Issue #6501, opened 2022-04-03
[BUG] Code of conduct and privacy policy disappears in firefox mobile view


anmode:
## Comportamiento Esperado
Code of conduct and privacy policy should be visible at bottom in footer in mobile view
Just like in Desktop version.
![webview1](https://user-images.githubusercontent.com/88332977/161428743-6ab3f4f9-4670-4364-a93d-102cde51579d.png)

## Comportamiento Actual
Code of conduc and privacy policy hyperlinks disappear in mobile view.

## Posible Solución

Just have to do some changes in css. in view property.

## Pasos para reproducir (para bugs)
1. Open omegaup.com on mobile or change the dimensions of your browser to Mobile style
2. Move to footer
3. You will unable to see code of conduct and privacy policy

## Contexto
Cannot see privacy polivy and code of conduct link because they disappears

## Tu ambiente

*Incluye los detalles relevantes acerca del ambiente en el que reprodujiste
este bug*

* Nombre del navegador (e.g. Firefox):
* Sistema Operativo y versión (Ubuntu 21.10l):
* Link al sitio en el que este problema pasa:https://omegaup.com/
* Captura de pantalla:
![mobileview1](https://user-images.githubusercontent.com/88332977/161428753-c0dc7e8a-b48f-4438-a199-838b11b0863e.png)

carlosabcs:
I'm able to see it on my mobile. Using Brave browser.
![](https://user-images.githubusercontent.com/20785082/161449467-d18711f8-4719-44e7-b526-926cafd885f1.jpeg)

anmode:
> I'm able to see it on my mobile. Using Brave browser. ![](https://user-images.githubusercontent.com/20785082/161449467-d18711f8-4719-44e7-b526-926cafd885f1.jpeg)

If you reduce window size more nothing dissappear but these two link will.
Try to reduce screen size in brave.

anmode:
Yes this problem is coming in firefox only, what to do now?
:(

arishmaurya:
@tvanessa is problem is in mobile view of all browsers at least in chrome and firefox it is there i want to fork this repo but github is showing fork limit is over can you please help me how should i fork this repo i will fix this issue in 1-2 days

anmode:
> @tvanessa is problem is in mobile view of all browsers at least in chrome and firefox it is there i want to fork this repo but github is showing fork limit is over can you please help me how should i fork this repo i will fix this issue in 1-2 days

[...]

PR #6538, opened 2022-04-15
[Fix] Disappering of code of conduct and privacy policy in mobile mode in some browser.

anmode:
# Descripción
In some browser the code of conduct and privacy policy is hidden, this can be replicated in desktop while reducing the screen size

Fixes: #6501 

# Checklist:

      omegaUp.

### DIFF
--- a/frontend/www/js/omegaup/components/common/Footer.vue
+++ b/frontend/www/js/omegaup/components/common/Footer.vue
@@ -159,24 +159,22 @@
             })
           }}
         </div>
-        <div class="d-none d-md-block d-flex">
-          <ul class="m-0 list-unstyled text-right">
-            <li>
-              <a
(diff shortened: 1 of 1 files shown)


frontend/www/js/omegaup/components/common/Footer.vue
```diff
@@ -159,7 +159,7 @@
             })
           }}
         </div>
-        <div class="d-none d-md-block">
+        <div class="d-none d-md-block d-flex">
```
lhchavez, line 162:
```suggestion
```

derp, the real culprit is `d-none`, which hides the content if the size of the viewport falls below a certain size. so we probably don't need the `<div>` at all!

anmode: oh yes! removing div does not affect anything, so d-none was creating the problem ohhhh got it! thanks

lhchavez (approved):

Issue #6539, opened 2022-04-15
[BUG] Footer is not at the bottom of some pages


strvight:
## Comportamiento Esperado

*Describe lo que originalmente estabas intentando hacer y cómo esperabas que se comportara la plataforma.*

The footer should be at the bottom of the page
![image](https://user-images.githubusercontent.com/63517269/163652563-d51889f8-b7fb-4a8d-bc2d-e7b83245f316.png)

## Comportamiento Actual

*Describe lo que en realidad pasó en lugar de lo que se esperaba*

The footer is not at the bottom of the page
![image](https://user-images.githubusercontent.com/63517269/163652614-092837f6-a15d-4343-9348-9b97645addc2.png)

## Posible Solución

*No es obligatorio, pero puedes sugerir una manera de arreglar el bug, o ideas
de cómo implementar la mejora*

Change footer css

## Pasos para reproducir (para bugs)

*Provee un link a un ejemplo, o una lista de pasos detallada de cómo reproducir
este bug.  Incluye el código que lo reproduce, si es relevante*

1. Go to https://omegaup.com/course/ or https://omegaup.com/scoreboardmerge/
2.  If on https://omegaup.com/scoreboardmerge/ click on enrolled courses when you have none enrolled

## Contexto

*Cómo te ha afectado este problema? Cuál es el objetivo que originalmente querías cumplir?*
*El contexto que nos proporciones nos ayuda a encontrar la solución adecuada*

## Tu ambiente

*Incluye los detalles relevantes acerca del ambiente en el que reprodujiste
este bug*

* Nombre del navegador (e.g. Chrome 39, node.js 5.4): Version 100.0.4896.88
* Sistema Operativo y versión (escritorio o móvil): Windows 10
* Link al sitio en el que este problema pasa: https://omegaup.com/course/ or https://omegaup.com/scoreboardmerge/

strvight:
#6540

strvight:
Sorry if I am doing something wrong, I am new to this.

anmode:
> Sorry if I am doing something wrong, I am new to this.

It is fine! I think if there is content in it.!

anmode:
> Sorry if I am doing something wrong, I am new to this.

It is fine! I think if there is content in it.!

strvight:
> 
Yes, there is content in the pull request I linked!

Mohdtalha12:
Hello strvight
Are you GSOC 2022 interested candidate?

[...]

PR #6540, opened 2022-04-15
css change to fix footer to bottom of page

strvight:
# Description

Fixes:  #6539 

# Comments

Added:

position: fixed;
bottom: 0px;
width: 100%;

to .common-footer 

It now looks like:
image

# Checklist:

- [ x] El código sigue la [guía de
      omegaUp.
- [x ] Se corrieron todas las pruebas y pasaron.
- [x ] Si se está agregando funcionalidad nueva, se agregaron pruebas.
- [x ] Si el cambio es grande (> 200 líneas), hay que intentar partirlo en

### DIFF
--- a/frontend/www/js/omegaup/components/common/Footer.vue
+++ b/frontend/www/js/omegaup/components/common/Footer.vue
@@ -216,6 +216,9 @@ export default class Footer extends Vue {
 .common-footer {
   background-color: $omegaup-primary--darker;
   color: $omegaup-white;
+  position: fixed;
+  bottom: 0px;
+  width: 100%;
 
   .footer-navigation {
     .footer-brand {
--- a/frontend/www/js/omegaup/components/common/Footer.vue
+++ b/frontend/www/js/omegaup/components/common/Footer.vue
@@ -216,9 +216,6 @@ export default class Footer extends Vue {
 .common-footer {
   background-color: $omegaup-primary--darker;
   color: $omegaup-white;
-  position: fixed;
-  bottom: 0px;
-  width: 100%;
 
   .footer-navigation {
     .footer-brand {
--- a/frontend/templates/template.tpl
+++ b/frontend/templates/template.tpl
@@ -53,7 +53,7 @@
       <div id="common-navbar"></div>
       {% jsInclude 'common_navbar' omitRuntime %}
     {% endif %}
-    <main role="main" {% if not fullWidth %}class="container-lg p-5"{% endif %}>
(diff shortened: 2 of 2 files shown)


frontend/www/js/omegaup/components/common/Footer.vue
```diff
@@ -216,6 +216,9 @@ export default class Footer extends Vue {
 .common-footer {
   background-color: $omegaup-primary--darker;
   color: $omegaup-white;
+  position: fixed;
```
lhchavez, line 219: note that `position: fixed` does not reserve any space for this element: https://developer.mozilla.org/en-US/docs/Web/CSS/position#syntax

so that means that the containing element would need to have a padding of height equal to the height of this element for this technique to work. but we can't know upfront what the height is, which means that we can't quite use this solution :(

maybe it's better to have the main content use `flex` sizing so that the height expands if it's smaller than the screen, but keep its height if it's larger.

strvight: I see. I played around with it and it seems if I add flex-grow: 1 to the main content container it will fix the issue. I will test it a little bit more and see. Thanks for your help!

Mohdtalha12:
This issue is fake
It is not appearing
Please strvight don't make Fake issues

Prabhat-10m01:
Yes ,@Mohdtalha12  you are right.

@strvight you are wrong. This is not an issue.

strvight:
No, this is a real issue. Please see below:

image

Prabhat-10m01:
@strvight ,this is a fake issue.

Please look at @anmode comment.
Then you will come to know about it.

Prabhat-10m01:
@lhchavez 

@strvight is wasting your time for a fake issue.
He wants to clear phase 2 that's why he is making fake issue.

@strvight solve good first issue inorder to pass phase 2 
Don't make fake issue

12345Garvit:
Please strvight 
Don't make false issues
This is not fair  especially for other GSOC candidates

If you want to clear phase 2
Solve good first issue
Don't make false issue. 
This is cheating

lhchavez:
@12345Garvit @Mohdtalha12 @Prabhat-10m01 please refrain from posting comments not related to the PR at hand. this is not the right forum for that (that night be Discord), not the kind of behavior we want from GSoC candidates, and is not the right attitude towards a fellow GSoC candidate (or a fellow human being for that matter), since everybody's trying to do the right thing. please focus on your own work and progress.

we will decide whether this qualifies as a good issue for the purposes of the program.

lhchavez:
https://github.com/omegaup/omegaup/pull/6548 managed to fix this first

Issue #6490, opened 2022-04-03
[FEATURE] (ArenaV2) Fix Arena V2 pagination.


heduenas:
We always show several pages even regardless of how many pages there are actually.
![Screen Shot 2022-04-02 at 8 37 11 PM](https://user-images.githubusercontent.com/597815/161410269-55c38f7c-634d-40c5-b63c-21597cc25449.png)
We should query the number of pages from the API and limit the pagination to that number of pages like we do for problems for instance: https://omegaup.com/problem/collection/problemLevelAdvancedSpecializedTopics/. 

Steps to reproduce:
* Go to: https://omegaup.com/arenav2/?query=lakfjsdajsdfkl#past

anmode:
Please assign this issue to me!

PR #6546, opened 2022-04-18
Fix the arenav2 pagination

anmode:
# Descripción

Pagination before is just 10 pages (static/limited)
Screenshot from 2022-04-03 17-42-00

Pagination now are dynamic
!Screenshot from 2022-04-18 23-35-37

Fixes: #6490

# Comentarios
I do not have the data to check whether it's working fine or not! but I think this could be the solution of this problem

# Checklist:

      omegaUp.

### DIFF
--- a/frontend/www/js/omegaup/arena/contestStore.ts
+++ b/frontend/www/js/omegaup/arena/contestStore.ts
@@ -8,7 +8,8 @@ Vue.use(Vuex);
 export interface ContestState {
   // The map of contest lists.
   contests: types.TimeTypeContests;
-}
+  countContests:{[key: string]: number };
+  }
 
 export interface NamedContestListRequest {
   name: string;
@@ -23,10 +24,14 @@ interface NamedContestListResponse {
 export const contestStoreConfig = {
   state: {
     contests: {},
+    countContests: {},
   },
   mutations: {
     updateAll(state: ContestState, payloadContests: types.TimeTypeContests) {
       state.contests = { ...state.contests, ...payloadContests };
+    },
+    updateAllcounts(state: ContestState, payloadcountContests:{[key: string]: number }) {
+      state.countContests = { ...state.countContests, ...payloadcountContests };
     },
     updateList(
       state: ContestState,
--- a/frontend/www/js/omegaup/arena/contestStore.ts
+++ b/frontend/www/js/omegaup/arena/contestStore.ts
@@ -30,11 +30,11 @@ export const contestStoreConfig = {
     updateAll(state: ContestState, payloadContests: types.TimeTypeContests) {
       state.contests = { ...state.contests, ...payloadContests };
     },
-    updateAllcounts(
+    updateAllCounts(
       state: ContestState,
-      payloadcountContests: { [key: string]: number },
+      payloadCountContests: { [key: string]: number },
     ) {
-      state.countContests = { ...state.countContests, ...payloadcountContests };
+      state.countContests = { ...state.countContests, ...payloadCountContests };
     },
     updateList(
       state: ContestState,
--- a/frontend/www/js/omegaup/arena/contestStore.ts
+++ b/frontend/www/js/omegaup/arena/contestStore.ts
@@ -8,8 +8,8 @@ Vue.use(Vuex);
 export interface ContestState {
   // The map of contest lists.
   contests: types.TimeTypeContests;
-  countContests:{[key: string]: number };
-  }
+  countContests: { [key: string]: number };
+}
 
 export interface NamedContestListRequest {
   name: string;
@@ -30,7 +30,10 @@ export const contestStoreConfig = {
     updateAll(state: ContestState, payloadContests: types.TimeTypeContests) {
       state.contests = { ...state.contests, ...payloadContests };
     },
-    updateAllcounts(state: ContestState, payloadcountContests:{[key: string]: number }) {
+    updateAllcounts(
+      state: ContestState,
+      payloadcountContests: { [key: string]: number },
+    ) {
       state.countContests = { ...state.countContests, ...payloadcountContests };
     },
     updateList(
--- a/frontend/www/js/omegaup/components/arena/ContestListv2.test.ts
+++ b/frontend/www/js/omegaup/components/arena/ContestListv2.test.ts
@@ -235,23 +235,38 @@ describe('ContestListv2.vue', () => {
     expect(pastContestTab.text()).toContain('Past Contest 1');
   });
 
-  it('Should handle paginator', async () => {
-    const wrapper = mount(arena_ContestList, {
-      propsData: {
-        contests,
-        countContests: {
-          current: 24,
-          future: 0,
-          past: 56,
+  const periodMapping = [
+    {
+      tab: ContestTab.Current,
+      expectedValue: '‹123›',
+    },
+    {
+      tab: ContestTab.Future,
+      expectedValue: '‹1›',
+    },
+    {
+      tab: ContestTab.Past,
+      expectedValue: '‹123456›',
+    },
+  ];
+  each(periodMapping).it(
+    'Should handle paginator when "%s" field is selected',
+    async ({ tab, expectedValue }) => {
+      const wrapper = mount(arena_ContestList, {
+        propsData: {
+          contests,
+          countContests: {
+            current: 24,
+            future: 0,
+            past: 56,
+          },
+          tab,
         },
-        tab: ContestTab.Future,
-      },
-    });
-
-    const paginator = wrapper.findComponent({ ref: 'paginator' });
-
-    expect(paginator.text()).toBe('ddfd');
-  });
+      });
+      const paginator = wrapper.findComponent({ ref: 'paginator' });
+      expect(paginator.text()).toBe(expectedValue);
+    },
+  );
 
   const dropdownMapping = [
     [{ value: T.contestOrderByTitle }],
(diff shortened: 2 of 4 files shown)


frontend/www/js/omegaup/components/arena/ContestListv2.vue
```diff
@@ -278,10 +278,13 @@
       </b-tabs>
       <b-pagination-nav
         v-model="currentPage"
+        :number-of-pages="pages"
```
lhchavez, line 281: this is not defined anywhere, so this is going to fail :( can you add a test in `ContestListv2.test.ts`?

every PR should have tests to ensure that the change works as expected.

anmode: There is not content so i think that's why it doesn't show last, so first page is visible only. 
Actually idk how to add test! If this falls what to then.
To fix pagination , any advice please?

lhchavez: take a look at the other tests in https://github.com/omegaup/omegaup/blob/main/frontend/www/js/omegaup/components/arena/ContestListv2.test.ts to begin. there should be one new test that passes in 100 contests and checks that pagination items for pages 1, 2, 3, and 10 are displayed.

if you have any more questions, please ask them on Discord!

anmode:
I tested it by looking at url changes! And hopefully it changes as expected.

frontend/www/js/omegaup/components/arena/ContestListv2.vue
```diff
@@ -365,7 +368,11 @@ export default class ArenaContestList extends Vue {
       return ['text-center', 'title-link'];
     }
   }
-
+  data() {
+    return {
+      pages: 10,
+    };
+  }
```
lhchavez, line 375: we don't use the `data` function since we use https://class-component.vuejs.org/ . furthermore, `pages` is a constant, so it's better to leave `:number-of-pages="10"` as it was before.

lhchavez (commented):
can you add a test?

frontend/www/js/omegaup/components/arena/ContestListv2.vue
anmode: > we don't use the `data` function since we use https://class-component.vuejs.org/ . furthermore, `pages` is a constant, so it's better to leave `:number-of-pages="10"` as it was before.

Ok I'll remove it

anmode:
> can you add a test?

Actually i m not familiar with jest, I'll try to add test.

frontend/www/js/omegaup/components/arena/ContestListv2.vue
```diff
@@ -278,10 +278,13 @@
...
       <b-pagination-nav
         v-model="currentPage"
-        size="lg"
+        base-url="#"
+        first-number
+        last-number
+        :number-of-pages="10"
         align="center"
+        size="lg"
         :link-gen="linkGen"
```
lhchavez, line 287: can these items preserve the original order?

```suggestion
        base-url="#"
        first-number
        last-number
        size="lg"
        align="center"
        :link-gen="linkGen"
        :number-of-pages="10"
```

frontend/www/js/omegaup/components/arena/ContestListv2.vue
```diff
@@ -365,7 +368,6 @@ export default class ArenaContestList extends Vue {
       return ['text-center', 'title-link'];
     }
   }
-
```
lhchavez, line 368: can this newline be preserved?

lhchavez (commented):
all right! this is not changing any functionality (only how it displays), so it's fine to not have a test. just a final round of small details.

frontend/www/js/omegaup/components/arena/ContestListv2.vue
```diff
@@ -278,10 +278,13 @@
       </b-tabs>
       <b-pagination-nav
         v-model="currentPage"
-        size="lg"
+        base-url="#"
+        first-number
+        last-number
+        :number-of-pages="10"
```
pabo99, line 284: When PR #6549 is merged with main branch, you can get access to a new property in the payload: `countContests` which is a mapping of the number of contest for every period. e.g.:
```js
{
  current: 2,
  future: 4,
  past: 9,
}
``` 
So, you need create a method to calculate the number of pages to show, I made this one and it worked:
```js

  numberOfPages(tab: ContestTab): number {
    if (!this.countContests[tab]) {
      // Default value when there is no contests in the list 
      return 1;
    }
    const numberOfPages = Math.ceil(this.countContests[tab] / 10);
    return numberOfPages;
  }

```
Another thing to do is send the property in the file `contest_listv2.ts`, as `contests` property is sent, maybe you need to add `countContests` into the `contestStore` .

If you have any doubt, don't hesitate to ask me.

frontend/www/js/omegaup/components/arena/ContestListv2.vue
anmode: Sure!

frontend/www/js/omegaup/components/arena/ContestListv2.vue
anmode: Sure!

frontend/www/js/omegaup/components/arena/ContestListv2.vue
```diff
@@ -369,6 +369,15 @@ export default class ArenaContestList extends Vue {
     }
   }
 
+  numberOfPages(tab: ContestTab): number {
+    if (!this.countContests[tab]) {
```
lhchavez, line 373: now there is functionality, so there should be a jest test. please ask in Discord for guidance!

anmode: Okk sure!!

frontend/www/js/omegaup/components/arena/ContestListv2.vue
anmode: sure i will! Thanks

frontend/www/js/omegaup/arena/contestStore.ts
```diff
@@ -23,11 +24,18 @@ interface NamedContestListResponse {
 export const contestStoreConfig = {
   state: {
     contests: {},
+    countContests: {},
   },
   mutations: {
     updateAll(state: ContestState, payloadContests: types.TimeTypeContests) {
       state.contests = { ...state.contests, ...payloadContests };
     },
+    updateAllcounts(
```
pabo99, line 33: nit:
```suggestion
    updateAllCounts(
```

frontend/www/js/omegaup/arena/contestStore.ts
```diff
@@ -23,11 +24,18 @@ interface NamedContestListResponse {
...
+      state: ContestState,
+      payloadcountContests: { [key: string]: number },
+    ) {
+      state.countContests = { ...state.countContests, ...payloadcountContests };
```
pabo99, line 37:
```suggestion
      state.countContests = { ...state.countContests, ...payloadCountContests };
```

frontend/www/js/omegaup/components/arena/ContestListv2.vue
```diff
@@ -339,6 +343,7 @@ export enum ContestOrder {
   },
 })
 export default class ArenaContestList extends Vue {
+  @Prop() countContests!: { [key: string]: number };
```
pabo99, line 346: To avoid modify the existing tests, it should be a good idea set null as default value:
```suggestion
  @Prop({ default: null }) countContests!: { [key: string]: number } | null;
```

frontend/www/js/omegaup/components/arena/ContestListv2.vue
```diff
@@ -366,6 +371,15 @@ export default class ArenaContestList extends Vue {
     }
   }
 
+  numberOfPages(tab: ContestTab) {
```
pabo99, line 374: This function will return a number, so you need to define it:
```suggestion
  numberOfPages(tab: ContestTab): number {
```

frontend/www/js/omegaup/components/arena/ContestListv2.test.ts
```diff
@@ -235,6 +235,24 @@ describe('ContestListv2.vue', () => {
...
+          past: 56,
+        },
+        tab: ContestTab.Future,
+      },
+    });
+
+    const paginator = wrapper.findComponent({ ref: 'paginator' });
+
+    expect(paginator.text()).toBe('ddfd');
+  });
```
pabo99, line 254: Here you can use a [parameterised test](https://www.npmjs.com/package/jest-each) to cover all the functionallity:
```suggestion
  const periodMapping = [
    {
      tab: ContestTab.Current,
      expectedValue: '‹123›',
    },
    {
      tab: ContestTab.Future,
      expectedValue: '‹1›',
    },
    {
      tab: ContestTab.Past,
      expectedValue: '‹123456›',
    },
  ];

  each(periodMapping).it(
    'Should handle paginator when "%s" field is selected', 
    async ({tab, expectedValue}) => {
    const wrapper = mount(arena_ContestList, {
      propsData: {
        contests,
        countContests: {
          current: 24,
          future: 0,
          past: 56,
        },
        tab,
      },
    });

    const paginator = wrapper.findComponent({ ref: 'paginator' });

    expect(paginator.text()).toBe(expectedValue);
  });

```

pabo99 (approved):
LGTM. Defer to @lhchavez

PR #6703, opened 2022-07-30
plagiarism detector

RohitYandigeri:
I have built this detector using python flask


lhchavez:
just a caveat: we don't like having external servers. anything that should be interacted with by the users should be written in PHP, not in Python+Flask.

lhchavez:
superseded by https://github.com/omegaup/omegaup/pull/6720

PR #6695, opened 2022-07-22
Created a birthdate picker in sign up form

anmode:
Related to mainly Frontend 

Fixes: #6666

Screenshot 2023-03-24 at 1 05 01 PM

# Checklist:

- [x] El código sigue la [guía de
      estilo](https://github.com/omegaup/omegaup/wiki/Coding-guidelines) de
      omegaUp.
- [x] Se corrieron todas las pruebas y pasaron.
- [x] Si se está agregando funcionalidad nueva, se agregaron pruebas.
- [x] Si el cambio es grande (> 200 líneas), hay que intentar partirlo en
      varios pull requests. De preferencia uno para los controladores + phpunit
      y luego otro para la interfaz.

PR #6710, opened 2022-08-03
[Account for children] Add `parent_email_id` field to DB

anmode:
Add the `parent_email_id` field to database for the PR #6695


frontend/database/00212_add_parent_email.sql
```diff
@@ -0,0 +1,3 @@
+-- Emails table
+ALTER TABLE `Emails`
+ADD COLUMN `parentEmail` varchar(100) DEFAULT NULL
```
pabo99, line 3: Nit:
```suggestion
ADD COLUMN `parent_email` varchar(100) DEFAULT NULL;

```

frontend/database/00212_add_parent_email.sql
```diff
@@ -0,0 +1,3 @@
+-- Emails table
+ALTER TABLE `Emails`
+ADD COLUMN `parent_email` varchar(100) DEFAULT NULL;
```
pabo99, line 3: Now that I think well, this field could be a boolean, and it going to be true when U13 users register their parent's email
```suggestion
ADD COLUMN `parent_email` tinyint(1) DEFAULT NULL;
```

lhchavez:
```suggestion
ADD COLUMN `is_parent_email` tinyint(1) DEFAULT NULL;
```

but why is this needed? isn't it sufficient to know that the user is U13, which means that the email associated with that account is their parent's?

tvanessa: @heduenas ^

heduenas: We need to store the parent's email address in a separate field because we need the parent to be able to create their own account so they can monitor their children's activity on omegaup. Otherwise the email value will be repeated in the DB for the children and the parent.

lhchavez: if anything, we should instead store the parent email id in the `Users` table as `parent_email_id` instead of creating a new row in Emails with a flag.

anmode: So then should i close this pR? and create a pr with parent email id in `Users` table as `parent_email_id` ?

lhchavez: you can use this same PR to make the change to the `Users` table to avoid closing this and creating a new one.

anmode: Okayyy!

frontend/database/00212_add_parent_email.sql
```diff
@@ -0,0 +1,3 @@
+-- Users table
```
lhchavez, line 1: this needs to be renamed to start with `00213` and we should be good!

lhchavez (approved):

PR #6908, opened 2023-02-22
mypy Error fix

anmode:
# Descripción
Screenshot 2023-02-22 at 1 45 38 PM

[click here](https://github.com/omegaup/omegaup/actions/runs/4240266370/jobs/7369117708#step:9:214)

1.In the _CustomJsonFormatter class, there is an error in the call to the super().add_fields() method. The third argument should be message_dict instead of dict(message_dict).
2. In the init function, there is a missing import for the Union type hint. You can add it with from typing import Union.
3. There is a missing whitespace after a comma in the call to logging.Formatter in the else branch of the if args.log_json statement.
4. There are two lines that exceed the recommended maximum line length of 79 characters. You can either break the lines or 
use string concatenation to split them into multiple lines.
5.In the init function, there is an error with the indentation in the else branch of the if args.log_json statement. The logging.basicConfig call should be indented one level more.

# Comentarios

1. The first error indicates that the add_fields method of the JsonFormatter class is expecting a Dict[str, Any] type for the third argument, but it's receiving a Mapping[str, Any] type instead. This can be fixed by explicitly converting the mapping object to a dictionary object using the dict() function

2. The second error suggests that there is an untyped function call to _CustomJsonFormatter().

# Checklist:

      omegaUp.

### DIFF
--- a/stuff/lib/logs.py
+++ b/stuff/lib/logs.py
@@ -10,7 +10,7 @@ import argparse
 import datetime
 import logging
 
-from typing import Any, Dict, Mapping
+from typing import Any, Dict, Mapping, Union
 
 from pythonjsonlogger import jsonlogger  # type: ignore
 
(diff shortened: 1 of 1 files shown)


anmode:
In local it is working.

Issue #6915, opened 2023-02-28
[FEATURE] La solución de Google Signin que se utiliza en omegaUp quedará obsoleta en poco tiempo


pabo99:
Según lo descrito en la siguiente página:

https://developers.google.com/identity/sign-in/web/sign-in

El método de autenticación que se utiliza en omegaUp está cerca de quedar obsoleto, 
así que tenemos que migrar a la nueva versión antes del 31 de marzo del presente 
año (2023).

El link con la documentación de este nuevo método es el siguiente:

https://developers.google.com/identity/gsi/web/guides/overview

anmode:
@pabo99 I think we should add sign-in with GitHub. As it is coding platform i think it will good to have GitHub login/signin
Just a suggestion!

heduenas:
@anmode good suggestion can you plese file that as a separate issue so we can track it there?

anmode:
Sure!!!

heduenas:
These are the instructions for migration to the new thing: https://developers.google.com/identity/oauth2/web/guides/migration-to-gis

More specifically, I think we need to follow this example: https://developers.google.com/identity/oauth2/web/guides/migration-to-gis#the_new_way_2

heduenas:
@anmode offered to give this issue a try. Thank you Anmol and feel free to ping @pabo99 or me if you need help.

anmode:
> @anmode offered to give this issue a try. Thank you Anmol and feel free to ping @pabo99 or me if you need help.

sure!!!

PR #6923, opened 2023-03-06
[Important_Feature]Migrate to Google Identity Services

anmode:
1. Removed the old library `gapi.auth2`

fixes: #6915

Just made this pR. So to track that it is going in right way.

### DIFF
--- a/frontend/server/src/Controllers/User.php
+++ b/frontend/server/src/Controllers/User.php
@@ -604,7 +604,7 @@ class User extends \OmegaUp\Controllers\Controller {
                 ],
                 'title' => new \OmegaUp\TranslationString('omegaupTitleLogin'),
                 'scripts' => [
-                    'https://apis.google.com/js/platform.js?onload=init',
+                    'https://accounts.google.com/gsi/client?onload=init',
                 ],
             ],
             'entrypoint' => 'login_signin',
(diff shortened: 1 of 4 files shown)


heduenas:
@anmode please let us know when this is ready for review!

anmode:
> @anmode please let us know when this is ready for review!

Sure! Ig there would be descent changes🥲. I have to take help.

anmode:
I'm Writing some of my research and what i figure out as changes would be descent!
files need to be modified:
https://github.com/anmode/omegaup/blob/google/frontend/www/js/omegaup/login/logout.ts#L22 - Here we used old method as `gapi.auth2.init` and this should be updated by `initTokenClient() or initCodeClient()`
https://github.com/anmode/omegaup/blob/google/frontend/www/js/omegaup/components/login/Login.vue#L84

Ig this should remain as it is - https://github.com/anmode/omegaup/blob/google/frontend/server/src/Controllers/Session.php#L509

I have already changes the script as here - https://github.com/anmode/omegaup/blob/google/frontend/server/src/Controllers/User.php#L606
https://github.com/anmode/omegaup/blob/google/frontend/www/logout.php#L16

I think only these two files need to be changed to adapt `logout.ts` and `login.vue`.

anmode:
closing as pR opened for perfect solution- #6936 Thanks!

Issue #6943, opened 2023-03-19
[BUG] Disabling test_client_contest_with_duplicated_codes to prevent failures in Python tests.


devbrianmedina:
## Comportamiento Esperado
I made a pull request and expected all tests to pass.

## Comportamiento Actual
I made a pull request and the Python test test_client_contest_with_duplicated_codes is failing.
![image](https://user-images.githubusercontent.com/38367679/226157230-2164e5e9-15d4-4d6f-a9fd-ade5f4a79383.png)

## Posible Solución
Disable the test_client_contest_with_duplicated_codes test.

## Pasos para reproducir (para bugs)
The tests have failed in the last pull requests. (Examples #6940 #6941)

## Contexto
The problem is that it prevents the tests from being successfully fulfilled.

## Tu ambiente
* Nombre del navegador Google Chrome
* Sistema Operativo y versión Windows 11

devbrianmedina:
I would like to make changes regarding this issue, to temporarily disable the Python test 'test_client_contest_with_duplicated_codes'. (Username on omegaup: DevBrianMedina)

anmode:
@devbrianmedina we have an ongoing issue it would be great if we can continue to report all flaky tests there and not create another issue.
here's - #6927

PR #6944, opened 2023-03-19
Disabling python test to prevent failures

devbrianmedina:
# Descripción
Some Python tests were temporarily disabled to prevent errors with the PRs.

Fixes: #6943 

# Checklist:

      omegaUp.
- [x] Si el cambio es grande (> 200 líneas), hay que intentar partirlo en


heduenas:
@devbrianmedina  I don't think think we need to disable these tests at the moment. Yes they are flaky, but their flakiness rate is not that high at the moment. So I would say let's wait and see, if they were flakier then let's disable them later on.

devbrianmedina:
@heduenas Yeah, that sounds good.

PR #6989, opened 2023-04-10
FIX ISSUE :#6975 Remove all references to ArenaV1

subhajit20:
Description: Remove all references to ArenaV1 FIX-https://github.com/omegaup/omegaup/issues/6975

- [x] Delete the record of the file nginx.rewritesnamed: arenav1.
- [x] Delete the file /arena/index.phpthat will no longer be used.
- [x] Delete the controller function Contestcalled: getContestListDetailsForTypeScript.
- [x] Delete the record of the file webpack.config-frontend.jsnamed: arena_contest_list.
- [x] Delete the file /frontend/www/js/omegaup/arena/contest_list.tsthat will no longer be used.
- [x] Delete the component /components/arena/ContestList.vuethat will no longer be used.
- [x] Delete all tests that call anything from what was previously deleted.
- [x] Delete unused translation strings.

Fixes: #6975


subhajit20:
@pabo99  can you tell me how to fix this check error ?

pabo99:
"Sure, the "CI / lint" check is displaying the following error:

```shell
Error: Translation string 'arenaCurrentContests' not used
Error: Translation string 'arenaCurrentPublicContests' not used
Error: Translation string 'arenaFutureContests' not used
Error: Translation string 'arenaMyActiveContests' not used
Error: Translation string 'arenaOldContests' not used
Error: Translation string 'arenaPageIntroduction' not used
Error: Translation string 'arenaRecommendedCurrentContests' not used
Error: Translation string 'arenaRecommendedOldContests' not used
Error: Translation string 'frontPageIntroduction' not used
Error: Translation string 'frontPageIntroductionButton' not used
Error: Process completed with exit code 1.
```
To fix this error, you need to search for these translation strings in the `.lang` files and remove them because they are no longer used."

webpack.config-frontend.js
```diff
@@ -8,253 +8,239 @@ const VueLoaderPlugin = require('vue-loader/lib/plugin');
 const defaultBadgeIcon = fs.readFileSync('./frontend/badges/default_icon.svg');
 
 module.exports = {
-  name: 'frontend',
+    name: 'frontend',
```
pabo99, line 11: Could you please revert the changes in this file? It's difficult to identify the important changes made in this way.

I think you are using a linter in your IDE. Could you please disable it in this file? Or perhaps it should be configured to use two spaces when tab is used instead of four.

subhajit20: @pabo99  sure! will do that

PR #7002, opened 2023-05-01
FIX ISSUE :#6975 Remove all references to ArenaV1

subhajit20:
Description: Remove all references to ArenaV1 FIX-https://github.com/omegaup/omegaup/issues/6975

Delete the record of the file nginx.rewritesnamed: arenav1.
Delete the file /arena/index.phpthat will no longer be used.
Delete the controller function Contestcalled: getContestListDetailsForTypeScript.
Delete the record of the file webpack.config-frontend.jsnamed: arena_contest_list.
Delete the file /frontend/www/js/omegaup/arena/contest_list.tsthat will no longer be used.
Delete the component /components/arena/ContestList.vue that will no longer be used.
Delete all tests that call anything from what was previously deleted.


subhajit20:
hey @pabo99  can you check the pr once I have pushed again with all the required changes and let me know if there is any further modifications needed

webpack.config-frontend.js
```diff
@@ -8,253 +8,239 @@ const VueLoaderPlugin = require('vue-loader/lib/plugin');
 const defaultBadgeIcon = fs.readFileSync('./frontend/badges/default_icon.svg');
 
 module.exports = {
-  name: 'frontend',
+    name: 'frontend',
```
pabo99, line 11: Could you please revert the changes made to this file? The changes are difficult to identify in the current format.

I noticed that your IDE is using a linter. Would it be possible for you to disable the linter in this file or configure it to use two spaces instead of four when a tab is used? 

This issue has occurred in a previous PR you opened, but I believe you were able to resolve it before.

PR #7411, opened 2024-01-24
Fix the possibility of creating a contest in the past

F4bbi:
# Description

I noticed that it is possible to create a contest with start date prior to the current date.
Since the `omegaup-datetimepicker` has a min property if `start` is not null, pass `start` to the `omegaup-datetimepicker` as a `new Date()` to trigger the min property.

## Screenshot

Screenshot

# Checklist:


heduenas:
Thank you for finding and fixing this bug!

F4bbi:
I have noted that in cypress tests, contests are created with start date prior to the current date, which is why the tests are failing. Wouldn't it make more sense for the contest creation date to be the day after today, for example?

pabo99:
> I have noted that in cypress tests, contests are created with start date prior to the current date, which is why the tests are failing. Wouldn't it make more sense for the contest creation date to be the day after today, for example?

It would be even better to set the current time rather than the day after today. Setting the start date to the day after today in the test causes it to continue failing because users are unable to join the contest.

pabo99:
Hi @F4bbi! I noticed that you didn't fix the errors in the cypress tests. If you need help to fix them, don't hesitate to ask us!

F4bbi:
> Hi @F4bbi! I noticed that you didn't fix the errors in the cypress tests. If you need help to fix them, don't hesitate to ask us!

Sorry for the absence and thank you for your availability! I think I've fixed the tests, but I've noticed that, because cypress takes a while to fill fields, it can happen that in the meantime the next minute arrives, causing the test to fail because the date entered is no longer valid. I solved this by subtracting 5 minutes from the minimum date to prevent errors, although I'm aware it's not an optimal solution. Would that be acceptable? Otherwise, every time cypress enters a date it should check that the date hasn't already passed and update it if necessary, but it could still happen that by the time it gets to the submit button, the next minute arrives, thus causing the test to fail :/

pabo99 (approved):
Excellent! Thank you for your collaboration!


It looks good to me!

heduenas (approved):

Issue #7450, opened 2024-02-26
[BUG] content in the ephemeral graderα editor should be copied over to the modal showing when making new submission


HNOONa-0:
## Current Behavior

when trying to make a new submission, a modal appears with empty textarea -area where you type code- and language is unset even though code is already present in the editor with specified language.
## Expected Behavior

i shouldn't have to copy paste code from editor to modal, instead contents of editor should be carried over to the modal when clicking the new submission button.
## If necessary, add the following information (otherwise delete this section):
here is a video showing what i mean, i have to copy code from editor, paste it in the modal and change coding language (recorder did not show tabs being opened at the end)

https://github.com/omegaup/omegaup/assets/59091536/e83686a9-dead-4299-870e-20aae062e7a2

*Relevant details about the environment in which you encountered this bug:*

	* Browser name (e.g., Chrome 39, node.js 5.4): chrome 121
	* Operating System and version (desktop or mobile): windows 10
	* Link to the site where this issue occurs: https://omegaup.com/arena/gsoc2024/#problems/IsItATree
	* Screenshot: video above is sufficient

HNOONa-0:
I would like this issue to be assigned to me

Vijaykv5:
> @HNOONa-0, Thank you for your help in creating this issue and for your interest in solving it. I recommend that you resolve all the problems in the [contest](https://omegaup.com/arena/gsoc2024) (I can see you have already resolved 2 out of 3 of them), and then we can assign this issue to you!

Hi @pabo99, may I can give it a try?
Also how can we see that we have attempted the contest. I solved 3 of them but where can I see its results? Or even O solved?

HNOONa-0:
hello @pabo99, I have finished that last problem after trying many different approaches.
I just want to add a small note, in phase 1 of [this link](https://github.com/omegaup/omegaup/wiki/Google-Summer-of-Code-2024#ideas-list), it says that solving 2 problems is enough. I suggest making small update saying that it's necessary to solve entire problemset.
I'm now ready to work on my first issue 🙏

pabo99:
Hello @HNOONa-0!

You're right, It was necesary solving only 2 problems. I think it's okay!

Now, this issue is assigned to you!

PR #7482, opened 2024-03-11
Added an event that copies content of editor into modal when mounted

HNOONa-0:
# Description

I triggered an event manually on **mounted** method of `CodeView`, this event is added using javascript window addEventListener method. This event updates an object to **have the same code and language of that of editor**, then I use that object to mount(init) the modal with the same data as the editor taking into consideration that these changes should be made visible to the parent of `CodeView`.
Here is a video of what I mean:

https://github.com/omegaup/omegaup/assets/59091536/729fce84-552d-47cf-881c-8d56c786b530

I have been trying a more `Vue` friendly approach using event bus or `Vuex` store but I couldn't get them to connect properly. This -I believe- is because the ephemeral editor is embedded where the `CodeView` is on one window while the editor is on another.

I submit this as a draft because I would like to know if I have made any subtle mistakes with naming or otherwise, your feedback is appreciated!

Fixes: #7450

# Comments

# Checklist:


frontend/www/js/omegaup/components/arena/CodeView.vue
```diff
@@ -116,6 +116,19 @@ export default class CodeView extends Vue {
   onLanguageChange(newLanguage: string) {
     this.mode = languageModeMap[newLanguage];
   }
+  mounted() {
+    const codeAndLanguage = {
+      code: this.value,
+      language: !this.language ? 'cpp17-gcc' : this.language,
```
pabo99, line 122: If you require a default value, you can define the property's default value as follows:
```ts
  @Prop({ default: 'cpp17-gcc' }) language!: string;
```
Afterward, there's no need for the ternary operator in this context. You can simply assign the property's value like this:
```suggestion
      language: this.language,
```

pabo99 (commented):
Just one comment.

pabo99 (approved):
Thank you!

frontend/www/js/omegaup/components/arena/CodeView.vue
```diff
@@ -80,7 +80,7 @@ export interface EditorOptions {
 export default class CodeView extends Vue {
   @Prop() language!: string;
   @Prop({ default: false }) readonly!: boolean;
-  @Prop() value!: string;
+  @Prop({ default: 'cpp17-gcc' }) value!: string;
```
pabo99, line 83: Hmm, the default value is for the property `language`, not for the property `value`.

pabo99 (commented):
To resolve the error in the cypress tests, the simplest solution is to select the first element in the failing selector 

Before:
```
cy.get('.CodeMirror-line').type(fileContent);
```
After:
```
cy.get('.CodeMirror-line').first().type(fileContent);
```
Additionally, if you encounter similar occurrences elsewhere in the code, it's advisable to refactor those as well.

frontend/www/js/omegaup/components/arena/CodeView.vue
HNOONa-0: oooooooh thank you, i should pay more attention, let me fix that.

cypress/support/pageObjects/coursePage.ts
```diff
@@ -239,7 +239,7 @@ export class CoursePage {
     cy.get('[data-notification-list]').last().click();
     cy.get('[data-run-feedback]').should('contain', feedback);
     cy.get('.CodeMirror-lines').should('be.visible');
-    cy.get('.CodeMirror-line').then((rawHTMLElements) => {
+    cy.get('.CodeMirror-line').first().then((rawHTMLElements) => {
```
pabo99, line 242: I believe this change is unnecessary because we already have content in the CodeMirror. Could you please revert it?

pabo99 (commented):
One more comment

HNOONa-0:
@pabo99 I believe i made too many commits, would you like if we delete this pr and start another one having fewer commits or just continue?

pabo99:
> @pabo99 I believe i made too many commits, would you like if we delete this pr and start another one having fewer commits or just continue?

It's perfectly fine to make multiple commits. Otherwise, we risk losing all the valuable comments and suggestions, which can be helpful for other contributors.

heduenas (approved):

HNOONa-0:
Hello @heduenas, thank you for approving my pr. Is there something extra I should do to merge my pr? Do I need to update my branch to be in sync with main again?

pabo99:
@HNOONa-0 Auto-merge is enabled for this PR. So, you need update your branch and then push the new commit.

HNOONa-0:
> @HNOONa-0 Auto-merge is enabled for this PR. So, you need update your branch and then push the new commit.

I just updated my branch, no newer commits on my side. Should smoothly merge now I think.

Issue #7471, opened 2024-03-08
[BUG] UI Enhancement : Add Space between Icons


Vijaykv5:
## Current Behavior
The Date Icons and the no of people's (group's icon) are conjusted, disturbing the UI experiance

## Expected Behavior

Instead we can have a `4px` on distance of x axis on both the date as well as group icon

## If necessary, add the following information (otherwise delete this section):
*Provide a link to an example, or a detailed list of steps to reproduce this bug.*  
  
	1. click in competitions from navbar
	2. select see contest
	3. now you'll notice the bug here
*Relevant details about the environment in which you encountered this bug:*

	* Browser name (e.g., Chrome 39, node.js 5.4): Chrome
	* Operating System and version (desktop or mobile): Mac. OS
	* Link to the site where this issue occurs: https://omegaup.com/arena/
	* Screenshot: 
<img width="989" alt="Screenshot 2024-03-09 at 12 11 33 AM" src="https://github.com/omegaup/omegaup/assets/94985341/a8ff2a5a-0c87-41d5-a8f3-279ca965c3be">

Vijaykv5:
I tried to reproduce locally but I figured out that this issue is only in production.
It's working fine locally!

<img width="1299" alt="Screenshot 2024-03-09 at 5 26 18 PM" src="https://github.com/omegaup/omegaup/assets/94985341/1fd1854b-f95d-413d-a455-020475430bfc">

RingoTC:
I can do this.

pabo99:
> I can do this.

@RingoTC: Assigned to you!

PR #7502, opened 2024-03-22
Fix4spacing

RingoTC:
# Description #7471 

## The cause of this bug
Because the previous code is as follows:
```jsx
<b-col class="col-md-3 col-sm-12 p-1 text-center">
   <b-card-text>
       <font-awesome-icon class="mr-1" icon="clipboard-list" />
       {{ contest.organizer }}
       </b-card-text>
</b-col>
```
it will cause a space between icon and text like that:
image
which is not expected and may cause unpredictable problems.

A better way to handle this, is wrap text with p tag and using flex-box to style this component. Thus I change this codeblock as follows:
```jsx
<b-col class="col-md-3 col-sm-12 p-1 text-center">
  <b-card-text class="d-flex justify-content-center align-items-center">
    <font-awesome-icon class="mr-1" icon="clipboard-list" />
    <p class="m-0">{{ contest.organizer }}</p>
  </b-card-text>
</b-col>
```
And the final display as we want:
image

# lint
run `./stuff/lint.sh fix -- frontend/www/js/omegaup/components/arena/ContestCard.vue` for vue.

# ref
[Paragraphs, Lines, and Phrases](https://www.w3.org/TR/REC-html40/struct/text.html#h-9.1)


pabo99 (approved):
LGTM!

Issue #7492, opened 2024-03-17
adding row animation while submission of a problem


antedotee:
## Current Behavior

![CleanShot 2024-03-17 at 15 52 02@2x](https://github.com/omegaup/omegaup/assets/151192805/46c5368d-4a0e-4570-9309-51b1727e73cd)

## Expected Behavior

https://github.com/omegaup/omegaup/assets/151192805/d65f1801-c797-4cb2-a7db-32e8cd55f424

antedotee:
Can anyone assign this issue to me?

Vijaykv5:
May I can give it a try @antedotee ?

antedotee:
I would love to,but I have already worked on this issue.I think there are plenty of other issues to solve.Hope you have a nice day.

PR #7504, opened 2024-03-22
Added SkeletonLoader while submission of problems

antedotee:
# Description

Earlier,user was unable to know whether the code is running or not.So,I added a skeleton loader and its component so that user can know that program is running.

Here is the video regarding skeleton loader:

https://github.com/omegaup/omegaup/assets/151192805/58b17d56-9eff-4b02-bb03-6e78dba65a49

Fixes: #7492

# Checklist:

### DIFF
--- a/cypress/support/pageObjects/coursePage.ts
+++ b/cypress/support/pageObjects/coursePage.ts
@@ -433,7 +433,6 @@ export class CoursePage {
   }
 
   verifyProblemRun(status: string): void {
-    cy.get('[data-run-status] > span', { timeout: 10000 }).should('be.visible').first().should('have.text', status);
     cy.get('div.line').should('be.visible');
 
     cy.intercept({ method: 'POST', url: '/api/run/status/' }).as('runStatus');


pabo99:
Could you upload a video or GIF animation showing how the SkeletonLoader works?

I've tried testing it in my local environment, but I can't get it.

frontend/www/js/omegaup/components/arena/RunsForCourses.vue
```diff
@@ -243,215 +243,226 @@
...
-                  :data-content="statusHelp(run)"
-                  data-toggle="popover"
-                  data-trigger="focus"
-                  class="btn-outline-dark btn-sm"
-                  @click="showVerdictHelp"
+              <template v-if="run.status !== 'ready'">
+                <td colspan="16">
+                  <div class="line"></div>
+                </td>
+              </template>
```
pabo99, line 250: Here you don't need a template because you are using a single element, so it can work like this:
```suggestion
              <td v-if="run.status !== 'ready'" colspan="16">
                <div class="line"></div>
              </td>
```

frontend/www/js/omegaup/components/arena/RunsForCourses.vue
```diff
@@ -1136,4 +1147,21 @@ export default class Runs extends Vue {
   background: var(--arena-runs-table-status-ce-background-color);
   color: var(--arena-runs-table-status-ce-font-color);
 }
+.line {
+  height: 49px;
+  background: #e0e0e0;
```
pabo99, line 1152: Could you add this colors as variables? 

You can check in this file how we are assigning the values to all colors

pabo99 (commented):
I've discovered a solution for the errors in the Cypress tests:

You need to search for the following command in the codebase:
```ts
cy.get('[data-run-status] > span').first().should('have.text', 'new');
```
and replace it with:
```ts
cy.get('div.line').should('be.visible');
```

Additionally, it's worth noting that I can only see the animation in `ArenaProblem`. When I submit a run in `ArenaCourse`, I'm unable to see it. This component is supposed to be used in both views mentioned above. What's even stranger is that when I run the Cypress tests for the course spec, I can actually see the animation

frontend/www/js/omegaup/components/arena/RunsForCourses.vue
```diff
@@ -219,7 +219,7 @@
           </thead>
           <tfoot v-if="problemAlias != null">
             <tr>
-              <td colspan="10" data-new-run>
```
pabo99, line 222: This data-attribute is used in the cypress tests. So, it could be better, revert the change in ths line

antedotee:
Regarding video,I attached the video in the description of the issue linked to this pull request.

antedotee:
It is still giving giving the same error🥲

frontend/www/js/omegaup/components/arena/RunsForCourses.vue
```diff
@@ -243,215 +243,223 @@
           </tfoot>
           <tbody>
             <tr v-for="run in paginatedRuns" :key="run.guid">
-              <td>{{ time.formatDateLocalHHMM(run.time) }}</td>
-              <td v-show="showGUID">
-                <acronym :title="run.guid" data-run-guid>
-                  <tt>{{ getShortGuid(run.guid) }}</tt>
-                </acronym>
+              <td v-if="run.status !== 'ready'" colspan="8" data-new-run>
```
pabo99, line 246: The attribute `data-new-run` corresponds to the new-submit button, and here you are displaying the SkeletonLoader. So it can be removed:
```suggestion
              <td v-if="run.status !== 'ready'" colspan="8">
```

pabo99:
> It is still giving giving the same error🥲

Unfortunately, I'm unable to reproduce the errors in my local environment. Let me check what else we can do to resolve these errors.

cypress/support/pageObjects/coursePage.ts
```diff
@@ -433,7 +433,8 @@ export class CoursePage {
   }
 
   verifyProblemRun(status: string): void {
-    cy.get('[data-run-status] > span').first().should('have.text', 'new');
+    cy.get('[data-run-status] > span', { timeout: 10000 }).should('be.visible').first().should('have.text', status);
```
pabo99, line 436: In this test, you are adding the skeleton loader component as well, so you don't need to get `[data-run-status]` element. This line can be removed:
```suggestion
```

pabo99 (commented):
"I've left a comment for you that helps resolve the error in the Cypress test

antedotee:
now the test cases are only failing in course.cy.ts, earlier it was failing in 3 files.I think I am very close to fixing this but can't get past this error.

PR #7506, opened 2024-03-23
Added a Trigger so whenever a problem gets inserted to a problemset the order field gets updated to be pushed to the end of the list.

mohannedah:
# Description

Added a Trigger to update the order of the problem before being added to the problem set.

Fixes: #6682 

# Comments
What I did was basically adding a Trigger, so, whenever we try to add a problem related to a specific problem_set the order gets updated to the count of the total problems in a problem_set + 1.
Screenshot 2024-03-22 232916
That is before inserting the problem named "filas" into the problemset whose id is equal to 7.
Before 2 1
and that's the case after inserting the problem "filas".
After 2 2
I tried reordering and saving and I realized when it comes to updates it works just fine because whenever we try to have a new order. we send a request to /api/course/updateProblemsOrder and the body of the request contains a list of the aliases of the problems ordered by the new order assigned in the frontend. 
the controller handles the logic of updating the order.
Screenshot 2024-03-23 021209

# Checklist:


pabo99:
Issue #6682 was fixed on PR #7507, so this PR can be closed

PR #7507, opened 2024-03-23
Added a Trigger to update the order of the problem before being added to the problem set.

mohannedah:
# Description

What I did was basically adding a Trigger, so, whenever we try to add a problem related to a specific problem_set the order gets updated to the count of the total problems in a problem_set + 1.

Fixes: #6682 

# Comments

What I did was basically adding a Trigger, so, whenever we try to add a problem related to a specific problem_set the order gets updated to the count of the total problems in a problem_set + 1.
Screenshot 2024-03-22 232916
That is before inserting the problem named "filas" into the problemset whose id is equal to 7.
Before 2 1
and that's the case after inserting the problem "filas".
After 2 2

I tried reordering and saving and I realized when it comes to updates it works just fine because whenever we try to have a new order. we send a request to /api/course/updateProblemsOrder and the body of the request contains a list of the aliases of the problems ordered by the new order assigned in the frontend.
the controller handles the logic of updating the order.
Screenshot 2024-03-23 021209


frontend/database/00009_acls.sql
```diff
@@ -164,4 +164,23 @@ ALTER TABLE `Group_Roles`
...
+--     SELECT COUNT(*) INTO row_count FROM Problemset_Problems;
+
+--     UPDATE Problemset_Problems
+--     SET Problemset_Problems.order = row_count
+--     WHERE problem_id = NEW.problem_id;
+-- END;
+-- //
+
+-- DELIMITER ;
+
```
pabo99, line 184: Could you revert all this commented code:
```suggestion
```
?

frontend/database/00009_acls.sql
```diff
@@ -164,4 +164,23 @@ ALTER TABLE `Group_Roles`
...
 -- vim: set expandtab:ts=2:sw=2
+
```
pabo99, line 186: Nit:
```suggestion
```

frontend/database/00049_add_identity_table.sql
```diff
@@ -32,7 +32,7 @@ ALTER TABLE `Users`
 START TRANSACTION;
 
 INSERT INTO `Identities` (
-  `identity_id`,
+  `identity_id`,
```
pabo99, line 35: Nit:
```suggestion
  `identity_id`,
```

frontend/server/src/DAO/ProblemsetProblems.php
```diff
@@ -223,7 +223,7 @@ final public static function getProblemsByAssignmentAlias(
                 INNER JOIN
                     Assignments a
                 ON
-                    a.problemset_id = pp.problemset_id
+                    a.problemset_id = ps.problemset_id
```
pabo99, line 226: What is the difference between joining with the `Problemsets` table instead of `Problemset_Problems`?

frontend/database/00008_problemsets.sql
```diff
@@ -15,6 +15,21 @@ CREATE TABLE `Problemset_Problems` (
...
+BEGIN
+    DECLARE max_order INT;
+    SELECT MAX(`order`) INTO max_order FROM Problemset_Problems where Problemset_Problems.problemset_id = NEW.problemset_id;
+    SET NEW.`order` = COALESCE(max_order, 0) + 1;
+END;
+//
+DELIMITER ;
+
+
+
```
pabo99, line 32: Previously, we used some triggers, but at the end we decided that all the logic should be centralized on the server because maintaining those triggers was more complicated. So, could you revert the changes in this file?
```suggestion
```

pabo99 (commented):
I have been looking for a way to solve the problem, and in my research, I found the following:

- When adding problems during assignment creation, the sorting works correctly because the elements are iterated and the order is incremented for each iteration:
https://github.com/omegaup/omegaup/blob/e18e34de31c9b62ab62a01b67410af289aaea797/frontend/server/src/Controllers/Course.php#L997-L1009

- Manually sorting the problems also works correctly, due to the same aforementioned case: the elements are iterated and the order is incremented for each iteration:
https://github.com/omegaup/omegaup/blob/e18e34de31c9b62ab62a01b67410af289aaea797/frontend/server/src/Controllers/Course.php#L1529-L1548

- The problem occurs when adding a problem to the assignment. Although the `addProblemToAssignment` function receives the `order` parameter, this data is not being sent here:
https://github.com/omegaup/omegaup/blob/e18e34de31c9b62ab62a01b67410af289aaea797/frontend/server/src/Controllers/Course.php#L1419-L1427

So, what needs to be done before calling the aforementioned function is to obtain the next position in the order, just as is done with the assignments:
https://github.com/omegaup/omegaup/blob/e18e34de31c9b62ab62a01b67410af289aaea797/frontend/server/src/DAO/Assignments.php#L396-L412

Or even better, to avoid creating a new function, I found a function that retrieves the number of problems in a problemset:
https://github.com/omegaup/omegaup/blob/e18e34de31c9b62ab62a01b67410af289aaea797/frontend/server/src/DAO/ProblemsetProblems.php#L95-L109
You can call that function and then increment 1 to the total number of problems in the problemset.

frontend/server/src/DAO/ProblemsetProblems.php
mohannedah: I assume, I've deleted this line accidentally and then I had to write the query again, so that's why.

frontend/database/00008_problemsets.sql
mohannedah: So, my task is, instead of handling the logic on the database layer, we should handle it while inserting the records through our API. That will work fine, but the problem is what if we decided to seed data into the database directly without our API being involved, we will have the same issue, the order of the records will have a default value of 1.

I will revert the changes and I will try to handle this task on the server-side.

frontend/server/src/Controllers/Course.php
```diff
@@ -1408,7 +1408,8 @@ public static function apiAddProblem(\OmegaUp\Request $r): array {
                 'problemsetNotFound'
             );
         }
-
+        
+        $cntProblems = \OmegaUp\DAO\ProblemsetProblems::countProblemsetProblems($problemset);
```
pabo99, line 1412: --Could you use the whole word:
```suggestion
        $countProblems = \OmegaUp\DAO\ProblemsetProblems::countProblemsetProblems($problemset);
```
?

pabo99 (commented):
Don't forget execute  `./stuff/lint.sh` before pushing  your   changes

pabo99 (approved):
LGTM!

heduenas (approved):

Issue #7516, opened 2024-03-24
[FEATURE] Remove the nested scrollers.


iqbalcodes6602:
*Provee un título para esta sugerencia con el siguiente formato:*

The contest page has unnecessary nested scrollers which somewhat dirrupts the UI of the website it would be better to have a dynamic height whihc would adjust according to the number of participants.

![image](https://github.com/omegaup/omegaup/assets/111554475/af861459-0ba8-45b0-88c6-2a8521bd715b)

PR #7518, opened 2024-03-24
Remove scroll CSS styles from ContestListv2.vue

iqbalcodes6602:
# Description

After reading some of the previous commits this was also asked to be fixed previously, but wasn't done correctly. I have removed the styles for scroll-content, hence enablin the height to be auto.

image

here it is displaying 10 contest yet no scrolls

Fixes: #7516 

# Comments

# Checklist:


pabo99:
The reason for deciding to set a maximum height for the contest list was precisely to ensure that the search, sorting, and filtering controls were always visible. However, this wasn't being achieved.

Is there a way to stick the control bar to the top of the page even when scrolling down? We previously attempted to use CSS styles with position: sticky, but perhaps we were doing it incorrectly.

Could you try adding that behavior?

iqbalcodes6602:
@pabo99  done as you  said

image

pabo99 (approved):
Excellent! Thank you so much!

iqbalcodes6602:
another thing currently the problems page is also having the same nested scrollers, should i also do that in another PR,. 
The main challenge though  in this is the that we need to fix the pagination part in the bottom and search and all at top.
do u have any suggestion regarding this.
image

iqbalcodes6602:
> Excellent! Thank you so much!

@pabo99 
another thing currently the problems page is also having the same nested scrollers, should i also do that in another PR,.
The main challenge though in this is the that we need to fix the pagination part in the bottom and search and all at top.
do u have any suggestion regarding this.
image

pabo99:
> > Excellent! Thank you so much!
> 
> @pabo99 another thing currently the problems page is also having the same nested scrollers, should i also do that in another PR,. The main challenge though in this is the that we need to fix the pagination part in the bottom and search and all at top. do u have any suggestion regarding this. image

Please assist us in creating the issue. We will continue all discussions in the new thread. Regarding the pagination, I don't think it's necessary to keep it sticky.

heduenas (approved):

Issue #7512, opened 2024-03-23
[BUG] The most recent notifications should be displayed at the top of the list


pabo99:
## Current Behavior

When viewing the list of notifications, I notice that the latest ones are displayed at the top.

![image](https://github.com/omegaup/omegaup/assets/3230352/4a671f3f-685d-4925-8678-3339c77d0f94)

## Expected Behavior

The behavior should be the opposite: Display the most recent ones at the top.

Ashish-003:
@pabo99 I can start with this

pabo99:
> @pabo99 I can start with this

Excellent. The issue is all yours!

PR #7519, opened 2024-03-24
Bug Fix # 7512: Reversed Notification Array

Ashish-003:
# Description
Reversed the notification array before showing it to the user. Most recent notification is on top

image

Fixes: #7512 

# Comments
Preferred reversing in place of sorting by time because of O(n) vs O(NlogN) and there wont be many race conditions

# Checklist:


pabo99 (commented):
Okay. The problem comes in the Cypress test because it's getting the last element in the dropdown, but you are inverting the order of the elements. 

The solution can be the following .in the function `verifyFeedback` you will find this line:
https://github.com/omegaup/omegaup/blob/1f71ef4d93616a8522977860c633d54095351ee0/cypress/support/pageObjects/coursePage.ts#L239

So, you should replace it with:
```ts
cy.get('[data-notification-list]').filter(':visible').first().click();
```
in order to get the first visible element in the dropdown.

pabo99 (approved):
LGTM!

heduenas (approved):

Issue #7530, opened 2024-03-26
[BUG] Active tab gets changed when refreshed - Contest edit page


Aritra8438:
## Current Behavior

As a result of PR #7505, a new bug was introduced where the active tab of the contest edit page gets changed when refreshed.

Please refer to [this](https://github.com/omegaup/omegaup/pull/7527#pullrequestreview-1959453919) comment of @pabo99. 

## Expected Behavior

When the page is refreshed, the selected tab should remain active, similar to the behaviour on the CourseEdit page.

##  Steps to reproduce this bug.  
[...]

PR #7531, opened 2024-03-26
Render initial tab based on anchor - Contest Edit Page

Aritra8438:
# Description

This PR fixes the bug where the active tab gets changed when refreshed on the contest edit page.

Fixes: #7530 

https://github.com/omegaup/omegaup/assets/64671908/dc9f02b3-be6a-484f-81c0-d724d7795fbd

# Checklist:

### DIFF
--- a/cypress/support/pageObjects/contestPage.ts
+++ b/cypress/support/pageObjects/contestPage.ts
@@ -134,8 +134,8 @@ export class ContestPage {
 
   createContest(contestOptions: ContestOptions, users: Array<string>, shouldShowIntro: boolean = true): void {
     cy.createContest(contestOptions, shouldShowIntro);
-
     cy.location('href').should('include', contestOptions.contestAlias);
+    cy.find('a[data-contest-new_form]').trigger('click');
     cy.get('[name="title"]').should('have.value', contestOptions.contestAlias);
     cy.get('[name="alias"]').should('have.value', contestOptions.contestAlias);
     cy.get('[name="description"]').should(
--- a/cypress/support/pageObjects/contestPage.ts
+++ b/cypress/support/pageObjects/contestPage.ts
@@ -135,7 +135,7 @@ export class ContestPage {
   createContest(contestOptions: ContestOptions, users: Array<string>, shouldShowIntro: boolean = true): void {
     cy.createContest(contestOptions, shouldShowIntro);
     cy.location('href').should('include', contestOptions.contestAlias);
-    cy.find('a[data-contest-new_form]').trigger('click');
+    cy.get('a[data-contest-new-form]').trigger('click');
     cy.get('[name="title"]').should('have.value', contestOptions.contestAlias);
     cy.get('[name="alias"]').should('have.value', contestOptions.contestAlias);
     cy.get('[name="description"]').should(
--- a/frontend/www/js/omegaup/components/contest/Edit.test.ts
+++ b/frontend/www/js/omegaup/components/contest/Edit.test.ts
@@ -46,6 +46,7 @@ describe('Edit.vue', () => {
   const propsData: {
     admins: types.ContestAdmin[];
     details: types.ContestAdminDetails;
+    initialTab: string;
     groups: types.ContestGroup[];
     groupAdmins: types.ContestGroupAdmin[];
     problems: types.ProblemsetProblemWithVersions[];
@@ -56,6 +57,7 @@ describe('Edit.vue', () => {
   } = {
     admins: [],
     details,
+    initialTab: 'clone',
     groupAdmins: [],
     groups: [],
     problems: [],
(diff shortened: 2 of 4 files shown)


cypress/support/pageObjects/contestPage.ts
```diff
@@ -134,8 +134,8 @@ export class ContestPage {
 
   createContest(contestOptions: ContestOptions, users: Array<string>, shouldShowIntro: boolean = true): void {
     cy.createContest(contestOptions, shouldShowIntro);
-
     cy.location('href').should('include', contestOptions.contestAlias);
+    cy.find('a[data-contest-new_form]').trigger('click');
```
pabo99, line 138: The correct funcion should be `cy.get`:
```suggestion
    cy.get('a[data-contest-new-form]').trigger('click');
```

frontend/www/js/omegaup/components/contest/Edit.vue
```diff
@@ -21,6 +21,7 @@
         <a
           v-if="!virtual"
           href="#new_form"
+          data-contest-new_form
```
pabo99, line 24: Nit:
```suggestion
          data-contest-new-form
```
just for consistency with the rest of data-attributes.

frontend/www/js/omegaup/contest/edit.ts
```diff
@@ -25,6 +25,7 @@ OmegaUp.on('ready', () => {
     data: () => ({
       admins: payload.admins,
       details: payload.details,
+      initialTab: window.location.hash ? window.location.hash.substring(1) : '',
```
pabo99, line 28: Here you can simplify as follows:
```suggestion
      initialTab: window.location.hash?.substring(1) || '',
```
and it might be a good idea to always set the initial tab as 'problems'. What do you think?

pabo99 (commented):
Just a few comments

frontend/www/js/omegaup/contest/edit.ts
Aritra8438: Then, when refreshed, would the tab not change to `problem` every time?

pabo99: Not really, only when the URL doesn't contain any hash, the initial tab will be `problems`.

Aritra8438: I see. Sure! That's a great idea, contestants expect to view the problems when they click on the contest.

cypress/support/pageObjects/contestPage.ts
Aritra8438: This was the reason for cypress test failure :3
Thank you.

Aritra8438:
Done @pabo99! PTAL!

frontend/www/js/omegaup/components/contest/Edit.vue
```diff
@@ -363,13 +365,16 @@ export default class Edit extends Vue {
...
+      return this.initialTab;
+    }
     if (!ui.isVirtual(this.details)) {
       return 'new_form';
     }
     if (this.originalContestAdmissionMode != 'private') {
       return 'contestants';
     }
-    return 'links';
+    return 'problems';
```
pabo99, line 377: Okay, I knew this value had a reason.

We have two kinds of virtual contests:

- Those in which the original contest's admission mode is public, allowing contestants to be added:
image

- Those in which the original contest's admission mode is private, restricting the addition of contestants:
image

So, the return value should be `links`
```suggestion
    return 'links';
```

frontend/www/js/omegaup/components/contest/Edit.test.ts
```diff
@@ -85,11 +88,12 @@ describe('Edit.vue', () => {
...
   it('Should handle a virtual contest from an original private contest', () => {
     propsData.details.rerun_id = 2;
+    propsData.initialTab = '';
     propsData.originalContestAdmissionMode = 'private';
     const wrapper = shallowMount(contest_Edit, {
       propsData,
     });
 
-    expect(wrapper.vm.showTab).toBe('links');
+    expect(wrapper.vm.showTab).toBe('problems');
```
pabo99, line 97: According the previous comment, this tab should be `links`:
```suggestion
    expect(wrapper.vm.showTab).toBe('links');
```

pabo99 (commented):
The last suggestions (I hope).

Thank you for your patience.

frontend/www/js/omegaup/contest/edit.ts
pabo99: I realized that in the component, we are setting the default value according to each type of contest. So, just forget the previous messages in the thread.

frontend/www/js/omegaup/components/contest/Edit.vue
Aritra8438: No problem, @pabo99.

pabo99 (approved):
LGTM!

heduenas (approved):

PR #7535, opened 2024-03-28
fix #7523 notification overflow when browser size increases

### LINKED ISSUE #7523: [BUG]
## Current Behavior
The Notification Window becomes weird when window length is >= 1000
image

## Expected Behavior
Should look like this
image

## If necessary, add the following information (otherwise delete this section):
*Provide a link to an example, or a detailed list of steps to reproduce this bug.*  
  
	1.  Change window length while notification window is opened

	* Browser name (e.g., Chrome 39, node.js 5.4): Mozilla Firefox 116.0.2 (64-bit)
	* Operating System and version (desktop or mobile): Linux Desktop
	* Link to the site where this issue occurs: https://omegaup.com/
	* Screenshot: Pasted Above

RingoTC:
# Description
The previous style of .notification-dropdown is
```css
.notification-dropdown {
  max-width: 100vw;
  max-height: 600px;
  overflow-y: auto;
}
```
it may cause the overflow problem if its child elements wider than normal like that:
image

To solve this problem I modify this style setting as follows:
```css
.navbar-expand-lg .navbar-nav .dropdown-menu {
  min-width: 55vw;
  @media only screen and (min-width: 767px) {
    min-width: 35rem;
  }
}

.notification-dropdown {
  max-width: 75vw;
  max-height: 600px;
  overflow-y: auto;
}

```

Fixes: #7523


pabo99 (approved):

heduenas (approved):

heduenas:
@RingoTC can you include an screenshot of how it looks after your change?

Issue #7524, opened 2024-03-24
[BUG] ephemeral.js is not reflecting changes after successful compilation


HNOONa-0:
## Current Behavior

Making any change to `www/js/omegaup/grader/ephemeral.js` takes a long time to be reflected. Only files in the `grader` directory seem to have this issue, below in the video I tried to log a line, and after successful compilation, no lines were logged! (video was edited because compilation took too long)

https://github.com/omegaup/omegaup/assets/59091536/cfcdd230-c7de-46c5-8077-2cc64ead0bd1

## Expected Behavior

The changes should be reflected immediately

## If necessary, add the following information (otherwise delete this section):
*Provide a link to an example, or a detailed list of steps to reproduce this bug.*  
  
	1. go `ephemeral.js` and make any local changes.
	2. save and recompile. find out that changes did not take place.

iqbalcodes6602:
We need to hard refresh the Brower for seeing the changes, this means re loading the page without cache, in chrome it can be down by pressing ctrl + f5 Or ctrl+ shift + r

RingoTC:
I will try to solve this problem.

HNOONa-0:
> I will try to solve this problem.

Thanks but I'm trying to solve it at the moment.

PR #7539, opened 2024-03-29
Webpack cache busting grader directory

### LINKED ISSUE #7493: [BUG] The editor does not load properly because of "reading properties of null" exception.
## Current Behavior
Sometimes when loading a problem, the editor bugs out.

editor-error

## Expected Behavior

The editor should load normally at all times without bugging.

## If necessary, add the following information (otherwise delete this section):
*Provide a link to an example, or a detailed list of steps to reproduce this bug.*  
  
	1. go to any problem
	2. refresh a couple of times and this bug occurs
	3. 
	4. 

	* Browser name (e.g., Chrome 39, node.js 5.4):
	* Operating System and version (desktop or mobile):
	* Link to the site where this issue occurs:
	* Screenshot:

HNOONa-0:
# Description

Problem:
Browser was caching `ephemeral.js`. So, any changes made on `ephemeral.js`  will not be reflected as the browser will **return cached version** of `ephemeral.js` not updated one.
grader-cache-1
Why is it doing that for `ephemeral.js`?
I think it's because it's a large file (9 MB) so the browser decides to cache that file, but I'm not sure.
Solution:
Implement cache busting for `webpack.config-grader` file. Put simply, cache busting is a way to tell the browser **not to  cache certain files/assets when we don't want them to**, like in this scenario. What I did was I moved `index.html` and `index-light.html` to a separate directory called `templates`. Then, I use the `HtmlWebpackPlugin` to init the `index.html` and `index-light.html` with actual hashed `ephemeral.js` script. Now it works properly.

Fixes: #7524

# Comments
1) Each time we update `ephemeral.js` and new hashed script is created in `dist` directory. We need to find a way to `clean` previous hashed when creating new one.
2) Put resulting `index.html` and `index-light.html` created by the plugin in `.gitignore` file.
3) Webpack will `defer` the script, there is no `async` scriptLoading. I believe deferring the script fixes #7493
# Checklist:


RingoTC:
I think we can activate the "Disable cache" option in dev tools rather than modify the source code because cache is a useful mechanism for speeding up.

image

HNOONa-0:
> I think we can activate the "Disable cache" option in dev tools rather than modify the source code because cache is a useful mechanism for speeding up.
> 
> image

These changes will keep caching files as necessary, when compiling a file whose contents have changed the hash code will change and browser will load this new version.

If you don't make any new changes, the hash code wont change and browser will load cached version.

HNOONa-0:
This test is not deterministic it was first expecting 14 instead of 12 now its expecting 12 instead of 14. i have not touched any logic beside webpack config what is going on?

HNOONa-0:
Recap:
1) Added `index.html`, `index-light.html` to a separate directory called `template`.
2) Later use those template to init the `html` files with correct hashed version of `ephemeral.js` using `HtmlWebpackPlugin`. This is to preform cache busting.
3) Installed a new package `clean-webpack-plugin` that will delete unused scripts so that `dist` directory doesn't get over-populated with unused script files.

I installed the package using `npm`. This is a working fix. 
@pabo99 Please take a look.

pabo99:
Could you please resolve the conflicts?

HNOONa-0:
my bad i accidentally closed this pr

HNOONa-0:
> This test is not deterministic it was first expecting 14 instead of 12 now its expecting 12 instead of 14. i have not touched any logic beside webpack config what is going on? Screenshot_2024-04-04-22-31-28-65_40deb401b9ffe8e1df2f1cc5ba480b12

Hmm I think we should remove the line that validates submissions line count.

HNOONa-0:
I figured out why the this test fails. Sometimes, the modal mounts after the editor.
The modal mounts first->modal is empty->get 12 lines.
The modal mounts after the editor->editor code get copied over to the modal->test writes `main.cpp` contents-> not 12 lines anymore.

[Screencast from 04-12-2024 06:18:56 PM.webm](https://github.com/omegaup/omegaup/assets/59091536/c736e03a-d193-4a71-a130-6d90ecc9b755)

HNOONa-0:
d

HNOONa-0:
just to be transparent about the changes, here how the updated config file affect compilation:

[Screencast from 04-12-2024 08:24:30 PM.webm](https://github.com/omegaup/omegaup/assets/59091536/32f43971-0f9b-471f-93fd-96a67e1162ee)

cypress/support/pageObjects/coursePage.ts
```diff
@@ -248,7 +248,6 @@ export class CoursePage {
 
       cy.wrap(userCode).as('userCodeLines');
     });
-    cy.get('@userCodeLines').should('have.length', 12);
```
pabo99, line 251: Instead removing this line, do you think this can work:
```ts
cy.get('@userCodeLines').should('have.length.above', 11);
```
?

pabo99 (commented):
It works great! Thank you for resolving it!

Just one minor suggestion

cypress/support/pageObjects/coursePage.ts
HNOONa-0: Yes this works with either scenarios.

pabo99 (approved):
LGTM!

heduenas (approved):

Issue #7465, opened 2024-03-06
[FEATURE] [Code Reviews] Mejorar la UI/UX de la sección "Retroalimentación de un envío"


pabo99:
Como un creador de cursos, quiero que la interfaz de "Agregar retroalimentación" sea mas fluida para aprovechar de mejor forma la funcionalidad

Actualmente no encaja del todo, además de que cuenta con detalles de implementación, ya que en ese campo se guarda la solicitud de retroalimentación del usuario, tal como se puede ver en la imagen a continuación:

![image](https://github.com/omegaup/omegaup/assets/3230352/682851e8-5ec2-4aab-84a4-819fbb096a7a)

## Contexto

Para poder liberar la funcionalidad de "Code Reviews", esta parte debe quedar funcionando al 100%, debido a que las dos funcionalidades comparten la misma tabla en la base de datos

iqbalcodes6602:
also i have observed that the runs/judging for the submission is not correct, anything submitted reuslts in 100% accpetance rate even if its not correct, why is it so? do we need to fix that as well?

pabo:
I'm pabo. You want pabo99. Stop tagging me.

iqbalcodes6602:
> I'm pabo. You want pabo99. Stop tagging me.

i am really sorry

iqbalcodes6602:
hello @pabo99 can you please assign this to me?

pabo99:
FYI @iqbalcodes6602 

The intention of this issue is to make the code review feature more user-friendly. It should function similarly to code reviews on GitHub:
![FeedbackCode](https://github.com/omegaup/omegaup/assets/3230352/32882f1d-0e7c-4698-9d75-a7d196a304fd)
Once the Teaching Assistant has written all the comments/suggestions, they should press a button called "Finish review", and then a textarea should be displayed to write a general comment:
![image](https://github.com/omegaup/omegaup/assets/3230352/c61045f0-063e-4970-8ac7-62f3944ae2cd)


Currently, this textarea works independently of the feedback in the code.

shubham220420:
Hy @pabo99 is there a way to add testing data to environment, like at starting there is no initialized data in any user profile so is there a way to add them (  like contests, problems , courses for testing)?

iqbalcodes6602:
> Hy [@pabo99](https://github.com/pabo99) is there a way to add testing data to environment, like at starting there is no initialized data in any user profile so is there a way to add them ( like contests, problems , courses for testing)?

[...]

PR #7546, opened 2024-04-02
Fix showFeedbackForm sync issue in Course.vue and FeedbackCodeView.vue

iqbalcodes6602:
# Description

Added a global variable for changing the visiblity of textarea for adding review by the reviewer for feedback

Fixes: #7465

# Comments

# Checklist:


iqbalcodes6602:
hello @pabo99 
i had some doubt here, the functionning is wroking correctly just some minor fixes are remaining. When i am placing the textarea component for writing the comments it still appears at the top, else the its is working perfectly

iqbalcodes6602:
Also u think I should merge the two components for the review and textarea since they are using same array list and needs to work together?

pabo99:
> Also i think I should merge the two components for the review and textarea since they are using same array list and needs to work together?

Merging the two components seems like a good idea. How difficult do you think that proposed change might be? 

You can split the change into several parts if the number of modified lines exceeds the recommended size (200 lines)

iqbalcodes6602:
> > Also i think I should merge the two components for the review and textarea since they are using same array list and needs to work together?
> 
> Merging the two components seems like a good idea. How difficult do you think that proposed change might be? 
> 
> You can split the change into several parts if the number of modified lines exceeds the recommended size (200 lines)

Ya i can go for it I don't think it would take a lot of time, will definitely try for it.

iqbalcodes6602:
@pabo99 can u give it a look and provide some suggestion on how to fix these failing tests?

PR #7548, opened 2024-04-06
Enable editing problems when there are already maximum number of problems on `problemset`

Aritra8438:
# Description

Fix the bug where we can't update a `problem` if there are already the maximum number of problems in the `problemset`.

https://github.com/omegaup/omegaup/assets/64671908/d3c87a60-609b-41df-999d-15a1d149b53e

Fixes: #5974 

# Comments

- Added an extra condition to check if the problem is already in the `problemset`.
- If not, then validate the `MAX_PROBLEMS_IN_CONTEST` condition.

# Checklist:


Aritra8438:
Hi @pabo99. PTAL!

frontend/server/src/Controllers/Contest.php
```diff
@@ -3122,7 +3122,17 @@ public static function apiAddProblem(\OmegaUp\Request $r): array {
             );
         }
 
+        $existingProblemsetProblems = \OmegaUp\DAO\ProblemsetProblems::getByProblemset(
+            intval($problemset->problemset_id)
+        );
+
+        $existingProblemsetProblemsIds = array_map(function ($problemsetProblem): int {
+            return $problemsetProblem->problem_id;
+        }, $existingProblemsetProblems);
```
pabo99, line 3131: As I explained in the issue, we already obtain the `ProblemsetProblem` object in the `updateProblemsetProblem` function. Your code is adding an extra roundtrip to the DB. The solution would be to move
https://github.com/omegaup/omegaup/blob/a4eadba66ef725a44c83ad000e0744671e940b5f/frontend/server/src/Controllers/Problemset.php#L103-L106
(instead of `$oldProblemsetProblem` you can use `$originalProblemsetProblem`)

to this location and send the result of that function to the `updateProblemsetProblem` function as a parameter.

pabo99 (commented):
In addition to the comment I suggested, could you add a unit test in PHP to validate that this error will not occur again?

frontend/server/src/Controllers/Contest.php
Aritra8438: Done!

Aritra8438:
Done, @pabo99.
PTAL!

frontend/tests/controllers/ContestUpdateTest.php
```diff
@@ -950,7 +1070,7 @@ public function testUpdateFinishTimeInAContestWithWindowLength() {
     }
 
     /**
-     * Director extends time to user that joined the contest when it is almost is over
+     * Director extends time to user that joined the contest when it is almost over
```
pabo99, line 1073: Thanks for fixing this

frontend/tests/controllers/ContestUpdateTest.php
```diff
@@ -3,6 +3,36 @@
...
+     *
+     * @param array $problemData
+     * @param array $contestData
+     * @param \OmegaUp\Request $r
+     */
+    public static function assertProblemAddedToOrUpdatedInContest(
+        $problemData,
+        $contestData,
+        $r
+    ) {
```
pabo99, line 17: This function only utilizes the parameters 'problem alias', 'contest alias', 'points', and 'order'. Would it be possible to declare the correct type for each parameter as follows:
```php
    public static function assertProblemAddedToOrUpdatedInContest(
        $problemAlias,
        $contestAlias,
        $points,
        $order
    ) {
```
?

This way, the types for each parameter are explicitly defined.

frontend/tests/controllers/ContestUpdateTest.php
```diff
@@ -3,6 +3,36 @@
...
+        // Get problem and contest from DB
+        $problem = \OmegaUp\DAO\Problems::getByAlias(
+            $problemData['request']['problem_alias']
+        );
+        $contest = \OmegaUp\DAO\Contests::getByAlias(
+            $contestData['request']['alias']
+        );
+
+        // Get problem-contest and verify it
+        $problemset_problems = \OmegaUp\DAO\ProblemsetProblems::getByPK(
```
pabo99, line 27: Nit:
```suggestion
        $problemsetProblems = \OmegaUp\DAO\ProblemsetProblems::getByPK(
```
To maintain consistency with the other variable names

frontend/tests/controllers/ContestUpdateTest.php
```diff
@@ -123,6 +153,96 @@ public function testUpdateDefaultShowAllContestantsOption() {
...
+                $login
+            );
+
+            $r = new \OmegaUp\Request([
+                'auth_token' => $login->auth_token,
+                'contest_alias' => $contestData['contest']->alias,
+                'problem_alias' => $problemData['request']['problem_alias'],
+                'points' => 100,
+                'order_in_contest' => $i + 1,
+            ]);
```
pabo99, line 175: We have been trying to avoid the use of the $r variable in the tests. I understand that there are still many places where it is being used, but eventually, what we want is to pass the request inline. This way, we ensure that we are explicitly passing the necessary parameters:
```suggestion
            $response = \OmegaUp\Controllers\Contest::apiAddProblem(
                new \OmegaUp\Request([
                    'auth_token' => $login->auth_token,
                    'contest_alias' => $contestData['contest']->alias,
                    'problem_alias' => $problemData['request']['problem_alias'],
                    'points' => 100,
                    'order_in_contest' => $i + 1,
                ])
            );
```

frontend/tests/controllers/ContestUpdateTest.php
```diff
@@ -123,6 +153,96 @@ public function testUpdateDefaultShowAllContestantsOption() {
...
+        );
+
+        $r = new \OmegaUp\Request([
+            'auth_token' => $login->auth_token,
+            'contest_alias' => $contestData['contest']->alias,
+            'problem_alias' => $lastProblemData['request']['problem_alias'],
+            'points' => 100,
+            'order_in_contest' => MAX_PROBLEMS_IN_CONTEST,
+        ]);
+        $lastProblemResponse = \OmegaUp\Controllers\Contest::apiAddProblem($r);
```
pabo99, line 198: Same case here

frontend/tests/controllers/ContestUpdateTest.php
```diff
@@ -123,6 +153,96 @@ public function testUpdateDefaultShowAllContestantsOption() {
...
+        $r = new \OmegaUp\Request([
+            'auth_token' => $login->auth_token,
+            'contest_alias' => $contestData['contest']->alias,
+            'problem_alias' => $lastProblemData['request']['problem_alias'],
+            'points' => 50,
+            'order_in_contest' => MAX_PROBLEMS_IN_CONTEST,
+        ]);
+        $lastProblemUpdateResponse = \OmegaUp\Controllers\Contest::apiAddProblem(
+            $r
+        );
```
pabo99, line 237: And here

pabo99 (commented):
Just a few suggestions

frontend/tests/controllers/ContestUpdateTest.php
Aritra8438: Done!

frontend/tests/controllers/ContestUpdateTest.php
```diff
@@ -123,6 +155,96 @@ public function testUpdateDefaultShowAllContestantsOption() {
...
+        $lastProblemData = \OmegaUp\Test\Factories\Problem::createProblemWithAuthor(
+            $contestData['director'],
+            $login
+        );
+
+        $lastProblemResponse = \OmegaUp\Controllers\Contest::apiAddProblem(new \OmegaUp\Request([
+            'auth_token' => $login->auth_token,
+            'contest_alias' => $contestData['contest']->alias,
+            'problem_alias' => $lastProblemData['request']['problem_alias'],
+            'points' => 100,
```
heduenas, line 197: None of these values are actually changing. Can you change at least one of them?

heduenas: Nevermind, now I see that this line is adding the problem for the first time.

heduenas (approved):

Issue #7532, opened 2024-03-26
[BUG] Only One Page of latest submissions are shown


Ashish-003:
## Current Behavior
Only first 100 submissions are shown on View Problem Submssions page
![image](https://github.com/omegaup/omegaup/assets/73078401/8da5c8fc-9606-4bb4-a49f-376b26fb0270)

## Expected Behavior

Should contain more pages for submissions in last 24 hours.

## If necessary, add the following information (otherwise delete this section):
*Provide a link to an example, or a detailed list of steps to reproduce this bug.*  
  
	1. 
	2. 
	3. 
	4. 
*Relevant details about the environment in which you encountered this bug:*

	* Browser name (e.g., Chrome 39, node.js 5.4):
	* Operating System and version (desktop or mobile):
	* Link to the site where this issue occurs:
	* Screenshot:

Ashish-003:
@pabo99 It would be better to go with pagination right rather than returning a list of submissions

arnabnath01:
I think using infinity scrolling could resolve this.

pabo99:
+1 to the infinity scroll

aditi-star123:
@pabo99 please assign me this issue

PR #7553, opened 2024-04-10
Issue#7532 | Added Infinite Scroll

Ashish-003:
# Description

This is the submission loading UI
[Screencast from 23-04-24 05:13:59 PM IST.webm](https://github.com/omegaup/omegaup/assets/73078401/2b20485c-2fbf-404c-93d8-9aac541ac7b4)

Fixes: #7532 

# Comments

# Checklist:


frontend/www/js/omegaup/components/submissions/List.vue
```diff
@@ -17,7 +22,7 @@
       <div v-if="includeUser" class="card-body d-flex align-items-center">
         <omegaup-common-typeahead
           :existing-options="searchResultUsers"
-          :value.sync="searchedUsername"
+          :value.sync="searchedUsernameLocal"
```
pabo99, line 25: What is the reason for modifing this variable name? If there's no reason, could you please restore it as originally was

pabo99 (commented):
This PR will be blocked until PR #7555 or #7558 is approved and merged with main branch.

frontend/www/js/omegaup/components/submissions/List.vue
Ashish-003: Yeah I was trying something, but forgot to restore it back

pabo99 (changes requested):
With all the changes proposed in your PR, we have encountered an issue that can be easily resolved. Let me explain:

Previously, we rendered the submission list once, which was fine. However, now we render it every time we reach the bottom of the page and the API returns more submissions. Consequently, the iterator has a key. The purpose of using the key attribute is to assist Vue.js in efficiently updating the DOM when the underlying data changes.

Currently, we are using an index that starts from 0 every time we retrieve more data from the API.

The solution is to use an identifier for this key. Currently, we have a GUID for every submission created, which is passed from the server. All we need to do is include it in the type declarations to ensure it is recognized by TypeScript.

Don't forget to run the command:
```
./stuff/lint.sh
```
to make the changes visible in TypeScript.

frontend/www/js/omegaup/components/submissions/List.vue
```diff
@@ -1,5 +1,10 @@
 <template>
-  <div submissions-problem>
+  <div
+    v-infinite-scroll="() => $emit('fetch-more-data')"
```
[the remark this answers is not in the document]
Ashish-003: I thought we used arrow function when an event was triggered otherwise it takes it to be a method and invokes it when template is rendered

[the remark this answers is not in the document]
pabo99: You are right, the way you are doing it is correct. So let me mark this conversation as resolved.

Ashish-003:
> With all the changes proposed in your PR, we have encountered an issue that can be easily resolved. Let me explain:
> 
> Previously, we rendered the submission list once, which was fine. However, now we render it every time we reach the bottom of the page and the API returns more submissions. Consequently, the iterator has a key. The purpose of using the key attribute is to assist Vue.js in efficiently updating the DOM when the underlying data changes.
> 
> Currently, we are using an index that starts from 0 every time we retrieve more data from the API.
> 
> The solution is to use an identifier for this key. Currently, we have a GUID for every submission created, which is passed from the server. All we need to do is include it in the type declarations to ensure it is recognized by TypeScript.
> 
> Don't forget to run the command:
> 
> ```
> ./stuff/lint.sh
> ```
> 
> to make the changes visible in TypeScript.

Ashish-003:
> With all the changes proposed in your PR, we have encountered an issue that can be easily resolved. Let me explain:
> 
> Previously, we rendered the submission list once, which was fine. However, now we render it every time we reach the bottom of the page and the API returns more submissions. Consequently, the iterator has a key. The purpose of using the key attribute is to assist Vue.js in efficiently updating the DOM when the underlying data changes.
> 
> Currently, we are using an index that starts from 0 every time we retrieve more data from the API.
> 
> The solution is to use an identifier for this key. Currently, we have a GUID for every submission created, which is passed from the server. All we need to do is include it in the type declarations to ensure it is recognized by TypeScript.
> 
> Don't forget to run the command:
> 
> ```
> ./stuff/lint.sh
> ```
> 
> to make the changes visible in TypeScript.

The PR still shows 1 requested change. Can you mark the comment as resolved

frontend/www/js/omegaup/submissions/list.ts
```diff
@@ -38,6 +46,31 @@ OmegaUp.on('ready', () => {
               })
               .catch(ui.apiError);
           },
+          'fetch-more-data': (username: string) => {
+            if (this.loading || this.endOfResults) return;
+            this.loading = true;
+            this.searchedUsername = username ?? null;
+
+            api.Submission.list({
+              username: this.searchedUsername,
```
pabo99, line 55: Here we could directly obtain the username from the `payload`:
```suggestion
              username: payload.username,
```
For this, it would be necessary for you to add this data in the `getLatestUserSubmissionsForTypeScript` function right where the `payload` object is.

Additionally, you need to add this new field to the `SubmissionsListPayload` type declaration. The field to add should be something like: `username?: string`.

pabo99 (commented):
I noticed that there were still some suggestions to be resolved. I added a few comments; with them, it seems like my part is done.

Thank you for working on this issue.

frontend/www/js/omegaup/components/submissions/List.vue
```diff
@@ -1,5 +1,10 @@
 <template>
-  <div submissions-problem>
+  <div
+    v-infinite-scroll="() => $emit('fetch-more-data', submissions[0].username)"
```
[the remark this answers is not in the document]
Ashish-003: My bad for carelessness, 
I overlooked this.
Resolved it though

Ashish-003:
> I noticed that there were still some suggestions to be resolved. I added a few comments; with them, it seems like my part is done.
> 
> Thank you for working on this issue.

Did it, I just couldn't leave things halfway.

pabo99:
> > I noticed that there were still some suggestions to be resolved. I added a few comments; with them, it seems like my part is done.
> > Thank you for working on this issue.
> 
> Did it, I just couldn't leave things halfway.

We really appreciate it! Thank you so much!

frontend/server/src/Controllers/Submission.php
```diff
@@ -7,7 +7,7 @@
  *
  * @psalm-type PageItem=array{class: string, label: string, page: int, url?: string}
  * @psalm-type Submission=array{time: \OmegaUp\Timestamp, username: string, school_id: int|null, school_name: string|null, alias: string, title: string, language: string, verdict: string, runtime: int, memory: int}
- * @psalm-type SubmissionsListPayload=array{includeUser: bool, submissions: list<Submission>}
+ * @psalm-type SubmissionsListPayload=array{includeUser: bool, username: string|null, submissions: list<Submission>}
```
pabo99, line 10: Ok, it looks like there's an error in the type declaration that was just added. Let me explain the reason:

There is a difference between the type `username: string|null` and the type `username?: string`. While the first indicates that it will always be present even if its value is `null`, the second indicates that the element may or may not exist.

In this case, the `SubmissionsListPayload` declaration is shared between the two methods we've been using. In `getLatestSubmissionsForTypeScript`, the `username` is not present, but in `getLatestUserSubmissionsForTypeScript`, the `username` is sent.

There are two possible solutions:

Add `'username' => null` in the first function.
Declare this type as `username?: string`.

Both solutions should work.

pabo99 (commented):
One more suggestion!

frontend/server/src/Controllers/Submission.php
Ashish-003: Jinxed, I was just changing that

pabo99 (approved):
LGTM!

heduenas (approved):

PR #7557, opened 2024-04-12
Add language from payload to grader and use it on mount

HNOONa-0:
# Description

I added `preffered_langauge` from payload as a prop to details, passed it to `EphemeralGrader.vue` and select the preferred language on mount after the `iframe` loads.
video (ends at 0:53):

[Screencast from 04-12-2024 03:56:17 PM.webm](https://github.com/omegaup/omegaup/assets/59091536/b0340fc2-5a93-420e-b3c8-726fcda21a30)

Note that you have to close the problem tab and open it again to reflect changes in preferred language. That's because of the `sessionsStorage` variable.

Fixes: #7554

# Comments

# Checklist:


HNOONa-0:
Hello @pabo99, please check this out.

frontend/www/js/omegaup/components/problem/Details.vue
```diff
@@ -447,6 +448,7 @@ export default class ProblemDetails extends Vue {
   @Prop() problem!: types.ProblemInfo;
   @Prop() solvers!: types.BestSolvers[];
   @Prop() user!: types.UserInfoForProblem;
+  @Prop() userPreferredLanguage!: null | string;
```
pabo99, line 451: This property is no longer needed since we already store the user's preferred language. Actually in this file we have a computed property called `preferredLanguage` to obtain this value, you can use it for this purpose. According the PR #7386, the hierarchical order to define the programming language to display in the dropdown is the following:

- Programming language used in the last run submited to this problem
- If the user haven't submitted any run for this problem, then the selected language is the preferred one chosen by the user in the Preferences of their profile

pabo99 (commented):
Thank you for addressing this issue.

In addition to the suggested comment, I propose that this PR be considered for the inclusion of a Cypress test with the following steps:

1. The user selects a preferred language in their Profile Preferences.
2. They then navigate to a problem and validate that the selected language in the dropdown matches their previous preference.
3. The user changes the language and submits a new solution.
4. Upon returning to the same problem, they verify that the selected language now reflects the one used in their latest submission

HNOONa-0:
> Thank you for addressing this issue.
> 
> In addition to the suggested comment, I propose that this PR be considered for the inclusion of a Cypress test with the following steps:
> 
> 1. The user selects a preferred language in their Profile Preferences.
> 2. They then navigate to a problem and validate that the selected language in the dropdown matches their previous preference.
> 3. The user changes the language and submits a new solution.
> 4. Upon returning to the same problem, they verify that the selected language now reflects the one used in their latest submission

I added the test in this pr I did not pay attention to the comment about a separate pr, my bad. I can add it in a separate pr if you would like that.
Here is a recap of what I did:
1) I deleted `preferredUserLangauge` attribute and instead used `preferredLanguage` computed property.
2) I `defer`d the `grader_ephemeral` scripts inside of `index` and `index-light`. This is because I kept getting null pointer exception errors when adding the new test (fixes: #7493)
Screenshot from 2024-04-13 17-27-13
3) I added a new method that would updated `preferredLanguage`. There is already an existing method that updates all preferences for a user but that one validates the preferred language to always be `cpp-17`, I did not want to modify it so I just added this function instead.
4) I added the test inside of `coursePage.cy.ts` file as it felt like the closest place.

HNOONa-0:
Im having troubles debugging sessionStorage what does it do and what is it supposed to do?

HNOONa-0:
There was a logical error with computing `preferredLanguage`, the problem runs are not sorted by time.

HNOONa-0:
@pabo99 Here is a video to help with reviewing:

[preferred-language.webm](https://github.com/omegaup/omegaup/assets/59091536/c7402536-e1f4-420d-b260-7952406b7573)

frontend/www/js/omegaup/components/problem/Details.vue
```diff
@@ -523,9 +524,17 @@ export default class ProblemDetails extends Vue {
   }
 
   get preferredLanguage(): null | string {
-    const [lastRun] = this.runs.slice(-1);
-    if (lastRun) {
-      return lastRun.language;
+    // this will not work, runs are not fetched orderly by time
+    // const [lastRun] = this.runs.slice(-1);
```
pabo99, line 528: Since this code is no longer needed, we prefer not to comment unused code. So, can you remove it please?
```suggestion
```

frontend/www/js/omegaup/grader/ephemeral.js
```diff
@@ -364,6 +366,7 @@ let store = new Vuex.Store({
           languageExtensionMapping[state.sessionStorageSources.language]
         ];
       document.getElementById('language').value = state.request.language;
+      document.getElementById('language').dispatchEvent(new Event('change'));
```
pabo99, line 369: Here you are using the same element twice, could you use a variable (or constant) instead?

pabo99 (commented):
Just a few comments. Additionally, there are conflicts in some files. Could you resolve those as well?

pabo99 (approved):
LGTM!

HNOONa-0:
@heduenas I think this is ready for merging

heduenas (approved):
Thank you @HNOONa-0  for going the extra mile by adding a cypress test!

PR #7567, opened 2024-04-17
Modify activation threshold so that every problem may appear. 

Aritra8438:
# Description

Please include a description of the changes you have made. The text you put in this section will be used as the **commit message** when this PR is merged.

Fixes: #(issue)

# Comments

# Checklist:

### DIFF
--- a/frontend/www/js/omegaup/components/contest/AddProblem.vue
+++ b/frontend/www/js/omegaup/components/contest/AddProblem.vue
@@ -27,7 +27,7 @@
               <omegaup-common-typeahead
                 class="w-75"
                 :existing-options="searchResultProblems"
-                :activation-threshold="2"
+                :activation-threshold="1"
                 :value.sync="alias"
                 @update-existing-options="
                   (query) =>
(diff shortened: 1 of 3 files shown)


Aritra8438:
@pabo99, can you look into this pr once?

pabo99 (commented):
I noticed that we have a default value for the activation treshold parameter:
https://github.com/omegaup/omegaup/blob/21b47590c41bc9532a524b1b60c206853e7ffb90/frontend/www/js/omegaup/components/common/Typeahead.vue#L36

So,  could you review if we need to set the correct activation treshold value in other places where we use this component?

pabo99 (approved):
Defer to @heduenas 

What do you think about set the treshold activation value as "1"?

Aritra8438:
@heduenas, do you have any suggestion on this problem?

Aritra8438:
@heduenas, this issue.

heduenas:
Sorry that I missed this earlier. I think it would be better to force problem names and aliases to have at least 2 letters at creation time.

Aritra8438:
Okay, will do that @heduenas! However, there might be some problems with both alias and id as single characters.

Aritra8438:
This PR should not be merged, according to the suggestions, we might want to 
>restrict users to create a problem with both the problem name and alias set to a single character.

Issue #7570, opened 2024-04-19
[BUG] Feedback dismissal doesn't work on courses


heduenas:
Steps to reproduce:
1. Go to https://omegaup.com/course/introduccion_a_cpp/assignment/hola_mundo#problems/Hola-Mundo-c
2. Make any submission
3. Click the "X" to close the feedback modal.
4. Refresh the page and go to the problem again.
5. See the modal show up again.

5. Shouldn't happen once you have done 3.

HNOONa-0:
Hello @heduenas, I can work on this.

HNOONa-0:
Hmmm I could not reproduce this issue.

[a1.webm](https://github.com/omegaup/omegaup/assets/59091536/265824c0-38da-47d0-aa48-6d9026c2dc70)

heduenas:
@HNOONa-0 have you been able to reproduce the issue?

HNOONa-0:
> @HNOONa-0 have you been able to reproduce the issue?

Yes I did I actually managed to fix it it was a small silly error, I will try to make a pr now

PR #7571, opened 2024-04-19
emit dismiss event with arguments properly

HNOONa-0:
# Description

It was a small error where the `dismiss-promotion` event was not emitted properly. The event handler takes in **three separate parameters not a single object**.
Emitting before changes -> `handler({solved:true,tried:false,isDismissed:true}, undefined, undefined)`
After -> `handler(true, false, true)`
Here is a video proving that it works:

https://github.com/omegaup/omegaup/assets/59091536/aae1b904-3e4d-4552-948c-3ecee60e34e1

Fixes: #7570

# Comments

# Checklist:


HNOONa-0:
@heduenas Please take a look, is this what you meant?

frontend/www/js/omegaup/components/arena/Course.vue
```diff
@@ -94,11 +94,12 @@
               "
               @dismiss-promotion="
                 (qualityPromotionComponent, isDismissed) =>
-                  $emit('dismiss-promotion', {
-                    solved: qualityPromotionComponent.solved,
-                    tried: qualityPromotionComponent.tried,
+                  $emit(
```
pabo99, line 97: Alright, it's currently working fine. 

However, we'd like to establish a rule for emitting actions between components or to TypeScript files:

When passing more than one parameter in the `emit` function, we prefer to send a single object containing all the necessary parameters.

So, instead of these changes, please locate the method and modify it to accept only one parameter. It appears that the issue lies here:
https://github.com/omegaup/omegaup/blob/a216eeb481485854d980035ddd97f4068e20cef3/frontend/www/js/omegaup/arena/course.ts#L393-L397

If you can see, the method is expecting 3 parameters and we want to the function receives only one object with all of those parameters. You can do something like this:
https://github.com/omegaup/omegaup/blob/a216eeb481485854d980035ddd97f4068e20cef3/frontend/www/js/omegaup/arena/course.ts#L356-L366

pabo99 (approved):
LGTM

heduenas (approved):

Issue #7628, opened 2024-06-01
[BUG] Error thrown when switching to certain languages and re-loading the problem


HNOONa-0:
## Current Behavior

When switching to specific languages like `Kotlin`, an error is thrown. This error does not occur when we switch to a language like `c++`. Below are steps on how you can reproduce this bug.
![Screenshot from 2024-06-01 23-30-12](https://github.com/omegaup/omegaup/assets/59091536/ef00551e-10c6-41ba-8678-bd604f277a85)

## Expected Behavior

Should not throw for any language we support. I suspect it's because it's missing defined template **(initial piece of code for that language)** from template objects. 

## If necessary, add the following information (otherwise delete this section):
*Provide a link to an example, or a detailed list of steps to reproduce this bug.*  
  
	1. Go to any problem
	2. Switch to `Kotlin` language
	3. Reload the problem page
	4. Check console for error

PR #7635, opened 2024-06-03
Fix switch language error by adding remaining language templates

HNOONa-0:
# Description
The problem was that `sourceTemplates` for some languages were not defined, and so when trying to use `sourceTemplates` to init the editor with code, we get `undefined` as a result.
switch-language-bug

So I added the remaining templates and issue has been fixed.
Fixes: #7628

# Comments

# Checklist:


pabo99:
For interactive problems, the error remains the same:

image

Are there any templates missing?

HNOONa-0:
> For interactive problems, the error remains the same:
> 
> image
> 
> Are there any templates missing?

You actually need to clear application memory because that's where templates are stored. Even though templates are updated after recompilation changes are not reflected because old copy of templates persists.
Could you please try again?

heduenas (approved):

heduenas:
I think your change looks good as is and I don't think we need to add `idl` to the list of supported languages.

pabo99 (approved):

Issue #7640, opened 2024-06-04
Define remaining functionality from `util.js` inside `util.ts`


HNOONa-0:
There are some functions inside of `util.js` not present in `util.ts`. We should add **typed** version of these functions over to `util.ts`.

HNOONa-0:
Hey @pabo99, would you like to continue merging this [pr](https://github.com/omegaup/omegaup/pull/7489) or should I start a new one?

pabo99:
> Hey @pabo99, would you like to continue merging this [pr](https://github.com/omegaup/omegaup/pull/7489) or should I start a new one?

You can start a new PR since it has no revisions; it's okay

PR #7634, opened 2024-06-02
Create initial grader Typescript Vuex store

HNOONa-0:
# Description

I made an initial interface with attributes I noticed. I will discover remaining attributes and add them as well.
Then I will add other logic parts of Vuex store.

Part of: #7633

# Comments

Add any comments for the reviewers (e.g., indicating the beginning of the revision, something where the reviewer needs to pay special attention, etc.). If  there are no extra comments this section can be deleted.

# Checklist:

- [x] The code follows the [coding guidelines](https://github.com/omegaup/omegaup/wiki/Coding-guidelines) of omegaUp.
- [ ] The tests were executed and all of them passed.
- [ ] If you are creating a feature, the new tests were added.
- [ ] If the change is large (> 200 lines), this PR was split into various Pull Requests. It's preferred to create one PR for changes in controllers + unit tests in PHPUnit,  and then another Pull Request for UI + tests in Jest, Cypress or both.

PR #7641, opened 2024-06-05
Implement util.js functions inside util.ts

HNOONa-0:
# Description
Implement typed `util.js` functions for upcoming migration.

Fixes: #7640

# Comments

# Checklist:


frontend/www/js/omegaup/grader/util.ts
```diff
@@ -1,3 +1,49 @@
+export function vuexGet(store: any, name: string) {
```
pabo99, line 1: Is there any way to avoid using `any` type and instead use the correct type, both in this function and in the vuexSet function?

HNOONa-0: Yeah but the `store` interface is defined in this [pr](https://github.com/omegaup/omegaup/pull/7634) which has not been merged yet. I have stopped working on it because I'm still finding improvement and small bug fixes with current `store` implementation, take a look [here](https://github.com/omegaup/omegaup/pull/7634).

heduenas (approved):

Issue #7655, opened 2024-06-14
[FEATURE] Increase height of editor. Zoom in/out of code.


HNOONa-0:
The IDE height is currently too small. We can increase its height to make it more comfortable to use.
The IDE should also zoom in/out as needed to display an appropriate number of lines of code. Here is a video of what I mean, on other platforms, zooming out allows us to see more lines of code. However, this is not the case with our IDE, it's the same number of lines despite scaling.

[Screencast from 2024-06-14 14-04-42.webm](https://github.com/omegaup/omegaup/assets/59091536/04c30bfc-fbeb-42a7-ba38-2738edd1550a)

PR #7656, opened 2024-06-14
Increase editor height and dynamically scale it.

HNOONa-0:
# Description

I increased `min-width` of `iframe` to `1.5` its original size.
I added an event listener to `resize` event that would scale content of `iframe` based on `devicePixelRatio`. works on `firefox` and `chrome`. Here is a video:

[Screencast from 2024-06-14 16-37-12.webm](https://github.com/omegaup/omegaup/assets/59091536/c7a16470-8ba5-4c88-8db7-34159ddc9148)

Fixes: #7655

# Comments

This fix will be altered a little bit when we eventually get rid of the `iframe`.

# Checklist:


frontend/www/js/omegaup/components/arena/EphemeralGrader.vue
```diff
@@ -21,8 +21,22 @@ export default class EphemeralGrader extends Vue {
...
   }
+  handleWindowResize(): void {
+    const iframe = this.$refs.grader as HTMLIFrameElement;
 
+    // zoom in/out manually using device pixel ratio
+    // zoom property was recently supported on firefox browser
+    // so this approach works on most browsers
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    iframe.style.zoom = 1 / window.devicePixelRatio;
```
pabo99, line 38: Will it be possible to avoid the comment that ignores the TypeScript error?

If the problem is due to zoom not being available for all browsers, you can add an interface:

```ts

interface CSSStyleDeclaration {
  zoom?: string | number;
}
```

Additionally, you can indicate it here as follows:
```suggestion
    (iframe.style as CSSStyleDeclaration).zoom = String(
      1 / window.devicePixelRatio,
    );
```

pabo99 (commented):
Just two minor changes

frontend/www/grader/ephemeral/templates/index-light.html
```diff
@@ -1,5 +1,5 @@
 <!doctype html>
-<html class="h-75">
+<html class="h-100 border">
```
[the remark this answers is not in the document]
HNOONa-0: Yes. I have noticed the editor weird behavior when resizing and scaling with that styling, this is the best configuration I found that works. This is what I mean by weird behavior:

[Screencast from 2024-06-11 12-37-00.webm](https://github.com/omegaup/omegaup/assets/59091536/ee735614-6c17-4a59-9bd0-17a0506355fc)

frontend/www/js/omegaup/components/arena/EphemeralGrader.vue
HNOONa-0: Yes this makes sense.

frontend/www/grader/ephemeral/templates/index-light.html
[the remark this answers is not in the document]
pabo99: For now, let's leave it as it is then. Ideally, these templates should be replaced by a single Vue component that is prepared to support both modes.

frontend/www/js/omegaup/components/arena/EphemeralGrader.vue
```diff
@@ -7,6 +7,9 @@
 </template>
 
 <script lang="ts">
+interface CSSStyleDeclaration {
```
pabo99, line 10: Could you please move this interface declaration after the import statements? This is the standard we are using in other components:

https://github.com/omegaup/omegaup/blob/c8176969b063243edfb2619d806cb7f8cbd0046e/frontend/www/js/omegaup/components/problem/FinderWizard.vue#L84-L100

pabo99 (commented):
Just one minor change

pabo99 (approved):
Thank you!

heduenas (approved):

Issue #7657, opened 2024-06-14
[BUG] Deleting a case does not completely remove it


HNOONa-0:
## Current Behavior
Deleting a case does not delete it entirely, the current case data still persists. Here is a video of what I mean where I create a new case and delete it:

[Screencast from 2024-06-14 16-57-58.webm](https://github.com/omegaup/omegaup/assets/59091536/302421ad-b609-4fef-9bf1-91bca158b67f)

## Expected Behavior

Before deleting a case, make sure to switch to a different currently available case.

PR #7661, opened 2024-06-18
Switch to a different case when deleting one

HNOONa-0:
# Description

Before deletion of a case set current case to a random `case` in `cases` object. 

Fixes: #7657

# Checklist:


frontend/www/js/omegaup/grader/GraderStore.js
```diff
@@ -446,6 +446,18 @@ let store = new Vuex.Store({
...
         return;
+
+      let keys = Object.keys(store.state.request.input.cases);
+
+      // do not delete if there is only one test case
+      if (keys.length === 1) {
+        return;
+      }
+
+      // switch to a random case
```
pabo99, line 457: Nit:
```suggestion
      // switch to the closest available test case
```

pabo99 (commented):
Do you think it's possible to apply this same change in the TypeScript file? If this function hasn't been migrated yet, find a way to add a note so that we don't forget to include the change.

HNOONa-0:
> Do you think it's possible to apply this same change in the TypeScript file? If this function hasn't been migrated yet, find a way to add a note so that we don't forget to include the change.

Yes but I'm delaying it on purpose. I would like to first make all possible improvements to the `store` then add all those improvements in one go to Typescript version of the `store` in this [pr](https://github.com/omegaup/omegaup/pull/7634).

frontend/www/js/omegaup/grader/GraderStore.js
HNOONa-0: Cases are not sorted and when you insert new ones, the order might not be what you expect. That is why I wrote random case. I wanted to type a comment like that but I imagined that it would be confusing to future readers .

pabo99: It's okay! Let me mark this conversation as resolved

pabo99:
> > Do you think it's possible to apply this same change in the TypeScript file? If this function hasn't been migrated yet, find a way to add a note so that we don't forget to include the change.
> 
> Yes but I'm delaying it on purpose. I would like to first make all possible improvements to the `store` then add all those improvements in one go to Typescript version of the `store` in this [pr](https://github.com/omegaup/omegaup/pull/7634).

Excellent!

heduenas (approved):

Issue #7687, opened 2024-07-03
[FEATURE] Migrate TextEditorComponent to Typescript


HNOONa-0:
Refactor the `TextEditorComponent` to TypeScript in preparation for upcoming migration.

PR #7688, opened 2024-07-03
refactor text editor to class vue and typescript

HNOONa-0:
# Description

Fixes: #7687

# Checklist:


frontend/www/js/omegaup/graderv2/TextEditorComponent.vue
```diff
@@ -0,0 +1,65 @@
+<template>
```
HNOONa-0, line 1: https://github.com/omegaup/omegaup/blob/2bc420a29849fb5c354534ecf29d6335eb00a07a/frontend/www/js/omegaup/grader/TextEditorComponent.vue#L1-L83

frontend/www/js/omegaup/graderv2/TextEditorComponent.vue
```diff
@@ -0,0 +1,65 @@
...
+
+  get title(): string {
+    return this.filename;
+  }
+}
+</script>
+
+<style scoped>
+.textarea.vs-dark {
+  background: #222;
```
pabo99, line 53: Can you add this color to the `_variables.scss` file?

frontend/www/js/omegaup/graderv2/TextEditorComponent.vue
pabo99: What's the difference between this component and `MonacoEditorComponent`?

If you can add a screenshot or a screencast it would be great

frontend/www/js/omegaup/graderv2/TextEditorComponent.vue
HNOONa-0: `TextEditorComponent` is not actually an editor, it's those two readonly `textareas` at the end of the editor that display run information and errors as necessary. Maybe we should rename this component. 
image

frontend/www/js/omegaup/graderv2/TextEditorComponent.vue
HNOONa-0: `TextEditorComponent` is not actually an editor, it's those two readonly `textareas` at the end of the editor that display run information and errors as necessary. Maybe we should rename this component. 
image

pabo99: Oh, I got it now. It seems that the textarea allows text editing with the disabled property, so I don't see the need to change the component name (just a comment I made in another PR about removing the `Component` suffix). Thanks for the screenshot.

pabo99 (approved):
LGTM

heduenas (approved):

Issue #7700, opened 2024-07-07
Fix all the buttons on the problem creator page.


Aritra8438:
There are 3 buttons related to the zip file:
- [x] https://github.com/omegaup/omegaup/pull/7701
- [x] #7702
- [x] #7779 

We should make them work.

PR #7701, opened 2024-07-07
Enable users to reset the creator page

Aritra8438:
Add the logic to:
- Reset store.
- Reset CasesStore
- Reload the page.

Related to: #7700

https://github.com/omegaup/omegaup/assets/64671908/331062b7-21ae-453a-ae88-4705911ff98c

# Checklist:


frontend/www/js/omegaup/components/problem/creator/Header.vue
```diff
@@ -21,7 +21,7 @@
           {{ T.problemCreatorGenerateProblem }}</span
         >
       </b-button>
-      <b-button variant="warning" size="sm">
+      <b-button variant="warning" size="sm" @click="createNewProblem()">
```
carlosabcs, line 24: You need to pass a function reference, not a function call as the callback for the click event.
```suggestion
      <b-button variant="warning" size="sm" @click="createNewProblem">
```

Aritra8438: Hi @carlosabcs, may you look into this SO [answer](https://stackoverflow.com/questions/62828904/vue-testing-jest-button-triggerclick-not-working#:~:text=i%20have%20been%20struggling%20with%20this%20as%20well%2C%20it%20seems%20sometimes%20the%20test%20component%20has%20a%20hard%20time%20finding%20the%20emitting%2Fcalls%20to%20functions%20that%20do%20not%20have%20parentheses.)? For testing purposes, I added the parenthesis.

carlosabcs: Before accepting this, and because it looks weird. Could you try using `mount` instead of `shallowMount`? See reference [here](https://stackoverflow.com/questions/51536537/running-jest-with-bootstrap-vue)

carlosabcs: However, maybe you need to add:
`localVue.use(BootstrapVue)`.

Or did you try this: https://stackoverflow.com/a/53598621? Please try other options and if they don't finally work. Then let's use what you did because passing a function call seems uncommon.

frontend/www/js/omegaup/components/problem/creator/Header.test.ts
```diff
@@ -26,4 +27,47 @@ describe('Header.vue', () => {
...
+
+  it('Should reset the store on clicking the reset button', async () => {
+    const original = window.location;
+
+    Object.defineProperty(window, 'location', {
+      configurable: true,
+      value: { reload: jest.fn() },
+    });
+
+    const wrapper = mount(Header, { localVue, store });
```
Aritra8438, line 39: I have used `mount` only.

frontend/www/js/omegaup/components/problem/creator/Header.test.ts
```diff
@@ -10,7 +11,7 @@ localVue.use(IconsPlugin);
```
Aritra8438, line 11: Also I have used
```js
localVue.use(BootstrapVue);
```

Aritra8438 (commented):
Hi @carlosabcs, PTAL!

carlosabcs (approved):
LGTM

Aritra8438:
@heduenas, PTAL!

frontend/www/js/omegaup/components/problem/creator/Header.vue
carlosabcs: The problem with `shallowMount` is that it only renders the current component, the nested components are not "compiled". That's why BootstrapVue component's are not being correctly rendered.

heduenas (approved):

Issue #7710, opened 2024-07-12
[FEATURE] Add `ephemeral.ts` to manually test typescript components


HNOONa-0:
To test the recently approved typescript components CaseSelector, MonacoEditor, etc.., it would be easier to integrate them into typescript version of `ephemeral.js`.

The file `ephemeral.ts` will be largely similar to `ephemeral.js`. We won't spend time typing the code inside it because many sections will be removed, such as query selection of elements, when we eventually create an unembedded IDE. Instead, we will rely on Vue events, references, and directives to handle most of these functionalities.

PR #7711, opened 2024-07-12
initial ephemeral ts

HNOONa-0:
# Description
I downloaded the `goldenLayout` package  instead of using a script tag for importing. I imported the newly merged TypeScript components, everything is working as intended. I modified the components to import the Vuex `store` variable directly rather than passing it as a prop.
I added a new grader config file to compile the typescript version of the components.

https://github.com/user-attachments/assets/071c8820-4bcd-420a-9be2-bde5281af145

Fixes: #7710

# Checklist:


HNOONa-0:
@pabo99 PTAL!

HNOONa-0:
@pabo99 Why do we check for attribute presence using this method?
https://github.com/omegaup/omegaup/blob/ba91f88d551b84ec5b34199daaf33d60cc6b25ef/frontend/www/js/omegaup/grader/GraderStore.js#L279-L281
Are there advantages to using this method over **optional chaining**?

pabo99:
> @pabo99 Why do we check for attribute presence using this method?
> 
> https://github.com/omegaup/omegaup/blob/ba91f88d551b84ec5b34199daaf33d60cc6b25ef/frontend/www/js/omegaup/grader/GraderStore.js#L279-L281
> 
> 
> Are there advantages to using this method over **optional chaining**?

I guess that code is legacy, so using the **optional chaining** method should be the best alternative in this case.

HNOONa-0:
@pabo99 PTAL! This is an initial file that will help me work on other issues and testing with the store.

pabo99 (approved):
It seems that most of the comments will be related to the data types in the `ephemeral.ts` file, but since adding them is not being considered in this change, this PR can be merged into the main branch to unblock the following PRs you will be working on.

pabo99:
PTAL @heduenas !

heduenas (approved):

Issue #7712, opened 2024-07-12
[FEATURE] Create two separate golden layout configurations


HNOONa-0:
Create two separate layout for IDE, one for embedded layout the other for full screen layout.
Right now, there is only one configuration for the full screen editor, which we have to dynamically remove elements from if we are in embedded editor:

https://github.com/omegaup/omegaup/blob/9d8826f3581f2bfea0c299a97eb0d743e957fdba/frontend/www/js/omegaup/grader/ephemeral.js#L317-L362

We can just create two separated configs without having to remove and insert in our layout.

PR #7725, opened 2024-07-21
Add `EMBEDDED` and `UNEMBEDDED` configurations for golden layout

HNOONa-0:
# Description

Added a new file `GoldenLayoutConfigs` that contain both of these configurations. We can import these configs anywhere, inside `ephemeral.ts` I used the `isEmbedded` flag to decided on which one to use. Changes are working as intended.
 
Fixes: #7712

# Comments
We can eliminate some repetition of code in this file.
# Checklist:


HNOONa-0:
- [ ] #7711

☝️ Should be merged first.

frontend/www/js/omegaup/graderv2/GoldenLayoutConfigs.ts
```diff
@@ -0,0 +1,314 @@
+import GoldenLayout from 'golden-layout';
+
+export const TEXT_EDITOR_COMPONENT_NAME = 'text-editor-component';
```
HNOONa-0, line 3: @pabo99 I added extra code that would eliminate some of the redundancy between the two configs, I could do that for the rest of the components should I continue?

HNOONa-0:
@pabo99 PTAL!

frontend/www/js/omegaup/graderv2/GoldenLayoutConfigs.ts
pabo99: Sure, you can continue. 

However, I have a question: what code would you eliminate in the rest of the components?

Because what I understand is that we only have the two configurations you mentioned: for the embedded component and for the unembedded component.

frontend/www/js/omegaup/graderv2/ephemeral.ts
```diff
@@ -314,48 +113,18 @@ RegisterVueComponent(
...
-
-    // Also extend the case selector column a little bit so that it looks nicer.
-    caseSelectorColumn.container.setSize(Math.max(160, oldWidth), oldHeight);
-
     // Whenever a case is selected, show the cases tab.
     store.watch(
       Object.getOwnPropertyDescriptor(store.getters, 'currentCase').get,
       (value) => {
         if (store.getters.isUpdatingSettings) return;
+        let casesColumn = layout.root.getItemsById('cases-column')[0];
```
pabo99, line 127: Thank you for moving this line closer to where it will be used. Just a small comment: when a value is not going to change, we can use `const` instead of `let`.
```suggestion
        const casesColumn = layout.root.getItemsById('cases-column')[0];
```

pabo99 (approved):

frontend/www/js/omegaup/graderv2/GoldenLayoutConfigs.ts
HNOONa-0: > Sure, you can continue.
> 
> However, I have a question: what code would you eliminate in the rest of the components?
> 
> Because what I understand is that we only have the two configurations you mentioned: for the embedded component and for the unembedded component.

Oh I only meant to reduce redundancy not remove components, for example, the same `zip-viewer-component` is duplicated twice, what I did instead is create a function that would return that component object and call it instead copy pasting the components in both configs.

pabo99: Oh, I get it now. It looks better.

Thank you for reducing the redundant code.

One more question: How close are we to adding the types in this file?

I see that this PR has already grown enough, so I think it can be approved as it is.

HNOONa-0: > One more question: How close are we to adding the types in this file?
You mean adding types for each of the small components? `ComponentConfig` seems to be the appropriate type, we get intellisense for each of the components:

image

Issue #7713, opened 2024-07-16
[FEATURE] Move component logic to Vuex actions


HNOONa-0:
Currently, several components in our IDE directly perform state mutation operations on the Vuex store directly.

To improve the maintainability and testability of our code, we should refactor these components to dispatch actions instead. The actions will handle the logic and trigger the necessary mutations.

PR #7734, opened 2024-08-01
Refactor components to use actions for triggering mutations (Part 1)

HNOONa-0:
# Description
I modified four components `textEditor`, `zipViewer`, `DiffEditor` and `MonacoEditor` so that they would `dispatch` store actions rather than directly calling commit. The actions I implemented have the same name as their corresponding `getters` and `mutations`. This is for convenience as if they had another name, we would have to pass it as props.

refs: #7713

# Comments

# Checklist:


HNOONa-0:
@pabo99 PTAL!

frontend/www/js/omegaup/graderv2/TextEditor.vue
```diff
@@ -49,16 +48,15 @@ export default class TextEditor extends Vue {
 
 <style lang="scss" scoped>
 @import '../../../sass/main.scss';
+
```
pabo99, line 51: Now I realize we have some duplicated styles for the textarea element. 

If you review the `index.html` and `index-light.html` templates, you will find that we include the `frontend/www/grader/ephemeral/style.css` and `frontend/www/grader/ephemeral/style-light.css` files. Although this change is not related to the PR you are working on, can you work on unifying them so they remain in the appropriate component?

To avoid modifying this PR, can you fix it in a new one?

Ideally, both `.css` files should be removed.

pabo99 (approved):
I just left you one comment not related to this PR, so this one LGTM and it is approved.

Issue #7753, opened 2024-08-10
[BUG] Preloaded `sugar.js` script from `template.tpl` causes errors to be thrown in `monaco-editor`


HNOONa-0:
## Current Behavior

This script is what is causing languages of `monaco-editor` not to be loaded properly in https://github.com/omegaup/omegaup/pull/7750. Here is a video showing what I mean:

https://github.com/user-attachments/assets/33934c78-4432-4044-8e35-6e9f3b1c73fb

## Expected Behavior
The script should not interfere with `monaco-editor`. None of the scripts should interfere with any `webpack` assets.

HNOONa-0:
@pabo99 PTAL!

pabo99:
Ok.

The first thing we need to do is identify what we are using the `sugar.js` library for; It seems that we are using it to format local dates.

You can try to resolve the issue by following the steps mentioned below:

- Remove the script from the template
- Install `sugar.js` using yarn (there are two similar libraries: `sugar` and `sugar.js`. Check which one fits better).
- Import the library in the two files where it is used: `frontend/www/js/head.sugar_locale.js` and `frontend/www/js/omegaup/time.ts`
- Make the necessary adjustments for this new library to work.
- Run the tests in Jest to validate that there is no broken code due to the new change

HNOONa-0:
Sounds good!

PR #7756, opened 2024-08-14
Extend date object of `sugar.js` library

HNOONa-0:
# Description

I installed `sugar` library and "extended" the `Date` object so that it could have the same effect as `sugar.js` script.
I also added a `Jest` mock so that `Jest` could recognize this package, its method and function calls. Changes are working as expected.

Fixes: #7753 

# Comments

# Checklist:


pabo99 (commented):
Excellent!

One last comment.

We still have an issue to resolve:

In the file `head.sugar_locale.js`, the `locale` value is globally set to `en` when the `pseudo` language is chosen in user preferences because it is not a valid locale:


https://github.com/omegaup/omegaup/blob/b9451e18e132ca363a4315b48cd9a59fa65c8a2b/frontend/www/js/head.sugar_locale.js#L1-L7

This should not affect much anymore since we no longer display that option in the dropdown. However, we can still access this language via the following URL for instance:

http://localhost:8001/arena/?lang=pseudo

With the changes you made, that global variable is no longer stored. Therefore, when trying to set the `locale` value from the `time.ts` library, it attempts to set it again as `pseudo`.

I can think of two possible solutions (both allowing us to remove the `head.sugar_locale.js` file):

- Move the logic from this file to the `setSugarLocale` function, where you can access the `data-locale` attribute and change the `locale` value there.

- Similar to what you did for the Spanish and Portuguese languages, you can add a `locales/pseudo.ts` file containing the same information in English

HNOONa-0:
> Excellent!
> 
> One last comment.
> 
> We still have an issue to resolve:
> 
> In the file `head.sugar_locale.js`, the `locale` value is globally set to `en` when the `pseudo` language is chosen in user preferences because it is not a valid locale:
> 
> https://github.com/omegaup/omegaup/blob/b9451e18e132ca363a4315b48cd9a59fa65c8a2b/frontend/www/js/head.sugar_locale.js#L1-L7
> 
> This should not affect much anymore since we no longer display that option in the dropdown. However, we can still access this language via the following URL for instance:
> 
> http://localhost:8001/arena/?lang=pseudo
> 
> With the changes you made, that global variable is no longer stored. Therefore, when trying to set the `locale` value from the `time.ts` library, it attempts to set it again as `pseudo`.
> 
> I can think of two possible solutions (both allowing us to remove the `head.sugar_locale.js` file):
> 
> * Move the logic from this file to the `setSugarLocale` function, where you can access the `data-locale` attribute and change the `locale` value there.
> * Similar to what you did for the Spanish and Portuguese languages, you can add a `locales/pseudo.ts` file containing the same information in English

`sugar` ships with `en` locale by default, we can retrieve and apply that config to `pseudo` per documentation.

frontend/templates/template.tpl
```diff
@@ -17,7 +17,6 @@
     <script type="text/javascript" src="{% versionHash '/third_party/js/jquery-3.5.1.min.js' %}"></script>
     <script type="text/javascript" src="{% versionHash '/js/jquery_error_handler.js' %}"></script>
     <script type="text/javascript" src="{% versionHash '/third_party/js/highstock.js' %}" defer></script>
-    <script type="text/javascript" src="{% versionHash '/third_party/js/sugar.js' %}"></script>
```
pabo99, line 20: Is it possible to remove this file now? I think it's no longer needed after this change

pabo99 (commented):
The very last comment, I promise

pabo99 (approved):
Thank you!

PR #7795, opened 2024-09-03
handle edge case with 0 cases

HNOONa-0:
# Description
There was an edge case when a problem statement does not have I/O. I had to create my own problem to figure this one because locally created problems, even though the problem statement is empty, its `settings.cases` is not.

https://github.com/user-attachments/assets/8cabf7a2-e2c5-4958-8811-66e359b760e6

Fixes: #7669

# Comments

# Checklist:

### DIFF
--- a/frontend/www/js/omegaup/graderv2/GraderStore.ts
+++ b/frontend/www/js/omegaup/graderv2/GraderStore.ts
@@ -867,6 +867,22 @@ const storeOptions: StoreOptions<GraderStore> = {
       commit('Validator', settings.validator.name);
       commit('Tolerance', settings.validator.tolerance);
 
+      // when there are no problem statement I/O
+      // settings.cases become an empty object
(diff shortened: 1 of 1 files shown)


frontend/www/js/omegaup/graderv2/GraderStore.ts
```diff
@@ -867,6 +867,22 @@ const storeOptions: StoreOptions<GraderStore> = {
       commit('Validator', settings.validator.name);
       commit('Tolerance', settings.validator.tolerance);
 
+      // when there are no problem statement I/O
+      // settings.cases become an empty object
+      if (Object.keys(settings.cases).length === 0) {
```
pabo99, line 872: Nit:
```suggestion
      if (!Object.keys(settings.cases).length) {
```

pabo99 (approved):
LGTM

Issue #7810, opened 2024-09-11
[FEATURE] Unify the supported languages lists


pabo99:
Currently, we have a list of supported languages to solve problems on the omegaUp platform. The issue that may arise is inconsistency in different views because we have several places where the lists are stored. For example, we have this list in TypeScript: https://github.com/omegaup/omegaup/blob/3d5330701c8dfe3ff3cd828754974fe1d7150aea/frontend/www/js/omegaup/graderv2/util.ts#L63-L221

and we have this other one in PHP: https://github.com/omegaup/omegaup/blob/3d5330701c8dfe3ff3cd828754974fe1d7150aea/frontend/server/src/Controllers/Run.php#L19-L43

And there may be other places we haven’t identified.

The intention is to unify these two lists to display the same information in all places where these lists are shown.

HNOONa-0:
@pabo99 I have a couple of ideas:
1- Add a payload that would return all supported languages from `php` then store those languages and refer to them as necessary when the application loads.
2- Add a `languages.json` that would be "single source of truth" for supported languages. I'm more lenient towards this approach because we can comfortably hold more information about each language and no api calls are necessary.

PR #7819, opened 2024-09-15
Unify languages with `languages.json`

HNOONa-0:
# Description

I added `languages.json` file and two static methods in `Run.php` that would load languages on demand when called from that file. The file contents are only read once and stored in a static variable.
Fixes: #7810
# Checklist:


HNOONa-0:
@pabo99 PTAL!

frontend/server/src/Controllers/Contest.php
```diff
@@ -2904,11 +2904,13 @@ private static function validateCommonCreateOrUpdate(
             if (empty($language)) {
                 continue;
             }
+
+            /** @psalm-suppress RedundantCondition non-falsy-string for $language is always a string */
```
pabo99, line 2908: Is it possible to avoid suppressing the error? If necessary, instead of using the `validateInEnum` function, you can directly add the condition that is performed in that function

pabo99 (commented):
Just one suggestion

frontend/server/src/Controllers/Contest.php
HNOONa-0: I think this piece of code verify that `$languages` variable is a subset of the `\OmegaUp\Controllers\Run::SUPPORTED_LANGUAGES()`. I used array functions to do the same thing.

frontend/server/src/Controllers/Contest.php
```diff
@@ -2899,17 +2899,22 @@ private static function validateCommonCreateOrUpdate(
...
-            if (empty($language)) {
-                continue;
-            }
-            \OmegaUp\Validators::validateInEnum(
-                $language,
+        $supportedLanguages = array_keys(
+            \OmegaUp\Controllers\Run::SUPPORTED_LANGUAGES()
+        );
+        $diff = array_unique(array_diff($languages, $supportedLanguages));
+        $isSubset = $diff === [] || $diff === [''];
```
pabo99, line 2908: Could you add a test in `PHPUnit` to check if it works properly? I'm not sure if the `bad_elements` index displays the correct languages

pabo99 (commented):
One last comment!

frontend/server/src/Controllers/Contest.php
HNOONa-0: I will try. The older code throws an exception when the first language that do not belong to supported languages appears.
This one throws an array of languages that do not belong to supported languages.

HNOONa-0: I used another existing validation function that validate a subset instead of my own logic.

I have another test but I think its redundant, because it just confirms that the validate function is working as intended.

pabo99: Perfect, your solution was even better. Thank you for your patience applying the suggestions

pabo99 (approved):
LGTM

heduenas (approved):

HNOONa-0:
I had to resolve another conflict

HNOONa-0:
@heduenas PTAL!

Issue #7822, opened 2024-09-15
[FEATURE] Define `payload` for full screen IDE


HNOONa-0:
We can create an initial payload for full screen ide starting with `preferredLanguage` of the user and `acceptedLangauges`.

PR #7821, opened 2024-09-15
Add theme toggle button

HNOONa-0:
# Description

I have added `theme` variable to grader store that way grader components can listen to changes made on it. I have added functions that load the `stylesheet` based on theme:

https://github.com/user-attachments/assets/5cfe9129-45fe-4627-9027-9fe467a4488c

Fixes: #7820

# Comments
I think I will undo slot change that we included in https://github.com/omegaup/omegaup/pull/7799 for two reasons:
1- I cant consistently find if the editor is full screen or not. For example, in this picture we can see that a slot value exists but it returns `undefined`, so I can decide if the editor is full screen or not based on slot value
Screenshot from 2024-09-14 22-59-36
2- I would like to add default values for `problem` prop for example. If we keep the slot, we have to include the same default values at multiple different components which is redundant.
# Checklist:

- [x] The code follows the [coding guidelines](https://github.com/omegaup/omegaup/wiki/Coding-guidelines) of omegaUp.
- [x] The tests were executed and all of them passed.
- [ ] If you are creating a feature, the new tests were added.

PR #7830, opened 2024-09-18
Initial payload definition for full IDE

HNOONa-0:
# Description

I defined a new payload that have only two attributes. The first is `acceptedLanguages` and the second is `preferredLanguage`. The IDE renders with those two properties.

Fixes: #7822

# Comments

The line count is high from the other PR. This PR should be merged first https://github.com/omegaup/omegaup/pull/7821

# Checklist:


HNOONa-0:
The lint removed my definition for the payload, is that normal?

frontend/server/src/Controllers/Grader.php
```diff
@@ -25,19 +26,27 @@ public static function apiStatus(\OmegaUp\Request $r): array {
...
-     * @psalm-return array{templateProperties: array{title: \OmegaUp\TranslationString, fullWidth?: bool, hideFooterAndHeader?: bool, payload: array<string, mixed>}, entrypoint: string, inContest?: bool, navbarSection?: string}
+     * @return array{templateProperties: array{payload: FullIDEPayload, title: \OmegaUp\TranslationString, fullWidth?: bool, hideFooterAndHeader?: bool}, entrypoint: string}
+     * @omegaup-request-param null|string $auth_token
     */
     public static function getGraderForTypeScript(
         \OmegaUp\Request $r
     ): array {
+        $preferredLanguage = \OmegaUp\DAO\Users::getPreferredLanguage(
+            \OmegaUp\Controllers\Session::getCurrentSession()['user']->user_id ?? null
+        );
```
pabo99, line 37: Okay, we have a more sophisticated way to get the preferred language for a user/identity:
```suggestion
        try {
            $r->ensureIdentity();
        } catch (\OmegaUp\Exceptions\UnauthorizedException $_) {
            // Do nothing. Not logged user can access here
            $r->identity = null;
        }

        $preferredLanguage = \OmegaUp\Controllers\Identity::getPreferredLanguage(
            $r->identity,
            $r
        );
```

In the first part, it ensures an identity exists, but it’s not required to have an open session. So, the `catch` part allows Not-logged users to access the ephemeral grader page.

As you can see, in the second part we have the function `Identity::getPreferredLanguage` that receives two parameters:

- `$request`: If the request contains the parameter `lang` (meaning the language is provided via URL) it becomes the preferred language.  
- `$identity`: This parameter is useful for getting the language that the user selected in their profile settings

pabo99 (commented):
Just one suggestion.

pabo99:
One last thing: It contains changes from different branches. You need to update your branch to display only the changes related to the payload.

frontend/server/src/Controllers/Grader.php
HNOONa-0: Yes I noticed that function but I think this function returns a spoken language not a programming language, right?

pabo99: Apologies. You're right. It's kind of difficult when we use the same term for different things. 

Maybe in the future we need to rename "language" to "programming language" (I think we already use it in some places).

Regarding the suggestion, you can obtain the `user_id` in a different way, avoiding access to the current session:

You can use the first part of my original suggestion in order to get the logged user:
```php
        $r->user = null;
        try {
            $r->ensureIdentity();
        } catch (\OmegaUp\Exceptions\UnauthorizedException $_) {
            // Do nothing. Not logged user can access here
            $r->user = null;
        }
```
And then, you can use `$r->user?->user_id` in the function argument:
```php
        $preferredLanguage = \OmegaUp\DAO\Users::getPreferredLanguage(
            $r->user?->user_id
        );
```

This way, you don't need to define the request param `$auth_token`, so it can be removed

pabo99 (commented):
One more comment

pabo99 (approved):
Thank you!

heduenas:
@HNOONa-0 please resolve the conflicts that surfaced.

Issue #7785, opened 2024-08-31
[BUG] When contestants switch to a problem, the problem data is changed but not the runs


HNOONa-0:
## Current Behavior

In a contest, when the user switches between problems to solve, the problem data is updated properly, but not the associated runs with that problem. Here is a video:

https://github.com/user-attachments/assets/17779f26-af63-46e0-a952-37fef4b27ce0

When we switch to problem that has 3 runs it displays 3 runs, but then when we switch to a problem with 0 runs, it still displays 3 runs. Keeping runs in sync could be important for example, when we are tying to set the `preferredLangauge` properly

## Expected Behavior
As problem change, the runs should be in sync with that problem.

PR #7846, opened 2024-09-30
Get preferred langauge properly

HNOONa-0:
# Description

We were using the wrong runs property, instead we should use `runsByProblem` computed property. 

Fixes: #7785

# Checklist:


pabo99 (approved):
Thank you for finding this bug and fixing it. The change looks good.

I just have one question: does this change also fix the bug related to updating the problem submissions list? Or does it only focus on fixing the bug that updates the last used language for each problem on the ephemeral grader component?

If it only fixes the latter, could you update the PR description to indicate that this PR only solves part of the issue? This way, we can continue working on resolving the remaining parts.

HNOONa-0:
> Thank you for finding this bug and fixing it. The change looks good.
> 
> I just have one question: does this change also fix the bug related to updating the problem submissions list? Or does it only focus on fixing the bug that updates the last used language for each problem on the ephemeral grader component?
> 
> If it only fixes the latter, could you update the PR description to indicate that this PR only solves part of the issue? This way, we can continue working on resolving the remaining parts.

I was using the wrong computed property, there are multiple `runs` computed properties. This one, the `runsByProblem` seems to contain right information about submission per problem in contests, courses and problems pages.

heduenas (approved):

Issue #7603, opened 2024-05-19
Cambiar cadena "Fuente" por "Codigo"


heduenas:
<img width="1142" alt="Screenshot 2024-05-19 at 2 35 43 PM" src="https://github.com/omegaup/omegaup/assets/597815/ca2cb638-0db3-402e-a944-94d34f302316">

poojatalele:
Hey! @heduenas I would like to work on this issue, kindly assign

poojatalele:
Also, please guide me on how to start working on this issue.

Barfasushmita:
@heduenas Hello, I want to work on this issue please assign me this issue!

PR #7855, opened 2024-10-18
Changed 'Fuente' to 'Codigo' #7603

Barfasushmita:


heduenas (commented):
@Barfasushmita thank you for sending this PR. Unfortunately someone else sent a fix for the same issue before you, so we  merged their PR first and can't merge yours.

Issue #7901, opened 2024-11-15
[FEATURE] Username Not Updating Immediately After Saving Changes in Profile Settings


sudarsan2k5:
## Problem Description:
- When `User` editing the username and clicking the "Save Changes" button, the username does not immediately update in the UI. Instead, the updated username is reflected only after refreshing the page.

## Expected Behavior:
- After clicking the `Save Changes` button
- The username should update immediately in the UI without requiring a page refresh.

## Steps to Reproduce:

Go to this link:- https://omegaup.com/profile/#edit-basic-information
- Edit the `username` and Click the `Save Changes` button.
- Observe that the username does not update immediately in the UI.
- Refresh the page to see the updated username.

https://github.com/user-attachments/assets/8c21aff4-7619-49c6-8bc1-e185d61ea0ac

sudarsan2k5:
Hi @pabo99, @heduenas 

I’ve tried to solve the issue. The first approach that came to my mind was to trigger a re-render of the page after the user saves their changes. However, I’m not sure if this is the correct approach.

I’ve been working on this problem for two days but haven’t been able to resolve it yet. I also tried the method suggested in this StackOverflow thread: [Link](https://stackoverflow.com/questions/32106155/can-you-force-vue-js-to-reload-re-render,) but I still couldn’t get it to work.

Please suggest any resources or alternative approaches to fix this issue.

pabo99:
> Hi @pabo99, @heduenas
> 
> I’ve tried to solve the issue. The first approach that came to my mind was to trigger a re-render of the page after the user saves their changes. However, I’m not sure if this is the correct approach.
> 
> I’ve been working on this problem for two days but haven’t been able to resolve it yet. I also tried the method suggested in this StackOverflow thread: [Link](https://stackoverflow.com/questions/32106155/can-you-force-vue-js-to-reload-re-render,) but I still couldn’t get it to work.
> 
> Please suggest any resources or alternative approaches to fix this issue.

Very interesting issue you found, @sudarsan2k5 .

As mentioned in the link you provided, definitely refreshing the page is a horrible way to solve this issue.

But if we see that none of the suggested options work for us, we will have to consider the one that does work.

One question, do you have any progress on this issue that you can upload to a PR? That way we can help you check if anything can be fixed. It doesn't matter if it doesn't work yet

sudarsan2k5:
Thank you, @pabo99 and @heduenas, for your awesome feedback!

[...]

PR #7905, opened 2024-11-17
Update `username` in UI immediately after saving changes

sudarsan2k5:
# Description

It will refresh the page after the user clicks the `Save Changes` button and it will update the UI.

https://github.com/user-attachments/assets/ae77978b-c27d-4890-a2b6-b95553d171d0

Fixes: #7901

# Checklist:


sudarsan2k5:
I know this approach to solving this issue is very bad, and I would appreciate your suggestions on improving it.

Why I used setTimeout ?

With setTimeout:- 

```
setTimeout(() => {
   location.reload();
}, 1000);
```
I used the reload functionality inside the setTimeout because, after the user clicks the `Save Changes` button, a popup appears indicating `Profile successfully updated.` Without wrapping the reload in setTimeout, the popup and the page reload would overlap.


https://github.com/user-attachments/assets/37c00cb1-6ab8-4a2c-a00c-38b6c4fc389c


 
Without setTimeout:-
If I do not wrap the `location.reload()` inside the `setTimeout` it causes the popup and the reload to overlap, and the old username briefly appears again if I press Save Changes Butoon multiple times.


https://github.com/user-attachments/assets/5293dec4-64cd-4e12-8e63-fe9a20b6d1ab

frontend/www/js/omegaup/components/user/BasicInformationEdit.vue
```diff
@@ -107,6 +107,7 @@ export default class UserBasicInformationEdit extends Vue {
   birthDate = this.profile.birth_date
     ? time.convertLocalDateToGMTDate(this.profile.birth_date)
     : new Date('');
+  updateKey = 0;
```
heduenas, line 110: I'm having trouble understanding what this updateKey is for

heduenas:
does it refresh on changes to other fields as well (e.g. if you only change country)?

frontend/www/js/omegaup/components/user/BasicInformationEdit.vue
sudarsan2k5: Ahh, I actually forgot to remove that. I was using `updateKey` to force a component re-render by changing its key value. However, since the page reloads after submission, the `updateKey`-related code is no longer necessary and can be safely removed.

pabo99:
> Yes, it refreshes if I change the following fields like:- `Account name`, `Name`, `Gender`, `Country`, `State`, or `Birthdate`.
> 
>  Screen.Recording.2024-11-23.at.2.50.23.PM.mov

Is it possible to refresh the page only when the `Account name` field is updated? Regarding the country icon issue, I think it can be handled in a different way.

Could you take a look at how the refresh is applied for the language change? This way, the action would not be performed in the `Vue` component, but from the `.ts` file that handles this component.

heduenas:
> > Yes, it refreshes if I change the following fields like:- `Account name`, `Name`, `Gender`, `Country`, `State`, or `Birthdate`.
> > Screen.Recording.2024-11-23.at.2.50.23.PM.mov
> 
> Is it possible to refresh the page only when the `Account name` field is updated? Regarding the country icon issue, I think it can be handled in a different way.
> 
> Could you take a look at how the refresh is applied for the language change? This way, the action would not be performed in the `Vue` component, but from the `.ts` file that handles this component.

I agree with @pabo99 , we should only refresh when the username field changes.

pabo99 (commented):
Sorry for the late review of this PR. 


With the suggestions I provided, it will no longer be necessary to reload the page as we discussed earlier. We’ve also avoided using `localStorage`.

Please try to apply all the corrections. If something isn’t clear, feel free to reach out, and I’ll be happy to help you resolve any questions.

Once you apply all the suggestions, could you also try implementing the country flag change? That would be awesome!

I was also thinking we could add the avatar in the store, but since it only updates when the email is changed and we currently don’t allow email changes on the platform, it won’t be necessary for now.

sudarsan2k5:
Thank you so much, @pabo99 , for your incredibly detailed review and suggestions! Your explanations are always amazing and teach me something new every time. The way you describe things makes implementation much easier. Even someone who knows nothing about the topic could easily implement it just by reading your explanation. I genuinely appreciate the time and effort you put into guiding us! 🥹

As per your suggestions, I’ve made the necessary corrections, and everything is working fine now. However, there’s still one issue I’m facing : as I mentioned earlier, the popup message and page refresh are overlapping each other when the page refreshes. I attempted to fix this, but the solution I implemented doesn’t seem like the best approach.
Could you kindly provide your suggestion on the best way to address this? I would love to be able to refine my implementation with your guidance.

Once again, thank you so much 🙏

Overlapping problem:- 

https://github.com/user-attachments/assets/d4bbac08-a3d4-4e73-9a67-6f01b163b268

frontend/www/js/omegaup/components/user/Profile.vue
```diff
@@ -30,7 +30,8 @@
             :profile="profile"
             :countries="countries"
             @update-user-basic-information="
-              (request) => $emit('update-user-basic-information', request)
+              (request, usernameChanged) =>
+                $emit('update-user-basic-information', request, usernameChanged)
```
pabo99, line 34: Same as the previous comment
```suggestion
```

frontend/www/js/omegaup/user/profile.ts
```diff
@@ -78,7 +80,16 @@ OmegaUp.on('ready', () => {
...
+                mainStore.commit(
+                  'updateUsername',
+                  userBasicInformation.username,
+                );
                 ui.success(T.userEditSuccess);
+                if (
+                  userBasicInformation.username !== payload.profile.username
+                ) {
+                  window.location.reload();
+                }
```
pabo99, line 92: The commit added on line 83 should help update the username in the store. This field is also displaying the username in the navbar, so you can restore this block as well.
```suggestion
```

frontend/www/js/omegaup/user/profile.ts
```diff
@@ -78,6 +80,16 @@ OmegaUp.on('ready', () => {
...
+                mainStore.commit(
+                  'updateUsername',
+                  userBasicInformation.username,
+                );
+                if (userBasicInformation.username) {
+                  userProfile.profile = {
+                    ...userProfile.profile,
+                    username: userBasicInformation.username,
+                  };
+                }
```
sudarsan2k5, line 92: It updates the Account name on the  BasicInformation page.

sudarsan2k5: https://github.com/user-attachments/assets/8ba4f423-4613-47b3-8123-b4e6226a16fa

heduenas (approved):

frontend/www/js/omegaup/user/profile.ts
```diff
@@ -78,6 +80,16 @@ OmegaUp.on('ready', () => {
...
               .then(() => {
+                mainStore.commit(
+                  'updateUsername',
+                  userBasicInformation.username,
+                );
+                if (userBasicInformation.username) {
+                  userProfile.profile = {
+                    ...userProfile.profile,
+                    username: userBasicInformation.username,
+                  };
```
pabo99, line 91: If your intention is to update only the username field, you can do it directly by assigning the property:
```suggestion
                  userProfile.profile.username = userBasicInformation.username;
```

Another approach is to use the `username` property from the main store, which will update automatically this field. However, for now, this approach is fine.

pabo99 (commented):
One final comment, and it's fine with me.

frontend/www/js/omegaup/user/profile.ts
sudarsan2k5: Country flag update in user profile

https://github.com/user-attachments/assets/87772fc4-bde9-4abb-a1c3-bdc9dce5c4e3

pabo99: Awesome!!!

pabo99 (approved):
Excellent!

We truly appreciate your patience and dedication in resolving this issue.

pabo99:
@sudarsan2k5 The PR has been approved! The last step is to update your branch so it can be merged into the main branch.

sudarsan2k5:
Thanks for reviewing and merging my code!

PR #7915, opened 2024-12-07
Add state selection warning in profile page

### LINKED ISSUE #7904: [FEATURE]
_change the default states in the country

_ as a user i have to select the state i should not set to default states

_ if user is hurry in filling the data he may think  he selected the state also if the it is set to default so  we have to give the warning select the state 

Uploading Recording 2024-11-16 133126.mp4…

deepan190703:
Fixes #7904

Add validation and warning for state selection in user profile.

* Add validation mechanism in `cypress/support/pageObjects/profilePage.ts` to check if the state is selected.
* Display a warning message if the state is not selected in `cypress/support/pageObjects/profilePage.ts`.
* Update test cases in `cypress/e2e/navigation.cy.ts` to check for the warning related to state selection.


heduenas:
@deepan190703 this change is only adding tests? just checking if that's what you meant to do.

heduenas (approved):

pabo99:
@deepan190703, could you please update the branch? That way, it can be merged with the main branch since it has been approved.

Issue #7926, opened 2024-12-26
[FEATURE]Add Scrollbar to Navigation Dropdown Menu for Improved Content Accessibility


k-vijay-05:
In the `navigation dropdown` section, the content is hidden if it exceeded the visible area, making it inaccessible to users.

**Before**: The dropdown menu content was hidden if it overflowed, and users were unable to view the hidden content.

https://github.com/user-attachments/assets/648055f8-f4e5-4638-acbe-68e044a3f94a

**After**:  The dropdown menu now includes a scrollbar, allowing users to view and access all the content that previously overflowed.

https://github.com/user-attachments/assets/254a5aad-f301-400a-b4a2-578071a5e69f

I would like to hear your suggestion on this change. Can I go ahead and raise a PR for this improvement?

pabo99:
Hello @k-vijay-05, personally, I'm not a big fan of using scrollbars, but considering this bug might affect the user experience and may have gone unnoticed, your approach seems fine. Feel free to go ahead and create a new PR with the proposed solution.

k-vijay-05:
@pabo99 Thank you for your feedback! I understand the preference for avoiding scrollbars. To align with this, I've updated the approach to include a hidden scrollbar while still enabling smooth scrolling for better user experience

PR #7929, opened 2024-12-27
Fix dropdown menu overflow and add smooth scrolling for better user experience

k-vijay-05:
# Description

Added functionality to enable smooth scrolling within the dropdown menu, allowing users to access content that was previously hidden due to overflow.

Fixes: #7926 

Video: 

https://github.com/user-attachments/assets/2911df7d-7ecd-4398-b7fb-f2390e5ccacc

Video: (Mobile version)

https://github.com/user-attachments/assets/6143c595-11e6-4d60-9342-3c54dacb1b58

# Checklist:


frontend/www/js/omegaup/components/common/Navbar.vue
```diff
@@ -504,6 +504,12 @@ nav.navbar {
   .collapse-submenu .btn:focus {
     box-shadow: 0 0 0 0;
   }
+
+  .dropdown-menu {
+    overflow-y: auto;
+    max-height: 400 px;
```
heduenas, line 510: can you express this height in `rem` units?

k-vijay-05: Updated max-height to use rem units as suggested.

frontend/www/js/omegaup/components/common/Navbar.vue
```diff
@@ -504,6 +504,12 @@ nav.navbar {
   .collapse-submenu .btn:focus {
     box-shadow: 0 0 0 0;
   }
+
+  .dropdown-menu {
+    overflow-y: auto;
+    max-height: 25rem;
```
pabo99, line 510: I understand the need to make this change since it may be affecting users who cannot access all the content in the main menu. I just have one observation:

By limiting the maximum height of the dropdown menu, we might also affect users with high-resolution screens. Would it be possible to handle this style using the CSS property [`vh`](https://www.sitepoint.com/css-viewport-units-quick-start) and set it to a high percentage (75%, for example)?

heduenas: That's a good point

k-vijay-05: Thanks for the suggestion! Yes  Switching to 75vh is a better approach as it ensures the dropdown menu is fully visible on bigger screens while remaining scrollable on smaller screens. I've made this change accordingly.

pabo99 (approved):
Thanks @k-vijay-05 !

heduenas (approved):

Issue #7933, opened 2025-01-01
[FEATURE] Disabled `Save Changes` Button for No Updates on `Manage Schools` and `Preferences` Pages


sudarsan2k5:
There are two other pages, `Manage Schools` and `Preferences`, where if the user does not make any changes to the fields and clicks the `Save Changes button`, the `Profile successfully updated` popup still appears, which should not happen. To address this, we have disabled the Save Changes button for no updates. This approach helps avoid unnecessary validations on both the client and server sides.

## Before:- 
https://github.com/user-attachments/assets/85ea086a-e0db-47ae-ae3a-bd657fe0d19e

## After
https://github.com/user-attachments/assets/eb488d0a-da75-4763-8e38-ddc0f3769232

amaan-malik1:
I would love to work with this issue @Aritra8438 
The "Save Changes" button should be disabled by default and only enabled when the user makes changes to any fields.
This will prevent:
  Misleading success messages
  Unnecessary backend requests
 Extra validation checks

Aritra8438:
@amaan-malik1, have you appeared for the test?

amaan-malik1:
yesh i'm @Aritra8438

Aritra8438:
> not right now [@Aritra8438](https://github.com/Aritra8438?rgh-link-date=2025-03-13T09%3A50%3A55.000Z)

Please appear for the test first, as it's the first phase.

Anjola7:
@Aritra8438 I have done 2/3 in the programming test . How I can find some issue to work with?

PR #7935, opened 2025-01-02
Disabled Save Changes Button for No Updates on Manage Schools and Preferences Pages

sudarsan2k5:
# Description

Disable the `Save Changes Button` for No Updates on the `Manage Schools` and `Preferences Pages`. This helps avoid unnecessary validations on both the client and server sides.

Fixes: #7933

# Checklist:

### DIFF
--- a/cypress/support/pageObjects/profilePage.ts
+++ b/cypress/support/pageObjects/profilePage.ts
@@ -19,7 +19,7 @@ export class ProfilePage {
     cy.get('[data-nav-profile]').click();
     cy.get('a[href="/profile/#edit-preferences"]').click();
     cy.get('[data-preference-language]').select(preferredLanguage);
-    cy.get('[data-preference-save-button]').click();
+    cy.get('[data-preference-save-button]').click({ force: true });
   }
   updatePreferredProgrammingLanguage(preferredLanguage: string): void {
     cy.get('[data-nav-user]').click();
--- a/frontend/www/js/omegaup/components/user/ManageSchools.vue
+++ b/frontend/www/js/omegaup/components/user/ManageSchools.vue
@@ -54,6 +54,7 @@
         type="submit"
         class="btn btn-primary mr-2"
         data-save-school-changes
+        :disabled="!hasChanges"
       >
         {{ T.wordsSaveChanges }}
       </button>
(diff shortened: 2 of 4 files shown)


heduenas:
@sudarsan2k5 looks like your changes broke some tests, can you please take a look?

pabo99:
Ok, I think I found the error:

It seems that there is a test trying to submit the form, but it’s not working because it selects the initial values of the dropdowns. As a result, the submit button never gets enabled.

```
✕ Should emit correct objectives values when "learning" and "competitive" options are selected
```

The way to solve this issue is by removing this particular case:
https://github.com/omegaup/omegaup/blob/20549ed60c5fdf04f9f46c3f268632c6d09adc0c/frontend/www/js/omegaup/components/user/PreferencesEdit.test.ts#L140-L147

heduenas:
@sudarsan2k5 cypress test are still failing

cypress/support/pageObjects/profilePage.ts
```diff
@@ -18,8 +18,17 @@ export class ProfilePage {
...
+
+    // Get the current value and only change if different
+    cy.get('[data-preference-language]').then(($select) => {
+      const currentValue = $select.val();
+      if (currentValue !== preferredLanguage) {
+        cy.wrap($select).select(preferredLanguage);
+        cy.get('[data-preference-save-button]').click();
+        cy.get('[data-preference-save-button]').should('be.disabled');
+      }
+    });
```
sudarsan2k5, line 31: I verified in the Cypress UI that we are attempting to set the language to the same value as before, which does not trigger a change event, resulting in the "save button" remaining disabled. This update ensures that the language is only changed if it differs from the current setting, thereby triggering the change event and enabling the "save button" as expected.

please let me know if this is correct. Thank you!

sudarsan2k5:
image



Could you help me reproduce these failed tests on my local PC? All the test cases are passing on my system.

cypress/support/pageObjects/profilePage.ts
pabo99: Ok, I understand the reason behind this change. I just have a few observations:

It seems that the language displayed on the platform is English, but the one selected in the language preferences dropdown is Spanish.

This causes an issue where, during the comparison between the selected language string and the target language string, there is no difference, and the button click is skipped. However, the language displayed on the platform remains English.

What we can do to ensure these tests continue to work is to force the button click, regardless of its status.

To achieve this, you can revert the change to this file, and on line 22, add the argument to force the click as follows:

```ts
cy.get('[data-preference-save-button]').click({ force: true });
```

pabo99 (commented):
I left a comment to help resolve the failing tests. If you have any questions or need further clarification, feel free to reach out to me, and we can work on solving them together.

sudarsan2k5:
image


Hi @pabo99, as you suggested, I made the changes. They passed all the tests on my local machine, but the GitHub CI failed again. I am not sure why the GitHub CI is still failing. :(

pabo99:
Ok, I see.

It seems the error shown in the GitHub CI is not related to your changes. Updating the PR might help resolve the issue.

sudarsan2k5:
Thanks @pabo99 , I updated the PR, but it’s still failing. :(

pabo99:
> Thanks @pabo99 , I updated the PR, but it’s still failing. :(

Ok, we will need to look for another approach.

Issue #7956, opened 2025-01-28
Support ACL members should be able to set contests as recommended


heduenas:
The `setRecommended` [API endpoint](https://github.com/omegaup/omegaup/tree/main/frontend/server/src/Controllers#apicontestsetrecommended) can only be called by site-admins right now. Members of the support ACL should be able to call it as well.

PR #7960, opened 2025-01-29
Allow support team members to set contest recommended status

sudarsan2k5:
# Description

Now Members of the support ACL are also able to call it as well.

Fixes: #7956

# Comments

# Checklist:


sudarsan2k5:
I believe this is the method that grants permission to the support team, but I'm not entirely sure. Could you check my code and help me figure out how to test it on my local PC?

https://github.com/omegaup/omegaup/blob/4f9f4f6193c0957b6f68ed157f5206f339670bb6/frontend/server/src/Controllers/Contest.php#L5936-L5943


https://github.com/omegaup/omegaup/blob/4f9f4f6193c0957b6f68ed157f5206f339670bb6/frontend/server/src/Authorization.php#L533-L553

frontend/server/src/Controllers/Contest.php
```diff
@@ -5936,7 +5936,10 @@ public static function apiRole(\OmegaUp\Request $r): array {
     public static function apiSetRecommended(\OmegaUp\Request $r): array {
         $r->ensureIdentity();
 
-        if (!\OmegaUp\Authorization::isSystemAdmin($r->identity)) {
+        if (
+            !\OmegaUp\Authorization::isSystemAdmin($r->identity) &&
+            !\OmegaUp\Authorization::isSupportTeamMember($r->identity)
```
pabo99, line 5941: As you can see in the `isSupportTeamMember` method, the `sysAdmin` role is one level above the rest of the roles. This means that it implicitly has permission to perform all the actions that a support team member (or any other role) can do:

https://github.com/omegaup/omegaup/blob/4f9f4f6193c0957b6f68ed157f5206f339670bb6/frontend/server/src/Authorization.php#L545-L552

https://github.com/omegaup/omegaup/blob/4f9f4f6193c0957b6f68ed157f5206f339670bb6/frontend/server/src/Authorization.php#L603-L605

So, adding line 5941 should be enough.

To validate that this works correctly, a `PHPUnit` test should be added with the following actions:

Create an identity with the `sysAdmin` role
Create an identity with the `supportTeamMember` role
Create an identity with any other role
Create a contest
Add all three identities as contest administrators
Set the contest as recommended using each administrator
First two identities should be able to configure the contest
Third identity should receive an error due to permissions
Here is the existing test for the happy path of setting a contest as recommended:

https://github.com/omegaup/omegaup/blob/4f9f4f6193c0957b6f68ed157f5206f339670bb6/frontend/tests/controllers/ContestUpdateTest.php#L283-L315

You might use it as a reference to create the new test. Extra points if you use parameterized tests.

Let me know if you need more help.

sudarsan2k5:
Thanks🫡 @pabo99, as per the suggestion, I have added the test. Could you please have a quick review?

frontend/tests/controllers/ContestUpdateTest.php
```diff
@@ -1928,4 +1928,83 @@ public function testUpdatePenaltyTypeFromAContestWhenProblemVersionIsUpdated(
...
+            case 'admin':
+                ['identity' => $identity] = \OmegaUp\Test\Factories\User::createAdminUser();
+                break;
+            case 'support':
+                ['identity' => $identity] = \OmegaUp\Test\Factories\User::createUser();
+                \OmegaUp\Test\Factories\User::addSupportRole($identity);
+                break;
+            default:
+                ['identity' => $identity] = \OmegaUp\Test\Factories\User::createUser();
+        }
```
pabo99, line 1968: How about we modify the test behavior a bit?

Since it has already been validated in existing tests that the sysadmin can perform this action, let's remove it from this new test.

Instead of adding this switch, the identity should be created and added to its corresponding group (you can use the `addSupportRole` function as a reference). In this case, you can add more cases to the parameterized test, which you can take from this list (it can be all of them):

https://github.com/omegaup/omegaup/blob/4f9f4f6193c0957b6f68ed157f5206f339670bb6/frontend/server/src/Authorization.php#L85-L104

pabo99 (commented):
Great job, as usual! 

I added a suggestion. Once it's addressed, I can approve this PR

sudarsan2k5:
As you suggested, I implemented the changes, and it passed all the tests. However, when I updated the branch, the CI/Python (pull request) failed. :(

frontend/tests/controllers/ContestUpdateTest.php
```diff
@@ -1928,4 +1928,93 @@ public function testUpdatePenaltyTypeFromAContestWhenProblemVersionIsUpdated(
...
     }
+
+    /**
+     * A PHPUnit data provider for testing contest recommended flag permissions.
+     *
+     * @return list<array{0: string, 1: bool}>
+     */
+    public function contestRecommendedPermissionsProvider(): array {
+        return [
+            ['support', true],
```
pabo99, line 1939: Is it possible to use the group aliases here?
```suggestion
            [\OmegaUp\Authorization::SUPPORT_GROUP_ALIAS, true],
            ...
```
This way, we can avoid an extra step

frontend/tests/controllers/ContestUpdateTest.php
```diff
@@ -1928,4 +1928,93 @@ public function testUpdatePenaltyTypeFromAContestWhenProblemVersionIsUpdated(
...
+            ['quality-reviewer', false],
+            ['course-curator', false],
+            ['mentor', false],
+            ['group-identity-creator', false],
+            ['certificate-generator', false],
+            ['teaching-assistant', false],
+            ['normal', false],
```
pabo99, line 1946: This is not necessary since it isn't a valid group alias, so you can remove it
```suggestion
```

frontend/tests/controllers/ContestUpdateTest.php
```diff
@@ -1928,4 +1928,93 @@ public function testUpdatePenaltyTypeFromAContestWhenProblemVersionIsUpdated(
...
+
+        // Add user to the corresponding group based on type
+        if ($userType !== 'normal') {
+            $groupAlias = 'omegaup:' . $userType;
+            $group = \OmegaUp\DAO\Groups::findByAlias($groupAlias);
+            if (is_null($group)) {
+                throw new \OmegaUp\Exceptions\NotFoundException(
+                    'groupNotFound'
+                );
+            }
```
pabo99, line 1973: Now PHP 8 has integrated the [null-safe operator](https://php.watch/versions/8.0/null-safe-operator), so it's no longer necessary to add this condition:
```suggestion
            if (is_null($group)) {
                throw new \OmegaUp\Exceptions\NotFoundException(
                    'groupNotFound'
                );
            }
```
```suggestion
```
As you can use it like this:
```php
'group_id' => $group?->group_id,
```

pabo99 (commented):
Almost done!

frontend/tests/controllers/ContestUpdateTest.php
sudarsan2k5: Right 🫡

frontend/tests/controllers/ContestUpdateTest.php
sudarsan2k5: got it, Thanks 🙇🏼🙇🏼

sudarsan2k5:
I've updated the code based on it. Please let me know if any further changes are needed.

frontend/tests/controllers/ContestUpdateTest.php
```diff
@@ -1928,4 +1928,84 @@ public function testUpdatePenaltyTypeFromAContestWhenProblemVersionIsUpdated(
...
+        ['identity' => $identity] = \OmegaUp\Test\Factories\User::createUser();
+
+        // Add user to the group
+        $group = \OmegaUp\DAO\Groups::findByAlias($groupAlias);
+        \OmegaUp\DAO\GroupsIdentities::create(new \OmegaUp\DAO\VO\GroupsIdentities([
+            'group_id' => $group?->group_id,
+            'identity_id' => $identity->identity_id,
+        ]));
+
+        // Add user as contest admin
```
heduenas, line 1971: The support user shouldn't need to be added to the contest admin group

sudarsan2k5: DONE ✅, please have a look :)

pabo99 (approved):
Thank you so much, @sudarsan2k5 !

heduenas (approved):

Issue #7943, opened 2025-01-12
[BUG] Privacy policy link is broken


heduenas:
![Screenshot 2025-01-11 at 7 03 10 PM](https://github.com/user-attachments/assets/223ff3c9-288d-4ebb-87f2-11a8661d51b7)

The new URL is https://blog.omegaup.com/posts/privacy-policy/

kuldeeprathore05:
can i work on this issue

heduenas:
you got it @kuldeeprathore05! let us know if you have any questions

P-laxminarsimha:
@heduenas is this issue still open

KaranNegi20Feb:
Hey @heduenas can i work on this issue?

Kashaf1175:
Can i work on this issue

KaranNegi20Feb:
> [@KaranNegi20Feb](https://github.com/KaranNegi20Feb) [@Kashaf1175](https://github.com/Kashaf1175) have you completed the programming test https://omegaup.com/arena/gsoc2025/#problems ?

No @heduenas

KaranNegi20Feb:
is it possible that i can get this issue assigned since no one is working on it

Imkkrish:
> ![Screenshot 2025-01-11 at 7 03 10 PM](https://github.com/user-attachments/assets/223ff3c9-288d-4ebb-87f2-11a8661d51b7)
> 
> The new URL is https://blog.omegaup.com/posts/privacy-policy/

can i make it work again?

Kashaf1175:
> [@KaranNegi20Feb](https://github.com/KaranNegi20Feb) [@Kashaf1175](https://github.com/Kashaf1175) have you completed the programming test https://omegaup.com/arena/gsoc2025/#problems ?

not yet

Anjola7:
Hello! I am interested in contributing to omegaUp for GSoC.  
However, I see that most of the `Good First Issues` are already assigned.  
Could you suggest an issue that I could work on?  
Thank you!

heduenas:
@Imkkrish @KaranNegi20Feb @Anjola7 @KaranNegi20Feb @kuldeeprathore05 whoever solves 2 problems in the programming test first gets this bug https://omegaup.com/arena/gsoc2025-extension#problems For the rest, we can find a different issue to work on once you finish the programming test.

r6mez:
@heduenas 
I solved 2 problems in gsoc 25 test, my user name is R6mez

[...]

PR #7988, opened 2025-02-18
fixed-broken-link

ajey35:
Fixes: #7943 

Changes Made:

Added Privacy Policy link to the variable acceptPrivacyPolicy in the following language files:

en.lang
es.lang
pt.lang
pseudo.lang


pabo99:
Moving the conversation from the PR #7992 to this one:

> Hello @pabo99, Yes I already run that command, This is what I got :
> 
> image

Ok, this makes sense. It looks like the instruction needs to be clearer:

The command must be executed inside WSL for it to take effect.

> Hello @pabo99
> I followed your suggestion and ran the command; however, it is taking a considerable amount of time to execute. 
> Could you please provide an estimate of how long this process should typically take? Additionally, after stopping the
> previous execution and running the command again, there was still no response even though I waited for a significant > duration.
> 
> Screenshot 2025-02-25 220407
> 
> Thankyou!

Ok, according to the image you shared, it's important to indicate that this is the first time you run this command. It requires the `hook_tools` image for the linters to work properly. Since it couldn't find that image, it will try to download it (only this first time). The estimated download time depends on your internet connection speed.

The next time you run the command, it should execute considerably faster than the first time.

At the end of the execution, you should see the message `"ok"`. I'm not sure if the command accepts a debug argument, but it would be a good indicator if the command doesn’t freeze in the middle of a process.

Let us know if you still experience any issues when running the command again.

Thanks to you for your patience!

pabo99:
Close this PR in favour of #8012

Issue #7987, opened 2025-02-18
[FEATURE] Implement Logout Confirmation Feature


Saloni-2005:
Currently, clicking the logout icon results in an immediate logout without any confirmation. This can be problematic if a user inadvertently clicks the icon, leading to an unintentional logout and the inconvenience of having to log in again.

Proposed Solution:

To improve the user experience and prevent accidental logouts, I recommend implementing a confirmation prompt that asks users to verify their intention to log out. This simple addition will enhance usability and reduce frustration for our users.

Benefits:

- Prevents accidental logouts
- Enhances user experience by providing an additional layer of security
- Reduces user frustration and inconvenience

Steps to Implement:

Add a confirmation dialog that appears when the user clicks the logout icon.
The dialog should include confirm the logout action.
If the user confirms, proceed with the logout process.

@heduenas  Please assign this issue to me, as I would like to work on implementing this feature. Thank you!

heduenas:
@Saloni-2005 thank you for filing this. I've assigned it to you. When you implement your fix, please make sure that it works well on both desktop and mobile.

amaan-malik1:
In this issue we can also impliment the alert dialog box when the user try logOut. And add the bussiness logic so that user may think before he logout.

pabo99:
Hello @amaan-malik1, thanks for your willingness to collaborate with omegaUp!

It looks like @Saloni-2005 has already been working on creating the dialog box, which is included in PR #7992 and is currently under review.

Could you look for another issue where you can contribute so we can assign it to you?

amaan-malik1:
yeah sure @pabo99 i'll check it and provide the resolved output for the issue.

PR #7992, opened 2025-02-19
[FEATURE]  Added logout confirmation dialog 

Saloni-2005:
Added a logout confirmation dialog to enhance user security and prevent accidental logouts. When users click the logout icon, a confirmation modal appears requesting them to confirm their action before signing out of the application.

Key changes:

- Implemented a confirmation dialog that appears when the logout icon is clicked
- Added "Cancel" and "Logout" buttons in the dialog
- User session remains active if "Cancel" is clicked
- Logout process executes only after user confirms

https://github.com/user-attachments/assets/452bf3d2-29b9-4611-a3a6-c42f5b8dfc75

Fixes: #7987 

The main changes are in the UI components where the logout confirmation dialog was implemented. The confirmation modal f includes both "Cancel" and "Confirm" actions.

@heduenas Please review this PR.

# Checklist:


frontend/www/js/omegaup/components/common/LogoutConfirmation.vue
```diff
@@ -0,0 +1,65 @@
+<!-- components/LogoutConfirmation.vue -->
```
pabo99, line 1: We currently use [Bootstrap 4](https://getbootstrap.com/docs/4.6/getting-started/introduction/) toolkit within omegaUp.

We also use [bootstrap-vue](https://bootstrap-vue.org/), which already includes a set of components that we can import to avoid re-creating new ones. In fact, there is a component specifically for displaying modals, which you're implementing in this component.

What do you think about using that instead? You can find the documentation for implementing it [here](https://bootstrap-vue.org/docs/components/modal#using-v-model-property)


We also have some examples in this [file](https://github.com/omegaup/omegaup/blob/main/frontend/www/js/omegaup/components/problem/creator/cases/CaseEdit.vue) that might help clarify its usage.

frontend/www/js/omegaup/components/common/LogoutConfirmation.vue
```diff
@@ -0,0 +1,65 @@
...
+    tabindex="-1"
+    role="dialog"
+    aria-labelledby="logoutConfirmationModalLabel"
+    aria-hidden="true"
+  >
+    <div class="modal-dialog" role="document">
+      <div class="modal-content">
+        <div class="modal-header">
+          <h5 class="modal-title" id="logoutConfirmationModalLabel">
+            {{ T.logoutConfirmationTitle }}
```
pabo99, line 15: The correct way to add translation strings is as follows:

Open the files: `frontend/templates/en.lang`, `frontend/templates/es.lang`, `frontend/templates/pt.lang`
Write the text in each language (just as you did in the frontend/www/js/omegaup/lang.en.ts file).
Once you have written the text in all three languages and saved the changes, run the following command:
```bash
./stuff/lint.sh
``` 

This script will propagate the texts to the files used in the components, ensuring that they are displayed in the user's preferred language.

frontend/www/js/omegaup/components/common/LogoutConfirmation.vue
```diff
@@ -0,0 +1,65 @@
...
+        <div class="modal-body">
+          {{ T.logoutConfirmationMessage }}
+        </div>
+        <div class="modal-footer">
+          <button
+            type="button"
+            class="btn btn-secondary"
+            data-dismiss="modal"
+          >
+            {{ T.no }}
```
pabo99, line 35: In this particular case, we already have a translation string located in the index `wordsNo`. You can use it to avoid creating a duplicate value.

frontend/www/js/omegaup/components/common/LogoutConfirmation.vue
```diff
@@ -0,0 +1,65 @@
...
+          </button>
+          <button
+            type="button"
+            class="btn btn-primary"
+            @click="confirmLogout"
+          >
+            {{ T.yes }}
```
pabo99, line 42: Same case, you can use `wordsYes` index

pabo99 (commented):
I left you some comments to help simplify the creation of the new feature. Let us know if you have any questions.

Saloni-2005:
Hello @pabo99, thank you for your review and suggestions.
I have made the suggested changes.
Kindly review them and let me know if any further modifications are needed.

frontend/www/js/omegaup/components/common/Navbar.vue
```diff
@@ -444,6 +448,7 @@ export default class Navbar extends Vue {
   @Prop() isUnder13User!: boolean;
   @Prop() mentorCanChooseCoder!: boolean;
   @Prop() userVerificationDeadline!: Date | null;
+  @Prop() logoutModalVisible: boolean = false;
```
pabo99, line 451: In Vue, component properties are immutable, which means that most of the time they should not be able to modify their value.

To handle this, we can use dynamic data. You can declare it as follows, and it should work:
```suggestion
  logoutModalVisible = false;
```

Right below line 454.

frontend/www/js/omegaup/components/common/LogoutConfirmation.vue
```diff
@@ -0,0 +1,40 @@
+<!-- components/LogoutConfirmation.vue -->
```
pabo99, line 1: I’m still considering whether it’s worth creating a component to include the existing modal component. In any case, it doesn’t affect performance, so we can leave it as you have it

pabo99 (commented):
Just a few suggestions, and this PR looks good to me.

frontend/www/js/omegaup/components/common/LogoutConfirmation.vue
Saloni-2005: Hello @pabo99, thank you so much for your review and suggestions. It helped me a lot and I have made the suggested changes.
Kindly review them and let me know if any further modifications are needed.

pabo99:
The PR looks good. You just need to run the following command (outside the container):
```bash
./stuff/lint.sh
``` 
If you have already run it, please let us know, and preferably share the output after executing the command.

This will help us diagnose if something is going on, as it would be the second case of contributors being unable to run the command successfully.

Saloni-2005:
Hello @pabo99, Yes I already run that command, Below is what I got : 

image

pabo99:
> Hello @pabo99, Yes I already run that command, This is what I got :
> 
> image

Ok, this makes sense. It looks like the instruction needs to be clearer:

The command must be executed inside WSL for it to take effect.

Saloni-2005:
Hello @pabo99
 I followed your suggestion and ran the command; however, it is taking a considerable amount of time to execute. Could you please provide an estimate of how long this process should typically take? Additionally, after stopping the previous execution and running the command again, there was still no response even though I waited for a significant duration.

Screenshot 2025-02-25 220407

Thankyou!

pabo99:
Ok, according to the image you shared, it's important to indicate that this is the first time you run this command. It requires the `hook_tools` image for the linters to work properly. Since it couldn't find that image, it will try to download it (only this first time). The estimated download time depends on your internet connection speed.

The next time you run the command, it should execute considerably faster than the first time.

At the end of the execution, you should see the message `"ok"`. I'm not sure if the command accepts a debug argument, but it would be a good indicator if the command doesn’t freeze in the middle of a process.

Let us know if you still experience any issues when running the command again.

Thanks to you for your patience!

Saloni-2005:
Hello @pabo99,
I followed your suggestion and successfully passed the command. However, the test cases are still not passing. I noticed that the tests which were failing are not considering the current content; instead, they are testing against the previous content. Could you please explain why this is happening?

This is the previous which is being considered : 
image


Whereas , this my current content : 
image

Thankyou!

pabo99:
Please run the following command:

```bash
./stuff/lint.sh fix -- frontend/www/js/omegaup/components/common/LogoutConfirmation.vue
```

It will be necessary to check if the Cypress tests are working properly after applying your changes.

Saloni-2005:
Thank you so much, @pabo99, for your guidance and time. I want to confirm if it's expected behavior for around 73 files to update when I ran the command to fix LogoutConfirmation.vue. Should I proceed with pushing them?


image

frontend/www/js/omegaup/components/common/Navbar.vue
```diff
@@ -316,10 +316,12 @@
                   }}</a>
                 </template>
                 <div class="dropdown-divider"></div>
-                <a class="dropdown-item" href="/logout/" data-logout-button>
-                  <font-awesome-icon :icon="['fas', 'sign-out-alt']" />
-                  {{ T.navLogOut }}
```
pabo99, line 321: Now you are no longer using this text string, which causes the linter to complain because all localization strings must be used at least once in the code. To fix this error, you need to go to the translation files and remove the entry for this string.

You can delete only the three files you modified (es.lang, en.lang, and pt.lang), and then run:
```bash
./stuff/lint.sh
```
so that the change is applied to the rest of the files.

pabo99 (commented):
Alright, we've made some progress. But there's still one linter error left before we can proceed with the rest of the issues.

pabo99:
You're very close to resolving this issue! There's just one detail with the Cypress tests.

Most of them use the logout option, which breaks due to the change you're proposing.

The issue is specifically in this function:

https://github.com/omegaup/omegaup/blob/40e7e529776f46c65e9a22c84c2f4ba1f4d21cda/cypress/support/commands.ts#L36-L41

First, it clicks on the button in the main navbar, then it clicks on the logout option, and finally, it waits for the URL to be the home page, which is where the logout redirects you.

Now, you need to add the intermediate step to select the button that appears in the confirmation modal to accept the logout.

This is a bit tricky since we are using an external component, but we can still solve it.

Considering that in omegaUp we try to avoid using the id attribute in component elements, and after reading the documentation:

https://bootstrap-vue.github.io/bootstrap-vue/docs/components/modal#comp-ref-b-modal-props

It looks like we could use the `footer-class` property, where the `b-modal` buttons are located. You could add a class to this property, for example:

```ts
footer-class="logout-confirmation-modal"
```
This way, you can now add the click action to the confirm button like this:

```ts
  cy.get('footer.logout-confirmation-modal>button.btn-primary').click();
  ```
 This should be executed right after clicking on the "Logout" menu option.

**Review by pabo99 — APPROVED** [review]
Thank you so much for all your patience! This PR is approved by me!

**Review by heduenas — APPROVED** [review]

File: frontend/templates/en.lang [hunk 8]
```diff
@@ -824,6 +824,8 @@ loginRequired = "You need to login to perform that action."
 loginSignUp = "Register"
 loginSignupHeader = "Create an omegaUp account, it's quick and easy!"
 loginThroughThirdParty = "Your account was created through a third-party login (Google, Facebook, etc.). Please login using that mechanism again in order to set a password."
+logoutConfirmationMessage = "Do you really want to logout?"
```
tvanessa, line 827: I suggest removing this text, so only display `title` and `buttons`

frontend/www/js/omegaup/components/common/LogoutConfirmation.vue
```diff
@@ -0,0 +1,40 @@
+<template>
+  <b-modal
+    v-model="showModal"
+    :title="T.logoutConfirmationTitle"
+    :ok-title="T.wordsYes"
+    :no-title="T.wordsNo"
```
pabo99, line 6: I realized something: the button that should display "No" was displaying "Cancel" in all three languages. So, I checked the documentation and noticed that the attribute has a different name:
```suggestion
    :cancel-title="T.wordsNo"
```
With this change, it should now display the correct text.

frontend/www/js/omegaup/components/common/LogoutConfirmation.vue
```diff
@@ -0,0 +1,40 @@
...
+    ok-variant="primary"
+    no-variant="secondary"
```
pabo99, line 8: Same case as the previous comment:
```suggestion
    cancel-variant="secondary"
```

pabo99 (commented):
Two more comments and that's good to me

frontend/templates/en.lang
pabo99: This suggestion hasn't been addressed yet.

frontend/www/js/omegaup/components/common/LogoutConfirmation.vue
pabo99: This suggestion still needs to be addressed as well

pabo99:
@tvanessa, is this PR good to go for you? Since @heduenas and I have approved it, @Saloni-2005 just needs your approval before the PR can be merged.

tvanessa (approved):

PR #8004, opened 2025-02-27
Add user profile information display in contest intro component

iqbalcodes6602:
# Description

Added the suggested solution of showing the fields which are not filled.

image

Fixes: #7780 7780

# Checklist:


iqbalcodes6602:
hi @pabo99 , i have implemented the changes as you said. please give it a look.

frontend/www/js/omegaup/components/contest/Intro.vue
```diff
@@ -186,6 +228,7 @@ export default class ContestIntro extends Vue {
   @Prop() statement!: types.PrivacyStatement;
   @Prop({ default: false })
   shouldShowModalToLoginWithRegisteredIdentity!: boolean;
+  @Prop() userBasicInformation!: any;
```
pabo99, line 231: In `TypeScript`, we try to avoid using the `any` type (just as in `PHP+PSALM`, we try to avoid the `mixed` type). This is because it’s a way to bypass types, and if we use `TypeScript` in that way, it defeats the purpose.

So, there are two possible solutions:

- Assign the correct type. You already have it, it was automatically created in the `api_types.ts` file.

- Or better, create an instance that you can call here. Here’s how to achieve that:
In the `Contest` controller, you added `userBasicInformation` to `ContestIntroPayload`. The type for this new property can also be added as a new type: `UserBasicInformation`, and for that new type, you can assign:
```php
 * @psalm-type UserBasicInformation=array{country: null|string, state: null|string, school: null|int}
```
Now you can assign the new type as follows: `userBasicInformation: UserBasicInformation` in `ContestIntroPayload`.

Then, when you run `./studd/lint.sh`, you can check if the new instance was created, and you can use it here as: `types.UserBasicInformation`.

frontend/www/js/omegaup/components/contest/Intro.vue
```diff
@@ -39,7 +39,49 @@
...
+                v-if="needsBasicInformation"
+                :class="['list-unstyled', 'text-muted', 'font-weight-light']"
+              >
+                <li>
+                  {{ T.wordsCountry }}:
+                  <span
+                    :class="
+                      userBasicInformation.country
+                        ? 'text-success'
+                        : 'text-danger'
```
pabo99, line 54: Since you are going to apply corrections, could you convert this into a computed property?

The same for the other two items.

frontend/server/src/Controllers/Contest.php
```diff
@@ -1003,14 +1001,22 @@ public static function getContestDetailsForTypeScript(
...
+        $userBasicInformation = [
+            'country' => $r->identity->country_id ?: null,
+            'state' => $r->identity->state_id ?: null,
+            'school' => $r->identity->current_identity_school_id ?: null,
+        ];
+
+        $needsBasicInformation = $needsBasicInformation && (
+            is_null($userBasicInformation['country']) ||
+            is_null($userBasicInformation['state']) ||
+            is_null($userBasicInformation['school'])
```
pabo99, line 1013: Excellent! Now it looks more readable! :tada:

frontend/server/src/Controllers/Contest.php
```diff
@@ -1003,14 +1001,22 @@ public static function getContestDetailsForTypeScript(
...
         ] = \OmegaUp\DAO\Contests::getNeedsInformation($contest->problemset_id);
 
-        $result['templateProperties']['payload']['requestsUserInformation'] = $requestsUserInformation;
-        $result['templateProperties']['payload']['needsBasicInformation'] =
-            $needsBasicInformation &&
-            (
-                !$r->identity->country_id || !$r->identity->state_id ||
-                is_null($r->identity->current_identity_school_id)
+        $userBasicInformation = [
+            'country' => $r->identity->country_id ?: null,
```
pabo99, line 1005: This would be evaluating if the user's `country_id` is not `null`, and if so, you would be assigning the `null` value, which is redundant. Or am I missing something important? If not, can you directly assign the value without using the "Elvis operator"?

The same applies to the other two properties.

pabo99 (commented):
I suggested a few corrections, and with that, it's good for me

frontend/www/js/omegaup/components/contest/Intro.vue
iqbalcodes6602: converted to computed property

frontend/server/src/Controllers/Contest.php
iqbalcodes6602: yeah, i just thought we can add some null check, but what you are saying is better so i have implemented that.

pabo99 (approved):
It LGTM! Thank you @iqbalcodes6602 !

heduenas (approved):
This is great, thank you for fixing this issue!

PR #8013, opened 2025-03-05
Allow downloading all submissions from a completed contest

iqbalcodes6602:
# Description

Added the bulk action button the contest submission tab for the admin to download all the codes of different user.
The zip file will contain the list of folders by the problem name as in the contest and inside each folder there would be a list of submissions denoted by the user name and extension would be language of the code.

image

Fixes: #893 

# Checklist:


frontend/server/src/Controllers/Run.php
```diff
@@ -1515,6 +1515,128 @@ private static function getGraderResource(
         return $result;
     }
 
+    public static function apiDownloadMultipleRuns(\OmegaUp\Request $r): void {
```
heduenas, line 1518: You need to call `isProblemAdmin` like we do for the single-download function.

iqbalcodes6602: > You need to call `isProblemAdmin` like we do for the single-download function.
Okay, so we need to allow only admins to download all the files? Sure will do it.

iqbalcodes6602: i had one question regarding this. this functionality is at contest level so we need to allow only the contest's admins to download all the submissions.
But as you are saying allowing isProblemAdmin to download would allow non admins of the contest to download the runs as well?

Am i correct regarding this? can you please clarify this.

thank you

iqbalcodes6602:
hi @heduenas  , i tried doing what you said, but its now not letting the contest admin to download all the files. Instead what i have done is allowing the canViewSubmission check for downloading all the files. do let me know if it is correct.

iqbalcodes6602:
hi @heduenas , i tried doing what you said, but its now not letting the contest admin to download all the files. Instead what i have done is allowing the canViewSubmission check for downloading all the files. do let me know if it is correct.

iqbalcodes6602:
hi @heduenas  , i tried doing what you said, but its now not letting the contest admin to download all the files. Instead what i have done is allowing the canViewSubmission check for downloading all the files. do let me know if it is correct.

pabo99:
I believe I understand what @heduenas is suggesting: contest admins should only be allowed to download files corresponding to submissions for problems they manage. Is that correct?

frontend/server/src/Controllers/Run.php
```diff
@@ -1532,6 +1532,127 @@ private static function getGraderResource(
...
+                \ZipArchive::CREATE | \ZipArchive::OVERWRITE
+            ) !== true
+        ) {
+            throw new \OmegaUp\Exceptions\InternalServerErrorException(
+                'zipCreationFailed'
+            );
+        }
+
+        foreach ($runs as $run) {
+            if (
```
pabo99, line 1565: Here you could add the condition to check if the user is a problem admin; otherwise, you should skip the download, just like this validation does.

iqbalcodes6602: ohhh, yes, now i finally get it now.
Thank you so much for helping me.
I have added the check in  ```getSourceCode``` function on line no. 1636

pabo99 (approved):
It looks good to me! Thank you!

frontend/server/src/Controllers/Run.php
heduenas: Actually, the validation should be that the user is admin of the contest.

It seems that this funciton takes a list of runs? Maybe it should take a contest id and then query all the runs for it and download them all.

And we do need unit tests to make sure that users that are not contest admins cannot download the codes.

iqbalcodes6602: okay, now i completely understand what needs to be done. Will update accordingly
thank you.

Issue #8014, opened 2025-03-06
Create index for column `Name` in table `Schools` and use it in DAO


heduenas:
1. Create index for column `Name` in table `Schools`.
2. Change [this query](https://github.com/omegaup/omegaup/blob/main/frontend/server/src/DAO/Schools.php#L31) to do `WHERE s.name LIKE CONCAT(?, \'%\')`, that way the index will be utilized.

Kashaf1175:
can i work on this issue

pushkardev123:
can i get assigned to this?

isabelccc:
https://github.com/omegaup/omegaup/pull/8144

antedotee:
This issue has been inactive for a long time. Can I work on this, I have contributed to omegaup in the past.
Thanks and regards.

pabo99:
Thanks for your interest in contributing, @antedotee! You can follow the discussion in PR #8016, which addresses the first part of the issue, specifically the creation of the index. 

Some improvements to the query are still pending. Feel free to ask @sudarsan2k5 in the PR whether he’s still working on that part or if you could give it a try yourself

sudarsan2k5:
I was working on this issue earlier, but since another contributor showed interest, I decided not to continue. If @antedotee is interested in taking it up, please go ahead.

PR #8016, opened 2025-03-06
Add index on Schools.name to optimize search performance

sudarsan2k5:
# Description

Create index for column Name in table Schools.
Changed [this query](https://github.com/omegaup/omegaup/blob/main/frontend/server/src/DAO/Schools.php#L31) to do WHERE s.name LIKE CONCAT(?, \'%\'), that way the index will be utilized.

Part of: #8014


sudarsan2k5:
Could you please check why this test is failing?

frontend/server/src/DAO/Schools.php
```diff
@@ -28,7 +28,7 @@ public static function findByName($name) {
             FROM
                 Schools s
             WHERE
-                s.name LIKE CONCAT(\'%\', ?, \'%\')
+                s.name LIKE CONCAT(?, \'%\')
```
pabo99, line 31: Could you please elaborate on the reason for this change?

pabo99 (commented):
As the error indicates:

```bash
'frontend/database/schema.sql' and 'frontend/database/dao_schema.sql' cannot be modified in the same commit.
``` 

These files cannot be modified in the same commit (or PR, actually). So, as long as this PR contains only the changes in the `schema.sql` file, it should be fine.

sudarsan2k5:
OK, why changing the query from ` s.name LIKE CONCAT(\'%\', ?, \'%\')` to `s.name LIKE CONCAT(?, \'%\')` is important for performance when using an index.

#### The Performance Issue with LIKE `'%...%'`

- When use a pattern like `LIKE '%term%'` (which searches for a term anywhere in the string), MySQL cannot effectively use an index on that column. Here's why:

- Database indexes (typically B-tree indexes) work by sorting the values and allowing the database to quickly find entries based on this sorted order.
- When we have a wildcard `(%)` at the beginning of a `LIKE pattern`, the database can't use the sorted nature of the index because it doesn't know where to start in the index. It has to scan every entry to see if the pattern matches anywhere in the string.

#### Why `LIKE 'term%'` Is Better
- When change to `LIKE 'term%'` (which searches for strings that start with the term):

1. Index Can Be Used:
The database can use the index to quickly find all entries that start with `"term"` because the index is sorted. It knows exactly where to start looking in the index.

2. Significant Performance Improvement:
This can make the query orders of magnitude faster, especially on large tables. Instead of scanning every row, it can jump directly to the relevant section of the index.

#### But it has a downside.

The downside of this change is that it alters the functionality:
 For example:- 
- Original query `(LIKE '%term%')`: Finds schools with the `term` anywhere in their name (e.g., "University of California", "California State University")
- New query `(LIKE 'term%')`: Only finds schools where the name starts with the `term` (e.g., "University of California" but not "California State University")



If my changes cause any issues, I will revert them?

pabo99:
Great explanation! Could you add it as a comment in the query? Or as a PHP comment before the query? Either option is valid.

frontend/server/src/DAO/Schools.php
```diff
@@ -28,6 +28,9 @@ public static function findByName($name) {
             FROM
                 Schools s
             WHERE
+                /*
+                Using LIKE CONCAT(?, \'%\') instead of LIKE CONCAT(\'%\', ?, \'%\') to utilize the index on the name column.
+                 */
```
pabo99, line 33:
```suggestion
                -- Using LIKE CONCAT(?, \'%\') instead of LIKE CONCAT(\'%\', ?, \'%\')
                -- to utilize the index on the name column.
```

pabo99:
It looks like the linter doesn’t allow adding `?` in queries because it expects the same number of arguments to be passed. The ideal solution would be to add the comment from PHP.

sudarsan2k5:
As the  suggestion, I tried that and ran the test locally, but it also failed. So, I thoroughly went through the codebase and understood why the tests are failing. The tests are failing because the changes to the `findByName` function have altered its behavior. The function now only finds schools whose names start with the search term, rather than schools whose names contain the search term anywhere.


Screenshot 2025-03-13 at 12 33 42 PM


https://github.com/omegaup/omegaup/blob/4d4257282fe1d0035896dd643a1b7c0295480f41/frontend/tests/controllers/SchoolCreateTest.php#L94-L109

- in `testSchoolsList` , it expects searching for "school" to find "all school", "all new school", and "all old school" but With LIKE `term%` , it would only find schools that start with "school", not those that contain it elsewhere so we have to revert the functionality

pabo99 (approved):
Considering that we couldn't change the filtering query, it looks good to me (LGTM).

heduenas (approved):

heduenas:
Thank you @sudarsan2k5 for adding this index. I will still keep the issue open because we will still need to change the PHP code and tests to use the new query pattern. Let me know if you want to do that part as well, otherwise I can try to find someone else interested in it.

sudarsan2k5:
Okay, I'm working on this.

pabo99:
Now that the index has been created in the `schema.sql` file, it's time to add it to the `dao_schema.sql` file.

You can create a new issue and work on the solution. In fact, you can copy all the content from `schema.sql` into `dao_schema.sql`. It doesn't matter if additional changes are applied, as this file might be outdated

Issue #7974, opened 2025-02-02
[FEATURE] Como un mantenedor de la plataforma quiero tener un reporte con la lista de usuarios por perfil


pabo99:
A lo largo del tiempo se ha tenido la necesidad de crear distintos perfiles para realizar acciones que deben ser restringidas.

A su vez se han agregado usuarios a los ACLs a tal grado de no saber si dichos usuarios siguen utilizando la plataforma.

Lo que se espera es crear un reporte que indique el listado de perfiles y los usuarios que pertenecen a ellos. Así mismo se podría incluir el listado de acciones que puede realizar cada perfil.

## Contexto

Para poder tomar decisiones de que perfiles siguen siendo necesarios y que usuarios deberían pertenecer a cada perfil es importante tener la mayor cantidad de información posible

iqbalcodes6602:
@pabo99 , hi I would like to give it a try, can you please assign it to me?

pabo99:
How about starting to work on the solution for issue #7973, which was recently assigned to you?

brauliorivas:
Hey @pabo99, I would like to work on this.

iqbalcodes6602:
> Hey @pabo99, I would like to work on this.

Hi @brauliorivas i would be working on this, since I have completed the work assigned to me.

heduenas:
The UX I would suggest is adding a adding an item "My ACLs" in the username dropdown. When clicked it would take you to a view where you will be able to see which users are part of the ACLs you yourself are part of (only the ones you are part of, for privacy). Users that are not part of any ACL do not see the "My ACL" item in the dropdown. @pabo99 @iqbalcodes6602 , what do you think?

iqbalcodes6602:
> The UX I would suggest is adding a adding an item "My ACLs" in the username dropdown. When clicked it would take you to a view where you will be able to see which users are part of the ACLs you yourself are part of (only the ones you are part of, for privacy). Users that are not part of any ACL do not see the "My ACL" item in the dropdown. [@pabo99](https://github.com/pabo99) [@iqbalcodes6602](https://github.com/iqbalcodes6602) , what do you think?

(only the ones you are part of, for privacy).

In My Acls page we can show the user, list of users who are part of ACL created by the user, ie all the ACL which the user is owner of for privacy?

PR #8021, opened 2025-03-09
Add userProfileReport API and ACL retrieval methods for Groups, Cours…

iqbalcodes6602:
…es, Problems, and Contests

# Description

Made the api for getting a list of users with their roles assigned to various ACLs. 

Fixes: #7974 

# Comments

Will be adding the UI update in another PR.

# Checklist:


iqbalcodes6602:
hi @heduenas , i have changed the ux for this feature as you asked, could you give it a look.

frontend/server/src/Controllers/ACL.php
```diff
@@ -62,4 +62,118 @@ public static function removeGroup(
...
+        if ($contest = \OmegaUp\DAO\Contests::getByAclId($aclId)) {
+            return ['type' => 'contest', 'alias' => $contest->alias ?? null];
+        } elseif ($course = \OmegaUp\DAO\Courses::getByAclId($aclId)) {
+            return ['type' => 'course', 'alias' => $course->alias ?? null];
+        } elseif ($problem = \OmegaUp\DAO\Problems::getByAclId($aclId)) {
+            return ['type' => 'problem', 'alias' => $problem->alias ?? null];
+        } elseif ($group = \OmegaUp\DAO\Groups::getByAclId($aclId)) {
+            return ['type' => 'group', 'alias' => $group->alias ?? null];
+        }
+        return ['type' => 'unknown', 'alias' => null];
```
pabo99, line 177: Is there any way to retrieve this information in a single round trip to the DB? I'm concerned that getting a `group-type` `ACL` requires four separate queries. Additionally, it's being called inside a `foreach` loop.

This definitely needs to be restructured to improve how the information is retrieved.

Perhaps you could create a function in the `ACL` `DAO` with a query that joins each table and only fetches the alias from each of them, since that seems to be the only field used here

frontend/server/src/Controllers/ACL.php
```diff
@@ -62,4 +62,118 @@ public static function removeGroup(
...
+        foreach ($allACLs as $acl) {
+            if (
+                !isset(
+                    $acl->acl_id,
+                    $acl->owner_id
+                ) || $acl->owner_id !== $userId
+            ) {
+                continue;
+            }
+            $aclData = self::getAclType($acl->acl_id);
```
pabo99, line 96: To avoid this `foreach`, it would be necessary for the query I suggested in the `getACLType` function to handle retrieving all the information, if possible.

To ensure that your new query is returning the expected data, I recommend adding a test where you assert that the retrieved data is correct.

frontend/server/src/Controllers/ACL.php
```diff
@@ -62,4 +62,118 @@ public static function removeGroup(
...
+            'acls' => array_values($userACLs),
+            'roles' => $aclRoles,
+        ];
+    }
+
+    /**
+     * Determines the type and alias of an ACL.
+     *
+     * @param int $aclId The ACL ID.
+     * @return array{type: string, alias: ?string}
```
pabo99, line 165: Is there a way to ensure that the alias is not null? Since it’s the identifier that will be used, it would be too confusing to understand what it refers to from the UI.

frontend/server/src/Controllers/ACL.php
```diff
@@ -62,4 +62,118 @@ public static function removeGroup(
...
+            }
+
+            // Get user identity
+            $identity = \OmegaUp\DAO\Identities::findByUserId(
+                $userRole->user_id
+            );
+            $username = $identity->username ?? 'unknown';
+
+            // Get role name & description
+            $roleData = $roleMap[$userRole->role_id] ?? ['name' => 'Unknown', 'description' => 'Unknown role'];
```
pabo99, line 143: It's not good idea to use magic values for `usernames`. In what cases would the `username` be `null`?
In the worst case, that data shouldn't be displayed at all.

But first, I want to understand when this situation would occur.

frontend/server/src/Controllers/ACL.php
```diff
@@ -62,4 +62,118 @@ public static function removeGroup(
...
+                    $userACLs[$userRole->acl_id]
+                )
+            ) {
+                continue;
+            }
+
+            // Get user identity
+            $identity = \OmegaUp\DAO\Identities::findByUserId(
+                $userRole->user_id
+            );
```
pabo99, line 139: Here as well, it will be necessary to create a query that retrieves all the users assigned to that role, in order to avoid additional database roundtrips.

frontend/server/src/Controllers/ACL.php
```diff
@@ -62,4 +62,118 @@ public static function removeGroup(
...
+     *
+     * @param \OmegaUp\Request $r The request object containing user session information.
+     * @throws \OmegaUp\Exceptions\NotFoundException If the user is not found.
+     * @return array{acls: list<array{acl_id: int, type: string, alias: ?string}>, roles: list<array{acl_id: int, user_id: int, username: string, role_id: int, role_name: string, role_description: string}>}
+     */
+    public static function apiUserOwnedAclReport(\OmegaUp\Request $r): array {
+        // Ensure user is authenticated
+        $r->ensureMainUserIdentity();
+
+        /** @var null|\OmegaUp\DAO\VO\Identities $user */
```
pabo99, line 77: In what case would Identities be `null`?

In line 75, you're ensuring that the user is authenticated, and also that it's the user's main identity. That implies there must be both a user and an identity.

frontend/server/src/Controllers/ACL.php
```diff
@@ -62,4 +62,118 @@ public static function removeGroup(
...
+        $user = $r->identity;
```
pabo99, line 78: For consistency:
```suggestion
        $identity = $r->identity;
```
If you're interested in the user, you can access it directly via `$r->user`

frontend/server/src/Controllers/ACL.php
iqbalcodes6602: yes updated in the lastes commit

frontend/server/src/Controllers/ACL.php
iqbalcodes6602: yes you are correct the username would never be null, i just wanted to be cautious, well now i have removed these checks in the latest commit

frontend/server/src/Controllers/ACL.php
iqbalcodes6602: i couldn't get what do you mean by adding test. Do you mean the PHP tests?

frontend/server/src/Controllers/ACL.php
iqbalcodes6602: i have added another function for this purpose in the ACL DAO file in the lastest commit

iqbalcodes6602:
hi @pabo99 , i tried to remove the redundant for loops and database round trips.
I have also rezstructured the response.
Earlier we were sending the acls and the users in the those acls seperately, and i planned to merge them in the frontend.
But now i am sending the users associted with each acl inside that very object only. Hence only one extra for loop is required now.
I have also updated the description to show the same updated response as well

frontend/server/src/DAO/ACLs.php
```diff
@@ -10,4 +10,43 @@
...
+        $params = [$userId, $userId, $userId, $userId];
+
+        $rows = \OmegaUp\MySQLConnection::getInstance()->GetAll($sql, $params);
+
+        return array_map(fn($row) => [
+            'acl_id' => intval($row['acl_id']),
+            'type' => strval($row['type']),
+            'alias' => strval($row['alias']),
+            'users' => [],
+        ], $rows);
```
pabo99, line 50: There's no need to assert the type of each element here if MySQL already knows each of them:
```suggestion
        return array_map(fn($row) => [
            'acl_id' => $row['acl_id'],
            'type' => $row['type'],
            'alias' => $row['alias'],
            'users' => [],
        ], $rows);
```
So instead, you can simply specify the return type of the `GetAll()` function like this:
```php
/** @var list<array{acl_id: int, alias: string, type: string}> */
```

Even better, why not return only the result of the query in this function? That way, you can remove the `users` field, which isn't actually used here.

frontend/server/src/DAO/ACLs.php
```diff
@@ -10,4 +10,43 @@
...
+            FROM ACLs a
+            INNER JOIN Problems p ON p.acl_id = a.acl_id
+            WHERE a.owner_id = ?
+            UNION
+            SELECT a.acl_id, g.alias, "group" AS type
+            FROM ACLs a
+            INNER JOIN Groups_ g ON g.acl_id = a.acl_id
+            WHERE a.owner_id = ?
+        ';
+        $params = [$userId, $userId, $userId, $userId];
```
pabo99, line 41: If you want to avoid repeating the same variable multiple times:
```suggestion
        $params = array_fill(0, 4, $userId);
```

frontend/server/src/Controllers/ACL.php
```diff
@@ -62,4 +62,88 @@ public static function removeGroup(
...
+                is_null(
+                    $role->role_id
+                ) || is_null(
+                    $role->name
+                ) || is_null(
+                    $role->description
+                )
+            ) {
+                continue;
+            }
```
pabo99, line 103: No field in the Roles table allows null values, so this can be removed.
```suggestion
```

frontend/server/src/Controllers/ACL.php
```diff
@@ -62,4 +62,88 @@ public static function removeGroup(
...
+            $roleMap[$role->role_id] = [
+                'name' => $role->name,
+                'description' => $role->description,
+            ];
+        }
+
+        // Get all user roles for these ACLs
+        $userRoles = \OmegaUp\DAO\UserRoles::getByAclIds($aclIds);
```
pabo99, line 111: What do you think about retrieving the usernames right here and avoiding an extra round trip to the DB in `getUsernamesByUserIds`?

frontend/server/src/Controllers/ACL.php
```diff
@@ -62,4 +62,88 @@ public static function removeGroup(
...
+        $usernames = \OmegaUp\DAO\Identities::getUsernamesByUserIds($userIds);
+
+        foreach ($userRoles as $userRole) {
+            if (
+                is_null($userRole->acl_id) ||  // Changed === null to is_null()
+                is_null($userRole->user_id) || // Changed === null to is_null()
+                is_null($userRole->role_id)   // Changed === null to is_null()
+            ) {
+                continue;
+            }
```
pabo99, line 131: In the `UserRoles` table, all fields are `non-nullable`. Could this condition be removed?

frontend/server/src/Controllers/ACL.php
```diff
@@ -62,4 +62,88 @@ public static function removeGroup(
...
+            if (!isset($aclMap[$aclId])) {
+                continue;
+            }
+
+            $aclMap[$aclId]['users'][] = [
+                'user_id' => $userRole->user_id,
+                'username' => $usernames[$userRole->user_id],
+                'role_id' => $userRole->role_id,
+                'role_name' => $roleMap[$userRole->role_id]['name'] ?? 'Unknown',
+                'role_description' => $roleMap[$userRole->role_id]['description'] ?? 'Unknown',
```
pabo99, line 143: I still don't understand in what case this could be unknown. If you can reproduce this in a PHPUnit test, I can figure out the reason you're using it

pabo99 (commented):
Many unnecessary round trips to the DB have already been reduced! 

But there are still things that can be improved. I’ve left you some comments that will help.

frontend/server/src/DAO/ACLs.php
iqbalcodes6602: i have removed the type assertion.
But we still need the emtpy users array for each  acl type since we need to append the user in each of the acl.
If we do not initialised the empty users array here then we need to run a for each loop again in order to have an empty users array to each acl of the user

frontend/server/src/Controllers/ACL.php
iqbalcodes6602: when i try to remove this, the psalm test are giving errors

```
::error file=frontend/server/src/Controllers/ACL.php,line=93,col=13,title=PossiblyNullArrayOffset::frontend/server/src/Controllers/ACL.php:93:13: PossiblyNullArrayOffset: Cannot access value on variable $roleMap[$role->role_id] using possibly null offset int|null (see https://psalm.dev/125)
```

frontend/server/src/Controllers/ACL.php
iqbalcodes6602: same as above, if  i dont use the null checks the psalm test gives errors

::error file=frontend/server/src/Controllers/ACL.php,line=138,col=13,title=PossiblyNullArrayOffset::frontend/server/src/Controllers/ACL.php:138:13: PossiblyNullArrayOffset: Cannot access value on variable $aclMap[$aclId] using possibly null offset int|null (see https://psalm.dev/125)
::error file=frontend/server/src/Controllers/ACL.php,line=140,col=31,title=PossiblyNullArrayOffset::frontend/server/src/Controllers/ACL.php:140:31: PossiblyNullArrayOffset: Cannot access value on variable $usernames using possibly null offset int|null (see https://psalm.dev/125)

frontend/server/src/Controllers/ACL.php
iqbalcodes6602: i think its because in this file ```/frontend/server/src/DAO/VO/Roles.php```
line numbers - 
56 - @var int|null 
63 - @var string|null
70 - @var string|null

defines the type with null

iqbalcodes6602:
> Many unnecessary round trips to the DB have already been reduced!
> 
> But there are still things that can be improved. I’ve left you some comments that will help.

i have updated the code accordingly.

frontend/server/src/Controllers/ACL.php
iqbalcodes6602: sure i have removed the unknown check.

frontend/server/src/Controllers/ACL.php
pabo99: You're right, let's keep the validations.

frontend/server/src/DAO/UserRoles.php
```diff
@@ -331,4 +331,44 @@ public static function getSystemGroups(int $identityId): array {
...
+        $rows = \OmegaUp\MySQLConnection::getInstance()->GetAll($sql, $aclIds);
+
+        /** @var list<array{acl_id: int, user_id: int, role_id: int, username: string}> */
+        $result = [];
+        foreach ($rows as $row) {
+            $result[] = [
+                'acl_id' => intval($row['acl_id']),
+                'user_id' => intval($row['user_id']),
+                'role_id' => intval($row['role_id']),
+                'username' => strval($row['username']),
```
pabo99, line 368: You can avoid parsing all this data if you put the type declaration in the right place (L358)

frontend/server/src/Controllers/ACL.php
pabo99: It's okay

frontend/server/src/Controllers/ACL.php
```diff
@@ -62,4 +62,78 @@ public static function removeGroup(
...
+        }
+
+        // Get all user roles for these ACLs
+        $userRoles = \OmegaUp\DAO\UserRoles::getByAclIds($aclIds);
+        $userIds = array_filter(
+            array_column(
+                $userRoles,
+                'user_id'
+            ),
+            fn($id) => !is_null($id) // Changed !== null to !is_null()
```
pabo99, line 117: nit:
```suggestion
            fn($id) => !is_null($id)
```

frontend/server/src/DAO/ACLs.php
```diff
@@ -10,4 +10,44 @@
  * @access public
  */
 class ACLs extends \OmegaUp\DAO\Base\ACLs {
+        /**
+     * Get all ACLs owned by a user along with their type, alias, and an empty users array.
+     *
+     * @param int $userId The user ID whose ACLs are being fetched.
+     * @return list<array{acl_id: int, type: string, alias: string, users: list<array<string, never>>}>
```
pabo99, line 17: Ok, está perfecto que se declare desde inicio `users`. Lo que faltaría es agregar el tipo correcto, si ya sabes que es lo que contendrá este elemento:
```suggestion
     * @return list<array{acl_id: int, type: string, alias: string, users: list<array{user_id: int, username: string, role_id: int, role_name: string, role_description: string}>}>
```

frontend/server/src/Controllers/ACL.php
```diff
@@ -62,4 +62,78 @@ public static function removeGroup(
...
+        );
+        /** @var list<int> $userIds */
+        $userIds = array_map('intval', array_values(array_unique($userIds)));
```
pabo99, line 120: What happens if you avoid parsing to integer?
```suggestion
        $userIds = array_values(array_unique($userIds));
```

At this point, we already know that the `$userIds` elements are of type `int`.

pabo99 (commented):
Just a few minor changes

frontend/server/src/DAO/UserRoles.php
iqbalcodes6602: updated in lastest commit

frontend/server/src/Controllers/ACL.php
iqbalcodes6602: updated in lastest commit

frontend/server/src/DAO/ACLs.php
iqbalcodes6602: updated in lastest commit

frontend/server/src/Controllers/ACL.php
iqbalcodes6602: updated in lastest commit

iqbalcodes6602:
> Just a few minor changes

Hi @pabo99 , i have fixed everything as told.

pabo99 (approved):
It LGTM! Thank you!

PR #8029, opened 2025-03-13
UI for a report with the list of users by profile

iqbalcodes6602:
# Description

Added the ui for displaying the report of list of users with acl roles assigned to them

image

Fixes: #7974

# Comments:
This pr is intended just for the ui part. Please review the api pr first and merge it before this one. It also mentioned in the same issue. So this way we won't be having much changes in this pr.

# Checklist:


heduenas:
@iqbalcodes6602 I think what the issue is asking for for a way to see the list of users that are part of of each ACL. If I understand correctly your change adds a way to see the ACLs a user belongs to.  In the bug I added details about the UX I expect to see in the fix to this issue: https://github.com/omegaup/omegaup/issues/7974

iqbalcodes6602:
> @iqbalcodes6602 I think what the issue is asking for for a way to see the list of users that are part of of each ACL. If I understand correctly your change adds a way to see the ACLs a user belongs to. In the issue I added details about the UX I expect to see in the fix: #7974

hi @heduenas , apologies for the confusion. i think i misunderstood the issue, i thought, we need to show the admin only the the list of users and what acl role they have, and the admin can decide which user should have access to which acl role.

Will work on the issue based on the UX you asked for.

Issue #7478, opened 2024-03-10
[BUG][LOW PRIORITY] Date picker offsets the day by one when setting birth, graduation dates


HNOONa-0:
## Current Behavior

Here is a video showing what I mean, you can see that after I select a select a date and click save, the day gets decremented by 1:

https://github.com/omegaup/omegaup/assets/59091536/94dfa9e0-9eac-453c-bf55-f83929fdc51f

## Expected Behavior

The date should exactly what the user selects.
This is easily fixed by using the date portion (year,month,day) and leaving out the time portion of the date object.

pushkardev123:
please assign me this

pabo99:
Thank you for your interest, @pushkardev123! The issue is now assigned to you!

pushkardev123:
Is there any guide for starting the local dev environment of the repo?

HNOONa-0:
> Is there any guide for starting the local dev environment of the repo?

There was a video on how to setup the project locally.also join the discord server.

pushkardev123:
frontend-1     | ERROR 1410 (42000) at line 1: You are not allowed to create a user with GRANT
frontend-1     | 2025-03-01 18:19:12,972 INFO exited: developer-environment (exit status 1; not expected)
frontend-1     | 2025-03-01 18:19:13,974 INFO spawned: 'developer-environment' with pid 2552


i am getting this error while setting up the local environment, and ive ensured the code is running inside the container yet i get this error are there any other troubleshoot options?

heduenas:
Does this bug still reproduce? What browser/OS does it reproduce in?
I couldn't reproduce this issue in Firefox on MacOS.

pushkardev123:
## Case Study: Timezone Offset Issue in Date of Birth Handling

### Overview
The issue arises when a user in Indian Standard Time (IST, UTC+5:30) or any other comparable timezone selects a date for their Date of Birth (DOB). Due to timezone conversion logic in our application, the selected date is shifted by one day when it is converted to UTC before being sent to the backend. This discrepancy can cause the stored or displayed date to be incorrect.

### Detailed Explanation

1. **User Input in IST:**
   - When a user selects a date (for example, "2004-07-24") using our date picker, it is interpreted as "2004-07-24 00:00:00" in IST.
   - In IST, midnight of "2004-07-24" actually corresponds to "2004-07-23 18:30:00" in UTC because IST is 5 hours and 30 minutes ahead of UTC.

2. **Conversion to UTC:**
[...]

PR #8039, opened 2025-03-15
Fixed birthdate offset issue and improved validation

pushkardev123:
# **Description**  

This PR fixes the issue where the birthdate offset was incorrectly set when updating a user's profile. The validation logic has been improved to ensure accuracy in handling date transformations. Additionally, the implementation now correctly converts and stores the birthdate without introducing an unintended offset.  

### **Changes made:**  
- Fixed incorrect birthdate offset issue.  
- Improved date validation and conversion logic.  
- Ensured compatibility with existing data formats.  

Fixes:  #7478  

# **Comments**  

- Reviewers should verify that date transformations work correctly across different time zones.  
- Please ensure the change does not break existing user data.  

# **Checklist:**  

- [x] The code follows the [[coding guidelines](https://github.com/omegaup/omegaup/wiki/Coding-guidelines)](https://github.com/omegaup/omegaup/wiki/Coding-guidelines) of omegaUp.  
- [ ] If this is a new feature, new tests were added.  
- [ ] If the change is large (> 200 lines), it was split into multiple PRs (e.g., controllers + PHPUnit tests in one, UI + Jest/Cypress tests in another).


heduenas:
@pushkardev123 I couldn't reproduce this bug on my end, but if it still reproduces we need to fix in the frontend code.

pushkardev123:
Actually the code had a problem with timezones, your timezone doesnt reflect the offset, but if the server is fetched from elsewhere it offsets the DOB by 1 due to timezone conversions. I have fixed the bug in the frontend controller by ensuring the conversion is done right. This should work fine for everyone now.

pushkardev123:
Im adding a new commit to this, i found out the actual error is with the frontend fetching the dob from database and converting it as per local-time, will look into it and update soon.

pushkardev123:
### Issue Summary  
The main problem was related to timezone differences affecting the Date of Birth (DOB) value. Initially, the backend stored the correct DOB, but due to timezone conversions when fetching and displaying the data on the frontend, the date was getting offset by one day in certain cases. This happened because the frontend was interpreting the DOB based on local timezone settings rather than as a UTC date.

### How I Fixed It  
- Ensured that the frontend correctly converts and displays the DOB as a UTC date rather than applying local timezone offsets.  
- Adjusted the conversion logic in the frontend controller to normalize the date handling across different timezones.  
- Added proper validation to prevent incorrect date shifts when fetching from the database.

### Why This Fix Works  
By enforcing consistent date handling in UTC across the frontend, we eliminate unintended shifts caused by timezone discrepancies. Now, regardless of where the data is fetched or displayed, the DOB remains accurate for all users.

This should now work correctly for everyone. Let me know if any further change is required!

frontend/server/src/Controllers/User.php
```diff
@@ -2523,7 +2523,7 @@ public static function apiUpdate(\OmegaUp\Request $r): array {
                 );
             }
             $r['birth_date'] = $birthDate;
-        }
+        }
```
heduenas, line 2526: the linter will probably complain about these spaces, please remove.

frontend/www/js/omegaup/components/user/BasicInformationEdit.vue
```diff
@@ -165,13 +165,22 @@ export default class UserBasicInformationEdit extends Vue {
       return;
     }
 
+    // Ensure birthDate is converted to a UTC Date object or set to null if invalid
```
heduenas, line 168: For future reference, can you extend this comment to mention what happens if the date is not converted to UTC? Add a link to the bug https://github.com/omegaup/omegaup/issues/7478

heduenas:
> ### Issue Summary
> 
> The main problem was related to timezone differences affecting the Date of Birth (DOB) value. Initially, the backend stored the correct DOB, but due to timezone conversions when fetching and displaying the data on the frontend, the date was getting offset by one day in certain cases. This happened because the frontend was interpreting the DOB based on local timezone settings rather than as a UTC date.
> ### How I Fixed It
> 
>     * Ensured that the frontend correctly converts and displays the DOB as a UTC date rather than applying local timezone offsets.
> 
>     * Adjusted the conversion logic in the frontend controller to normalize the date handling across different timezones.
> 
>     * Added proper validation to prevent incorrect date shifts when fetching from the database.
> 
> 
> ### Why This Fix Works
> 
> By enforcing consistent date handling in UTC across the frontend, we eliminate unintended shifts caused by timezone discrepancies. Now, regardless of where the data is fetched or displayed, the DOB remains accurate for all users.
> 
> This should now work correctly for everyone. Let me know if any further change is required!

Thank you @pushkardev123 for that detailed explanation. It makes sense. I left a couple of minor comments. After that we can merge your PR.

pushkardev123:
resolved the comments, pr is good to merge now i guess. @heduenas

pabo99:
> resolved the comments, pr is good to merge now i guess. @heduenas

It seems like the linter is still complaining. Could you please run the following command to fix it?

```bash
./stuff/lint.sh fix -- frontend/www/js/omegaup/components/user/BasicInformationEdit.vue
```

We have noticed that the process for running the linter is still not entirely clear.

This command must be executed outside the Docker container, but it also needs to be run from a Linux instance. If you are working on Windows, you will need to run it from WSL2. If you are using a different setup, please let us know so we can guide you on how to run this command.

pushkardev123:
I ran this command in linux terminal outside the container and the changes made by linter are included in the latest commit.

frontend/www/js/omegaup/components/user/BasicInformationEdit.vue
```diff
@@ -165,13 +165,27 @@ export default class UserBasicInformationEdit extends Vue {
...
+    // Otherwise, timezone discrepancies may cause incorrect DOB display.
+    // See bug report: https://github.com/omegaup/omegaup/issues/7478
+
+    const formattedBirthDate = this.birthDate
```
pabo99, line 172: Would it be possible to handle everything related to date formatting within the `frontend/www/js/omegaup/time.ts` file?

In fact, as you can see, the `DatePicker` component calls the `parseDateLocal` function every time the input value containing the date changes:
https://github.com/omegaup/omegaup/blob/ce308208bf8b0d3bb5d988e8218c5b8c7a43aa05/frontend/www/js/omegaup/components/DatePicker.vue#L68-L75

And this is the function responsible for parsing:
https://github.com/omegaup/omegaup/blob/ce308208bf8b0d3bb5d988e8218c5b8c7a43aa05/frontend/www/js/omegaup/time.ts#L99-L117

This function might also be important since it formats the date received in the `DatePicker` component:
https://github.com/omegaup/omegaup/blob/ce308208bf8b0d3bb5d988e8218c5b8c7a43aa05/frontend/www/js/omegaup/time.ts#L86-L97

pabo99 (commented):
Excellent job! I left you a comment.

In addition to that, would it be possible for you to add a screencast showing how the date selection currently fails? And another screencast showing how it works after the fix?

This will be very helpful since, in many local environments, including mine, it hasn't been possible to reproduce the issue.

pushkardev123:
## Case Study: Timezone Offset Issue in Date of Birth Handling

### Overview
The issue arises when a user in Indian Standard Time (IST, UTC+5:30) or any other comparable timezone selects a date for their Date of Birth (DOB). Due to timezone conversion logic in our application, the selected date is shifted by one day when it is converted to UTC before being sent to the backend. This discrepancy can cause the stored or displayed date to be incorrect.

### Detailed Explanation

1. **User Input in IST:**
   - When a user selects a date (for example, "2004-07-24") using our date picker, it is interpreted as "2004-07-24 00:00:00" in IST.
   - In IST, midnight of "2004-07-24" actually corresponds to "2004-07-23 18:30:00" in UTC because IST is 5 hours and 30 minutes ahead of UTC.

2. **Conversion to UTC:**
   - Our current conversion logic (using functions like `strtotime()` or equivalent date parsing routines) converts the local IST date to its UTC equivalent.
   - As a result, "2004-07-24 00:00:00 IST" becomes "2004-07-23 18:30:00 UTC".
   - If this conversion is not properly accounted for when storing or displaying dates, the system will interpret and show the date as "2004-07-23" instead of the intended "2004-07-24".

3. **Environment Variability:**
   - The issue might not be reproducible in every environment because it depends on the server’s timezone settings and how the date conversion is handled in the frontend.
   - Reviewers in environments with different local time settings (or with servers configured directly to UTC) may not observe this issue, which makes it particularly problematic for users in IST or any other comparable timezone.
  
- **Backend Logging:**  
  Server logs (using `error_log`) confirmed that the date was being stored as a UTC timestamp that corresponds to the day before the date chosen in IST.


https://github.com/user-attachments/assets/32657c2a-49bb-4465-9f6b-4d865a84562b

frontend/www/js/omegaup/components/user/BasicInformationEdit.vue
pushkardev123: Actually the problem might be with datePicker.vue which assumes the UTC, as the comments mention, it assumes UTC irrespective of the local timezone, as IST is UTC+5:30 i was able to recognise this issue. As i enter assumingly 12th December 2004, its actually 11th December 2024 17:30 Hours in UTC as Indian Standard Time is UTC+5:30.

pabo99 (approved):
Great! 

If this change fixes the bug, we can approve the PR.

heduenas (approved):

Issue #8067, opened 2025-03-20
[BUG]Incorrect messages in problem solution


SammanSarkar:
## Current Behavior

in the see Solution section for any problem, if the user is not logged in it should show message "You must login to unlock/see this solution." instead it shows "You have 0 available views left, out of a total of 5 that you can make per day."

## Expected Behavior

 If the user is not logged in it should show message "You must login to unlock/see this solution."

*Provide a link to an example, or a detailed list of steps to reproduce this bug.*  
  
	1. Open any problem when logged out
	2. Click on "See Solution"
	3. it says "You have 0 available views left, out of a total of 5 that you can make per day."

*Relevant details about the environment in which you encountered this bug:*

	* Browser name (e.g., Chrome 39, node.js 5.4):All Browsers
	* Operating System and version (desktop or mobile):All OS
	* Link to the site where this issue occurs:https://omegaup.com/arena/problem/Los-Romanos-de-Roman/#solution
	* Screenshot:
![Image](https://github.com/user-attachments/assets/b1bdc79d-2945-442b-a607-510dafe8a29c)

SammanSarkar:
I am working on this @pabo99 please assign it to me

pabo99:
Done!

PR #8077, opened 2025-03-21
Fixed the issue of wrong message in solution page when not logged in

SammanSarkar:
# Description

Previously, the solution tab did not show intended message when not logged in and instead showed "You have 0 available views left, out of a total of 5 that you can make per day." Now, it correctly displays "You must login to unlock/see this solution"
image

The solution tab showed no message when not logged in because status was null. This PR fixes it by adding computedSolutionStatus in Details.vue, ensuring it returns 'not_logged_in' when user.loggedIn is false.

Fixes: #8067

# Checklist:


frontend/www/js/omegaup/components/problem/Details.vue
```diff
@@ -490,7 +490,10 @@ export default class ProblemDetails extends Vue {
   @Prop({ default: true }) useNewVerdictTable!: boolean;
 
   @Ref('statement-markdown') readonly statementMarkdown!: omegaup_Markdown;
-
+  get computedSolutionStatus(): string {
+    if (!this.user.loggedIn) return 'not_logged_in';
+    return this.solutionStatus || 'locked';
+  }
```
pabo99, line 496: According to the previous comment, the ideal approach is to always receive the solution status from the payload. So this computed property could be removed.
```suggestion
```

frontend/www/js/omegaup/components/problem/Solution.vue
```diff
@@ -18,7 +18,7 @@
     <div v-else class="interstitial">
       <omegaup-markdown :markdown="statusMessage"></omegaup-markdown>
       <omegaup-markdown
-        v-show="allowedSolutionsToSee !== null"
+        v-show="allowedSolutionsToSee !== null && status !== 'not_logged_in'"
```
pabo99, line 21: In this case, could you add a computed property that validates both conditions?

frontend/www/js/omegaup/problem/details.ts
```diff
@@ -68,7 +68,7 @@ OmegaUp.on('ready', async () => {
     data: () => ({
       popupDisplayed,
       runDetailsData: runDetails,
-      solutionStatus: payload.solutionStatus,
+      solutionStatus: payload.solutionStatus || 'not_logged_in',
```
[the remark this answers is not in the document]
SammanSarkar: Hi @pabo99, thanks for the feedback! I initially removed `solutionStatus` from `data` and passed it as a prop per your suggestion, but I ran into an issue in the `'unlock-solution'` handler (line 303). When the user unlocks a solution, we need to update `solutionStatus` to `'unlocked'` reactively so the UI reflects the change. Since props are read-only, it needs to be in `data` to be mutable.

I agree the initial value should come from the backend without a frontend default (`|| 'not_logged_in'`). However, I noticed `payload.solutionStatus` is `null` when not logged in, which broke the original behavior my PR fixed (showing a login prompt). To address this without a computed property, I kept it in `data` for reactivity and handled the `null` case in `Details.vue`’s template as `:status="user.loggedIn && solutionStatus ? solutionStatus : 'not_logged_in'`.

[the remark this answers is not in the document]
pabo99: > Hi @pabo99, thanks for the feedback! I initially removed `solutionStatus` from `data` and passed it as a prop per your suggestion, but I ran into an issue in the `'unlock-solution'` handler (line 303). When the user unlocks a solution, we need to update `solutionStatus` to `'unlocked'` reactively so the UI reflects the change. Since props are read-only, it needs to be in `data` to be mutable.

You're right. I forgot that the status could change when the user unlocks a solution. So, this change should be reverted. Thanks for clarifying!

> I agree the initial value should come from the backend without a frontend default (`|| 'not_logged_in'`). However, I noticed `payload.solutionStatus` is `null` when not logged in, which broke the original behavior my PR fixed (showing a login prompt). To address this without a computed property, I kept it in `data` for reactivity and handled the `null` case in `Details.vue`’s template as `:status="user.loggedIn && solutionStatus ? solutionStatus : 'not_logged_in'`.

Exactly! Now you need to fix the behavior in the controller. solutionStatus shouldn't be null in any case. You need to find the right place to set this property and send it via the payload.

SammanSarkar:
@pabo99 adding solution status to initial payload in Problem.php solved the issue. I also added the computed property in Solution.vue that validates both conditions

frontend/server/src/Controllers/Problem.php
```diff
@@ -4670,7 +4670,8 @@ public static function getProblemDetailsForTypeScript(
                         'admin' => false,
                         'reviewer' => false,
                     ],
-                    'allowedSolutionsToSee' => $allowedSolutionsToSee
+                    'allowedSolutionsToSee' => $allowedSolutionsToSee,
+                    'solutionStatus' => self::SOLUTION_NOT_LOGGED_IN,
```
pabo99, line 4674: Now that you are sending the `solutionStatus` parameter in all scenarios, it might be a good idea to update the type declaration for `solutionStatus` in `ProblemDetailsPayload` (Line 44).

Since its presence is now mandatory rather than optional, you can simply change `solutionStatus?: string` to `solutionStatus: string`.

Once you've made this change, you can verify that PSALM does not return any errors by running the following command inside the `Docker` container:
```bash
find frontend/ \
    -name *.php \
    -and -not -wholename 'frontend/server/libs/third_party/*' \
    -and -not -wholename 'frontend/tests/badges/*' \
    -and -not -wholename 'frontend/tests/controllers/*' \
    -and -not -wholename 'frontend/tests/runfiles/*' \
    -and -not -wholename 'frontend/www/preguntas/*' \
  | xargs ./vendor/bin/psalm \
    --long-progress \
    --show-info=false
```
If there are no errors, everything is good to go.

The last step is to run the following command:

```bash
./stuff/lint.sh
```
This will update the `ProblemDetailsPayload` declaration in `TypeScript` within the `api_types.ts` file.

pabo99 (commented):
The change looks good!

I left you one last comment!

pabo99:
Almost done.

You forgot to run the command:
```bash
 ./stuff/lint.sh fix -- frontend/www/js/omegaup/api_types.ts
 ```

This should help the remaining check passes.

pabo99 (approved):
Thank you @SammanSarkar !

SammanSarkar:
Thank you @pabo99! Appreciate your help and feedback.

pabo99:
This PR is almost ready to be merged into the main branch. It just needs to be updated

PR #8095, opened 2025-03-25
Improving the way to retrieve URL parameters after signing in

SammanSarkar:
# Description

After logging in, the page redirects to the previous page.
Fixed Incorrect handling of the redirect function in signin.ts. Updated how the function processes the URL path to ensure correct redirection.

# Checklist:


SammanSarkar:
@pabo99 @heduenas kindly review this and let me know your thoughts. Thanks!

SammanSarkar:
Updated URL redirect: Login now correctly redirects to the contest. Also works similarly when the user is taken to the login page while trying to submit a problem solution.
ScreenRecording2025-03-26002241-ezgif com-video-to-gif-converter

pabo99 (commented):
Thank you for helping with the update. Now, this code is much more readable.

Although, I’m noticing that this is not the final solution to the issue.

If we don't find the root cause or simply cannot reproduce it in the local environment, we can approve this PR without directly linking it to the reported issue.

Here’s what I’ve noticed so far:

The case I’m reporting has the peculiarity that when the user logs in, the platform redirects them to the `/profile/...` page.

After reviewing the code, this condition should only be triggered for newly created users (`isAccountCreation`).

What’s strange is that this happens every time I log in with my omegaUp account using Gmail. Curiously, when I log in with my personal Gmail account, it redirects me correctly according to the conditions in this function.

Could you try to reproduce this behavior in your local environment?

SammanSarkar:
I am unsure what might be going wrong.  
What could be going wrong is when we login, `isAccountCreation` calls `findByEmail` in `Identities.php`. `findByEmail` has a join condition where `u.main_identity_id = i.identity_id`. This could be causing the problem. I removed this and it worked when I logged in from native omegaup sign-in, but I am not sure how to check it for Google sign-in.  
I am removing the issue number on this. I will take that up in a separate PR. Please feel free to approve this when you have a moment—thank you!

pabo99 (approved):

heduenas (approved):

Issue #8069, opened 2025-03-21
Create the necessary tables for (the a streak of consecutive days solving a problem)


Asdts:
Table should contain following:

challenge_id | challenge_name | difficulty | start_time | end_time | completed_users

Key tasks:

-Design tables to store challenge details.
-Create tables to track user competency levels.
-Design a table for streak tracking.
-Implement database migrations to apply changes.

Master issue: #7949 

for more information can refer the below doc:
https://docs.google.com/document/d/1GsKZGJ9ngBurQzd7JWiHJr6YIRR6lo2hXpKynJNZx-A/edit?usp=sharing

Issue #7949, opened 2025-01-15
[FEATURE] Have a streak of consecutive days solving a problem


ChascaNeperiana:
_As a [trainer], I want [a Duolingo-like streaks] and challenges for students to stay engaged with solving problems_

## Context

It's challenging for teachers to keep the  students motivated. I think implementing a streak of consecutive days solving a at least one problem will keep students engaged. Daily or weekly challenges (i.e. to solve a binary search problem) could add to this, also encouraging students to not keep strikes solving only easy problems

abhishekdubey369:
i would like to work on this assign me

pabo99:
> i would like to work on this assign me

@abhishekdubey369 Thank you for your interest in collaborating with the omegaUp project. This issue could become quite extensive, so the first step is to break it down into simple tasks. That way, we can start working on each one individually. 

Would you be able to draft a document outlining your proposal for reaching a solution?

We currently have a feature on the platform for generating badges. We can start by creating an issue for each badge we want to award, for example:


- 5 problems in a row
- The most active user of the month

Here is an example of how to create new badges:

https://github.com/omegaup/omegaup/pull/2615/

abhishekdubey369:
Sure , I will get back to you with documentation soon.

abhishekdubey369:
sorry for the delay here is the doc for the propsed solution let me know the feedback and Im to excited to begin my journey .
https://docs.google.com/document/d/1GsKZGJ9ngBurQzd7JWiHJr6YIRR6lo2hXpKynJNZx-A/edit?usp=sharing

pabo99:
Sure,

The first thing to do is to create separate issues to split all the functionality:

- Create the necessary tables for the new feature.
- Challenge creation and question selection based on the user's competency level.
- Streak API creation and test case implementation to validate its functionality.
- Frontend integration.
- Final testing of the streak feature.

Once you create the issues, you can start working on the database schema changes.

I want to ask you something: Are you interested in participating in the GSoC program? If so, have you solved the contest (at least two out of three problems)?

Asdts:
Hii @pabo99 ,
Thanks for assigning me . I have created separate issues to split functionality. 
I am planning to take part in GSOC and would love your guidance for the same. I have not appeared contest yet . I will do the same till tomorrow and our database schema will be pushed till monday as well .

Asdts:
Thanks @pabo99

Asdts:
[...]

PR #8099, opened 2025-03-26
Table creation for streak system #8069 

Asdts:
# Description

This PR introduces schema for :
Streak System
Challenges (Daily/Weekly)
Streak Freeze Option

Fixes: #8069 for first step of #7949 as mentioned in the doc : https://docs.google.com/document/d/1GsKZGJ9ngBurQzd7JWiHJr6YIRR6lo2hXpKynJNZx-A/edit?tab=t.0

-  The code follows the [coding guidelines](https://github.com/omegaup/omegaup/wiki/Coding-guidelines) of omegaUp.
- Table generation is tested and it get auto deleted if user is deleted from user table
- Next step is implementing DAO and Controllers

### DIFF
--- a/frontend/database/schema.sql
+++ b/frontend/database/schema.sql
@@ -1262,8 +1262,48 @@ CREATE TABLE `Users_Experiments` (
 /*!40101 SET character_set_client = @saved_cs_client */;
 /*!40103 SET TIME_ZONE=@OLD_TIME_ZONE */;
 
+CREATE TABLE `Users_Streaks` (
+  `user_id` INT NOT NULL COMMENT 'Identificador de usuario',
+  `streak_count` INT NOT NULL DEFAULT 0 COMMENT 'Total de días de racha',
+  `submission_time` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT 'Último envío de problema',
+  `easy_solved` INT NOT NULL DEFAULT 0 COMMENT 'Total fácil problema resuelto',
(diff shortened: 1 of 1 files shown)


frontend/database/schema.sql
```diff
@@ -1262,8 +1262,48 @@ CREATE TABLE `Users_Experiments` (
 /*!40101 SET character_set_client = @saved_cs_client */;
 /*!40103 SET TIME_ZONE=@OLD_TIME_ZONE */;
 
+CREATE TABLE `Users_Streaks` (
```
pabo99, line 1265: These changes should be added in a migration script, just like it was done in the following PR:

https://github.com/omegaup/omegaup/pull/8016

The identifier number of the file should be the next consecutive one after the last existing file in the repository

In fact, the changes in the `schema.sql` file will be regenerated once you try to `git push` your branch. It's important to run the `git push` command from the terminal so that all `pre-push` `hook_tools` are executed.

Asdts: so i have to create VO and DAO first and it will get regenerated once i pushed?

pabo99: Not entirely, the VOs DAOs are generated automatically once the changes in `dao_schema.sql` are applied. But we haven't reached that point yet.

For now, in this PR, you only need to add the migration script and the changes that are auto-generated in `schema.sql`.

By the way, since you're going to apply some fixes, it's a good time to mention that we try to avoid adding changes to the main branch of our repos. So if you can create a new branch, that would be great.

Here you can find the explanation for this comment:
https://github.com/omegaup/omegaup/wiki/How-to-Make-a-Pull-Request-(English)

Issue #8080, opened 2025-03-22
[FEATURE]Add activity heatmap to user profile to visualize problem-solving patterns


SammanSarkar:
As a competitive programmer, I want a contribution heatmap in my profile so that I can visualize my problem-solving activity patterns over time

## Context

When using platforms like GitHub, Codeforces, or LeetCode, I find the activity heatmap to be an extremely useful visualization that helps me track my consistency in solving problems. Currently, omegaUp shows statistics about solved problems, but lacks a calendar-style visualization that would help users:

Identify patterns in their problem-solving habits
Maintain motivation through visual feedback on consistency
Showcase their dedication to competitive programming
Set and track goals for regular practice
The existing Charts tab in the user profile already shows various statistics, but adding a heatmap would provide a more intuitive way to visualize activity over time. This would be particularly valuable for students trying to maintain consistent practice schedules and for educators who want to monitor student engagement patterns.

Proposed Solution
Add a GitHub/Codeforces-style contribution heatmap to the Charts tab in the user profile that:

Shows problem-solving activity over the past year
Uses color intensity to represent the number of problems solved each day
Provides tooltips with exact counts when hovering over days
Respects existing privacy settings (only visible for own profile or non-private profiles)
Uses the existing Highcharts library already implemented in the codebase
This feature would enhance user engagement by gamifying consistency and providing visual feedback on progress patterns.

SammanSarkar:
I’d like to take this up, @pabo99. Let me know if you have any guidance or expectations.

pabo99:
+1 to this.

If you consider this change could be managed as a project since it could grow too much. I recommend to create a   document with your proposal, folowing all the guidelines we have. If not, creating a simple One Pager would be enough.

[Proposal template](https://docs.google.com/document/d/1_FKfpc2M3VLDVYqvT8ZgsgwIJ3zaZnyUVmSm-H3h6UQ/edit?usp=sharing)

One pager should include:
- Introduction
- Design Overview
- Alternatives considered
- Estimated costs
- Status

You can use the same format of Proposl template in the One Pager document

SammanSarkar:
@pabo99 Here's the one pager doc for the proposed feature: https://docs.google.com/document/d/1DvEgMQRBRBCITIvEpKBhFga3r5wdd_jmYC0vhDYUbKU/edit?usp=sharing

pabo99:
Excellent! Feel free to create issues if needed. This way, revisions can be  easier and faster

PR #8110, opened 2025-03-28
Heatmap feature 

SammanSarkar:
# Description

Created a UserHeatmap component that visualizes user submission activity with a color-coded yearly calendar. Integrated it with the profile system, added translations, and implemented filtering by year.
Positioning in User profile:
image

on hovering over a cell it shows submissions on dates that have more than 1:
image

color scheme:
#dddddd (light grey) for no submissions.

#739DE3 (light blue) for values between 1 and 4.

#5588DD (medium blue) for values between 5 and 9.

#4670B5 (dark blue) for values between 10 and 1000.
image

The dropdown shows options for year in descending order from current year to year of account creation

The "Total submissions" text is currently hardcoded into UserHeatmap.vue rather than begin retreived from translation files.
Fixes: #8080


frontend/www/js/omegaup/components/user/UserHeatmap.vue
```diff
@@ -0,0 +1,337 @@
...
+  created?: string;
+}
+
+interface UserStatsResponse {
+  heatmap?: HeatmapDataPoint[];
+}
+
+@Component({
+  components: {
+    highcharts: Chart,
```
pabo99, line 46: You're including the highcharts-vue component here, but you're not using it. Could you try using it to display the chart? If it works, you could remove the div where it's currently being rendered.

Here's the documentation: https://github.com/highcharts/highcharts-vue

pabo99 (commented):
Thank you for resolving this issue so quickly!

Unfortunately, this creates a bit of a problem for the review process due to the large number of lines that need to be reviewed.

We have a standard of ~200 lines of code per PR, so if possible, it would be great to split it up.

In this case, it might help to break it down into three parts:

- **Server changes:** This could include adding a PHPUnit test.
- **Changes in the new component**
- **Changes in the `.ts` file:** which in this case would be in frontend/www/js/omegaup/user/profile.ts. These changes could include (based on the suggestions I made):
  - Sending all the new data to the component, obtained from the `payload`
  - Adding actions to retrieve information by year for the user's submissions
  - If possible, moving the `Highcharts.options`
   
If you have any questions about what I mentioned, I’d be happy to help you resolve them.

PR #8117, opened 2025-03-29
Updated DatePicker for Editing Birthday in User Profile

vaidya45:
# Description

This PR updates the DatePicker used to edit the birthday information within the user profile section.

**Updated UI:** Shown within red box

update_ui

**Demo Video** 

https://github.com/user-attachments/assets/5798ef0a-0924-4cb3-9eb2-30d3c20848bf

Fixes: #4017 

# Comments

1. To ensure that the birthday is easily formatted, I provided the user with the option to either edit it directly in the text box or select it using the calendar on the right.
2. The original implementation is buggy when the text is edited (the year defaults to 1902).
3. I ensured that the user cannot enter incorrectly formatted strings.

# Checklist:

### DIFF
--- a/frontend/www/js/omegaup/components/UpdatedDatePicker.vue
+++ b/frontend/www/js/omegaup/components/UpdatedDatePicker.vue
@@ -1,0 +1,90 @@
+<template>
+  <b-input-group size="16">
+    <!-- Editable Text Input -->
+    <b-form-input
+      v-model="stringValue"
+      :name="name"
+      :disabled="!enabled"
+      placeholder="YYYY-MM-DD"
+      :class="{ 'is-invalid': isInvalid }"
+      required
+      @change="onInputChange"
+    />
+    
+    <!-- Calendar Input -->
+    <b-form-datepicker
+      v-model="stringValue"
+      :min="minDateStr"
+      :max="maxDateStr"
+      button-only
+      right
+      :disabled="!enabled"
+      @input="onDateSelected"
+    />
+  </b-input-group>
+</template>
+
+<script lang="ts">
+import { Vue, Component, Watch, Prop } from 'vue-property-decorator';
+import T from '../lang';
+import * as time from '../time';
+import { BFormDatepicker, BFormInput, BInputGroup } from 'bootstrap-vue';
+
+@Component({
+  components: { BFormDatepicker, BFormInput, BInputGroup },
+})
+export default class UpdatedDatePicker extends Vue {
+  T = T;
+
+  @Prop({ default: '' }) name!: string;
+  @Prop() value!: Date;
+  @Prop({ default: true }) enabled!: boolean;
+  @Prop({ default: T.datePickerFormat }) format!: string;
+  @Prop({ default: false }) isInvalid!: boolean;
+  @Prop({ default: null }) min!: Date | null;
+  @Prop({ default: null }) max!: Date | null;
+
+  private stringValue: string = '';
+
+  get minDateStr() {
+    return this.min ? time.formatDateLocal(this.min) : '';
+  }
+
+  get maxDateStr() {
+    return this.max ? time.formatDateLocal(this.max) : '';
+  }
+
+  mounted() {
+    this.updateStringValue();
+  }
+
+  @Watch('value', { immediate: true })
+  onValueChanged(newValue: Date) {
+    if (newValue && time.formatDateLocal(newValue) !== this.stringValue) {
+      this.stringValue = time.formatDateLocal(newValue);
+    }
+  }
+
+  updateStringValue() {
+    if (this.value && this.value instanceof Date && !isNaN(this.value.getTime())) {
+      this.stringValue = time.formatDateLocal(this.value);
+    } else {
+      this.stringValue = '';
+    }
+  }
+
+  onInputChange() {
+    const parsedDate = time.parseDateLocal(this.stringValue);
+    if (parsedDate) {
+      this.$emit('input', parsedDate);
+    }
+  }
+
+  onDateSelected(newDate: Date) {
+    if (newDate) {
+      this.$emit('input', newDate);
+      this.stringValue = time.formatDateLocal(newDate);
+    }
+  }
+}
+</script>
--- a/frontend/www/js/omegaup/components/user/BasicInformationEdit.vue
+++ b/frontend/www/js/omegaup/components/user/BasicInformationEdit.vue
@@ -88,11 +88,11 @@ import { dao, types } from '../../api_types';
 import T from '../../lang';
 import * as time from '../../time';
 import * as iso3166 from '@/third_party/js/iso-3166-2.js/iso3166.min.js';
-import DatePicker from '../DatePicker.vue';
+import UpdatedDatePicker from '../UpdatedDatePicker.vue';
 
 @Component({
   components: {
-    'omegaup-datepicker': DatePicker,
+    'omegaup-datepicker': UpdatedDatePicker,
   },
 })
 export default class UserBasicInformationEdit extends Vue {


frontend/www/js/omegaup/components/UpdatedDatePicker.vue
```diff
@@ -0,0 +1,90 @@
+<template>
+  <b-input-group size="16">
```
pabo99, line 2: It looks like that size is not being applied because the styles from `b-input-group` are being used, which cause the content to expand across the entire container div. So it can be removed:
```suggestion
  <b-input-group >
```

frontend/www/js/omegaup/components/UpdatedDatePicker.vue
```diff
@@ -0,0 +1,90 @@
...
+      :name="name"
+      :disabled="!enabled"
+      placeholder="YYYY-MM-DD"
+      :class="{ 'is-invalid': isInvalid }"
+      required
+      @change="onInputChange"
+    />
+    
+    <!-- Calendar Input -->
+    <b-form-datepicker
```
pabo99, line 15: To group these two elements in the UI(the input and the calendar button), `BootstrapVue` provides a wrapper called [`b-input-group-append`](https://bootstrap-vue.org/docs/components/input-group/#using-sub-components).

So if you wrap this element with the wrapper, you will see the result.

frontend/www/js/omegaup/components/UpdatedDatePicker.vue
```diff
@@ -0,0 +1,90 @@
...
+      v-model="stringValue"
+      :min="minDateStr"
+      :max="maxDateStr"
+      button-only
+      right
```
pabo99, line 20: Here in this component, the texts need to be configured according to the user's preferred language.

According to the [documentación](https://bootstrap-vue.org/docs/components/form-datepicker#comp-ref-b-form-datepicker-props), you can pass the `locale` property here to update those texts correctly.

You can retrieve this information from the `profile.locale` object, which is sent to the parent component—the same place where the user's birthdate is sent.

frontend/www/js/omegaup/components/UpdatedDatePicker.vue
```diff
@@ -0,0 +1,90 @@
...
+      :disabled="!enabled"
```
pabo99, line 21: I also noticed that the component includes a `help` label for using the controls with the keyboard, and it has a property to change the text called `label-help`.

To keep this change simple, you can set that property to an empty string so that it doesn't appear in the datepicker.

frontend/www/js/omegaup/components/UpdatedDatePicker.vue
```diff
@@ -0,0 +1,90 @@
...
+
+  @Prop({ default: '' }) name!: string;
+  @Prop() value!: Date;
+  @Prop({ default: true }) enabled!: boolean;
+  @Prop({ default: T.datePickerFormat }) format!: string;
+  @Prop({ default: false }) isInvalid!: boolean;
+  @Prop({ default: null }) min!: Date | null;
+  @Prop({ default: null }) max!: Date | null;
+
+  private stringValue: string = '';
```
pabo99, line 47: To avoid using magic values, can it be declared like this?
```suggestion
  private stringValue: string | null = null;
```

frontend/www/js/omegaup/components/UpdatedDatePicker.vue
```diff
@@ -0,0 +1,90 @@
...
+
+  get minDateStr() {
+    return this.min ? time.formatDateLocal(this.min) : '';
```
pabo99, line 50: Here it would be::
```suggestion
    return this.min ? time.formatDateLocal(this.min) : null;
```

frontend/www/js/omegaup/components/UpdatedDatePicker.vue
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  get maxDateStr() {
+    return this.max ? time.formatDateLocal(this.max) : '';
```
pabo99, line 54: same case:
```suggestion
    return this.max ? time.formatDateLocal(this.max) : null;
```

In these two computed properties, it would be ideal to specify their return type, which in this case would change to `string | null`

frontend/www/js/omegaup/components/UpdatedDatePicker.vue
```diff
@@ -0,0 +1,90 @@
...
+    }
+  }
+
+  updateStringValue() {
+    if (this.value && this.value instanceof Date && !isNaN(this.value.getTime())) {
+      this.stringValue = time.formatDateLocal(this.value);
+    } else {
+      this.stringValue = '';
+    }
+  }
```
pabo99, line 74: I see that this function is only used to format the user's birthdate. Is there a way to get this value already formatted directly from the `.ts` file that controls it? In this case, it would be `profile.ts`.

This way, you won't need this function, nor the `mounted` one. Both can be removed.

frontend/www/js/omegaup/components/UpdatedDatePicker.vue
```diff
@@ -0,0 +1,90 @@
...
+
+  onInputChange() {
+    const parsedDate = time.parseDateLocal(this.stringValue);
```
pabo99, line 77: With the proposed changes, this function could fail, because now `stringValue` accepts null values.

To handle this, you can use the guard clause pattern right above:
```ts
    if (!this.stringValue) {
      return;
    }
```
This ensures that by the time it reaches this line, the data contains valid information.

frontend/www/js/omegaup/components/UpdatedDatePicker.vue
```diff
@@ -0,0 +1,90 @@
...
+    if (parsedDate) {
+      this.$emit('input', parsedDate);
+    }
+  }
+
+  onDateSelected(newDate: Date) {
+    if (newDate) {
+      this.$emit('input', newDate);
+      this.stringValue = time.formatDateLocal(newDate);
```
pabo99, line 86: Here, the action needs to be emitted with the new formatted date, so it could be simplified to:
```suggestion
      this.$emit('input', time.parseDateLocal(newDate));
```

pabo99 (commented):
This still doesn't fully address the problem reported in the original issue (a datepicker that makes it easier to select the user's birthdate), but it's a good start. 

With the comments I suggest, some bugs and other visual aspects of the component will be fixed.

frontend/www/js/omegaup/components/UpdatedDatePicker.vue
vaidya45: Wrapped this element with b-input-group-append

frontend/www/js/omegaup/components/UpdatedDatePicker.vue
vaidya45: Removed helper label

frontend/www/js/omegaup/components/UpdatedDatePicker.vue
vaidya45: If the user does not already have a stored birthdate, my goal is to display a placeholder text (which happens when the string is empty, not null). This is the easiest way I could implement this

frontend/www/js/omegaup/components/UpdatedDatePicker.vue
vaidya45: Added the guard clause

frontend/www/js/omegaup/components/UpdatedDatePicker.vue
vaidya45: Updated input type to string, and updated it to the proposed solution

frontend/www/js/omegaup/components/UpdatedDatePicker.vue
vaidya45: Updated to null

frontend/www/js/omegaup/components/UpdatedDatePicker.vue
vaidya45: Updated function to return null

frontend/www/js/omegaup/components/UpdatedDatePicker.vue
vaidya45: Updated function to return null

PR #8142, opened 2025-04-03
Clarification rank

SammanSarkar:
# Description

Added author_classname field to clarification API and frontend components. Modified SQL queries to retrieve user rank data, updated API types, and enhanced Username component integration to display properly colored usernames in clarifications.
image

Fixes: # #8104


SammanSarkar:
@pabo99 @heduenas Let me know if you'd like any changes.

frontend/www/js/omegaup/components/arena/Clarification.vue
```diff
@@ -113,15 +121,21 @@ import T from '../../lang';
...
+  },
+})
 export default class ArenaClarification extends Vue {
   @Prop() clarification!: types.Clarification;
   @Prop({ default: false }) isAdmin!: boolean;
   @Prop({ default: false }) selected!: boolean;
 
   T = T;
   time = time;
+  ui = ui;
```
pabo99, line 138: What is this being used for? If it's not being used, can it be removed?

frontend/www/js/omegaup/components/arena/Clarification.vue
```diff
@@ -180,12 +208,19 @@ export default class ArenaClarification extends Vue {
...
       author: this.clarification.author,
+      author_classname: this.clarification.author_classname,
       answer: this.responseText,
       public: this.isPublic,
       message: this.message,
       problem_alias: this.clarification.problem_alias,
       time: new Date(),
     };
+
+    if (this.clarification.receiver && this.clarification.receiver_classname) {
```
pabo99, line 219: In the future, we’ll need to create an interface that contains:
```ts
{
  username: string,
  classname: string,
}
```
because it’s becoming increasingly difficult to validate the existence of one or the other, when both should inherently exist.

frontend/www/js/omegaup/components/arena/Clarification.vue
```diff
@@ -150,6 +164,20 @@ export default class ArenaClarification extends Vue {
...
+      author: { username: string; classname: string };
+      receiver: { username: string; classname: string };
+    },
+  ): string {
+    const authorSpan = `<span class="${params.author.classname} font-weight-bold">${params.author.username}</span>`;
+    const receiverSpan = `<span class="${params.receiver.classname} font-weight-bold">${params.receiver.username}</span>`;
+    return template
+      .replace('%(author)', authorSpan)
+      .replace('%(receiver)', receiverSpan);
+  }
```
pabo99, line 179: Is there a way to add these elements directly in the template? Creating them programmatically makes it much harder to understand how they work

pabo99 (commented):
Oh, this change was bigger than I expected.

That explains why it was the last change related to displaying the ranking colors in the usernames.

Well, for now, the change looks good. I added some comments that will help make the code a bit more readable.

In addition to those, there are a few other cases I noticed with details that can be improved:

image

- In "Question 7", `test_user_4` should be shown in yellow. This clarification was shown via the broadcaster (meaning it was shown right when the message was sent; it gets fixed after reloading the page).
- In "Question 1", `test_user_1` should be shown in black because they are not ranked. It seems to be taking the same color used for resolved clarifications text.
- In "Answer 1" y "Answer 2", the ranking color isn’t being displayed for any of the users. To reproduce those clarifications, you need to create a new clarification from the contest admin’s account.
- In "Question 4", the color for `test_user_5` is actually correct. The issue is that it uses the same color as the resolved clarification text. Ideally, all text should be in black so the rank color stands out.

PR #8170, opened 2025-04-07
Feature/contest recommendation 4802

zikunz:
# Description

Added functionality to the admin support dashboard allowing support team members to mark contests as recommended. This feature streamlines the administration workflow by providing a single interface where contests can be searched and their recommendation status can be toggled directly. The implementation includes proper UI feedback, API integration, and full translation support.

Fixes: #4802

# Comments

The implementation required changes to several files, each serving a specific purpose:

1. **`frontend/www/js/omegaup/admin/support.ts`**:
   - Added new state variables (`contestAlias`, `contestTitle`, `contestFound`, `isContestRecommended`)
   - Implemented event handlers for contest search, recommendation toggle, and state reset
   - Integrated with Contest API endpoints to fetch details and update recommendation status
   - Added proper error handling and success notifications

2. **`frontend/www/js/omegaup/components/admin/Support.vue`**:
   - Added contest search input field in the same pattern as existing user search
   - Implemented "Options" section with checkbox for toggling recommendation status
   - Added display of contest title and alias in the dashboard header
   - Used v-model for reactive form binding and component properties for state management

3. **`frontend/www/js/omegaup/api_types.ts`**:
   - Added `recommended?: boolean` property to the ContestDetails interface
   - Used optional property to maintain backward compatibility

4. **Language files (`lang.en.ts`, `lang.es.ts`, `lang.pt.ts`)**:
   - Added translations for all UI elements in all three supported languages
   - Used consistent naming scheme for translation keys
   - Ensured language-appropriate phrasing for each string

The toggle implementation required special attention to ensure the success messages matched the user's action rather than the current state of the checkbox. This was addressed by having the component emit the current value after the v-model has updated it.

# Screenshots

## Pre-req: adding a contest (already implemented)
Screen Shot 2025-04-08 at 6 23 04 AM

## Searching for a non-existent contest
Screen Shot 2025-04-08 at 6 25 53 AM

## Searching for an existing contest
Screen Shot 2025-04-08 at 7 44 51 AM

## Toggling recommendation status (ON)
Screen Shot 2025-04-08 at 7 45 36 AM

## Toggling recommendation status (OFF)
Screen Shot 2025-04-08 at 7 45 45 AM

# Checklist:

- [] The tests were executed and all of them passed.
- [] If you are creating a feature, the new tests were added.

Note and Questions:
1. If I run `yarn test` in the main branch (where no code changes are made), some tests fail. Hence, I did not tick the box for "The tests were executed and all of them passed." above.
2. Do I need to write tests for this new feature in addition to testing it manually?
3. Can I modify `lang.en.ts`, `lang.es.ts`, `lang.pt.ts` or they are changed somewhere else and automatically generated from there?


frontend/www/js/omegaup/lang.en.ts
```diff
@@ -2089,6 +2089,11 @@ const translations: { [key: string]: string; } = {
   wordsWarningProblem: "Keep with a warning",
   wordsYes: "Yes",
   zipPrepare: "Prepare zip file",
+  supportTypeContestAlias: "Type contest alias",
+  supportOptions: "Options",
+  supportSetAsRecommended: "Set as recommended",
+  supportContestSetAsRecommended: "Contest has been set as recommended",
+  supportContestRemovedFromRecommended: "Contest has been removed from recommended",
```
pabo99, line 2096: The wording seems fine. Just one comment — these should be added in the following files:
- `frontend/templates/en.lang`
- `frontend/templates/es.lang`
- `frontend/templates/pt.lang`

Once you’ve added the translations in those files, you need to run the following command:
```bash
./stuff/lint.sh
``` 
This will take care of propagating the changes to the files used in TypeScript (including the ones you modified).

Let me know if the instructions are clear.

pabo99 (commented):
Good first contribution, congratulations!

I left you a comment; once you apply the changes, I’ll start making more suggestions if needed.

frontend/www/js/omegaup/lang.en.ts
zikunz: Thank you so much for your code review! I completely understand the instructions given. Let me incorporate the changes now. :)

pabo99:
Ok, it looks like the second part is still missing, which is running the  `./stuff/lint.sh` command.

Most likely you did run it, but maybe not in the correct place. Let me explain:

For Linux users, it's enough to just run it in the terminal since Docker uses Linux kernel features. But for Windows, the project must be set up inside a WSL2 instance. This [video](https://youtu.be/H1PG4Dvje88?si=mhaCwWORMBFh6EZi) explains how to configure your computer to work with omegaUp on Windows.

So when you run the command, it has to be from within that WSL2 instance.

If something needs fixing, the output will look like this:
```bash
juan@juan-HP-Laptop-xx-xxxxx:~/Desarrollo/omegaup$ ./stuff/lint.sh 
Fixing frontend/server/src/Controllers/Contest.php (trailing whitespace)
Fixing frontend/server/src/Controllers/Contest.php (php)
Fixing frontend/www/js/omegaup/vue-html.d.ts (typescript)
Fixing frontend/www/js/omegaup/components/contest/NewForm.vue (vue)
Files written to working directory. Please commit them before pushing.
```
When everything is good, the output will look like this:
```bash
juan@juan-HP-Laptop-xx-xxxxx:~/Desarrollo/omegaup$ ./stuff/lint.sh 
OK
```

zikunz:
> Ok, it looks like the second part is still missing, which is running the `./stuff/lint.sh` command.
> 
> Most likely you did run it, but maybe not in the correct place. Let me explain:
> 
> For Linux users, it's enough to just run it in the terminal since Docker uses Linux kernel features. But for Windows, the project must be set up inside a WSL2 instance. This [video](https://youtu.be/H1PG4Dvje88?si=mhaCwWORMBFh6EZi) explains how to configure your computer to work with omegaUp on Windows.
> 
> So when you run the command, it has to be from within that WSL2 instance.
> 
> If something needs fixing, the output will look like this:
> 
> ```shell
> juan@juan-HP-Laptop-xx-xxxxx:~/Desarrollo/omegaup$ ./stuff/lint.sh 
> Fixing frontend/server/src/Controllers/Contest.php (trailing whitespace)
> Fixing frontend/server/src/Controllers/Contest.php (php)
> Fixing frontend/www/js/omegaup/vue-html.d.ts (typescript)
> Fixing frontend/www/js/omegaup/components/contest/NewForm.vue (vue)
> Files written to working directory. Please commit them before pushing.
> ```
> 
> When everything is good, the output will look like this:
> 
> ```shell
> juan@juan-HP-Laptop-xx-xxxxx:~/Desarrollo/omegaup$ ./stuff/lint.sh 
> OK
> ```

Thank you so much for your message (I am using both MacOS and Windows machines), the resources you provided are really helpful! I stopped half way because I ran into some issues after executing 
```shell
./stuff/lint.sh
```
, I just had more time to take a look at it. It still has one issue, I am not sure why it gives me this error
Screen Shot 2025-04-08 at 8 37 03 PM
is it because what I added to `frontend/templates/en.lang`, `frontend/templates/pt.lang` and `frontend/templates/es.lang` has wrong format?

```shell
supportContestRemovedFromRecommended = "Contest has been removed from recommended" // frontend/templates/en.lang
```

```shell
supportContestRemovedFromRecommended = "O concurso foi removido dos recomendados" // frontend/templates/pt.lang
```

```shell
supportContestRemovedFromRecommended = "El concurso ha sido eliminado de recomendados" // frontend/templates/es.lang
```

Thank you!

pabo99:
Remember to run the command `./stuff/lint.sh` after making any changes to frontend files (including translation files)

zikunz:
> Remember to run the command `./stuff/lint.sh` after making any changes to frontend files (including translation files)

@pabo99  I did it but saw missing translation string which I already added to `frontend/templates/en.lang`, `frontend/templates/pt.lang` and `frontend/templates/es.lang`. Do you know what causes this issue?

Is it because what I added to frontend/templates/en.lang, frontend/templates/pt.lang and frontend/templates/es.lang has wrong format?

Screen Shot 2025-04-08 at 9 28 00 PM

zikunz:
@pabo99 Thank you so much! Meanwhile, let me also investigate the issue!

pabo99:
I think I found the error. It seems to be caused by a bug in the `lint.sh` script:

When you add a new entry to the translation files, and you place it in the last line of this file, you must add a blank line below the new entry. Otherwise, if the blank line is not added, for some reason, the last line with information gets deleted.

frontend/server/src/Controllers/Contest.php
```diff
@@ -1901,6 +1901,7 @@ function () use ($contest) {
                     'penalty_type',
                     'points_decay_factor',
                     'problemset_id',
+                    'recommended',
```
pabo99, line 1904: Great! This line works so you can get the value of the `recommended` field from the contest.

But now TypeScript needs to know that this field is being received on the client side.

What you need to do is add this new element to the type declaration of this variable (line 1887).

You also need to add it to the declaration of `ContestDetails`, which is what this function returns.

Once you’ve made these changes, you should run:
```bash
./stuff/lint.sh
```
So the changes propagate to TypeScript.

frontend/www/js/omegaup/admin/support.ts
```diff
@@ -118,6 +126,40 @@ OmegaUp.on('ready', () => {
...
+            adminSupport.contestFound = false;
+            adminSupport.contestTitle = null;
+            adminSupport.isContestRecommended = false;
+
+            api.Contest.details({ contest_alias: contestAlias })
+              .then((data) => {
+                adminSupport.contestAlias = contestAlias;
+                adminSupport.contestTitle = data.title;
+                adminSupport.contestFound = true;
+                adminSupport.isContestRecommended = !!data.recommended;
```
pabo99, line 139: Here, `data.recommended` is already being sent as a boolean, so there's no need to convert it
```suggestion
                adminSupport.isContestRecommended = data.recommended;
```

frontend/www/js/omegaup/components/admin/Support.vue
```diff
@@ -4,6 +4,9 @@
       <div class="card-title h4">
         {{ T.omegaupTitleSupportDashboard }}
         <span v-if="username != null">- {{ username }} ({{ email }})</span>
+        <span v-if="contestAlias != null && contestTitle != null"
```
pabo99, line 7: Here you only need to check if `contestAlias` is not null. The rest can be removed. Even better, you can ensure that only one of the two texts (user or contest) is shown, so this can be changed to a `v-else-if`
```suggestion
        <span v-else-if="contestAlias != null"
```

frontend/www/js/omegaup/components/admin/Support.vue
```diff
@@ -220,6 +277,11 @@ export default class AdminSupport extends Vue {
   @Prop() roles!: string[];
   @Prop() roleNamesWithDescription!: types.UserRole[];
 
+  @Prop() contestAlias!: string;
```
pabo99, line 280: This value, just like `isContestRecommended`, may change from its initial value at some point in the component's lifecycle.

Since props in VueJS are immutable, this would generate a series of warnings in the client's console. To avoid these warnings, you can create 
```ts
currentContestAlias = this.contestAlias;
```
 right after the props declaration. The same should be done for `isContestRecommended`.

One more thing to consider: in every place where you use these props, you’ll need to replace them with the new variables you created.

frontend/www/js/omegaup/components/admin/Support.vue
```diff
@@ -270,5 +332,21 @@ export default class AdminSupport extends Vue {
       selected: (ev.target as HTMLInputElement).checked,
     };
   }
+
```
pabo99, line 335: I'm noticing that the `recommended` checkbox is not enabled when displaying a contest that already has that setting in the database.

This happens for two main reasons:

- That value is obtained from the parent component (in this case from `support.ts`). So in this file, you need to add a watcher to detect changes coming from that source. The simplest way to do it is as follows:
```ts
  @Watch('isContestRecommended')
  onContestRecommendedChange(newValue: boolean) {
    this.currentRecommended = newValue;
  }
```
- Contest details are being cached. Since the cache is invalidated every time the contest is updated, it's very likely that the `setRecommended` API isn't performing this action. What you need to do is add:
```php
        \OmegaUp\Cache::deleteFromCache(
            \OmegaUp\Cache::CONTEST_INFO,
            $contestAlias
        );
```
in the `apiSetRecommended` function right after updating the contest.

pabo99 (commented):
With the comments I left you, everything should now work correctly. However, there are still a few aesthetic details missing, such as:

- Hiding everything related to contests when searching for a user.

- Hiding everything related to users when searching for a contest.

frontend/www/js/omegaup/components/admin/Support.vue
zikunz: Thank you so much! I have created local data properties `currentContestAlias` and `currentIsContestRecommended` to track these values internally, initialized with the prop values.

I have also updated all the `v-model` bindings and references throughout the component to use these local variables instead of the props. This should prevent the Vue warnings about mutating props directly.

frontend/www/js/omegaup/components/admin/Support.vue
zikunz: Thank you so much for your feedback. This is my oversight. 

I have fixed the two issues mentioned:
1. The checkbox does not show the correct state when first loading a contest that is already recommended in the database (fix: Add a watcher for the `isContestRecommended` property)
2. When toggling the checkbox, the cache might not be invalidated properly, causing stale data to be shown (fix: Add cache invalidation in the `apiSetRecommended` function)

I have verified it locally and now it works correctly:

Set a contest as recommended:
Screen Shot 2025-04-13 at 2 06 33 PM

Refresh the page and the contest is indeed shown as recommended:
Screen Shot 2025-04-13 at 2 07 16 PM

Set the same contest as not recommended:
Screen Shot 2025-04-13 at 2 07 46 PM

Refresh the page and the contest is indeed shown as not recommended:
Screen Shot 2025-04-13 at 2 08 08 PM

frontend/server/src/Controllers/Contest.php
zikunz: Thank you so much! Now I understand what is the complete, proper way to update TypeScript type definitions to match the PHP backend changes. Changes made :)

frontend/www/js/omegaup/components/admin/Support.vue
zikunz: You are right, make me make the changes. Thank you!

zikunz:
Hi @pabo99, I just addressed all the comments you raised during the code review and updated the PR content. However, I do not quite understand what is meant by the following. May I ask whether you could elaborate?

> With the comments I left you, everything should now work correctly. However, there are still a few aesthetic details missing, such as:
> 
> * Hiding everything related to contests when searching for a user.
> * Hiding everything related to users when searching for a contest.

Do you mean there are additional UI changes that should be done for #4802?

Also, I notice that if I change from
```bash
adminSupport.isContestRecommended = !!data.recommended;
``` 

to

```bash
adminSupport.isContestRecommended = data.recommended;
```

I get the following error:
Screen Shot 2025-04-13 at 2 51 05 PM

If I run 
```bash
yarn run webpack --mode=development
```
to rebuild all TypeScript definitions after making changes to api_types.ts, below is what I see:
```bash
yarn run v1.22.22
$ /Users/zhuzikun/Desktop/omegaup/omegaup/node_modules/.bin/webpack --mode=development
Browserslist: caniuse-lite is outdated. Please run:
npx browserslist@latest --update-db

Why you should do it regularly:
https://github.com/browserslist/browserslist#browsers-data-updating
(node:51151) [DEP_WEBPACK_COMPILATION_ASSETS] DeprecationWarning: Compilation.assets will be frozen in future, all modifications are deprecated.
BREAKING CHANGE: No more changes should happen to Compilation.assets after sealing the Compilation.
        Do changes to assets earlier, e. g. in Compilation.hooks.processAssets.
        Make sure to select an appropriate stage from Compilation.PROCESS_ASSETS_STAGE_*.
(Use `node --trace-deprecation ...` to show where the warning was created)
[webpack-cli] Compilation finished
[webpack-cli] Compilation finished
frontend:
  assets by path js/dist/*.js 49.4 MiB
    assets by status 43.2 MiB [big]
      assets by chunk 14.2 MiB (id hint: vendor) 7 assets
      assets by chunk 2.97 MiB (id hint: default)
        asset js/dist/default-vendors-node_modules_monaco-editor_esm_vs_language_css_cssMode_js.js 1.59 MiB [compared for emit] [big] (id hint: default, vendors) 1 related asset
        asset js/dist/default-vendors-node_modules_monaco-editor_esm_vs_language_html_htmlMode_js.js 938 KiB [compared for emit] [big] (id hint: default, vendors) 1 related asset
        asset js/dist/default-vendors-node_modules_monaco-editor_esm_vs_language_json_jsonMode_js.js 472 KiB [compared for emit] [big] (id hint: default, vendors) 1 related asset
  assets by path js/dist/*.json 42.6 KiB 86 assets
  assets by path media/dist/badges/*.svg 178 KiB 20 assets
  asset js/dist/codicon.ttf 59.6 KiB [compared for emit] [from: node_modules/monaco-editor/esm/vs/base/browser/ui/codicons/codicon/codicon.ttf] (auxiliary name: npm.monaco-editor) (auxiliary id hint: vendor)
  asset wmd-buttons.png?c2040b12f2d6cb568ddc33587a070a29a18fed20099260c3ecbbfdbd62042f7b 7.29 KiB [compared for emit] [from: frontend/www/third_party/js/pagedown/wmd-buttons.png] (auxiliary name: problem_creator, problem_edit, problem_statement)
  orphan modules 16.8 KiB [orphan] 11 modules
  runtime modules 7.19 KiB 10 modules
  modules by path ./node_modules/ 16.6 MiB (javascript) 59.6 KiB (asset) 2549 modules
  modules by path ./frontend/ 6.39 MiB 1706 modules

  ERROR in frontend/www/js/omegaup/admin/support.ts:139:58
  TS2339: Property 'recommended' does not exist on type 'ContestDetails'.
      137 |                 adminSupport.contestTitle = data.title;
      138 |                 adminSupport.contestFound = true;
    > 139 |                 adminSupport.isContestRecommended = data.recommended;
          |                                                          ^^^^^^^^^^^
      140 |               })
      141 |               .catch(ui.apiError);
      142 |           },

  frontend (webpack 5.8.0) compiled with 1 error in 15623 ms

style:
  asset css/dist/omegaup_styles.css 15 KiB [compared for emit] (name: omegaup_styles)
  Entrypoint omegaup_styles 15 KiB (0 bytes) = css/dist/omegaup_styles.css 15 KiB omegaup_styles.js 0 bytes 1 auxiliary asset
  runtime modules 274 bytes 1 module
  ./frontend/www/sass/main.scss 50 bytes [built] [code generated]
  css ./node_modules/css-loader/dist/cjs.js!./node_modules/sass-loader/dist/cjs.js!./frontend/www/sass/main.scss 15 KiB [code generated]
  style (webpack 5.8.0) compiled successfully in 4485 ms
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Can I ask what is the best practice to solve this problem? Thank you so much!

frontend/server/src/Controllers/Contest.php
pabo99: This suggestion still hasn't been addressed

pabo99:
Ok, let’s start by clarifying what’s causing the error:
```ts
  ERROR in frontend/www/js/omegaup/admin/support.ts:139:58
  TS2339: Property 'recommended' does not exist on type 'ContestDetails'.
      137 |                 adminSupport.contestTitle = data.title;
      138 |                 adminSupport.contestFound = true;
    > 139 |                 adminSupport.isContestRecommended = data.recommended;
          |                                                          ^^^^^^^^^^^
      140 |               })
      141 |               .catch(ui.apiError);
      142 |           },
```
This is happening because the recommended field does not exist in the `ContestDetails` type. The reason for this is that `PHP` hasn’t informed TypeScript that this property exists.
The way to fix it is by applying the changes I suggested in a previous comment that hasn’t been addressed yet. Basically, it means that the types you declare in `PHP` will be translated into `TypeScript` once you add them to the `PHPDoc blocks` and then run the following command:
```bash
./stuff/lint.sh
```

pabo99:
Regarding the UI topic, we can leave that for another PR. That way, this change can be approved once you've applied the suggestions

zikunz:
> Ok, let’s start by clarifying what’s causing the error:
> 
> ```ts
>   ERROR in frontend/www/js/omegaup/admin/support.ts:139:58
>   TS2339: Property 'recommended' does not exist on type 'ContestDetails'.
>       137 |                 adminSupport.contestTitle = data.title;
>       138 |                 adminSupport.contestFound = true;
>     > 139 |                 adminSupport.isContestRecommended = data.recommended;
>           |                                                          ^^^^^^^^^^^
>       140 |               })
>       141 |               .catch(ui.apiError);
>       142 |           },
> ```
> 
> This is happening because the recommended field does not exist in the `ContestDetails` type. The reason for this is that `PHP` hasn’t informed TypeScript that this property exists. The way to fix it is by applying the changes I suggested in a previous comment that hasn’t been addressed yet. Basically, it means that the types you declare in `PHP` will be translated into `TypeScript` once you add them to the `PHPDoc blocks` and then run the following command:
> 
> ```shell
> ./stuff/lint.sh
> ```

Sorry for the late reply and thank you for your guidance! I will be sure to make the code changes on Monday and update you :)

zikunz:
> Regarding the UI topic, we can leave that for another PR. That way, this change can be approved once you've applied the suggestions

I get it! Please let me know if I can also work on this UI topic after this PR is approved and merged. I will clarify the requirements and implement the UI if given the chance to work on it. Thank you!

zikunz:
> Ok, let’s start by clarifying what’s causing the error:
> 
> ```ts
>   ERROR in frontend/www/js/omegaup/admin/support.ts:139:58
>   TS2339: Property 'recommended' does not exist on type 'ContestDetails'.
>       137 |                 adminSupport.contestTitle = data.title;
>       138 |                 adminSupport.contestFound = true;
>     > 139 |                 adminSupport.isContestRecommended = data.recommended;
>           |                                                          ^^^^^^^^^^^
>       140 |               })
>       141 |               .catch(ui.apiError);
>       142 |           },
> ```
> 
> This is happening because the recommended field does not exist in the `ContestDetails` type. The reason for this is that `PHP` hasn’t informed TypeScript that this property exists. The way to fix it is by applying the changes I suggested in a previous comment that hasn’t been addressed yet. Basically, it means that the types you declare in `PHP` will be translated into `TypeScript` once you add them to the `PHPDoc blocks` and then run the following command:
> 
> ```shell
> ./stuff/lint.sh
> ```

I am sorry, in the latest commit, I updated the PHP docblock to add `recommended: bool` to the `ContestDetails` type definition and then ran 
```shell
./stuff/lint.sh
```
but encounter the following lint and php errors:
https://github.com/omegaup/omegaup/actions/runs/14578791716/job/40890721643?pr=8170

Does this mean there is more code I should change or is my approach in the latest commit wrong?

I am stuck for one whole day, may I ask if you could give me some more guidance? Thank you so much!

pabo99:
> > Regarding the UI topic, we can leave that for another PR. That way, this change can be approved once you've applied the suggestions
> 
> I get it! Please let me know if I can also work on this UI topic after this PR is approved and merged. I will clarify the requirements and implement the UI if given the chance to work on it. Thank you!

Sure, you can work on this in a new PR

pabo99:
> > Ok, let’s start by clarifying what’s causing the error:
> > ```ts
> >   ERROR in frontend/www/js/omegaup/admin/support.ts:139:58
> >   TS2339: Property 'recommended' does not exist on type 'ContestDetails'.
> >       137 |                 adminSupport.contestTitle = data.title;
> >       138 |                 adminSupport.contestFound = true;
> >     > 139 |                 adminSupport.isContestRecommended = data.recommended;
> >           |                                                          ^^^^^^^^^^^
> >       140 |               })
> >       141 |               .catch(ui.apiError);
> >       142 |           },
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This is happening because the recommended field does not exist in the `ContestDetails` type. The reason for this is that `PHP` hasn’t informed TypeScript that this property exists. The way to fix it is by applying the changes I suggested in a previous comment that hasn’t been addressed yet. Basically, it means that the types you declare in `PHP` will be translated into `TypeScript` once you add them to the `PHPDoc blocks` and then run the following command:
> > ```shell
> > ./stuff/lint.sh
> > ```
> 
> @pabo99 I am sorry, in the latest commit, I updated the PHP docblock to add `recommended: bool` to the `ContestDetails` type definition and then ran
> 
> ```shell
> ./stuff/lint.sh
> ```
> 
> but encounter the following lint and php errors: https://github.com/omegaup/omegaup/actions/runs/14578791716/job/40890721643?pr=8170
> 
> Does this mean there is more code I should change or is my approach in the latest commit wrong?
> 
> I am stuck for one whole day, may I ask if you could give me some more guidance? Thank you so much!

You're very close to solving it. Let me explain — I hope it's clear:

In your recent commits, you added the `recommended` property to `ContestDetails`, which was great. But if you look at line 1885, there's a callback that returns the same `ContestDetails` type, and this in turn converts the `Contest` object to an array. This array doesn't know what properties the object contains, so an annotation is added on line 1887. There, you'll also need to include the `recommended` property.

That's all. Would you like to give it a try? Once you push the change, we'll see if the error in the github actions is gone. In fact, you can verify it before pushing by running inside the container:

```
find frontend/ \
    -name *.php \
    -and -not -wholename 'frontend/server/libs/third_party/*' \
    -and -not -wholename 'frontend/tests/badges/*' \
    -and -not -wholename 'frontend/tests/controllers/*' \
    -and -not -wholename 'frontend/tests/runfiles/*' \
    -and -not -wholename 'frontend/www/preguntas/*' \
  | xargs ./vendor/bin/psalm \
    --long-progress \
    --show-info=false
```

frontend/server/src/Controllers/Contest.php
```diff
@@ -1994,6 +1995,7 @@ function () use ($contest) {
                 $result['admin'] = false;
                 $result['opened'] = false;
                 $result['has_submissions'] = $hasSubmissions;
+                $result['recommended'] = boolval($contest->recommended);
```
pabo99, line 1998: This line shouldn't be necessary since you're already getting the `recommended` field when calling `$contest->asFilteredArray`, so this line can be removed:
```suggestion
```
Usually, these manually added fields are only needed when they don't belong to the object, in this case, `Contest`.

pabo99 (commented):
Just one last detail:

It seems you're accessing an `API` where the support team member doesn’t have access: `Contest::apiDetails`.

I found that the `getContestDetails` function is restricting access to check and store the last access time, which is correct, since this user is not officially part of the contest.

So, we need to add a validation to exit the function before reaching the error. It looks like there’s already a place where this check is performed:
https://github.com/omegaup/omegaup/blob/70f98829ed21ea807a3aa22607f910dadd072a56/frontend/server/src/Controllers/Contest.php#L2074-L2078

zikunz:
> > > Ok, let’s start by clarifying what’s causing the error:
> > > ```ts
> > >   ERROR in frontend/www/js/omegaup/admin/support.ts:139:58
> > >   TS2339: Property 'recommended' does not exist on type 'ContestDetails'.
> > >       137 |                 adminSupport.contestTitle = data.title;
> > >       138 |                 adminSupport.contestFound = true;
> > >     > 139 |                 adminSupport.isContestRecommended = data.recommended;
> > >           |                                                          ^^^^^^^^^^^
> > >       140 |               })
> > >       141 |               .catch(ui.apiError);
> > >       142 |           },
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This is happening because the recommended field does not exist in the `ContestDetails` type. The reason for this is that `PHP` hasn’t informed TypeScript that this property exists. The way to fix it is by applying the changes I suggested in a previous comment that hasn’t been addressed yet. Basically, it means that the types you declare in `PHP` will be translated into `TypeScript` once you add them to the `PHPDoc blocks` and then run the following command:
> > > ```shell
> > > ./stuff/lint.sh
> > > ```
> > 
> > 
> > @pabo99 I am sorry, in the latest commit, I updated the PHP docblock to add `recommended: bool` to the `ContestDetails` type definition and then ran
> > ```shell
> > ./stuff/lint.sh
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > but encounter the following lint and php errors: https://github.com/omegaup/omegaup/actions/runs/14578791716/job/40890721643?pr=8170
> > Does this mean there is more code I should change or is my approach in the latest commit wrong?
> > I am stuck for one whole day, may I ask if you could give me some more guidance? Thank you so much!
> 
> You're very close to solving it. Let me explain — I hope it's clear:
> 
> In your recent commits, you added the `recommended` property to `ContestDetails`, which was great. But if you look at line 1885, there's a callback that returns the same `ContestDetails` type, and this in turn converts the `Contest` object to an array. This array doesn't know what properties the object contains, so an annotation is added on line 1887. There, you'll also need to include the `recommended` property.
> 
> That's all. Would you like to give it a try? Once you push the change, we'll see if the error in the github actions is gone. In fact, you can verify it before pushing by running inside the container:
> 
> ```
> find frontend/ \
>     -name *.php \
>     -and -not -wholename 'frontend/server/libs/third_party/*' \
>     -and -not -wholename 'frontend/tests/badges/*' \
>     -and -not -wholename 'frontend/tests/controllers/*' \
>     -and -not -wholename 'frontend/tests/runfiles/*' \
>     -and -not -wholename 'frontend/www/preguntas/*' \
>   | xargs ./vendor/bin/psalm \
>     --long-progress \
>     --show-info=false
> ```

@pabo99 Thank you for the clear explanation! I understand now, the issue is twofold:

I added the `recommended` property to the `ContestDetails` type definition, but I also needed to update the PHPDoc annotation on line 1887 for the callback that returns `ContestDetails`

I have implemented the first part by adding `$result['recommended'] = boolval($contest->recommended);` before the return statement in the callback function. I have also updated the PHPDoc annotation to include the `recommended` property in the array type.

I appreciate your guidance on how to verify the fix locally using Psalm. This helped ensure the changes are correct before pushing!

All tests have been passed. May I ask if it is okay to merge to main? :)

pabo99:
I added a couple of comments. Once you address them, I'll be able to approve this PR

frontend/server/src/Controllers/Contest.php
zikunz: Sorry, I think it is addressed now.

zikunz:
@pabo99 I know saw one additional comment added just now and I removed the unnecessary line of code. May I ask whether there are other problem(s) to fix?

pabo99:
> @pabo99 I know saw one additional comment added just now and I removed the unnecessary line of code. May I ask whether there are other problem(s) to fix?

Sure, there's still the following comment I made a few moments ago that needs to be addressed:

> Just one last detail:
>
> It seems you're accessing an `API` where the support team member doesn’t have access: `Contest::apiDetails`.
>
>I found that the `getContestDetails` function is restricting access to check and store the last access time, which is correct, since this user is not officially part of the contest.
>
>So, we need to add a validation to exit the function before reaching the error. It looks like there’s already a place where this check is performed:
>https://github.com/omegaup/omegaup/blob/70f98829ed21ea807a3aa22607f910dadd072a56/frontend/server/src/Controllers/Contest.php#L2074-L2078

If you have any questions, I'd be happy to help you resolve them

zikunz:
> > @pabo99 I know saw one additional comment added just now and I removed the unnecessary line of code. May I ask whether there are other problem(s) to fix?
> 
> Sure, there's still the following comment I made a few moments ago that needs to be addressed:
> 
> > Just one last detail:
> > It seems you're accessing an `API` where the support team member doesn’t have access: `Contest::apiDetails`.
> > I found that the `getContestDetails` function is restricting access to check and store the last access time, which is correct, since this user is not officially part of the contest.
> > So, we need to add a validation to exit the function before reaching the error. It looks like there’s already a place where this check is performed:
> > https://github.com/omegaup/omegaup/blob/70f98829ed21ea807a3aa22607f910dadd072a56/frontend/server/src/Controllers/Contest.php#L2074-L2078
> 
> If you have any questions, I'd be happy to help you resolve them

@pabo99 Oh yes, I am sorry, I only checked the code review commentsunder Files changed. I just made changes to add `|| \OmegaUp\Authorization::isSupportTeamMember($identity)` to the existing if-statement, which allows support team members to bypass the access time check completely, just like admins and practice mode users do. I also added a test case to verify support team members can access contest details. May I ask if there are other problems that should be fixed?

frontend/tests/controllers/ContestDetailsTest.php
```diff
@@ -1197,4 +1197,52 @@ public function testContestPrintDetails() {
...
+        );
+
+        // Create a support team member
+        ['identity' => $supportTeamMember] = \OmegaUp\Test\Factories\User::createSupportUser();
+
+        // Assert the log is empty
+        $this->assertSame(0, count(\OmegaUp\DAO\ProblemsetAccessLog::getByProblemsetIdentityId(
+            $contestData['contest']->problemset_id,
+            $supportTeamMember->identity_id
+        )));
```
pabo99, line 1221: What do you think about using the assertEmpty function, which is closer to what is described in the comment:
```suggestion
        $this->assertEmpty(\OmegaUp\DAO\ProblemsetAccessLog::getByProblemsetIdentityId(
            $contestData['contest']->problemset_id,
            $supportTeamMember->identity_id
        ));
```
?

frontend/tests/controllers/ContestDetailsTest.php
```diff
@@ -1197,4 +1197,52 @@ public function testContestPrintDetails() {
...
+
+        // Prepare our request
+        $login = self::login($supportTeamMember);
+        $r = new \OmegaUp\Request([
+            'contest_alias' => $contestData['request']['alias'],
+            'auth_token' => $login->auth_token,
+        ]);
+
+        // Call api directly (without explicitly joining the contest)
+        $response = \OmegaUp\Controllers\Contest::apiDetails($r);
```
pabo99, line 1231: If you are only using the `$r` variable in this place, could you instead add the request inline?
```suggestion
        $login = self::login($supportTeamMember);

        // Call api directly (without explicitly joining the contest)
        $response = \OmegaUp\Controllers\Contest::apiDetails(
            new \OmegaUp\Request([
                'contest_alias' => $contestData['request']['alias'],
                'auth_token' => $login->auth_token,
            ])
        );
```
In fact, we prefer to always use it inline to send only the parameters required for the test.

frontend/tests/controllers/ContestDetailsTest.php
```diff
@@ -1197,4 +1197,52 @@ public function testContestPrintDetails() {
...
+            $contestData['contest']->problemset_id,
+            $supportTeamMember->identity_id
+        );
+        $this->assertNull($problemsetIdentity);
+
+        // Assert the log is not empty (should still log access)
+        $this->assertSame(1, count(\OmegaUp\DAO\ProblemsetAccessLog::getByProblemsetIdentityId(
+            $contestData['contest']->problemset_id,
+            $supportTeamMember->identity_id
+        )));
```
pabo99, line 1246: There is a function that is a better fit for this assertion:
```suggestion
        $this->assertCount(1, \OmegaUp\DAO\ProblemsetAccessLog::getByProblemsetIdentityId(
            $contestData['contest']->problemset_id,
            $supportTeamMember->identity_id
        ));
```

pabo99 (approved):
Thank you very much for fixing this, and also for adding a test that ensures the feature works correctly. It will also help prevent possible regressions.

I left you a few suggestions in the test; if you don't manage to apply them, that's okay. The test's behavior would still remain the same.

From my side, the change looks great!

frontend/tests/controllers/ContestDetailsTest.php
zikunz: You are right, it is a more elegant way to express the same test logic. We should update the test to use this approach instead.

frontend/tests/controllers/ContestDetailsTest.php
zikunz: I get it now!

frontend/tests/controllers/ContestDetailsTest.php
zikunz: I see!

heduenas:
cc @IsmaelSantanaR

PR #8173, opened 2025-04-08
Controller for managing the carousel items

iqbalcodes6602:
# Description
Currently the carousel items are static which are described in the carousel.config.ts file

 - We had already created the tables, dao and vo files for a new table Carousel_Items.
 - In this PR i have created a new contoller for managing the carousel items through the ui
 - The fetch, add, delete, update operationsa can be performed through this contoller.

Fixes: #6916 

# Comments

# Checklist:


frontend/server/src/Controllers/CarouselItems.php
```diff
@@ -0,0 +1,183 @@
...
+     *
+     * @omegaup-request-param int $carousel_item_id
+     *
+     * @return array{status: string}
+     */
+    public static function apiDelete(\OmegaUp\Request $r): array {
+        $r->ensureMainUserIdentity();
+        self::validateAdmin($r);
+
+        $carouselItemId = $r->ensureInt('carousel_item_id');
```
pabo99, line 57: We usually don't use IDs to perform these actions. Instead, we typically rely on aliases.

Is there a simple way to create a naming convention for the carousel items so they can be removed/updated based on that?

If it's not straightforward, we can leave it as is, no problem.

frontend/server/src/Controllers/CarouselItems.php
```diff
@@ -0,0 +1,183 @@
...
+                ],
+                array_values($activeItems)
+            ),
+        ];
+    }
+
+    /**
+     * @throws \OmegaUp\Exceptions\ForbiddenAccessException
+     */
+    private static function validateAdmin(\OmegaUp\Request $r): void {
```
pabo99, line 177: Is there a way for this function to receive something other than `request`? `identity` comes to mind.

The reason is that it's harder to control the parameters coming from a `request`, and therefore more difficult to guarantee that the types are actually what we expect.

Usually, we only use `request` in endpoints or entrypoints.

iqbalcodes6602: defenitely, i have updated this in the latest commit

frontend/database/00232_add_sample_news.sql
```diff
@@ -1,6 +1,6 @@
 -- Creating the Carousel_Items table with JSON support for multilingual text
 CREATE TABLE `Carousel_Items` (
-    `corousel_item_id` INT AUTO_INCREMENT PRIMARY KEY,
+    `carousel_item_id` INT AUTO_INCREMENT PRIMARY KEY,
```
pabo99, line 3: Oh no! This was a mistake during the review process of the PR where this was merged. I sincerely apologize.

Unfortunately, it's not possible to modify scripts that have already been deployed to production. To make this change, you'll need to create a new script with the necessary schema alterations.

After that, you'll need to follow the usual process for any schema changes.

pabo99 (commented):
I left you a comment to help fix the GitHub Actions error shown in the PHP check

frontend/database/00232_add_sample_news.sql
iqbalcodes6602: oh yes, when we were merging the Carousal table PR, you suggested me change in the name of id column, and there we got it wrongly spelled, since i just committed the changes which you have done this was pushed in the PR.
I am really sorry for this blunder, i should have checked from my side as well.

iqbalcodes6602:
> I left you a comment to help fix the GitHub Actions error shown in the PHP check

I had looked into this issue and pushed the latest commit with ```corousel_item_id``` everywhere.
If we keep ```corousel_item_id``` everywhere, then we dont need to change the database schema and dont need to add any new migration file as well, just the typo would be there.

But if you want me to fix the database and all i can do that as well. 
What do you think should i change the db schema and do the migration fix or just use the ```corousel_item_id``` everywhere?

pabo99:
> > I left you a comment to help fix the GitHub Actions error shown in the PHP check
> 
> hi @pabo99 , I had looked into this issue and pushed the latest commit with `corousel_item_id` everywhere. If we keep `corousel_item_id` everywhere, then we dont need to change the database schema and dont need to add any new migration file as well, just the typo would be there.
> 
> But if you want me to fix the database and all i can do that as well. What do you think should i change the db schema and do the migration fix or just use the `corousel_item_id` everywhere?

Yes, preferably we should fix the typo. It will be a tedious task, but in the long run, it will save us from other types of errors.

Let me know if you need any help

pabo99:
Sure, it would be best to address that change in a separate PR.

Just a reminder: you should create two separate PRs — one with the script and the `schema.sql` file, and another with the changes in the `dao_schema.sql` file and the autogenerated `DAO VO` and `DAO Base` files

frontend/server/src/Controllers/CarouselItems.php
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    /**
+     * @throws \OmegaUp\Exceptions\ForbiddenAccessException
+     */
+    private static function validateAdmin(\OmegaUp\DAO\VO\Identities $identity): void {
+        if (!\OmegaUp\Authorization::isSystemAdmin($identity)) {
+            throw new \OmegaUp\Exceptions\ForbiddenAccessException();
+        }
+    }
```
pabo99: Could you move this function to `Authorization` class?

This way we can reuse it in other places where we validate the sysadmin role

frontend/server/src/Controllers/CarouselItems.php
```diff
@@ -0,0 +1,182 @@
...
+        $carouselItem->image_url = $r->ensureString('image_url');
+        $carouselItem->link = $r->ensureString('link');
+        $carouselItem->button_title = $r->ensureString('buttonTitle');
+
+        $expiration = $r->ensureOptionalString('expiration_date');
+        $carouselItem->expiration_date = is_null($expiration)
+            ? null
+            : new \OmegaUp\Timestamp(strtotime($expiration));
+
+        $carouselItem->status = $r['status'] ? 'true' : 'false';
```
pabo99: This field is declared as a string:
https://github.com/omegaup/omegaup/blob/bc307989e299b91ed4552c7e64ee2fb7545e3040/frontend/server/src/DAO/VO/CarouselItems.php#L176-L181
, so you should handle it the same way as the other parameters.

frontend/server/src/Controllers/CarouselItems.php
```diff
@@ -0,0 +1,182 @@
...
+        );
+        if (is_null($carouselItem)) {
+            throw new \OmegaUp\Exceptions\NotFoundException(
+                'carouselItemNotFound'
+            );
+        }
+
+        \OmegaUp\DAO\Base\CarouselItems::delete($carouselItem);
+        return ['status' => 'ok'];
+    }
```
pabo99, line 75: Normally we only perform logical deletion, which means that changing the status should be enough, so this endpoint is not necessary unless there is a solid reason to keep it.

frontend/server/src/Controllers/CarouselItems.php
```diff
@@ -0,0 +1,182 @@
...
+    public static function apiListActive(\OmegaUp\Request $r): array {
+        $allItems = \OmegaUp\DAO\Base\CarouselItems::getAll();
+        $now = new \OmegaUp\Timestamp(\OmegaUp\Time::get());
+
+        $activeItems = array_filter($allItems, function ($item) use ($now) {
+            /** @var \OmegaUp\DAO\VO\CarouselItems $item */
+            return boolval($item->status) && (
+                is_null($item->expiration_date) || $item->expiration_date >= $now
+            );
+        });
```
pabo99: Could a function be added in the DAOs that contains the query with these filters?

frontend/server/src/Controllers/CarouselItems.php
iqbalcodes6602: ok

frontend/server/src/Controllers/CarouselItems.php
iqbalcodes6602: actually a funciton already exist i wil use that only

frontend/server/src/Controllers/CarouselItems.php
iqbalcodes6602: so we need to remove the ` public static function apiDelete(\OmegaUp\Request $r): array {` entirely
 and instead call the 
`public static function apiUpdate(\OmegaUp\Request $r): array {`  and pass status as `inactive`?

Or is it something else?

frontend/server/src/Controllers/CarouselItems.php
iqbalcodes6602: i didn't get it can you give the file path which you are talking about

frontend/server/src/Controllers/CarouselItems.php
```diff
@@ -0,0 +1,181 @@
...
+        $carouselItem->title = $r->ensureString('title');
+        $carouselItem->excerpt = $r->ensureString('excerpt');
+        $carouselItem->image_url = $r->ensureString('image_url');
+        $carouselItem->link = $r->ensureString('link');
+        $carouselItem->button_title = $r->ensureString('buttonTitle');
+
+        $expiration = $r->ensureOptionalString('expiration_date');
+        $carouselItem->expiration_date = is_null($expiration)
+            ? null
+            : new \OmegaUp\Timestamp(strtotime($expiration));
```
pabo99: Here you might find useful the function `\OmegaUp\DAO\DAO::fromMySQLTimestamp`
```suggestion
        $carouselItem->expiration_date = \OmegaUp\DAO\DAO::fromMySQLTimestamp(
            $r->ensureOptionalString('expiration_date')
        );
```
This function already handles the case of receiving null values, and for the rest it attempts to convert them to a Timestamp

frontend/server/src/Controllers/CarouselItems.php
```diff
@@ -0,0 +1,181 @@
...
+
+        $carouselItem->status = $r['status'] ? 'active' : 'inactive';
```
pabo99: Ideally, the UI should send the status as a string to avoid type conversions. But if it’s necessary for it to be received as a boolean, here you can use:
```suggestion
        $carouselItem->status = $r->ensureOptionalBool(
            'status'
        ) ? 'active' : 'inactive';
```

frontend/server/src/Controllers/CarouselItems.php
pabo99: It can be either of the two options:

- Keep the `apiDelete` endpoint and have it only update the status
- Use the option you propose of calling `apiUpdate` with just that parameter

Either option is valid. Just avoid the direct use of:
```php
\OmegaUp\DAO\Base\CarouselItems::delete($carouselItem)
```

frontend/server/src/Controllers/CarouselItems.php
pabo99: Sure, the path for this file is `frontend/server/src/DAO/CarouselItems.php`

It may not exist yet because this table has just been created. In this new file, you can add all the functions to retrieve data from the table that are not found in `frontend/server/src/DAO/Base/CarouselItems.php`

To create it, you can use the file `frontend/server/src/DAO/ACLs.php` as a reference, only updating the class name

frontend/server/src/Controllers/CarouselItems.php
iqbalcodes6602: okay i went with first approach and updated the code,

frontend/server/src/Controllers/CarouselItems.php
iqbalcodes6602: added the file also i went through the other files, so instead i just used the sql to fetch only active ones, instead of fetching all then filtering out

frontend/server/src/DAO/CarouselItems.php
```diff
@@ -0,0 +1,49 @@
...
+            \OmegaUp\DAO\VO\CarouselItems::FIELD_NAMES,
+            'ci'
+        ) . '
+            FROM
+                `Carousel_Items` ci
+            WHERE
+                ci.status = "active"
+                AND (
+                    ci.expiration_date IS NULL
+                    OR ci.expiration_date >= ?
```
pabo99: Could you use the MySQL function here to get the current date?
```suggestion
                    OR ci.expiration_date >= NOW()
```
This way, you don't need to add parameters

iqbalcodes6602: done

pabo99 (commented):
Great!

PR #8239, opened 2025-04-20
easy , medium and hard problem solver badge if user solve more than 10 problem in 7 day streak 

Asdts:
# Description

This allot a easy , medium and hard solver badge to user who solver more that 10 problem in past 7 days of easy , medium and hard category respectively

Fixes: #7949 

# Comments

have some issue while writing test case for this particular badge \OmegaUp\Test\Factories\ProblemParams need to have difficulty params which is not present need help in writing test case for same and guidance can I change Factories or not.

# Checklist:

- [ The code follows the [coding guidelines](https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/Coding-guidelines.md) of omegaUp.
- The tests were executed and all of them passed.
- The new tests were added.


frontend/badges/easySolver/query.sql
```diff
@@ -0,0 +1,67 @@
...
+  FROM
+    StreakCounts
+  WHERE
+    streak_days = 7
+    AND problems_solved >= 10
+)
+
+SELECT
+  user_id,
+  'Easy' AS badge
```
pabo99, line 65: Oh no, it looks like this is breaking the tests!

In the `getSortedResults` function, it expects only a list of `user_id`s who obtained the badge. But in the query, it also retrieves the badge along with the difficulty of the problems.

Is it really necessary to add this last field? I understand it's meant to identify which badge it corresponds to, but that should already be determined since this query is specifically for easy problems. Or am I missing something important?

frontend/badges/hardSolver/query.sql
```diff
@@ -0,0 +1,67 @@
...
+  FROM
+    StreakCounts
+  WHERE
+    streak_days = 7
+    AND problems_solved >= 10
+)
+
+SELECT
+  user_id,
+  'Hard' AS badge
```
pabo99, line 65: Same comment as the previous one.

frontend/badges/mediumSolver/query.sql
```diff
@@ -0,0 +1,67 @@
...
+  FROM
+    StreakCounts
+  WHERE
+    streak_days = 7
+    AND problems_solved >= 10
+)
+
+SELECT
+  user_id,
+  'Medium' AS badge
```
pabo99, line 65: Here as well.

Issue #8084, opened 2025-03-23
[BUG] UI Enhancement: Add Hover Effect to Header and Coder of the Month Sections


NiharikaPanda09:
## Current Behavior
. The header and the "Coder of the Month" sections (for Her, Overall, and School) do not change background color on hover.

. This makes the UI feel static and less interactive.

*Describe what happened instead of what was expected.*

*Current Behavior
. The header and the "Coder of the Month" sections do not change background color on hover.

. The UI remains static, making it feel less interactive.

*Expected Behavior
. When hovering over the header, the background color should change smoothly.

. When hovering over any "Coder of the Month" section, the background color should also change smoothly.

. The transition should be smooth and visually appealing.

## Expected Behavior
. When hovering over any "Coder of the Month" section, the background color should change smoothly.

. When hovering over the header, the background color should also change smoothly.

. The effect should have a smooth transition.

*Describe what you were originally trying to do and how you expected the platform to behave.*

	* Screenshot:

![Image](https://github.com/user-attachments/assets/3d2a14c9-943b-44c8-9c1a-98cb9df97121)

Sarthakv007:
heyy I want to fix this issue is it still there?

Saloni-2005:
Hello @pabo99 Could you assign me if its open to add hover effect to header of the main page of the website.

pabo99:
Now the issue is assigned to you, @Saloni-2005

HarshYadv5554:
@NiharikaPanda09 if the issue is not yet resolved, can it be assigned to me

pabo99:
> [@NiharikaPanda09](https://github.com/NiharikaPanda09) if the issue is not yet resolved, can it be assigned to me

Sure! Actually, the issue has been assigned to you since March 23rd 😃

NiharikaPanda09:
@pabo99 yea im so sorry..i got scared..this was my first contribution so..i did a lot of file changes and couldn't redo it..i tried many times still couldn't do it ..so i gave up .... ..you can assign to some one...sorry for late reply..i was literally scared😞🥺

HarshYadv5554:
@pabo99 Can u please assign this task to me ??

pabo99:
Hello @HarshYadv5554! Sorry about that — I mistakenly thought it was @NiharikaPanda09 who had requested the issue assignment instead of you. I've now assigned it to you.

HarshYadv5554:
@pabo99 Hover effect added , give it a check

[...]

PR #8246, opened 2025-05-01
feat: Add hover effects to Coder of the Month sections and header

HarshYadv5554:
# Description

This PR adds hover effects and smooth transitions to the Coder of the Month sections and header to improve user interaction and visual feedback. The changes make the UI more modern and interactive while maintaining the existing functionality.

Key changes include:
- Added smooth background color transitions on hover for all sections
- Implemented elevation effects with box-shadows for better depth perception
- Added subtle transform animations for interactive feedback
- Created a dedicated CSS file for better organization of styles
- Updated Vue components to use the new styling system

Fixes: #8084 

# Comments

Key areas for review:
1. The new `coder-of-month.css` file organization and structure
2. Transition timings and smoothness of animations
3. Color choices for hover states
4. Accessibility considerations with the interactive elements

## Files Changed:
1. `frontend/www/css/coder-of-month.css` (new file)
2. `frontend/www/css/style.css`
3. `frontend/www/js/omegaup/components/coderofthemonth/List.vue`
4. `frontend/www/js/omegaup/components/coderofthemonth/CodersList.vue`


pabo99:
Thank you very much, @HarshYadv5554!

Before starting the review, could you add a screencast of the changes you applied? This way, we will know exactly what was implemented in this PR.

frontend/www/js/omegaup/components/coderofthemonth/CodersList.vue
```diff
@@ -1,42 +1,45 @@
...
-            :country="coder.country_id"
-          ></omegaup-countryflag>
-        </td>
-        <td class="text-center align-middle">
-          {{ coder.date }}
-        </td>
-      </tr>
-    </tbody>
-  </table>
+  <div class="coder-of-month-section">
```
pabo99, line 2: If you only created this div as a wrapper, is it possible to remove it and add the class directly to the table?

frontend/www/css/style.css
```diff
@@ -1,3 +1,5 @@
+@import url('coder-of-month.css');
+
```
pabo99, line 2: It seems that this file is only required to display an error when a config file doesn't exist:

https://github.com/search?q=repo%3Aomegaup%2Fomegaup+style.css&type=code

So it shouldn't be necessary to add this import.

```suggestion
```

frontend/www/css/coder-of-month.css
```diff
@@ -0,0 +1,98 @@
...
+
+/* Header hover effects */
+.main-header {
+    transition: background-color 0.3s ease;
+    background-color: #5c7cba;
+}
+
+.main-header:hover {
+    background-color: #4a6ba9;
+}
```
pabo99, line 41: It seems that none of these styles are being used anywhere. And if they are, ideally they should be added directly in the component where they are needed.

If the reason for adding them across multiple components is reuse, perhaps those components could be refactored to use slots instead.

For now, can these be removed from this file?
```suggestion
```

frontend/www/css/coder-of-month.css
```diff
@@ -0,0 +1,98 @@
...
+    text-decoration: none;
+    display: block;
+    text-align: center;
+    margin-top: 15px;
+    transition: color 0.3s ease;
+}
+
+.see-full-list:hover {
+    color: white;
+}
```
pabo99, line 98: It looks like this isn't being used anywhere either. Can it be removed from this file?
```suggestion
```

frontend/www/css/coder-of-month.css
```diff
@@ -0,0 +1,98 @@
...
+    border-radius: 50%;
+    margin: 0 auto 15px;
+    display: block;
+    border: 3px solid rgba(255, 255, 255, 0.2);
+    transition: border-color 0.3s ease;
+}
+
+.coder-of-month-section:hover .coder-profile-image {
+    border-color: rgba(255, 255, 255, 0.4);
+}
```
pabo99, line 69: From what I can see, only these styles (and the ones on line 2) are actually used, and they only appear in the `CodersList.vue` component. Would it be possible to include them directly in that component? That way, adding this file wouldn’t be necessary.

pabo99 (commented):
I left some comments to help make the PR clearer.

Great first contribution!

frontend/www/js/omegaup/components/coderofthemonth/CodersList.vue
HarshYadv5554: @pabo99 thanks, would be taking it into consideration

frontend/www/js/omegaup/components/coderofthemonth/CodersList.vue
```diff
@@ -58,3 +59,74 @@ export default class CoderOfTheMonthList extends Vue {
...
+.coder-of-month-section:hover {
+  box-shadow: 0 4px 8px rgba(0, 0, 0, 0.15);
+  transform: translateY(-2px);
+  transition: all 0.3s ease;
+}
+
+/* Profile image styles */
+.coder-profile-image {
+  width: 100px;
+  height: 100px;
```
pabo99, line 112: Could you avoid using static sizes for the image?
```suggestion
```
If you want the image to be displayed at 100px dimensions, you can do the following:
- Identify where the Gravatar image is obtained:
https://github.com/omegaup/omegaup/blob/b59b3717cee2a550be32994122c7e4828c719f5b/frontend/server/src/Controllers/User.php#L4758
- As you can see, it is possible to send a parameter `s` to indicate the required image size. In that location, you can specify that you need the image at `100`.
To complete the change, you can rename the parameter—from `gravatar_32` to `gravatar_100`.

If you notice that it is used in multiple places, you can add this new parameter and simply update it in the component you are working on.

frontend/www/js/omegaup/components/coderofthemonth/CodersList.vue
```diff
@@ -58,3 +59,74 @@ export default class CoderOfTheMonthList extends Vue {
...
+  transition: all 0.3s ease;
+}
+
+.coder-row:hover {
+  background-color: rgba(0, 0, 0, 0.05);
+  transform: translateX(5px);
+}
+
+.coder-date {
+  color: #666;
```
pabo99, line 74: Can you add this (and all the other colors) as variables? Here’s an example of how to do it:
https://github.com/omegaup/omegaup/blob/b59b3717cee2a550be32994122c7e4828c719f5b/frontend/www/js/omegaup/components/coderofthemonth/TopCodersList.vue#L73-L76

 This way, it will be easier to build the color palette when dark mode is implemented in omegaUp.

pabo99 (commented):
A few comments left

HarshYadv5554:
@pabo99 improved CodersList component with Bootstrap classes and CSS variables for better maintainability

frontend/server/src/Controllers/User.php
```diff
@@ -4755,11 +4755,11 @@ private static function processCodersList(array $coders): array {
         $response = [];
         foreach ($coders as $coder) {
             $hashEmail = md5($coder['email'] ?? '');
-            $avatar = "https://secure.gravatar.com/avatar/{$hashEmail}?s=32";
+            $avatar = "https://secure.gravatar.com/avatar/{$hashEmail}?s=100";
             $response[] = [
                 'username' => $coder['username'],
                 'country_id' => $coder['country_id'],
-                'gravatar_32' => $avatar,
+                'gravatar_100' => $avatar,
```
pabo99, line 4762: I'm noticing that `gravatar_32` is still being used in a couple of different places. To avoid affecting the behavior of other components, could you add both sizes?

Please review all the instances where `gravatar_32` is used so that you replicate the same change with `gravatar_100`.

There will be one place where you don't need to modify it manually: https://github.com/omegaup/omegaup/blob/main/frontend/www/js/omegaup/api_types.ts

You can update that file by running:

```
$ ./stuff/lint.sh
```

from outside the container. That script will also take care of a variety of other changes regarding styles, indentation, and more.

pabo99 (commented):
It's almost done.

frontend/www/js/omegaup/components/coderofthemonth/List.vue
```diff
@@ -146,7 +146,10 @@ export default class CoderOfTheMonthList extends Vue {
   border-right: 0.0625rem solid #dee2e6;
   border-top-left-radius: 0rem;
   border-top-right-radius: 0rem;
+  background-color: rgba(0, 0, 0, 0.05);
```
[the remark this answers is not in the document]
HarshYadv5554: Added new variable --border-color-light with the value #dee2e6 in the _variables.scss file and also updated   List.vue file to use the CSS variables:

HarshYadv5554:
@pabo99 Your feedback has been lovingly turned into code. Changes are in—let me know if I nailed it or if I need to go back to the code cave 😄

frontend/www/js/omegaup/components/coderofthemonth/List.vue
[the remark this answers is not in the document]
pabo99: I think you're missing push some changes, as the PR doesn't show any modifications made to the `_variables.scss` file

frontend/server/src/Controllers/User.php
pabo99: The only thing left is to add `gravatar_100` to the `CoderOfTheMonthList` type declaration found on line 16 of this same file. Once you've added it, you’ll need to run the command I shared earlier from outside the container but within the WSL2 instance, assuming you're working on Windows.

pabo99 (commented):
Just one last comment and this PR is fine to me!

PR #8254, opened 2025-05-14
added a section in getTranslatedTag which converts the hardcoded to tag to the required one 

rajputdivyanshu81:
Fixes: #(8244)

# Description

  added a new script by which it redirects to the required required tag 

image

### DIFF
--- a/frontend/www/js/omegaup/components/problem/BaseList.vue
+++ b/frontend/www/js/omegaup/components/problem/BaseList.vue
@@ -152,7 +152,7 @@
                 :icon="['fas', 'eye-slash']"
               />
               <a
-                v-for="tag in problem.tags"
+                          v-for="tag in problem.tags"
                 :key="tag.name"
                 :class="`badge custom-badge custom-badge-${
                   tag.source.includes('quality') ? 'owner' : tag.source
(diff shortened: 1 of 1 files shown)


pabo99:
Would it be possible to include a screenshot to help identify the UI changes introduced in this PR?

rajputdivyanshu81:
Resolved these blue tags which were shown in some problems these tags should not appear on frontend.

image

rajputdivyanshu81:
@heduenas sir, can u please review this PR

frontend/www/js/omegaup/components/problem/BaseList.vue
```diff
@@ -303,6 +298,24 @@ export default class BaseList extends Vue {
...
+    if (topicName === 'InputOutput') {
+      return T.problemTopicInputOutput || 'Input/Output';
+    }
+    if (topicName === 'BinarySearch') {
+      return T.problemTopicBinarySearch || 'Binary Search';
+    }
+
+    return topicName.replace(/([A-Z])/g, ' $1').trim();
+  }
+  return tag;
```
pabo99, line 317: Ok, it seems that none of the `problemTopic` tags have a translation string. So what needs to be done in this change is to add all the strings (you can check them in the `ALLOWED_TAGS` constant) to the translation files for the three languages (`en.lang`, `es.lang`, `pt.lang`), and that way you don’t need to modify this part anymore.

You can keep the validation that if the string doesn’t exist, it shows the topic name — I found that interesting

rajputdivyanshu81:
Sure , i will work on it, with this approach

Issue #8205, opened 2025-04-11
[FEATURE] Create English version of `Contests-API.md`


pabo99:
Create a file with the English translation of `Contests-API.md`. Additionally, ensure that the content is up-to-dated

## Context

This issue is created as part of the "Documentation translation to English" project.

Akarshkushwaha:
@pabo99 "I would like to work on updating the translation. Could you please assign it to me?"

pabo99:
@Akarshkushwaha Thank you for your interest in collaborating with omegaUp. The issue is now assigned to you!

Akarshkushwaha:
@pabo99 
PR_ID #8255 I remove the remove the files which are not related  to Translation part

PR #8255, opened 2025-05-14
Translate Contests-Api.md into English only

Akarshkushwaha:
# Description

Translates the Contents of Contests-Api.md file into English in Contests-Api-English.md

image

Fixes: #8205


frontend/www/docs/Contests-Api-English.md
```diff
@@ -0,0 +1,263 @@
+**NOTE**: This documentation is outdated.
```
pabo99: Could you rename this file to `Contests-API.md`? That way, we can also take the opportunity to remove the Spanish version, which we’ll no longer need.

Akarshkushwaha: @pabo99 I've renamed the file and removed the Spanish version as requested ✅

PR #8256, opened 2025-05-14
Fix/problem topic tags

HarshYadv5554:
Description:
This PR fixes an issue where certain problem tags (such as problemTopicInputOutput and problemTopicSorting) were being displayed as their raw string names on the problem collection page, rather than as user-friendly, human-readable values.
Changes made:
Added translations for problemTopicInputOutput ("Input and Output") and problemTopicSorting ("Sorting") in frontend/templates/en.lang.
Committed and pushed all local changes, ensuring the frontend now displays these tags with their proper, human-readable names.
Resolved build issues by restoring missing third-party dependencies and ensuring a successful build.
User Interface:
The tags on the collection page now display as “Input and Output” and “Sorting” instead of their raw string names.


pabo99:
It looks like these changes were meant to be in PR #8250, right?

Or are they two separate changes?

You'll need to add a screenshot to show what’s being fixed in this PR—it’ll make the review process easier.

One more thing, could you revert the changes made to the `package-lock.json` and `yarn.lock` files? It seems they aren’t necessary for this PR

Issue #8228, opened 2025-04-17
[BUG] `https://omegaup.com/problem/creator` shouldn't require log in


heduenas:
It doesn't really need the user to be authenticated to allow them to use the problem creator

Avadhutgiri:
hey @heduenas ,
Just to clarify—shouldn't access to /problem/creator require user authentication by design?
Allowing anonymous users to access it might pose risks in terms of abuse, spam .

RajvardhanT7747:
@heduenas can you assign this task to me ?

pabo99:
> hey [@heduenas](https://github.com/heduenas) , Just to clarify—shouldn't access to /problem/creator require user authentication by design? Allowing anonymous users to access it might pose risks in terms of abuse, spam .

For now, there shouldn't be any risk in terms of abuse or spam since the current behavior of the component only serves to create and/or modify the .zip file used in `/problem/new/`.

Is that right, @heduenas ?

Avadhutgiri:
Okay @pabo99 ,
Can You assign this issue to me?

pabo99:
> Okay [@pabo99](https://github.com/pabo99) , Can You assign this issue to me?

Sure!

Issue #8162, opened 2025-04-07
[BUG] Error starting Docker containers in GitHub workflows


pabo99:
## Current Behavior

Occasionally, GitHub workflows fail because Docker containers cannot be started

## Expected Behavior

Ideally, all workflows should run correctly to streamline the PR review process.

This is the error that appears:
```bash
Run docker compose up --no-build --detach
  docker compose up --no-build --detach
  
  # Add the container names as aliases to localhost
  cat << EOF | sudo tee --append /etc/hosts
  1[2](https://github.com/omegaup/omegaup/actions/runs/14309431730/job/40105003632?pr=8151#step:6:2)7.0.0.1 rabbitmq mysql
  EOF
  
  # Wait for some of the containers
  wait-for-it -t [3](https://github.com/omegaup/omegaup/actions/runs/14309431730/job/40105003632?pr=8151#step:6:3)0 mysql:13306
  retry -t 5 -d 5 -- docker compose exec -T rabbitmq rabbitmqctl await_startup
  
  for i in $(seq 10); do
    if [[ "$(mysql \
            -uomegaup -pomegaup --host=mysql --port=1330[6](https://github.com/omegaup/omegaup/actions/runs/14309431730/job/40105003632?pr=8151#step:6:6) --skip-column-names --batch \
            -e 'SELECT 1;' || echo 0)" == "1" ]]; then
      break
    fi
    sleep 1
  done
  shell: /usr/bin/bash -e {0}
 Network omegaup_default  Creating
 Network omegaup_default  Created
 Volume "omegaup_omegaupdata"  Creating
 Volume "omegaup_omegaupdata"  Created
 Volume "omegaup_dbdata"  Creating
 Volume "omegaup_dbdata"  Created
 Volume "omegaup_rabbitmq"  Creating
 Volume "omegaup_rabbitmq"  Created
 Volume "omegaup_redis"  Creating
 Volume "omegaup_redis"  Created
 Container omegaup-rabbitmq-1  Creating
 Container omegaup-redis-1  Creating
 Container omegaup-mysql-1  Creating
 Container omegaup-redis-1  Created
 Container omegaup-rabbitmq-1  Created
 Container omegaup-mysql-1  Created
 Container omegaup-grader-1  Creating
 Container omegaup-broadcaster-1  Creating
 Container omegaup-gitserver-1  Creating
 Container omegaup-broadcaster-1  Created
 Container omegaup-grader-1  Created
 Container omegaup-runner-1  Creating
[...]

PR #8326, opened 2025-06-06
Fix/allow unauthenticated problem creator

Avadhutgiri:
## Description
This PR fixes issue #8228 by allowing unauthenticated users to access the problem creator page.

## Changes
- Removed authentication check from the problem creator dropdown menu in the navbar
- Removed authentication check from creator.php in  the problems

## Testing
- Verified that unauthenticated users can access the problem creator page
- Confirmed that the dropdown menu is visible without login
- Tested that the form submission works correctly

## Screenshots
image

## Checklist
- [x] The PR title follows the [conventional commits](https://www.conventionalcommits.org/) format
- [x] The PR description clearly describes the changes and their purpose
- [x] The PR is linked to the relevant issue (#8228)
- [x] The changes have been tested locally
- [x] The code follows the project's style guidelines

### DIFF
--- a/frontend/www/js/omegaup/components/common/NavbarItems.vue
+++ b/frontend/www/js/omegaup/components/common/NavbarItems.vue
@@ -119,7 +119,6 @@
               T.navViewLatestSubmissions
             }}</a>
             <form
-              v-if="isLoggedIn && isMainUserIdentity && !isUnder13User"
               class="collapse-submenu"
             >
               <div class="btn-group d-flex">
--- a/frontend/www/js/omegaup/components/common/NavbarItems.vue
+++ b/frontend/www/js/omegaup/components/common/NavbarItems.vue
@@ -118,9 +118,7 @@
             <a class="dropdown-item" href="/submissions/">{{
               T.navViewLatestSubmissions
             }}</a>
-            <form
-              class="collapse-submenu"
-            >
+            <form class="collapse-submenu">
               <div class="btn-group d-flex">
                 <span class="dropdown-item">
                   {{ T.myproblemsListCreateProblem }}
--- a/frontend/www/problems/creator.php
+++ b/frontend/www/problems/creator.php
@@ -1,10 +1,10 @@
 <?php
 namespace OmegaUp;
 require_once(dirname(__DIR__, 2) . '/server/bootstrap.php');
-\OmegaUp\UITools::redirectToLoginIfNotLoggedIn();
 
 \OmegaUp\UITools::render(
     fn (\OmegaUp\Request $r) => \OmegaUp\Controllers\Problem::getCreatorForTypeScript(
         $r
     )
 );
+
--- a/frontend/www/problems/creator.php
+++ b/frontend/www/problems/creator.php
@@ -7,4 +7,3 @@ require_once(dirname(__DIR__, 2) . '/server/bootstrap.php');
         $r
     )
 );
-


pabo99 (commented):
Great! It seems to be working fine.

Just one more suggestion:

- Could you add a Cypress test? There might already be some tests related to this feature—my suggestion is to update the existing ones to simulate the process with a user who is not logged in.

Avadhutgiri:
Hey @pabo99 I am getting error while checking the cypress for add and edit layouts 
I tried fixing it but while making manual frontend changes there were no errors.
I would appreciate if you mentioned where I am doing mistake .
Thankyou!!

Avadhutgiri:
also is the error 
```
failed to mkdir /var/lib/docker/volumes/omegaup_omegaupdata/_data/problems.git: file exists
```

from my side ??
because I didnt make changes in any other file than /home/phoenix-heart/omegaup/cypress/e2e/problem_creator.cy.ts
Thanks!!

frontend/www/js/omegaup/components/common/NavbarItems.vue
```diff
@@ -118,10 +118,7 @@
             <a class="dropdown-item" href="/submissions/">{{
               T.navViewLatestSubmissions
             }}</a>
-            <form
-              v-if="isLoggedIn && isMainUserIdentity && !isUnder13User"
-              class="collapse-submenu"
-            >
+            <form class="collapse-submenu">
```
pabo99: It looks like this change enables both entries in the menu:

image

But we only want `Create zip file` to appear.

Is it possible for only that option to be visible? Ideally, it should appear without a sub-menu, and the text could be `Create zip file for problem`.

cypress/e2e/problem_creator.cy.ts
```diff
@@ -3,272 +3,299 @@ import T from '../../frontend/www/js/omegaup/lang';
 import { problemCreatorPage } from '../support/pageObjects/problemCreatorPage';
 import { profilePage } from '../support/pageObjects/profilePage';
 
-describe('Problem creator Test', () => {
+describe('Problem Creator Tests', () => {
```
pabo99: Oh, there were too many changes in this file, but it looks like the tests are working fine.

For future updates, take a look at the existing tests we have—you might find simpler ways to resolve these issues. In this case, you could have used parameterized tests, where you only needed to add the new scenario (unauthenticated users), and the rest of the code could have remained unchanged.
Either way, thanks for adjusting the tests!

pabo99 (commented):
It looks like the tests are working correctly. At least in my local environment, I didn’t find any errors.

You’ll just need to address the suggestion I left, and that should be it.

cypress/e2e/problem_creator.cy.ts
Avadhutgiri: Yeah you are right but the problem was whenever user was not authenticated the default language got set to english and in testing it was expecting in Espaniol so I modified it to include both.

Also the error was related  to cypress in add and modify the layout during testing of unauntheticated version
In previous part I had mistakenly skipped that part where the test runs for unautheticated user 

the problem is still arising Iit says missing button I have yet to grasp.

Avadhutgiri:
Sorry I guess after editing in the frontend I need to make changes in the cypress tests
I will try to do it today !!

cypress/e2e/problem_creator.cy.ts
pabo99: Ahh, I see. This is a good moment to mention something. Even when the user is not logged in, we have a way to force the language displayed in the browser. This allows us to run our tests in the expected language without causing them to fail due to language differences.

Let me explain in case you want to try it manually: in the browser’s address bar, you can add 
- `?lang=en` to display content in English
- `?lang=es` for Spanish
- `?lang=pt` for Portuguese. 
- `?lang=pseudo` you can also use this one  which we sometimes use to verify that all text strings are correctly passed through the internationalization files.

Back to the main topic— to fix the tests for unauthenticated users, you can update this function:

https://github.com/omegaup/omegaup/blob/ae7e646ac249136ad8687176f6eb68759b45daf0/cypress/support/pageObjects/problemCreatorPage.ts#L35-L40

by adding a lang parameter, and then include the following inside the function:

```ts
cy.location('pathname', { timeout: 10000 }).should(
  'eq',
  '/problem/creator/',
);
cy.location('search').then((search) => {
  if (!search.includes('lang') && lang) {
    cy.visit(`/problem/creator/?lang=${lang}`);
  }
});

```

That way, each test only needs to check if the user is not logged in to call the function with the proper language parameter.

Regarding the button you mentioned—this is something we need to fix in the tests. Right now, we're relying on indexes to select elements, but those can break when the DOM changes. 

In this case, removing a submenu (the navbar for unauthenticated users) seems to shift the positions of the other elements. The solution will be to add `data-attributes` to the elements we’re selecting in tests, so we can target them directly without guessing their position.

If the last part isn’t clear and you need help clarifying it, don’t hesitate to reach out to me.

pabo99:
> Sorry I guess after editing in the frontend I need to make changes in the cypress tests I will try to do it today !!

Thank you! Take your time.

Avadhutgiri:
Hey @pabo99  I'm occasionally seeing this Docker error during CI runs, even when I haven’t changed anything related to volumes:

```
Error response from daemon: failed to mkdir /var/lib/docker/volumes/omegaup_omegaupdata/_data/problems.git: file exists
```

It seems to be random and resolves on re-run, so maybe a race condition or Docker volume reuse issue?

Let me know if this is already tracked or if you'd prefer I open a separate issue for it. Just wanted to flag it here in case it affects other PRs too.

pabo99:
Oh, yes. That is a known issue. We have it documented in issue #8162

If you want to take it on, go ahead. Attempts have been made to fix it, but none have completely resolved the problem

For now, I'll rerun the failed tests for this PR, hoping the error disappears

Avadhutgiri:
Hey @pabo99 ,Can you check the cypress tests now?? I have fixed it .
I am open to suggestions , Thanks!!!

cypress/e2e/problem_creator.cy.ts
```diff
@@ -9,266 +9,290 @@ describe('Problem creator Test', () => {
...
+      name: 'Authenticated User',
+      isAuthenticated: true,
+      setup: () => {
+        cy.login(loginOptions);
+        profilePage.updatePreferredLanguage('es');
+      },
+      visit: () => {
+        cy.get('[data-nav-problems]').click();
+        cy.get('[data-nav-problems-create-options]').click();
+        cy.get('a[href="/problem/creator/"]').click();
```
pabo99: Okay, in the `problemCreatorPage.ts` file, you already have an intermediate `visit` function that allows you to format the URL you want to visit. So here, you would only need to call it if needed:

```suggestion
        problemCreatorPage.visit();
```

cypress/e2e/problem_creator.cy.ts
```diff
@@ -9,266 +9,290 @@ describe('Problem creator Test', () => {
...
+      setup: () => {
+        // No setup needed for unauthenticated user
+      },
+      visit: () => {
+        cy.get('[data-nav-problems]').click();
+        cy.get('a[href="/problem/creator/"]').click();
+        cy.location('search').then((search) => {
+          if (!search.includes('lang')) {
+            cy.visit('/problem/creator/?lang=es');
+          }
```
pabo99: Same case as the previous comment. Here you can access the intermediate function and pass it the lang parameter:
```suggestion
        problemCreatorPage.visit('es');
```

cypress/support/pageObjects/problemCreatorPage.ts
```diff
@@ -32,10 +32,23 @@ export class ProblemCreatorPage {
     cy.get('button[class="btn btn-success"]').eq(0).click();
   }
 
-  visit(): void {
+  visit(lang?: string): void {
```
pabo99, line 35: Ohh, I see that the behavior of this function will change depending on the parameters you send.

When you send the `lang `parameter, you are assuming that the user is not logged in, and when they are not logged in, the option in the menu appears directly. On the other hand, for logged-in users, it appears in a sub-menu.

The solution I suggest is the following, if you wrap this instruction in a condition:
```ts
cy.get('[data-nav-problems-create-options]').click();
```
where you validate that `lang` does not exist, it could look like this:
```ts
    if (!lang) {
      cy.get('[data-nav-problems-create-options]').click();
    }
```
the rest of the function would remain the same

pabo99 (commented):
Awesome! Excellent job! 

I just left a few minor comments, and this change is all set

Avadhutgiri:
Hey @pabo99 , You are right I was using but the unauthenticated tests were failing so I manually did it.
But yeah now I am sending lang parameter only with unauthenticated user
and this works fine Thanks!!

pabo99 (approved):
Great! Thank you for yor patience, @Avadhutgiri!

heduenas (approved):

PR #8347, opened 2025-06-25
AI Editorial Generator - DAO files

SammanSarkar:
# Description
Data access layer for AI editorial jobs: value object, base CRUD operations, and business logic methods.
# Checklist:

### DIFF
--- a/frontend/server/src/DAO/AiEditorialJobs.php
+++ b/frontend/server/src/DAO/AiEditorialJobs.php
@@ -1,0 +1,234 @@
+<?php
+
+namespace OmegaUp\DAO;
+
+/**
+ * AiEditorialJobs Data Access Object (DAO).
+ *
+ * Esta clase contiene toda la manipulacion de bases de datos que se necesita
+ * para almacenar de forma permanente y recuperar instancias de objetos
+ * {@link \OmegaUp\DAO\VO\AiEditorialJobs}.
+ *
+ * @access public
+ */
+class AiEditorialJobs extends \OmegaUp\DAO\Base\AiEditorialJobs {
+    /**
+     * Creates a new AI editorial job
+     *
+     * @return string The generated job ID (UUID)
+     */
+    public static function createJob(
+        int $problemId,
+        int $userId
+    ): string {
+        $jobId = self::generateUuid();
+
+        $job = new \OmegaUp\DAO\VO\AiEditorialJobs([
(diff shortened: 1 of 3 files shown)


frontend/server/src/DAO/AiEditorialJobs.php
```diff
@@ -0,0 +1,234 @@
...
+
+        $job = new \OmegaUp\DAO\VO\AiEditorialJobs([
+            'job_id' => $jobId,
+            'problem_id' => $problemId,
+            'user_id' => $userId,
+            'status' => 'queued',
+            'attempts' => 0,
+        ]);
+
+        self::save($job);
```
pabo99, line 38: Wow, I had never used this auto-generated function before. In fact, I think none of us had! 

Thanks for noticing it and putting the existing code to good use!

frontend/server/src/DAO/AiEditorialJobs.php
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    /**
+     * Gets a job by its UUID
+     *
+     * @return \OmegaUp\DAO\VO\AiEditorialJobs|null
+     */
+    public static function getJobByUuid(string $jobId): ?\OmegaUp\DAO\VO\AiEditorialJobs {
+        return self::getByPK($jobId);
+    }
```
pabo99: Is this function really necessary?

If it only contains the instruction to fetch the AiEditorialJob by its id, couldn’t we just call that function directly? If in the future we need to add more logic, then we could create a wrapper function at that point.

frontend/server/src/DAO/AiEditorialJobs.php
```diff
@@ -0,0 +1,234 @@
...
+
+        self::save($job);
+    }
+
+    /**
+     * Counts recent jobs by user for rate limiting
+     */
+    public static function countRecentJobsByUser(
+        int $userId,
+        int $hours = 1
```
pabo99: Can we create a constant to define the number of hours that determine whether a job is considered recent? That way, if we use it in multiple places, we’d only need to update the value in one spot.

pabo99 (commented):
Thanks for splitting the original PR. I've added some comments in this one!

SammanSarkar:
@pabo99 Updated the PR with your feedback. Let me know if anything else is needed.

Issue #8297, opened 2025-05-19
[FEATURE] Separate code into distinct functions based on `QualityNomination` type to be created


pabo99:
As a developer, I want the `QualityNomination` creation function to have clear and structured code.

To achieve this, each type of nomination should be handled in separate functions. This will ensure that the necessary parameters are correctly passed to each function, since the information is obtained from a JSON that could cause discrepancies when everything is managed within a single function divided by multiple `if/else` statements.

## Context

Currently, understanding the Quality Nomination process is challenging due to the large function size.

It is important that new collaborators can understand the code without difficulties.

Avadhutgiri:
Hey @pabo99 
Is this issue currently open??

pabo99:
> Hey [@pabo99](https://github.com/pabo99) Is this issue currently open??

Sure! Want to give it a try?

Avadhutgiri:
Thanks Will start it today!!!

Avadhutgiri:
Just to confirm before I begin:
This issue is more about refactoring and cleanup rather than adding any new feature or behavior or API fields, right?
So the goal is to split existing logic by nomination type into separate functions, without introducing any new functionality?

pabo99:
That’s correct! in this issue we’ll focus only on refactoring the code based on the nomination type. 

If we come across any bugs or functions that need fixing along the way, I’ll let you know so they can be addressed.


How do we know if there are issues? By reviewing the unit tests. All existing tests should pass. If they don’t, we’ll need to determine whether the refactor introduced a new issue or if the issue was already present in the original implementation.
Feel free to ask if you have any questions.

Avadhutgiri:
Okay thanks!!

antedotee:
I have solved the issue locally. If the issue is still unresolved, I would like to raise pull request. I have created seperated file for enums for better readability and dev experience.

thunderbird7413:
is this issue solved @antedotee?

pabo99:
The associated PR has the `Stale` label, so if you already have the version with the corrections, @antedotee, go ahead with the change.

thunderbird7413:
?

AbdulKreemShah2408:
Hey @pabo99, is this issue still open and available? If so, I would love to pick it up and start working on refactoring the QualityNomination creation logic.

pabo99:
Hi @AbdulKreemShah2408 

I assigned this issue to you since no one has opened a PR for it yet.

[...]

PR #8349, opened 2025-06-26
split QualityNomination creation logic by type

Avadhutgiri:
# Description

Refactored the `QualityNomination` controller to improve code clarity and maintainability.  
- **Main change:** Split the large `createNomination` function into smaller, type-specific functions for each nomination type (`suggestion`, `promotion`, `demotion`, `dismissal`, `quality_tag`).
- Each type now has its own validation and processing logic, making the code easier to understand and maintain.
- No changes to the user interface.

**Fixes:** #8297   

# Comments

- All existing tests were executed and passed, confirming that no regressions were introduced.
- Reviewers may want to pay special attention to the new helper functions to ensure all edge cases are still covered.
- No UI changes were made, so no screenshots or videos are included.

# Checklist:

- [ ] If you are creating a feature, the new tests were added. *(Not a new feature, so not applicable)*
- [x] If the change is large (> 200 lines), this PR was split into various Pull Requests. *(This PR only affects backend logic and tests)*


Avadhutgiri:
Hey @pabloaguilar Can You check if this PR satisfies the solution??

frontend/server/src/Controllers/QualityNomination.php
```diff
@@ -230,12 +230,83 @@ public static function createNomination(
         string $nominationType,
         array $contents
     ): \OmegaUp\DAO\VO\QualityNominations {
+        // Validate user permissions based on nomination type
```
pabo99, line 233: For each type of nomination, certain required parameters must exist. What do you think about deconstructing the `contents` array right from the start and only sending the necessary parameters for each validation?

frontend/server/src/Controllers/QualityNomination.php
```diff
@@ -230,12 +230,83 @@ public static function createNomination(
...
+    }
+
+    /**
+     * Validates user permissions based on nomination type and contents
+     */
+    public static function validateUserPermissions(
+        \OmegaUp\DAO\VO\Problems $problem,
+        \OmegaUp\DAO\VO\Identities $identity,
+        string $nominationType,
+        array $contents
```
pabo99, line 303: Based on the previous comment, here we would only need to receive `$beforeAC`, which would be of boolean type. That way, we can avoid parsing in line 308 and also skip line 307, since it will always exist.

The same approach should be applied to the rest of the validation functions.

frontend/server/src/Controllers/QualityNomination.php
```diff
@@ -230,12 +230,83 @@ public static function createNomination(
...
+    ): void {
```
pabo99, line 304: Here we could avoid a nested conditional by using the guard clause pattern, which means we can invert the condition to look like this:
```php
if ($nominationType === 'demotion'|| $nominationType === 'quality_tag') {
    return;
}
... // Rest of the function
```

frontend/server/src/Controllers/QualityNomination.php
```diff
@@ -230,12 +230,83 @@ public static function createNomination(
...
+            $identity,
+            $nominationType,
+            $contents
+        );
+
+        $qualityNomination = null;
+
+        // Validate and process nomination based on type
+        switch ($nominationType) {
+            case 'suggestion':
```
pabo99, line 245: Since you're already making this change, you could also add all these nomination types into an Enum. Fortunately, PHP version 8.1 now supports this feature: https://php.watch/versions/8.1/enums#enum-syntax

pabo99 (commented):
I left a few suggestions to make the code more readable.

Issue #8296, opened 2025-05-19
[FEATURE] Stop Using the `mixed` type in the `QualityNomination` Controller


pabo99:
As a developer, I want the type declarations in the `QualityNomination` controller in `PHP` to be as explicit as possible.

Using the `mixed` type is not very helpful, as it serves as a shortcut to avoid declaring the correct type and temporarily satisfies the type validator we currently use (in this case, `PSALM`).

So, for this change, all types declared as `mixed` should be updated, but only where possible. The goal is to minimize its usage as much as possible.

## Context

This affects runtime behavior when an incorrect type value is passed, which cannot be detected through validation tests.

This issue should only address the specified controller. Later, new issues will need to be created to work on eliminating `mixed` throughout the entire codebase.

isabelccc:
Hi, @pabo99 , Can you assign this to me

pabo99:
Sure!

PR #8354, opened 2025-07-03
Stop Using the mixed type in the QualityNomination Controller #8296

isabelccc:
# Description

Please include a description of the changes you have made. The text you put in this section will be used as the **commit message** when this PR is merged.

Fixes: #8296 

# Comments

# Checklist:

### DIFF
--- a/frontend/server/src/Controllers/QualityNomination.php
+++ b/frontend/server/src/Controllers/QualityNomination.php
@@ -221,7 +221,7 @@ class QualityNomination extends \OmegaUp\Controllers\Controller {
     ];
 
     /**
-     * @param array{tags?: list<string>, before_ac?: bool, difficulty?: int, quality?: int, statements?: array<string, array{markdown: string}>, source?: string, reason?: string, original?: string, tag?: string, quality_seal?: bool, level?: string, rationale?: string} $contents
+     * @param array $contents
      * @return \OmegaUp\DAO\VO\QualityNominations
      */
     public static function createNomination(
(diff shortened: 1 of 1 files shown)


frontend/server/src/Controllers/QualityNomination.php
```diff
@@ -221,7 +221,7 @@ class QualityNomination extends \OmegaUp\Controllers\Controller {
     ];
 
     /**
-     * @param array{tags?: mixed, before_ac?: mixed, difficulty?: mixed, quality?: mixed, statements?: mixed, source?: mixed, reason?: mixed, original?: mixed, tag?: mixed, quality_seal?: bool} $contents
+     * @param array $contents
```
pabo99, line 224: Okay, the motivation behind making this change is precisely to properly declare the types of each variable. When we use the `mixed` wildcard, we're telling `PSALM` that we don’t really care about the type.

The same happens when we use the `array` wildcard—`PSALM` knows the variable is an array, but it doesn't care about its contents.

So, the recommended approach here is to restore the elements that were there before, but try to identify their correct types. I’ve tried to infer the types in an automated way, but haven’t been able to so far, so doing it manually is the best option.

This seems to be the correct type:
 ```suggestion
      * @param array{tags?: list<string>, before_ac?: bool, difficulty?: int, quality?: int, statements?: array<string, array{markdown: string}>, source?: string, reason?: string, original?: string, quality_seal?: bool} $contents
```

Once you apply this change and run `PSALM`, you’ll start seeing some errors—this is because some explicit validations like `is_string`, `is_array`, etc., will no longer be necessary.

You'll also be able to remove forced assertions like `/** @psalm-assert ...*/`, since everything will already be properly typed.

pabo99 (commented):
Thanks for trying to resolve this issue. I left a comment, hoping it helps clarify a few points.

If you need any help, feel free to reach out to us.

pabo99:
Closed in favor of PR #8374

PR #8364, opened 2025-07-23
AI-Editorial-Generator Worker Components

SammanSarkar:
# Description
AI Editorial Generator worker components code - solution_handler and omegaup_api_client

### DIFF
--- a/stuff/ai_editorial_worker/omegaup_api_client.py
+++ b/stuff/ai_editorial_worker/omegaup_api_client.py
@@ -1,0 +1,433 @@
+"""
+omegaUp API Client - Handles all API interactions with omegaUp platform
+
+This client uses auth_token from user sessions (passed as 'ouat' parameter)
+instead of username/password authentication. This integrates seamlessly with
+omegaUp's session management system.
+"""
+
+import json
+import logging
+
+from typing import Dict, Any, List, Optional, Tuple
+
+import requests
+from requests.adapters import HTTPAdapter
+from urllib3.util.retry import Retry
+
+
+class OmegaUpAPIClient:
+    """
+    Production-ready omegaUp API client with auth_token authentication.
+
+    Features:
+    - Uses session auth_token (no username/password needed)
+    - Redis session caching for efficiency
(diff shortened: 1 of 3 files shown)


stuff/ai_editorial_worker/omegaup_api_client.py
```diff
@@ -0,0 +1,433 @@
+"""
```
heduenas: This file has no logic specific to the AI editorials, right? Can you move it under `/stuff/`?

stuff/ai_editorial_worker/solution_handler.py
```diff
@@ -0,0 +1,203 @@
...
+    def submit_solution(self,
+                        problem_alias: str,
+                        source_code: str,
+                        language: str = 'py3') -> Optional[str]:
+        """Submit solution and return run GUID."""
+        try:
+            # Check for Karel problems (special handling)
+            if 'karel' in problem_alias.lower():
+                logging.info("Karel problem detected, skipping submission")
+                return "KAREL_SKIP"
```
heduenas: Can you instead throw an exception and handle it in the calling code?

stuff/ai_editorial_worker/solution_handler.py
```diff
@@ -0,0 +1,203 @@
...
+        """Check run status with polling until completion."""
+        try:
+            start_time = time.time()
+
+            while time.time() - start_time < max_wait_time:
+                verdict, score, memory = (
+                    self.api_client.get_run_status_detailed(run_guid))
+
+                # Check if run is complete
+                if verdict in [
```
heduenas: what's the verdict returned when the run is not complete? Can you check for that instead? If in the future we decide to add another verdict type in the grader we will have to remember to include that here as well, and I would prefer to avoid that.

SammanSarkar: Makes sense, checking incomplete state(JE) instead of all complete verdicts

heduenas (approved):

heduenas (approved):

PR #8373, opened 2025-08-06
AI-Editorial-Generator Worker - Authentication

SammanSarkar:
# Description
Updates to worker and controller for authentication

### DIFF
--- a/frontend/tests/controllers/AiEditorialTest.php
+++ b/frontend/tests/controllers/AiEditorialTest.php
@@ -38,6 +38,7 @@ class AiEditorialTest extends \OmegaUp\Test\ControllerTestCase {
             ];
         }
     }
+
     /**
      * Test successful editorial generation
      */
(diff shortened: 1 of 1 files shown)


frontend/server/src/Controllers/AiEditorial.php
```diff
@@ -119,27 +138,110 @@ private static function queueJobToRedis(
...
+                OMEGAUP_ENVIRONMENT !== 'production' &&
+                OMEGAUP_ENVIRONMENT === 'test'
+            );
+
+            if ($isTestEnvironment) {
+                // In test mode, update job status to 'queued' and don't fail
+                try {
+                    \OmegaUp\DAO\AiEditorialJobs::updateJobStatus(
+                        $jobId,
+                        'queued',
```
heduenas: can you create 3 local variables, one for the 'queued', one for the 'Job queued ...' and one for the true. To avoid code repetition.

heduenas: Actually no. We should avoid having special logic for tests in controllers. In this case what you should do instead is directly update the status of the job in the DB after making the call to apiGenerate

heduenas (approved):

PR #8383, opened 2025-08-15
Create Problems-API(en).md

MishtiGarg250:
# Description

create a new file which contains english translation of Problems-API.md file


pabo99 (commented):
One minor suggestion and that's good to me

pabo99:
Ah, I forgot to mention:

The Spanish versions of these documents will gradually disappear. So it would be better if, in this same change, you could move the English content into the already existing file.

PR #8386, opened 2025-08-15
Changed Problems-API.md to english

MishtiGarg250:
Make conversion of language in Problems-API.md from spanish to english

### DIFF
--- a/frontend/www/docs/Problems-API.md
+++ b/frontend/www/docs/Problems-API.md
@@ -1,82 +1,90 @@
 ## POST `problem/create`
 
-### Descripción
-Crea un nuevo problema.
+### Description
+Creates a new problem.
 
-### Privilegios
-Usuario loggeado
+### Privileges
+- Must be a logged-in user.
 
-### Parámetros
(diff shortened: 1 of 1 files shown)


MishtiGarg250:
@heduenas sir, ok so you want me to delete the original file?

pabo99:
> @heduenas sir, ok so you want me to delete the original file?

I see now: what Hugo is saying is that it won’t be necessary to translate this document or any of the documents with the API suffix in the `/docs` directory, since the autogenerated documentation already exists at:

https://github.com/omegaup/omegaup/tree/main/frontend/server/src/Controllers

So, for this PR to be approved, you can look for the endpoints described in this document and add the description for each parameter. For example:

For `problem/create`, you can add the descriptions here:
https://github.com/omegaup/omegaup/blob/e3f449392e91a6f8f0bc5b8f89739def946d87d2/frontend/server/src/Controllers/Problem.php#L429-L448

And after applying the changes, you’ll need to run:

```bash
./stuff/lint.sh
```

Outside of the Docker container (inside your WSL instance if you’re on Windows) so that the changes are propagated to the autogenerated file.

PR #8398, opened 2025-08-22
Integrate `Teaching Assistant` with frontend and database

Nandini-Jaiswal:
# Description

https://github.com/user-attachments/assets/90442e44-6403-4c2a-8a09-6eecd5e8d71f

# Checklist:

### DIFF
--- a/frontend/database/dao_schema.sql
+++ b/frontend/database/dao_schema.sql
@@ -16,6 +16,30 @@ CREATE TABLE `ACLs` (
 /*!40101 SET character_set_client = @saved_cs_client */;
 /*!40101 SET @saved_cs_client     = @@character_set_client */;
 /*!50503 SET character_set_client = utf8mb4 */;
+CREATE TABLE `AI_Editorial_Jobs` (
+  `job_id` varchar(36) NOT NULL COMMENT 'UUID identificador único del trabajo',
+  `problem_id` int NOT NULL COMMENT 'Identificador del problema',
+  `user_id` int NOT NULL COMMENT 'Usuario que solicitó la generación',
+  `status` enum('queued','processing','completed','failed','approved','rejected') NOT NULL DEFAULT 'queued' COMMENT 'Estado actual del trabajo',
+  `error_message` text COMMENT 'Mensaje de error en caso de fallo',
+  `is_retriable` tinyint(1) NOT NULL DEFAULT '1' COMMENT 'Indica si el error permite reintentos (1 = sí, 0 = no)',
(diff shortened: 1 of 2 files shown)


frontend/server/src/Controllers/Course.php
```diff
@@ -6119,6 +6122,53 @@ function ($identity): int {
...
+
+        $r->ensureMainUserIdentity();
+        $courseAlias = $r->ensureString(
+            'course_alias',
+            fn (string $alias) => \OmegaUp\Validators::alias($alias)
+        );
+
+        $course = self::validateCourseExists($courseAlias);
+
+        if (
```
heduenas: can you extract this to a common function `ensureIsCourseAdmin`?

heduenas:
@Nandini-Jaiswal the frontend changes are the very last piece that should be checked in. Otherwise we will show the users buttons that do nothing. Can you separate the PHP changes into it's own PR? we can merge those already.

Nandini-Jaiswal:
> @Nandini-Jaiswal the frontend changes are the very last piece that should be checked in. Otherwise we will show the users buttons that do nothing. Can you separate the PHP changes into it's own PR? we can merge those already.

That completely makes sense; I will revert the `UI` changes, sorry!

frontend/server/src/Controllers/Course.php
Nandini-Jaiswal: Done!

Nandini-Jaiswal:
@heduenas @pabo99 PTAL!

frontend/templates/en.lang
```diff
@@ -1970,6 +1971,8 @@ wordsDetailed = "Detailed"
 wordsDetails = "Details"
 wordsDifference = "Difference"
 wordsDifficulty = "Difficulty"
+wordsDisableAITA = "Disable AI TA"
```
pabo99: This string is only used in UI changes. Could you please revert this line and the others in this PR?

The only string actually used here is `unableToUpdateTAEnabledField`.

pabo99 (commented):
Could you please add a new `PHPUnit` test to validate that this new field is updated correctly?

You can base it on some of the existing functions in `CourseCreateTest.php`.

frontend/templates/en.lang
heduenas: @Nandini-Jaiswal after this is done we should be ready to merge this PR

Nandini-Jaiswal: @heduenas @pabo99 
I have addressed the suggestions!
PTAL!

PR #8478, opened 2025-10-23
Exclude site-admins from CotM

sudarsan2k5:
# Description
Exclude site-admins from CotM eligibility.

Fixes: #8473

# Checklist:


stuff/cron/database/coder_of_the_month.py
```diff
@@ -262,6 +262,18 @@ def get_cotm_eligible_users(
                 s.verdict = 'AC' AND s.type= 'normal' AND s.time >= %s AND
                 s.time <= %s AND p.visibility >= 1 AND p.quality_seal = 1 AND
                 i.user_id IS NOT NULL
+                AND i.user_id NOT IN (
+                    SELECT ur.user_id
+                    FROM User_Roles ur
+                    WHERE ur.acl_id = 1 AND ur.role_id = 1
```
heduenas, line 270: please add comment mentioning what acl_id = 1 and role_id = 1 are

stuff/cron/database/coder_of_the_month.py
```diff
@@ -262,6 +262,18 @@ def get_cotm_eligible_users(
...
+                )
+                AND i.identity_id NOT IN (
```
heduenas: identities shouldn't be given sys admin access, so we can skip this check

heduenas (commented):
Thank you for working on this issue!

We have some integration tests for CotM somewhere in the PHP code. Can you add a test that verifies that a sys admin user that solves many problem loses CotM to a regular user that solves much less problems?

sudarsan2k5:
Thanks @pabo99, Should we also consider this  `And maybe it would also be good to exclude them from the user ranking so that they do not generate noise for users,` as you mentioned in the issue? Or should I keep these changes for now?

pabo99:
> Thanks @pabo99, Should we also consider this `And maybe it would also be good to exclude them from the user ranking so that they do not generate noise for users,` as you mentioned in the issue? Or should I keep these changes for now?

Let's handle that in a separate issue or PR. What do you think, @heduenas?

heduenas (approved):

heduenas:
> > Thanks @pabo99, Should we also consider this `And maybe it would also be good to exclude them from the user ranking so that they do not generate noise for users,` as you mentioned in the issue? Or should I keep these changes for now?
> 
> Let's handle that in a separate issue or PR. What do you think, @heduenas?

Yes let's do that in a separate PR. Thanks!

PR #8505, opened 2025-11-03
fixed the composing window size

### LINKED ISSUE #4850: [FEATURE] Aumentar el tamaño de la cajita de texto de la vista de editar redacción
Lo ideal sería que por default tuviera el mismo tamaño que el preview de la derecha.

image

antedotee:
Before:
image

After:
image

Fixes #4850

### DIFF
--- a/frontend/www/docs/How-to-use-Cypress-in-omegaUp.md
+++ b/frontend/www/docs/How-to-use-Cypress-in-omegaUp.md
@@ -75,13 +75,36 @@ You need to install the dependency:
 sudo apt-get install libasound2
 ```
 
+If you are using Ubuntu 24.04 or a newer version, install the required dependency with:
+```bash
+sudo apt install libasound2t64
+```
+
+**Note:** If you encounter any unexpected errors, try running:
+```bash
+sudo apt update
+```
+and then try the process again.
+
 ## Introduction
 **Note:** Currently, tests are run outside of the Docker container.
 
 Everything related to Cypress is located inside the `./cypress` folder at the root directory.
 
 ## GUI
-To run the Cypress GUI (which is great), use `npx cypress run` `./node_modules/.bin/cypress open`. This will open a screen with all available tests.
+To run Cypress, use one of the following commands:
+
+- To run the tests from the command line:
+  ```bash
+  npx cypress run
+  ```
+- To open the Cypress GUI:
+  ```bash
+  npx cypress open
+  # or
+  ./node_modules/.bin/cypress open
+  ```
+The GUI will open a window listing all available tests.
 
 ![image](https://github.com/user-attachments/assets/a3bf6b11-0e9a-4290-b5e5-fec6884be3e7)
 


frontend/www/js/omegaup/components/problem/creator/solution/SolutionTab.vue
```diff
@@ -169,4 +169,36 @@ export default class SolutionTab extends Vue {
 .wmd-button-bar {
   background-color: var(--wmd-button-bar-background-color);
 }
+
+.row {
+  display: flex;
```
pabo99: It looks like this style is already part of the row class in Bootstrap 4, so there's no need to add it here
```suggestion
```

frontend/www/js/omegaup/components/problem/creator/solution/SolutionTab.vue
```diff
@@ -169,4 +169,36 @@ export default class SolutionTab extends Vue {
...
+
+  .col-md-6 {
+    display: flex;
+    flex-direction: column;
```
pabo99: These styles can be handled using Bootstrap 4 classes:
```css
d-flex flex-column
``` 
You can add them directly in the places where you use `col-md-6`

frontend/www/js/omegaup/components/problem/creator/solution/SolutionTab.vue
```diff
@@ -169,4 +169,36 @@ export default class SolutionTab extends Vue {
...
+      .wmd-input {
+        flex: 1;
+        min-height: 400px;
+        height: auto !important;
+        resize: vertical;
+      }
+    }
+
+    &:last-child {
+      omegaup-markdown {
```
pabo99: The linter doesn't accept this as a valid selector. Instead, you can try using the generic `data-attribute` of the `Markdown` component:

```suggestion
     [data-markdown-statement] {
```
Or, if you want to be more specific, you can use the `data-attribute` assigned in this component. Both approaches yield the same result

frontend/www/js/omegaup/components/problem/creator/solution/SolutionTab.vue
```diff
@@ -169,4 +169,36 @@ export default class SolutionTab extends Vue {
...
+        flex: 1;
+        min-height: 400px;
+        overflow-y: auto;
+        border: 1px solid #ddd;
```
pabo99: Can this color be added to the variables file and used here just like it's done on line 170?

frontend/www/js/omegaup/components/problem/creator/statement/StatementTab.vue
```diff
@@ -105,4 +105,36 @@ export default class StatementTab extends Vue {
...
 }
+
+.row {
+  display: flex;
+
+  .col-md-6 {
+    display: flex;
+    flex-direction: column;
+
+    &:first-child {
```
pabo99: Is it necessary to specify that it only applies to the first child? It looks like you're already using a specific selector inside, which should only target that element.

Also, since you're using `scoped`, the styles in this file should be limited to this component and its children.

If it's not necessary, you can remove it and use the selector directly. The same approach could work where you're using  `last-child`

PR #8518, opened 2025-11-08
added the filter functionality to schools

### LINKED ISSUE #3381: [FEATURE] Agregar feature para filtrar escuelas
en https://omegaup.com/rank/schools/, analogo al de filtrar usuarios en el rank de usuarios.

antedotee:
I have added the filter functionality to schools based upon 3 criteria for now, by country or region and one filter by state. 

I was also unable to open my profile in local environment, so I fixed it by editing unused imports and then, I was able to open my profile.( If you need to create a seperate PR for that, I will do that).

Below are the screenshots:
image

image

image

Fixes #3381

### DIFF
--- a/frontend/server/src/Controllers/School.php
+++ b/frontend/server/src/Controllers/School.php
@@ -324,7 +324,10 @@ class School extends \OmegaUp\Controllers\Controller {
         $schoolRank = \OmegaUp\Cache::getFromCacheOrSet(
             \OmegaUp\Cache::SCHOOL_RANK,
             "{$offset}-{$rowCount}",
-            fn () => \OmegaUp\DAO\SchoolsRank::getFilteredRank($offset, $rowCount),
+            fn () => \OmegaUp\DAO\SchoolsRank::getFilteredRank(
+                $offset,
+                $rowCount
+            ),
             3600 // 1 hour
         );
 
--- a/frontend/server/src/DAO/SchoolsRank.php
+++ b/frontend/server/src/DAO/SchoolsRank.php
@@ -86,4 +86,3 @@ class SchoolsRank {
         ];
     }
 }
-
--- a/frontend/www/js/omegaup/api_types.ts
+++ b/frontend/www/js/omegaup/api_types.ts
@@ -4585,14 +4585,14 @@ export namespace types {
   }
 
   export interface SchoolRankPayload {
+    availableFilters?: { [key: string]: string };
+    filter?: string;
     length: number;
     page: number;
     pagerItems: types.PageItem[];
     rank: types.School[];
     showHeader: boolean;
     totalRows: number;
-    availableFilters?: { [key: string]: string };
-    filter?: string;
   }
 
   export interface SchoolUser {


frontend/server/src/Controllers/School.php
```diff
@@ -12,7 +12,7 @@
  * @psalm-type SchoolProblemsSolved=array{month: int, problems_solved: int, year: int}
  * @psalm-type SchoolUser=array{username: string, classname: string, created_problems: int, solved_problems: int, organized_contests: int}
  * @psalm-type SchoolProfileDetailsPayload=array{school_id: int, school_name: string, ranking: int, country: array{id: string, name: string}|null, state_name: string|null, monthly_solved_problems: list<SchoolProblemsSolved>, school_users: list<SchoolUser>, coders_of_the_month: list<SchoolCoderOfTheMonth>}
- * @psalm-type SchoolRankPayload=array{page: int, length: int, rank: list<School>, totalRows: int, showHeader: bool, pagerItems: list<PageItem>}
+ * @psalm-type SchoolRankPayload=array{page: int, length: int, rank: list<School>, totalRows: int, showHeader: bool, pagerItems: list<PageItem>, availableFilters?: array<string,string>, filter?: string}
```
pabo99, line 15: Can we make the type more specific? I'm thinking of something similar to what we have for users:

```suggestion
 * @psalm-type SchoolRankPayload=array{page: int, length: int, rank: list<School>, totalRows: int, showHeader: bool, pagerItems: list<PageItem>, availableFilters: array{country?: string, state?: string}, filter?: string}
```

frontend/server/src/Controllers/School.php
```diff
@@ -270,33 +270,79 @@ public static function getTopSchoolsOfTheMonth(
      *
      * @omegaup-request-param int $length
      * @omegaup-request-param int $page
+     * @omegaup-request-param string $filter
```
pabo99, line 273: This field is optional, so you could use:
```suggestion
     * @omegaup-request-param null|string $filter
```

frontend/server/src/Controllers/School.php
```diff
@@ -310,6 +356,101 @@ public static function getRankForTypeScript(\OmegaUp\Request $r): array {
...
+            $rankCacheName,
+            fn () => \OmegaUp\DAO\SchoolsRank::getFilteredRank(
+                $offset,
+                $rowCount,
+                'ranking',
+                'ASC',
+                $selectedFilter['filteredBy'],
+                $selectedFilter['value']
+            ),
+            3600 // 1 hour
```
pabo99, line 402: Would it be possible to use a constant? 

I'm thinking this cache could have a longer duration since the ranking is calculated every 24 hours. But for now, using a constant is enough

frontend/server/src/Controllers/School.php
```diff
@@ -310,6 +356,101 @@ public static function getRankForTypeScript(\OmegaUp\Request $r): array {
...
+    }
+
+    /**
+     * @return array{filteredBy: null|string, value: null|string|int}
+     */
+    private static function getSelectedFilter(
+        \OmegaUp\DAO\VO\Identities $identity,
+        string $filteredBy
+    ): array {
+        if ($filteredBy === 'country') {
```
pabo99, line 439: Can we use the values from the  `SchoolRankFilter` enum?

frontend/server/src/Controllers/School.php
```diff
@@ -403,6 +544,7 @@ public static function getSchoolOfTheMonth(string $date = null): array {
         }
 
         // Now get the school data
+        /** @var int $schoolOfTheMonthId */
```
pabo99, line 547: Why would this be necessary?

Considering you want to ensure that `$schoolOfTheMonthId` has an integer value, we already have the necessary validations.

For example, on line 526 we perform an early return if there are no records for schools of the month.
Then, on line 540, we validate that if no school Id exists, the function throws an exception

frontend/server/src/DAO/SchoolsRank.php
```diff
@@ -0,0 +1,88 @@
+<?php
```
pabo99, line 1: Hmm, it looks like there's no table named `Schools_Rank` as such. Maybe in the future it would be worth creating one so that rankings are completely independent from school records.

But for now, could you move this function to the `Schools` DAO?

frontend/server/src/Template/JsIncludeNode.php
```diff
@@ -105,8 +105,11 @@ private static function getJavaScriptDeps(string $entrypoint): array {
                 'Please run <tt style="background: #eee">cd /opt/omegaup && yarn install && yarn run dev-all</tt>.'
             );
         }
-        /** @var array{css: list<string>, js: list<string>} */
         $jsonContents = json_decode($textContents, associative: true);
+        if (is_null($jsonContents)) {
+            die('Invalid JSON in dependency file: ' . $jsonPath);
+        }
+        /** @var array{css: list<string>, js: list<string>} $jsonContents */
```
pabo99, line 112: Maybe the type declaration was in the right place. Could you please move it back there?

pabo99:
This PR contains conflicts. Are you still working on it, @antedotee ?

antedotee:
Yes. I will fix the merge conflicts as soon as possible.
Sorry for the delay.

pabo99:
No problem, take your time

PR #8541, opened 2025-11-21
Added the script and workflow for immutable SHA tags

### LINKED ISSUE #8539: The CI workflows uses version tags instead of immutable hashes
Currently, the project uses version tags. It would be better to use immutable SHA tags instead of version tags from a security perspective. 
More information can be found here:
https://www.stepsecurity.io/blog/pinning-github-actions-for-enhanced-security-a-complete-guide

antedotee:
Here are the detailed regarding the changes I implemented:

- `.gha-reversemap.yml` =>  A YAML file mapping GitHub Actions to their immutable commit hashes
- `hack/gha-reversemap.sh` => Adapted management script with verification operations
- `.github/workflows/verify-action-hashes.yml` => Now, this workflow will ensure the code user is pushing is containing immutable hashes or not

User can run this command `./hack/gha-reversemap.sh verify-mapusage` to ensure that no tags are there.

It can also run these commands:
```
# Update workflows to use commit hashes from reversemap (if any are out of sync)
./hack/gha-reversemap.sh apply-reversemap

# Update specific action to its latest version in the reversemap
./hack/gha-reversemap.sh update-action-version actions/checkout

# Update reversemap from current workflow files
./hack/gha-reversemap.sh update-reversemap
```

FIxes #8539


antedotee:
@pabo99  Please have a look. Also, the js tests are not related to the code I have modified. How should I fix this or should I leave it?

pabo99 (approved):

antedotee:
Are you talking about the tags changes to immutable hashes? If that is the case, I have already included all those changes in the PR as I first ran the commit locally to fix the tags.

pabo99:
Thanks for clarifying. Just to be clear, my suggestion was not about whether the tags are already immutable, but rather about adding a `--dry-run` mode to the script.

A `dry-run` is a common safeguard in programming (see [Wikipedia](https://en.wikipedia.org/wiki/Dry_run_(testing))) that lets contributors preview which files *would* be modified without actually applying changes. For example, tools like `rsync` use `--dry-run` to show the intended operations before executing them.

This would make the workflow safer and easier to trust. If you think this change could grow too much in scope, no worries, we can leave it for another occasion.

antedotee:
Hi, if the user ran the reversemap script locally, he can see the diff changes of tag to immutable hashes. In this way, there will be no need of dry run, I believe. I might be wrong still in the interpretation of the question you asked.

antedotee:
@pabo99 Added, Please have a look. Actually, I was thinking that how other users will use that, as there is not a dedicated documentation site for omegaup. Sorry, if I am mistaken. But, now, I will remember.

pabo99:
Hi! Could you please resolve the conflicts?

antedotee:
@pabo99 I have resolved the conflicts and updated the script to make sure local actions are skipped as they cannot be pinned to a commit hash. Looking forward to merging these changes for more security.

PR #8531, opened 2025-11-17
implemented the scorecard functionality

antedotee:
Here are the supporting screenshots:
image

image

Below is the screenshot of how the downloaded csv and xlsx looks:
image

image

Fixes #8088

PR #8545, opened 2025-11-23
Add server API for CSV/XLSX scoreboard export

antedotee:
Implements the backend API endpoint for exporting contest scoreboards.
Changes:
Adds exportScoreboard endpoint in Contest controller
Adds phpoffice/phpspreadsheet dependency for Excel generation
Updates ExitException to support file downloads
Part 2 of 3 — Split from #8531.

### DIFF
--- a/frontend/server/src/Controllers/Contest.php
+++ b/frontend/server/src/Controllers/Contest.php
@@ -6101,6 +6101,301 @@ class Contest extends \OmegaUp\Controllers\Controller {
         ];
     }
 
+    /**
+     * Downloads the scoreboard in CSV, XLSX, or ODS format
+     *
+     * @throws \OmegaUp\Exceptions\ForbiddenAccessException
+     *
+     * @omegaup-request-param string $contest_alias
+     * @omegaup-request-param string $format - One of: csv, xlsx, ods
+     */
+    public static function apiScoreboardDownload(\OmegaUp\Request $r): void {
+        $r->ensureIdentity();
+
+        $contestAlias = $r->ensureString(
+            'contest_alias',
+            fn (string $alias) => \OmegaUp\Validators::alias($alias)
+        );
+
+        $format = $r->ensureEnum(
+            'format',
+            ['csv', 'xlsx']
+        );
+
+        [
+            'contest' => $contest,
+            'problemset' => $problemset,
+        ] = self::validateBasicDetails($contestAlias);
+
+        if (is_null($contest->problemset_id)) {
+            throw new \OmegaUp\Exceptions\NotFoundException('contestNotFound');
+        }
+
+        // Check if user is contest admin
+        if (!\OmegaUp\Authorization::isContestAdmin($r->identity, $contest)) {
+            throw new \OmegaUp\Exceptions\ForbiddenAccessException(
+                'userNotAllowed'
+            );
+        }
+
+        // Get scoreboard data
+        $scoreboard = self::getScoreboard($contest, $problemset, $r->identity);
+
+        // Generate and download the file
+        self::generateScoreboardFile(
+            $scoreboard,
+            $format,
+            $contest->title ?? 'contest'
+        );
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


frontend/server/src/Controllers/Contest.php
```diff
@@ -6101,6 +6101,301 @@ public static function apiArchive(\OmegaUp\Request $r): array {
...
+    }
+
+    /**
+     * Generates and downloads scoreboard file in specified format
+     *
+     * @param Scoreboard $scoreboard
+     * @param string $format
+     * @param string $contestTitle
+     */
+    private static function generateScoreboardFile(
```
pabo99, line 6159: Would it be possible to move this and all the other private functions you created into a new class?

That way, this can be reused in courses without the need to duplicate code

frontend/server/src/Controllers/Contest.php
```diff
@@ -6101,6 +6101,301 @@ public static function apiArchive(\OmegaUp\Request $r): array {
         ];
     }
 
+    /**
+     * Downloads the scoreboard in CSV, XLSX, or ODS format
+     *
+     * @throws \OmegaUp\Exceptions\ForbiddenAccessException
+     *
+     * @omegaup-request-param string $contest_alias
+     * @omegaup-request-param string $format - One of: csv, xlsx, ods
```
pabo99, line 6110: If this PR doesn’t include the functionality to download the scoreboard in `ODS`, could you add a `TODO` comment in the code?

frontend/server/src/Controllers/Contest.php
```diff
@@ -6101,6 +6101,301 @@ public static function apiArchive(\OmegaUp\Request $r): array {
...
+     */
+    public static function apiScoreboardDownload(\OmegaUp\Request $r): void {
+        $r->ensureIdentity();
```
pabo99, line 6113: If only contest admins will be able to perform this action, could you use
```suggestion
        $r->ensureMainUserIdentity();
```
?

This way we further restrict the users who can access this API

frontend/server/src/Controllers/Contest.php
```diff
@@ -6101,6 +6101,301 @@ public static function apiArchive(\OmegaUp\Request $r): array {
...
+     */
+    private static function generateXLSX(
+        array $data,
+        string $filename,
+        string $contestTitle
+    ): void {
+        if (!class_exists('\PhpOffice\PhpSpreadsheet\Spreadsheet')) {
+            /** @psalm-suppress TranslationStringNotFound */
+            throw new \OmegaUp\Exceptions\NotFoundException(
+                'PhpSpreadsheet library not available'
```
pabo99, line 6281: Could we add this string to the localization files?


This way we avoid using psalm-suppress

frontend/server/src/Controllers/Contest.php
```diff
@@ -6101,6 +6101,301 @@ public static function apiArchive(\OmegaUp\Request $r): array {
...
+            header('Expires: 0');
+
+            $writer->save('php://output');
+            // In test environments, throw ExitException instead of exit
+            if (defined('IS_TEST') && IS_TEST === true) {
+                /** @psalm-suppress TranslationStringNotALiteralString */
+                throw new \OmegaUp\Exceptions\ExitException($headers);
+            }
+            exit;
+        } else {
```
pabo99, line 6387: This `else` is no longer necessary since the `if` statement has an explicit exit

frontend/server/src/Controllers/Contest.php
```diff
@@ -6101,6 +6101,301 @@ public static function apiArchive(\OmegaUp\Request $r): array {
...
+                self::generateXLSX($data, $filename, $contestTitle);
+                break;
+        }
+    }
+
+    /**
+     * Prepares scoreboard data for export format
+     *
+     * @param Scoreboard $scoreboard
+     * @return array<array<mixed>>
```
pabo99, line 6188: Would it be possible to declare a less general type? This is very ambiguous

PR #8550, opened 2025-11-24
UI redesign of create/edit contest

sudarsan2k5:
# Description

The UI of the contests was changed to use collapsibles and everything was organized into their respective categories

video:-

https://github.com/user-attachments/assets/9126ca80-2809-4ec0-8939-7ae6e1ca3c91

Fixes: #6354

# Checklist:


pabo99:
Looks like the failing checks are legit. How about slightly changing the approach for this PR?

Since this PR makes substantial changes to `NewForm.vue`, it might be easier to create a new component called `Form.vue`. This name fits better because it's used both for creating a new contest and editing an existing one.

This way, we can avoid the current test failures. Later, in a separate PR, we can:

- Add the necessary Jest tests
- Replace the old component with the new one
- Remove `NewForm.vue` and its corresponding test
- Update the Cypress test that’s currently failing

sudarsan2k5:
I did. There is no more present 😅

pabo99 (commented):
Oh, it looks like the Cypress test failures are legit.

What you need to do is review the place where contests are created/edited and check whether some of the elements that were hidden are being filled in from there. Once you find them, you'll need to trigger the click event on each of the collapsible menus.

I think I found the place:

https://github.com/omegaup/omegaup/blob/1f9107300e736509eafc3594df6c705711242f00/cypress/support/commands.ts#L209-L233

The element `[data-score-board-visible-time]` belongs to the `logistics` collapsible menu. So you’ll need to add:

```ts
cy.get('[data-target=".logistics"]').click();
```
before it’s used.

Then, you need to add:
```ts
cy.get('[data-target=".scoring-rules"]').click();
```
before `[data-score-mode]` is used.

And finally, you need to add:
```ts
cy.get('[data-target=".privacy"]').click();
```
before `[data-basic-information-required]` is used.

PR #8552, opened 2025-11-25
added the PR size labeler workflow

### LINKED ISSUE #8549: It would be better if the PR include label of size of the PR
Currently, when someone raise a PR, there is no label of how small or large it is. If somehow, we can add label of size using github actions, like XS for <+-10 lines, S for <+-25 lines and so on. It would be better for reviewer for reviewing.

antedotee:
Here is the demo screenshot of the github action working:
image

This is what I have kept as max lines for each PR:
xs -> <10
s -> <100
m -> <500
l -> <1000
xl -> >1000

Also, I have not included some files which shouldn't be included in the PR size count. Additionally, we can also add an automated comment if we want to prevent user raising a large PR(in case we need that).

Also, earlier, if I was making a subsequent commit, it was adding the tags, while ideally it should show the tag about the size of the state of the current PR. I have implemented this in my latest commit:
image

Fixes #8549


antedotee:
@pabloaguilar Please have a look.

.github/workflows/labeler.yml
```diff
@@ -0,0 +1,110 @@
...
+      - uses: codelytv/pr-size-labeler@v1
+        with:
+          xs_label: "size/XS"
+          xs_max_size: "10"
+          s_label: "size/S"
+          s_max_size: "100"
+          m_label: "size/M"
+          m_max_size: "500"
+          l_label: "size/L"
+          l_max_size: "1000"
```
pabo99: Actually, in the PR template we have a guideline that indicates the PR should ideally not exceed 200 lines. So, could we adjust the threshold values a bit?

I was thinking something like:

xs: 10  
s: 50  
m: 100  
l: 200  
xl: INF  

The message for the `xl` threshold looks fine, maybe we could include a note suggesting that if it’s possible to split the work into smaller changes, that would be great.

What do you think, @tvanessa?

antedotee:
@pabo99 Yes, for that to implement, like adding backend tag for backend changes and ci tag for changing github action files, we can use prow-github-actions to accomplish our task. I believe it will be better to handle that in a seperate PR.

antedotee:
@pabo99 Is there something left in this PR before merging?😅

PR #8565, opened 2025-11-29
implemented stale automation

### LINKED ISSUE #8553: It would be better to automatically close the issues and PR after a certain period of inactivity
Problem -> same as title

Also, If somehow we want to prevent some issues from staling, we can create an `exempt-from-stale` tag to prevent certain issues from staling. This way, the contributor can only see issues which still exist.

antedotee:
Here are the fixes I implemented:
- Issues and PRs are marked stale after 90 days of no activity
- The stale label is removed when issues/PRs are updated or commented on (remove-stale-when-updated: true)
- Set days-before-close: -1 to prevent automatic closing; issues remain open for manual closure

Fixes #8553

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -16,7 +16,7 @@ jobs:
         steps:
             - uses: actions/stale@v10
               with:
-                  # Mark issues as stale after 90 days of inactivity
+                  # Mark issues as stale after 45 days of inactivity
                   days-before-issue-stale: 45
                   # Mark PRs as stale after 45 days of inactivity
                   days-before-pr-stale: 45


.github/workflows/stale.yml
```diff
@@ -0,0 +1,40 @@
...
+                  # Messages to post when marking as stale
+                  stale-issue-message: "This issue has been marked as stale because it has been open for 90 days with no activity. Please comment or update this issue to remove the stale label."
+                  stale-pr-message: "This pull request has been marked as stale because it has been open for 90 days with no activity. Please comment or update this PR to remove the stale label."
+
+                  # Labels to use
+                  stale-issue-label: "Stale"
+                  stale-pr-label: "Stale"
+
+                  # Maximum operations per run to avoid rate limits
+                  operations-per-run: 30
```
pabo99: We currently have more than 500 issues and 140 PRs open. 

Can we start with a more aggressive number of operations? Once all of them have been processed, we can return to this amount

.github/workflows/stale.yml
```diff
@@ -0,0 +1,40 @@
...
+        runs-on: ubuntu-22.04
+        permissions:
+            actions: write
+            issues: write
+            pull-requests: write
+        steps:
+            - uses: actions/stale@v10
+              with:
+                  # Mark issues/PRs as stale after 90 days of inactivity
+                  days-before-stale: 90
```
pabo99: For PRs, could we use fewer days? Say, 45?

.github/workflows/stale.yml
```diff
@@ -16,8 +16,10 @@ jobs:
         steps:
             - uses: actions/stale@v10
               with:
-                  # Mark issues/PRs as stale after 90 days of inactivity
-                  days-before-stale: 90
+                  # Mark issues as stale after 90 days of inactivity
```
pabo99: nit:
```suggestion
                  # Mark issues as stale after 45 days of inactivity
```

pabo99 (approved):
Great!

Issue #8159, opened 2025-04-06
[FEATURE] Add Bookmark Functionality for Questions


k-vijay-05:
****Description****

This feature request suggests adding a bookmark functionality that allows users to save questions from the problem list that they find interesting or want to solve later

****Proposed Functionality****:

- When clicked, the question is saved to the user's bookmarks
- Create a "My Bookmarks" section in the user's profile view
- Allow users to view, manage, and access their bookmarked questions from their profile

I would like to hear your suggestion on this feature. Can I go ahead and raise a PR?

pabo99:
+1!

It sounds great! What do you think about the proposal, @heduenas @tvanessa ?

vijay-x-dev:
Hi @pabo99 

I’m thinking of splitting this feature into multiple PRs to keep the changes manageable:

1. Create the Bookmarks table and generate the corresponding DAO/VO files
2. Implement the backend controllers
3. Add the frontend changes for the bookmark functionality

Please let me know your thoughts on this breakdown

pabo99:
Hi @k-vijay-05 ! 

Your approach to splitting this issue makes a lot of sense. Just a couple of suggestions:

- Point 1 could be divided in two PRs: the first one adds the script along with the database schema changes, and the second one applies the changes to the DAO/VO.
- For the other two points, please make sure to include their corresponding tests.

Thanks in advance!

vijay-x-dev:
@pabo99 Sure! Thank you for the suggestions. I’ll follow this approach.

pabo99:
This issue remains open since the original PR was split into multiple smaller PRs

PR #8571, opened 2025-12-02
Add Problem_Bookmarks table for user problem bookmarking feature

vijay-x-dev:
# Description

This schema addition is part of the implementation for allowing users to bookmark problems and view them later in a dedicated section of their profile.

Part of: #8159 

# Checklist:


frontend/database/00254_bookmark_questions.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE `Problem_Bookmarks` (
+  `user_id` int NOT NULL,
```
pabo99: Could we use `identity_id` instead? That would reduce the complexity of the queries, since it would directly check the bookmarks of a single identity rather than reviewing the bookmarks of all identities associated with the user.

vijay-x-dev: Thanks for pointing this out! From this I understood that in omegaUp most things are tracked by identity_id instead of user_id, so using identity_id here lets us avoid the extra join through Users and makes the queries simpler.

vijay-x-dev:
Just to confirm: after this change I should also update `dao_schema.sql` and then run `./stuff/lint.sh` outside the container to generate the corresponding BASE and VO files, right?

frontend/database/00254_bookmark_questions.sql
pabo99: Yes, in practice, the` User` entity is mainly useful for grouping the results of all identities associated with it. This is particularly relevant for reward programs such as earning badges, coder of the month, or user rankings.

For everything else, using the `Identity` is enough

pabo99:
> Just to confirm: after this change I should also update `dao_schema.sql` and then run `./stuff/lint.sh` outside the container to generate the corresponding BASE and VO files, right?

Yes, ideally in a separate PR for the following reason: 

We currently don’t have a mechanism to revert changes in the database schema (basically, we don’t have a documented way to remove fields). 

The way we mitigate this issue is by first applying the changes directly to the database. Once the deploy is done, we verify that the change doesn't affect platform functionality. At that point, we can safely apply the schema changes without risk of regression

vijay-x-dev:
> > Just to confirm: after this change I should also update `dao_schema.sql` and then run `./stuff/lint.sh` outside the container to generate the corresponding BASE and VO files, right?
> 
> Yes, ideally in a separate PR for the following reason:
> 
> We currently don’t have a mechanism to revert changes in the database schema (basically, we don’t have a documented way to remove fields).
> 
> The way we mitigate this issue is by first applying the changes directly to the database. Once the deploy is done, we verify that the change doesn't affect platform functionality. At that point, we can safely apply the schema changes without risk of regression

Understood. Thank you for the clarification

vijay-x-dev:
> > Just to confirm: after this change I should also update `dao_schema.sql` and then run `./stuff/lint.sh` outside the container to generate the corresponding BASE and VO files, right?
> 
> Yes, ideally in a separate PR for the following reason:
> 
> We currently don’t have a mechanism to revert changes in the database schema (basically, we don’t have a documented way to remove fields).
> 
> The way we mitigate this issue is by first applying the changes directly to the database. Once the deploy is done, we verify that the change doesn't affect platform functionality. At that point, we can safely apply the schema changes without risk of regression

Hi @pabo99  I’m getting the following error when trying to push:
****'frontend/database/dao_schema.sql' can only have the same contents as 'frontend/database/schema.sql'.****

I had manually edited frontend/database/dao_schema.sql. Is there a specific script or command I should use to correctly regenerate/update dao_schema.sql from schema.sql instead of editing it directly?

pabo99:
Hi @k-vijay, this issue you mention is a known error that has already been reported to us. In that case, you can use the `--no-verify` flag, assuming that the change in `schema.sql` was applied in a different PR. 

Regarding how to perform the action, we don’t have a script that does it automatically, so the way you’re attempting it is the only option we have at the moment.

PR #8572, opened 2025-12-02
Implemented cache in the workflows

### LINKED ISSUE #8563: [FEATURE] Github action workflows doesn't use cache currently
We can optimize workflows by using composer dependencies cache, yarn dependencies cache, python pip cache. Additionally, we can also cache the docker layer. 

These changes can significantly reduce the time it takes for CI, I believe.

antedotee:
This PR adds caching for dependencies in CI workflows to reduce build times.
What's cached:
Composer dependencies — PHP packages (vendor/) cached based on composer.lock
Yarn dependencies — JavaScript packages (node_modules/) cached based on yarn.lock
Python packages — pip-installed packages cached based on requirements files
Docker layers — Docker image layers cached to speed up pulls and builds

Impact:
Faster CI runs when dependencies haven't changed
Reduced time spent downloading and installing dependencies
Lower CI costs from shorter job durations

Fixes #8563

### DIFF
--- a/.github/workflows/build-containers.yml
+++ b/.github/workflows/build-containers.yml
@@ -18,6 +18,17 @@ jobs:
         run: echo "branch=${GITHUB_REF#refs/heads/}" >> $GITHUB_OUTPUT
         id: extract_branch
 
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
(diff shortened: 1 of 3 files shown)


pabo99:
Could you please resolve the conflict?

Issue #8569, opened 2025-12-02
[FEATURE] Ability to create threads in discord regarding recently opened issues


antedotee:
Problem -> same as title

AshParmar:
Hey @antedotee I'd like to work on this issue

Ankitsinghsisodya:
Hi! I would like to work on this issue.

I have analyzed the requirements and propose implementing this feature using a lightweight **GitHub Action** workflow. This approach avoids external bot dependencies and keeps the integration native to the repository.

### Proposed Implementation Plan

**1. Workflow Configuration**
I will create a new workflow file (e.g., `.github/workflows/discord-issue-monitor.yml`) that triggers specifically on the `issues` event with the `opened` type.

**2. Discord Webhook Integration**
Instead of using heavy third-party actions, I will use a concise `curl` request within the workflow to hit the Discord Webhook API.
- The workflow will construct a JSON payload containing the **Issue Title**, **Number**, **URL**, and a snippet of the **Description**.
- I will utilize the `thread_name` parameter (or the compatible query parameter method) in the Webhook payload to ensure the message automatically generates a corresponding thread in the target channel.

**3. Security**
The Discord Webhook URL will be accessed securely via `{{ secrets.DISCORD_WEBHOOK_URL }}`, ensuring no credentials are exposed in the codebase.

**Outcome**
This implementation ensures that every new issue immediately spawns a discussion thread in Discord, keeping the community engaged without manual intervention.

Could you please assign this to me?

pabo99:
Is there a way to apply these actions only to issues that are created/edited with a specific tag?  

Perhaps we could start testing the functionality with the tags "GSoC" and "Good first issue", and later decide if we want to include more.

Assigned to @Ankitsinghsisodya !

Ankitsinghsisodya:
@pabo99 
Thanks for assigning this to me!

Yes, absolutely. We can restrict the notifications by adding a conditional check to the workflow.

I will configure the action to trigger on the `labeled` event and include a filter so the Discord webhook only fires if the applied tag matches "GSoC" or "good first issue". This will ensure we only create threads for the specific topics we want to prioritize.

I'll get started on the implementation now!

Ankitsinghsisodya:
@pabo99 
I have submitted a Pull Request for this issue. I look forward to your feedback.

Ankitsinghsisodya:
@pabo99 
Sir, kindly do the manually review. All the checks have passed.

[...]

Issue #7812, opened 2024-09-11
[BUG] Python test fails because of `mkdir`


Aritra8438:
## Current Behavior
Sometimes, the `python` and `cypress` test fails for the following reason.

```python
Error response from daemon: failed to mkdir /var/lib/docker/volumes/omegaup_omegaupdata/_data/problems.git: mkdir /var/lib/docker/volumes/omegaup_omegaupdata/_data/problems.git: file exists
```

<img width="639" alt="image" src="https://github.com/user-attachments/assets/b0222b90-0416-4e52-b559-b54f0e221915">

## Expected Behavior

- `Cypress` test isn't a required test for merging but `python` test failing blocks the pr to be merged when auto-merged is enabled.  

- [Here](https://github.com/omegaup/omegaup/actions/runs/10754323269/job/29828357668?pr=7801) is one such failed action.

Aritra8438:
@pabo99, here's the issue.

Ankitsinghsisodya:
Hi @pabo99 and @antedotee,
If you don’t mind, may I also work on fixing this issue? I believe it’s negatively impacting the developer experience, and I’d be happy to help address it.

pabo99:
Sure, feel free to work on fixing this issue, @Ankitsinghsisodya !

Ankitsinghsisodya:
@pabo99
I have submitted a Pull Request for this issue. I look forward to your feedback.

PR #8574, opened 2025-12-02
add Discord issue thread notification workflow

Ankitsinghsisodya:
# Description

Implemented a GitHub Action workflow (`discord-issue-monitor.yml`) to automatically send notifications to a Discord channel when an issue is created or tagged with specific labels.

**Key changes:**
* **Workflow Trigger:** Listens for `issues` events (specifically `opened` and `labeled`).
* **Filtering Logic:** The workflow strictly checks if the issue contains priority labels ("GSoC" or "good first issue") before executing.
* **Implementation:** Used `actions/github-script` (Node.js) instead of shell scripts to ensure robust JSON handling (preventing errors with special characters in issue titles) and to support creating threads in Discord via Webhook.
* **Outcome:** Sends a rich embed message to Discord and initiates a thread for discussion.

Fixes: #8569 

# Comments

**Notes for Reviewers:**
1.  **Prerequisite:** This workflow requires a repository secret named `DISCORD_WEBHOOK_URL` to be configured in the repository settings.
2.  **Tech Choice:** I opted for `actions/github-script` over a `curl` bash script to handle edge cases where issue titles contain quotes, backslashes, or emojis, ensuring valid JSON payloads every time.
3.  **Logic:** The script handles two scenarios:
    * An issue is **opened** with the required label already selected.
    * An existing issue is **labeled** later with the required label.

# Checklist:

- [x] The tests were executed and all of them passed. (Verified workflow syntax and logic manually).
- [ ] If you are creating a feature, the new tests were added. (N/A - Infrastructure/Workflow change).

### DIFF
--- a/.github/workflows/discord-issue-notification.yml
+++ b/.github/workflows/discord-issue-notification.yml
@@ -42,7 +42,7 @@ jobs:
 
       - name: Send Discord Notification
         if: steps.check-label.outputs.should_notify == 'true'
-        uses: actions/github-script@v7
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b
         env:
           DISCORD_WEBHOOK_URL: ${{ secrets.DISCORD_WEBHOOK_URL }}
         with:


.github/workflows/discord-issue-notification.yml
```diff
@@ -0,0 +1,128 @@
+name: Discord Issue Thread Notification
+
+on:
+  issues:
+    types: [labeled, opened]
+
+jobs:
+  notify-discord:
+    runs-on: ubuntu-latest
```
antedotee: I feel like this should be 22.04 as that is the version we use in most of our workflows.

.github/workflows/discord-issue-notification.yml
```diff
@@ -0,0 +1,128 @@
...
+                    console.log('❌ Discord API error:', res.statusCode, data);
+                    reject(new Error(`Discord API returned ${res.statusCode}`));
+                  }
+                });
+              });
+              
+              req.on('error', reject);
+              req.write(JSON.stringify(payload));
+              req.end();
+            });
```
antedotee: Add a newline at the end to remove this minus sign.

Ankitsinghsisodya:
Sorry @pabo99 sir, I noticed that emojis weren’t used in most parts of the project, so I removed them. However, later I saw that emojis were actually present in the get_user_problems.py file.

Since the last commit was only about removing emojis, I thought that deleting it would prevent the workflow from running again. That’s why I tried to delete the last commit.

I’m really, really sorry, sir. This is the first time something like this has happened with me. Sorry, sir.

.github/workflows/discord-issue-notification.yml
```diff
@@ -0,0 +1,134 @@
...
+        with:
+          script: |
+            const https = require('https');
+            
+            
+            const issue = context.payload.issue;
+            const label = '${{ steps.check-label.outputs.matched_label }}';
+            
+            // Truncate description
+            let description = issue.body || 'No description provided';
```
pabo99, line 161: Is it possible to sanitize the content of the description before truncating it? For example, removing images, code blocks, or converting links to text

.github/workflows/discord-issue-notification.yml
```diff
@@ -0,0 +1,134 @@
...
+            // Create thread name (max 100 chars)
+            let threadName = `Issue #${issue.number}: ${issue.title}`;
+            if (threadName.length > 100) {
+              threadName = threadName.substring(0, 97).replace(/[\uD800-\uDBFF]$/, '') + '...';
+            }
+            
+            const payload = {
+              thread_name: threadName,
+              embeds: [{
+                title: `🆕 Issue #${issue.number}: ${issue.title}`,
```
pabo99: Is there a way to display information from the label? Maybe an emoji for each label could help avoid making the title larger

.github/workflows/discord-issue-notification.yml
```diff
@@ -0,0 +1,134 @@
...
+                ],
+                footer: {
+                  text: 'omegaUp Issues'
+                },
+                timestamp: new Date().toISOString()
+              }]
+            };
+            
+            const webhookUrlString = process.env.DISCORD_WEBHOOK_URL;
+            if (!webhookUrlString) {
```
pabo99, line 234: Is it possible to add a validation for the URL format? According to the [documentation](https://discord.com/developers/docs/reference#api-reference-base-url), it must start with  `https://discord.com/api/webhooks/`

.github/workflows/discord-issue-notification.yml
```diff
@@ -0,0 +1,134 @@
...
+                  }
+                });
+              });
+              
+              req.on('error', reject);
+              req.write(JSON.stringify(payload));
+              req.end();
+            });
+            
+
```
pabo99: There should only be one blank line here, but without spaces so that this symbol stops appearing

Ankitsinghsisodya: okey

Ankitsinghsisodya:
@pabo99 Thank you for the suggestions! I have updated the workflow to address them:

* **Sanitization:** I added regex logic to strip images, code blocks, HTML tags, and raw URLs from the description before sending it to Discord.
* **Visuals:** I implemented an emoji mapping system so the thread title and embed now show a relevant emoji based on the label (e.g., ☀️ for GSoC, 👋 for good first issue).
* **Safety:** Added truncation checks to ensure the content stays within Discord's API character limits.

Let me know if this looks correct to you.

Ankitsinghsisodya:
@pabo99 

`Error response from daemon: failed to mkdir /var/lib/docker/volumes/omegaup_omegaupdata/_data/problems.git: mkdir /var/lib/docker/volumes/omegaup_omegaupdata/_data/problems.git: file exists`

I am not able to fix this issue. Can you guide me sir? Why is this happening?

pabo99:
> @pabo99
> 
> `Error response from daemon: failed to mkdir /var/lib/docker/volumes/omegaup_omegaupdata/_data/problems.git: mkdir /var/lib/docker/volumes/omegaup_omegaupdata/_data/problems.git: file exists`
> 
> I am not able to fix this issue. Can you guide me sir? Why is this happening?

This is an intermitent issue that sometimes occurs in our workflows, and unfortunately we haven’t been able to fully fix the root cause yet. The good news is that it doesn’t affect the workflow overall — simply re-running the job that failed is enough to continue.

We already have this problem tracked in the issue #7812, but the attempts we’ve made so far haven’t completely resolved it.

Ankitsinghsisodya:
@pabo99
Whenever you get a moment, could you please review this PR, sir? Thank you!

Ankitsinghsisodya:
Hi @pabo99, thanks again for the clarification regarding the intermittent CI/Docker error—I re-ran the jobs and the checks have passed successfully.

I’ve finalized the changes based on your previous feedback, specifically adding the sanitization regex, emoji mapping, and truncation logic. The recent commits were just to polish the whitespace and ensure the label logic matches the new requirements.

Could you please take a look when you have a moment? Thank you!

Ankitsinghsisodya:
@pabo99, I believe this is the right time to integrate this feature into omegaUp, especially with so many new contributors joining recently.

pabo99:
At the moment, I haven’t been able to reach the admin of our Discord server, which is preventing us from creating the webhook URL 😞

Ankitsinghsisodya:
Thanks for the update😅, @pabo99 sir. I understand, these administrative bottlenecks happen! No rush—just let me know whenever you're able to get a hold of them.

antedotee:
Has this change been integrated completely😅? I am unable to see the notifications in the discord channel?

pabo99:
Yes, the workflow is already running. The issue is that it’s throwing an error because threads can only be created in Forum-type channels, and the _**dev-training**_ channel is a regular channel. I also checked with the Discord server admin, and it’s not possible to change the channel type. So we’ll need to look for an alternative

antedotee:
Maybe we can create a seperate channel for that😅

pabo99:
Great idea! Can you help me come up with a channel name so we can request it quickly?

antedotee:
Maybe something as issue-tracker-channel for the channel name? What do you think?

antedotee:
Also, I think we need to think about using discord as a second place for conversation because it may lead to a lot of hassle rather than improving the experience, because as a maintainer, you may have to see the same issue conversation here on github as well as on discord. Maybe we should see and take reference from good orgs who utilise discord for their issue and PR tracking.🤔

pabo99:
Ok, good point. Still, let me give you some context. Discord is not meant to replace GitHub as the communication channel for issues, it’s not intended to be an issue tracker. The idea was to notify all candidates when new issues with the “GSoC” or “Good first issue” labels are created, so they can assign them, which would speed up the assignment process. The decision to use threads was so they could resolve any questions before assigning the issue.

What you mentioned is indeed a valid concern: the conversation could move to Discord and we might lose important information from what should be our source of truth (GitHub).

That said, I’m open to modifying the workflow so it sends a regular message (not a thread), and the interested candidate(s) can start the discussions on GitHub.

What are your thoughts, @Ankitsinghsisodya?

Ankitsinghsisodya:
Hi @pabo99 and @antedotee,

Thank you both for bringing this up and thinking it through! I completely agree with the concern about fragmenting the conversation. Keeping GitHub as the single source of truth for issue tracking and discussions is definitely the best approach.

Modifying the workflow to send a standard rich embed message (without starting a thread) sounds like the perfect solution. It will still achieve the original goal of quickly notifying candidates about new "GSoC" or "good first issue" tasks, while naturally directing them back to GitHub to ask questions and claim the issue.

Since this PR is already merged, would you like me to open a new PR to implement this change? It should be a quick update to remove the thread-creation parameters and send the notification as a regular message to the existing channel. Let me know!

pabo99:
I’d appreciate it if you could create a new PR to implement the change. There was also a pending issue with case sensitivity for “Good first issues,” in case you can include it in the same change or, if you prefer, open two separate PRs is okay.

Thanks in advance to both of you for your valuable help, @Ankitsinghsisodya @antedotee!

PR #8584, opened 2025-12-05
Add DAO Base and VO files for Problem_Bookmarks table

vijay-x-dev:
# Description

1. Updated `dao_schema.sql`
2. Added the corresponding DAO Base and VO files for the `Problem_Bookmarks` table

Part of: #8159 

# Checklist:


vijay-x-dev:
> Ok, I understand now. The reason you’re getting this error is the following:
> 
> It seems that the `schema.sql` file has more changes that are not being added to `dao_schema.sql`.
> 
> The most recommended approach in this case is to copy all the contents of the `schema.sql` file into `dao_schema.sql`. It doesn’t matter if the other changes are not related to the feature you’re implementing. That shouldn’t cause any issues.

Thanks for the guidance! I investigated further and found the root cause. The dao_schema.sql file was missing the following index from the User_Rank table:

****KEY `idx_user_rank_order` (`author_ranking`, `user_id`),****

After adding this missing line, the files match correctly.

pabo99 (approved):
Thank you, @k-vijay-05 !

vijay-x-dev:
@pabo99  It’s great to see you’ve already raised a PR for the automation to copy schema.sql into dao_schema.sql !

pabo99:
Thanks to you for raising the question about whether there was a tool to solve this — we almost had it ready! 

I’m just waiting for feedback now. Maybe there’s a reason why it wasn’t being done automatically

Issue #8582, opened 2025-12-03
[BUG] Timezone-dependent test failure in ManageSchools.test.ts: graduation_date differs in non-UTC environments


Ankitsinghsisodya:
## Current Behavior

When running `npm test` locally in a non-UTC timezone (e.g., IST), the test `ManageSchools.vue › Should emit user update schools` in `ManageSchools.test.ts` fails due to a mismatch in the `graduation_date` field. The test expects `2010-10-10T00:00:00.000Z` but receives `2010-10-09T18:30:00.000Z`, which is a timezone offset issue. The test passes when run with `TZ=UTC npm test`.

## Expected Behavior

Tests should pass regardless of the local timezone. The `graduation_date` value should be consistent and not depend on the developer's environment.

## Steps to Reproduce

1. Clone the repository.
2. Ensure your system timezone is not UTC (e.g., set to IST).
3. Run `npm test`.
4. Observe the failure in `ManageSchools.vue › Should emit user update schools`.

## Relevant details about the environment in which you encountered this bug:

* Browser name (e.g., Chrome 39, node.js 5.4): Node.js (version as per project requirements)
* Operating System and version (desktop or mobile): macOS (IST timezone)
* Link to the site where this issue occurs: N/A (local development)
* Screenshot:  

![Image](https://github.com/user-attachments/assets/59a4aac5-6098-4997-9b58-8d6aff917581)

I want to work on this issue.

pabo99:
Excellent! Go ahead, @Ankitsinghsisodya !

Ankitsinghsisodya:
@pabo99
I have submitted a Pull Request for this issue. I look forward to your feedback.

PR #8587, opened 2025-12-05
set timezone to UTC for Jest tests

Ankitsinghsisodya:
# Description

Updated the `test` script in `package.json` to enforce `TZ=UTC` when running Jest.

This fixes a failure in `ManageSchools.vue › Should emit user update schools` where the `graduation_date` was mismatching due to local timezone offsets (e.g., IST vs UTC). By forcing UTC globally for the test runner, we ensure `Date` objects are consistent across all developer environments and match the CI behavior.

Fixes: #8582 

# Comments

I encountered the issue locally where `ManageSchools.test.ts` expected `2010-10-10T00:00:00.000Z` but received a date offset by my local timezone (IST).

I chose to fix this via the `package.json` script rather than modifying the test file. This prevents similar "works on my machine" issues for other tests in the future by standardizing the test environment to UTC.

# Checklist:

### DIFF
--- a/frontend/www/docs/How-to-use-Cypress-in-omegaUp.md
+++ b/frontend/www/docs/How-to-use-Cypress-in-omegaUp.md
@@ -1,11 +1,14 @@
 ## Preliminary Steps
+
 You need to have `node` installed to use `cypress`:
+
 ```bash
 sudo apt-get install nodejs
 sudo apt-get install npm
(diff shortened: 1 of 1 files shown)


pabo99 (approved):
Great! Thanks, @Ankitsinghsisodya!

Issue #8536, opened 2025-11-19
[FEATURE] Modernize Contest List with Card Grid Layout


rahmanfaizur:
## Context
The current contest list uses a horizontal, row-based layout that doesn’t fully utilize available space on different screen sizes.  
Additionally, the **"Show more contests"** button appears even when there are only a few contests, which can be streamlined.

### 1. Card-Based Grid Layout
Introduce a responsive grid for contests:
- **Desktop:** 3 cards per row  
- **Tablet:** 2 cards per row  
- **Mobile:** 1 card per row  

**Card Layout:**  
Header (title + badge), body (organizer, date, duration, participants), footer (Enter, Details, Scoreboard).  
Subtle hover elevation for clearer interaction feedback.

### 2. Smart Pagination
Hide the “Show more contests” button when the contest count is below the visible threshold for each device size.

## Visual Comparison

**Before:**  
(Table-style layout)

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/e63f0d5c-1ffa-4065-891a-128fc925959d" />
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/6432ffcb-9de3-4f83-ad54-13b81280d960" />

**After:**  
(Grid layout with responsive, spacious cards)

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/181f9f06-85db-4f8e-b976-79415fe3bf9a" />
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/03fdcfe4-c667-4d80-8e6a-eb1406d99497" />

rahmanfaizur:
Hi @pabo99 ,
I’ve created this feature request with a proposed layout update.
I’d love to pick this up if it aligns with the project direction.
Open to any feedback or adjustments—would appreciate your thoughts whenever you get time!

pabo99:
I really like this change, @rahmanfaizur! Thanks for the mockups!

I feel the screen space could be used more effectively by utilizing the cards in the way you propose.

In fact, I would suggest that the three contest lists (current, past, and future) appear on separate rows so they are all visible at once. This way, we could avoid using the left sidebar, which also takes up space.

Each row could scroll horizontally, similar to streaming platforms, with a limit on the number of items shown. Additionally, there could be a “See more” or “View all” button.

What do you think of this proposal @tvanessa, @heduenas?

shubham220420:
[...]

PR #8560, opened 2025-11-28
Implement modernized contest list layout and UI updates

rahmanfaizur:
# Modernize Contest List UI

This PR updates the Contest List interface with a more modern, responsive, and user-friendly layout.

## ✨ Key Changes
- Replaced the old tabbed layout with **horizontally scrollable sections** for Current, Future, and Past contests.
- Added a **“View All” grid page** with responsive 3/2/1-column cards.
- Added **300ms search debounce** for smoother performance.
- Updated `ContestCard` styling and improved spacing/visual clarity.
- Improved store logic for resetting lists during search/filter updates.

## 📸 Screenshots

Screenshot from 2025-11-29 02-40-27
Screenshot from 2025-11-29 02-40-33
Screenshot from 2025-11-29 02-40-41

## ✔️ Checklist

- [x] Code follows the [omegaUp Coding Guidelines](https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/Coding-guidelines.md)
- [x] Linting and type checks pass
- [x] Manual testing completed (navigation, search, scrolling, grid view)
---
Fixes: #8536

PR #8592, opened 2025-12-05
Add ContestCard component, stories, and tests

rahmanfaizur:
# Description

Implemented the new [ContestCard](cci:2://file:///home/faizur/Documents/omega/omega/omegaup/frontend/www/js/omegaup/components/arena/ContestCard.vue:144:0-177:1) component to replace the legacy contest display.
- Created [ContestCard.vue](cci:7://file:///home/faizur/Documents/omega/omega/omegaup/frontend/www/js/omegaup/components/arena/ContestCard.vue:0:0-0:0) with a modern design.
- Added [ContestCard.stories.ts](cci:7://file:///home/faizur/Documents/omega/omega/omegaup/frontend/www/js/omegaup/components/arena/ContestCard.stories.ts:0:0-0:0) for Storybook support.
- Added unit tests in [ContestCard.test.ts](cci:7://file:///home/faizur/Documents/omega/omega/omegaup/frontend/www/js/omegaup/components/arena/ContestCard.test.ts:0:0-0:0).

# Checklist:

Fixes: #8536
Replaces: #8560


frontend/www/js/omegaup/components/arena/ContestCard.vue
```diff
@@ -1,123 +1,127 @@
...
-                ui.formatString(T.contestDuration, {
-                  duration: contestDuration,
-                })
-              }}
-            </p>
-          </b-card-text>
-        </b-col>
-        <b-col
-          class="col-md-4 col-sm-12 p-1 text-center d-flex justify-content-center"
+  <b-card class="mb-2 h-100 shadow-sm contest-card-vertical">
```
pabo99: I forgot to mention something in the suggestions:

Could you rename this component to `ContestCardv2.vue`? This way we won’t break the UI when the PR gets approved and merged into the main branch.

It looks like the test works fine for both versions of `ContestCard`, so you can leave that one exactly as it is.

frontend/www/js/omegaup/components/arena/ContestCard.vue
```diff
@@ -180,19 +184,22 @@ export default class ContestCard extends Vue {
 
 <style lang="scss" scoped>
 @import '../../../../sass/main.scss';
+.contest-card-vertical {
+  transition: transform 0.2s ease-in-out, box-shadow 0.2s ease-in-out;
+  border: 1px solid rgba(0, 0, 0, 0.125);
```
pabo99: Would it be possible to add this color and the one on line 193 as variables?

pabo99 (commented):
Just a couple of suggestions and then we’ll merge to move forward with the rest of the changes

rahmanfaizur:
Thanks for the review, @pabo99 ! I’ve resolved the suggested changes — please let me know if anything is missing.

pabo99 (commented):
Thank you for applying the suggested changes. I’ll need a few additional adjustments.

In the stories, could you add the three variants:
- **Current**
- **Future** (I think Upcoming sounds better, but we can create a separate issue to fix that later)
- **Past**

The goal is to check the behavior of each card, since each one has different elements.

In particular, I noticed something odd in this case:

Image

Here, the `Enter` button is shown twice. Also, in this new type of card I’m not too fond of having the dropdown — could you change it so that both buttons are always visible instead?

rahmanfaizur:
@pabo99 I've applied the requested changes. Let me know if any further changes are needed!

rahmanfaizur:
@pabo99 Just a gentle ping — I’ve addressed the feedback and requested a review. 
Whenever you get a chance, I’d appreciate another look. Thanks!

pabo99 (approved):
Thank you, @rahmanfaizur!

PR #8594, opened 2025-12-05
Added which labels to exempt from stale automation

antedotee:
I have added the labels which should be exempted from stale automation, both from issues as well as PRs.


antedotee:
@pabo99 Please have a look when you get time.

Issue #8598, opened 2025-12-06
[UI/UX] Improve Navbar: hover-effects,dropdown etc


PrasannaMishra001:
### Description
The current nav-bar and styles could be improved to provide a responsive UI.Like dropdowns require a click to open and the profile completion percentage is not explicitly shown textually.

### Proposed Changes
1. Enable dropdown menus to open automatically on hover for desktop users, improving navigation speed.
2. Display the exact profile completion percentage text next to the "My Profile" header.
- Add effects to dropdowns and hover states.
- Add effect to the profile picture and logo.
- Style the logout button 
4.  consistency whether logged in or not.

### Current Behavior
* Dropdowns require a click.
* Profile percentage is hidden (only a bar is shown).
* Visuals are flat opaque colors.

**Before:**

https://github.com/user-attachments/assets/fdfe9a75-9798-4d4a-9dd3-1c744e1e328a

PrasannaMishra001:
@pabo99 these were some small UI improvements I worked on for the navbar.  
If this direction looks good to u, I would love to continue contributing to other ui parts across the site as well.
I'm also open to making meaningful backend contributions if you suggest areas where help is needed.

PrasannaMishra001:
@pabo99 this is with reference to #7949 ,I checked the streak work in the repo and right now streaks only exist as badges like 7,15, 30 days which are calculated using SQL grouping.There is no live user streak counter and also no active Streak controller DAO API or frontend UI for showing current streak as i searched.
So badge based streaks are done but real time ongoing streak tracking and ui is missing like showing in navbar will be nice.

pabo99:
> [@pabo99](https://github.com/pabo99) this is with reference to [#7949](https://github.com/omegaup/omegaup/issues/7949) ,I checked the streak work in the repo and right now streaks only exist as badges like 7,15, 30 days which are calculated using SQL grouping.There is no live user streak counter and also no active Streak controller DAO API or frontend UI for showing current streak as i searched. So badge based streaks are done but real time ongoing streak tracking and ui is missing like showing in navbar will be nice.

That would be great. But unfortunately, for some reason this project was decided to be postponed. 

Let me check if it’s feasible to resume it; that could be continued in the issue you mentioned. 

For now, let’s leave it on stand by

PrasannaMishra001:
pr will be updated to close this issue soon today.

PR #8599, opened 2025-12-06
Improve Navbar hover effects,dropdown behavior,profile progress display

PrasannaMishra001:
# Description

This PR updates the navbar with a few small ui improvements to make it feel smoother and more modern. I added the numeric profile completion percentage inside the progress bar,fixed the alignment by making the container `position-relative`, and added hover effects to the logo,user avatar,nav links and logout button.The dropdown menus now have a subtle translucent background, and on desktop they open when hovered.I updated in both `Navbar.vue` and `NavbarItems.vue` to keep the look consistent whether the user is logged in or not.No logic or functionality will be affected.

Fixes: #8598

https://github.com/user-attachments/assets/77f42907-bce3-40c1-97c4-b337ac0e84f2

# Checklist:

- [x] The change is UI-only and does not exceed 200 lines.

### DIFF
--- a/frontend/www/js/omegaup/components/common/Navbar.vue
+++ b/frontend/www/js/omegaup/components/common/Navbar.vue
@@ -543,11 +543,11 @@ nav.navbar {
 }
 
 @media (min-width: 992px) {
-  .navbar-nav .dropdown {
+  .dropdown {
     position: relative;
   }
 
-  .navbar-nav .dropdown::before {
+  .dropdown::before {
     content: '';
     position: absolute;
     top: 100%;
@@ -556,20 +556,22 @@ nav.navbar {
     width: 100%;
   }
 
-  .navbar-nav .dropdown:not(.show):hover > .dropdown-menu {
+  .navbar-nav:not(:has(.dropdown.show)) .dropdown:hover > .dropdown-menu {
     display: block;
   }
 
-  .navbar-nav > .nav-item.dropdown > .dropdown-toggle {
-    cursor: default;
+  .navbar-right:not(:has(.dropdown.show)) .dropdown:hover > .dropdown-menu {
+    display: block;
   }
 
-  .navbar-collapse:has(.navbar-nav .dropdown:hover) .dropdown-menu.show {
-    display: none !important;
-  }
-
-  .navbar-nav .dropdown:hover > .dropdown-menu.show {
+  .dropdown.show > .dropdown-menu {
     display: block !important;
+  }
+
+  .navbar-collapse:has(.dropdown.show)
+    .dropdown:not(.show):hover
+    > .dropdown-menu {
+    display: none !important;
   }
 }
 


frontend/www/js/omegaup/components/common/Navbar.vue
```diff
@@ -484,13 +500,18 @@ nav.navbar {
     overflow-y: auto;
     max-height: 75vh;
     scrollbar-width: none;
+    background-color: rgba(255, 255, 255, 0.95);
```
pabo99: Can you add this and the other colors used as variables?
In this same file you can find some examples

PrasannaMishra001:
I’ve attached a short video demonstrating the updated navbar behavior.
It shows hover-based dropdown opening on desktop,automatic closing when moving to another menu, and the profile completion percentage displayed inside the progress bar.
Logout button hover styling and notification overflow behavior are also visible.

https://github.com/user-attachments/assets/ea749f7a-cdbd-4c9d-9748-414ba3fd83f6


The soln uses event listeners (mouseenter &mouseleave) scoped to desktop screens only, without modifying existing logic.
The aria-expanded state is updated manually to keep accessibility in sync with visual state, and the notification overflow was fixed using max-height instead of fixed widths.

pabo99 (commented):
Some notes:

- The behavior of the options in the right-hand menu should be consistent with those on the left-hand side to avoid the issue I show in the screencast:

https://github.com/user-attachments/assets/4490e2e3-f821-49e8-ac28-f3a92bc965bc

- It would be great if the options in the navbar had a hover effect


- Some Cypress tests were affected by the changes made in this PR. Could you check how we might fix them? If it’s too complicated, let me know and I’ll help review and fixthem

frontend/www/js/omegaup/components/common/NavbarItems.vue
```diff
@@ -247,5 +247,42 @@ export default class NavbarItems extends Vue {
...
+    const dropdowns = this.$el.querySelectorAll<HTMLElement>(
+      '.nav-item.dropdown',
+    );
+
+    dropdowns.forEach((dropdown) => {
+      const enter = (dropdown as any)._hoverEnter;
+      if (enter) dropdown.removeEventListener('mouseenter', enter);
+      delete (dropdown as any)._hoverEnter;
+    });
+  }
```
pabo99: This implementation has some issues:

- It only handles opening dropdowns (`mouseenter`) but doesn't close them (no `mouseleave` handler), so menus stay open after hovering
- When you hover over multiple dropdowns, they all stay open - there's no logic to close others when opening a new one
- It manually adds/removes Bootstrap classes (`.show`) which can conflict with Bootstrap's own click handlers
- Requires storing event handler references on DOM elements (`_hoverEnter`)  for cleanup, which is not ideal

What do you think about removing this block of code and using a different approach based on a pure CSS? It could be simpler for the following reasons:

- Native `:hover` pseudo-class handles both opening AND closing
- No manual event listener lifecycle management
- Browser handles hover state natively - better performance
- Works alongside Bootstrap's click behavior without conflicts

You only need to add something like:
```css
@media (min-width: 992px) {
  .navbar-nav .dropdown:not(.show):hover > .dropdown-menu {
    display: block;
  }
}
```

This would go in `Navbar.vue` instead of this component because Navbar is the parent component that encompasses both left-side items and right-side dropdowns. This way the CSS applies to all dropdowns consistently across the entire navbar.

Bootstrap's click behavior would still work normally.

What do you think?

PrasannaMishra001:
The issue was caused by Bootstrap’s click based dropdown (.show) coexisting with CSS hover behavior,which allowed one menu to remain open while hovering others.I am updating the implementation to make desktop dropdowns hover only by disabling click-based toggling via CSS on large screens,while keeping Bootstrap’s click behavior intact for mobile.This ensures only one dropdown is visible at a time, avoids JS and keeps consistent  behavior across the navbar.
Please let me know if this approach looks good to you @pabo99 .

PrasannaMishra001:
@pabo99 i looked at your suggestion  so i implemented hover plus click behavior similar to the w3schools example and spent some time exploring different approaches around it.

The current solution keeps bootstrap fully in control of click behavior and layers css hover on top in a safe way.Hover shows dropdowns temporarily on desktop while click toggles them open and closed as usual.When a dropdown is clicked open hover is automatically disabled on the others so only one menu can be visible at a time.Outside click still works as expected and the small hover gap issue is handled with an invisible buffer.

Personally I still find hover only the simplest option than this approach.Let me know if you would like me to adjust anything further. It looks good to me now.

PR #8601, opened 2025-12-07
implemented keep-alive to preserve component state when switching tabs

### LINKED ISSUE #7566: [BUG] We don't keep the `dynamic-components` alive!
## Current Behavior
When we edit a contest or a problem, we don't keep the tabs alive.
- We edit the contest in the `edit` tab.
- We go to any other tab without saving it.
- We go back to the `edit problem` tab to find that all unsaved changes have been reverted.
## Expected Behavior
We should keep the tabs alive.

In fact, Vue has a special built-in component `keep-alive` **that allows us to conditionally cache component instances when dynamically switching between multiple components.**
We can find details [here](https://vuejs.org/guide/built-ins/keep-alive.html).

## Screencast:

https://github.com/omegaup/omegaup/assets/64671908/8be54cdd-157c-4ed2-804d-6c0dbc9f7b41

antedotee:
Implemented keep-alive to preserve component state when switching tabs in the contest, problem, and course edit pages.

I have tested it locally(Below is an example from editing a problem preserving state while switching tabs, the title and problem source content didnt change):
image

Fixes #7566

### DIFF
--- a/frontend/www/js/omegaup/components/contest/Edit.vue
+++ b/frontend/www/js/omegaup/components/contest/Edit.vue
@@ -125,7 +125,11 @@
 
     <keep-alive>
       <div class="tab-content mt-2">
-        <div v-show="showTab === 'new_form'" class="tab-pane" :class="{ active: showTab === 'new_form' }">
+        <div
+          v-show="showTab === 'new_form'"
+          class="tab-pane"
+          :class="{ active: showTab === 'new_form' }"
+        >
           <omegaup-contest-new-form
             :admission-mode="details.admission_mode"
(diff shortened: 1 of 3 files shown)


pabo99 (commented):
The change looks good. I tested it in a couple of places and it seems to be working correctly.

However, there was one place where elements were moved. In the screencast, you can see how the content from `Add Admins` in a course was copied to all the other tabs:

https://github.com/user-attachments/assets/bd42fa49-a5c2-4818-bc1a-75e1eb511dcf

Could you check what's causing this behavior?

PR #8603, opened 2025-12-08
Backend implementation for problem bookmarking functionality

vijay-x-dev:
# Description

This PR adds the backend implementation for the problem bookmarking feature.

- **ProblemBookmark Controller**: Four API endpoints (create, delete, exists, list) 
-  **Backend Logic for Profile Display**: Cache invalidation and integration to show bookmarked problems in user profile

Part of : #8159 

# Checklist:


frontend/server/src/Controllers/ProblemBookmark.php
```diff
@@ -0,0 +1,190 @@
...
+namespace OmegaUp\Controllers;
+
+/**
+ * ProblemBookmarkController
+ *
+ * @psalm-type ProblemListItem=array{accepted: int, alias: string, can_be_removed?: bool, difficulty: float|null, difficulty_histogram: list<int>, points: float, problem_id: int, quality: float|null, quality_histogram: list<int>, quality_seal: bool, ratio: float, score: float, submissions: int, tags: list<array{name: string, source: string}>, title: string, visibility: int}
+ * @psalm-type BookmarkCreateResponse=array{success: bool}
+ * @psalm-type BookmarkDeleteResponse=array{success: bool}
+ * @psalm-type BookmarkExistsResponse=array{bookmarked: bool}
+ * @psalm-type BookmarkListResponse=array{problems: list<ProblemListItem>, total: int}
```
pabo99: You’re not using these anywhere, so they can be removed. 

```suggestion
```
Maybe you added them because they’ll be used in the upcoming UI changes, but don’t worry, they’re generated automatically

frontend/server/src/Controllers/ProblemBookmark.php
```diff
@@ -0,0 +1,190 @@
...
+        // Check if already bookmarked
+        $currentUserId = $r->identity->identity_id;
+        if (
+            \OmegaUp\DAO\ProblemBookmarks::isProblemBookmarked(
+                $currentUserId,
+                $targetProblem->problem_id
+            )
+        ) {
+            // Already bookmarked, return success
+            return ['success' => true];
```
pabo99: For consistency with the other endpoints, can you use here:
```suggestion
            return ['status' => 'ok'];
```
?

frontend/server/src/Controllers/ProblemBookmark.php
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        // Create bookmark
+        $newBookmark = new \OmegaUp\DAO\VO\ProblemBookmarks();
+        $newBookmark->identity_id = $currentUserId;
+        $newBookmark->problem_id = $targetProblem->problem_id;
+        $newBookmark->created_at = new \OmegaUp\Timestamp(\OmegaUp\Time::get());
+
+        \OmegaUp\DAO\ProblemBookmarks::create($newBookmark);
```
pabo99: If you’re not going to use it later, could you create the object inline?

```suggestion
        \OmegaUp\DAO\ProblemBookmarks::create(
            new \OmegaUp\DAO\VO\ProblemBookmarks([
                'identity_id' => $currentUserId,
                'problem_id' => $targetProblem->problem_id,
            ])
        );
```

frontend/server/src/Controllers/ProblemBookmark.php
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    /**
+     * Remove a bookmark for the current user
+     *
+     * @omegaup-request-param string $problem_alias
+     *
+     * @return array{success: bool}
+     */
+    public static function apiDelete(\OmegaUp\Request $r): array {
```
pabo99: Now that I’m noticing it, this endpoint receives the same parameter as `apiCreate`, and it does exactly the opposite without any extra validation and without throwing any exception.

What do you think about just creating` apiToggle` instead, so it can identify when to create the problem bookmark and when to remove it?

frontend/server/src/Controllers/ProblemBookmark.php
```diff
@@ -0,0 +1,190 @@
...
+        );
+
+        // Validate that the problem exists
+        $targetProblem = \OmegaUp\DAO\Problems::getByAlias($problemAlias);
+        if (is_null($targetProblem) || is_null($targetProblem->problem_id)) {
+            throw new \OmegaUp\Exceptions\NotFoundException('problemNotFound');
+        }
+
+        $currentUserId = $r->identity->identity_id;
+        $isBookmarked = \OmegaUp\DAO\ProblemBookmarks::isProblemBookmarked(
```
pabo99: I see that `isProblemBookmarked`only calls the `existsByPK` function. What do you think about calling it directly:
```suggestion
        $isBookmarked = \OmegaUp\DAO\ProblemBookmarks::existsByPK(
```
?

That way you can remove the intermediate function

frontend/server/src/Controllers/ProblemBookmark.php
```diff
@@ -0,0 +1,190 @@
...
+                'tags',
+                'title',
+                'visibility',
+                'quality_seal',
+            ]);
+            $problemData['tags'] = $shouldHideTags ? [] : \OmegaUp\DAO\Problems::getTagsForProblem(
+                $problemItem,
+                public: false,
+                showUserTags: $problemItem->allow_user_add_tags
+            );
```
pabo99: Is it really necessary to send all the problem information to the client? If you only plan to display the problem titles, you can even just fetch the `alias` and `title` from the query in `getAllBookmarkedProblems` function. 

That way we also avoid recursively executing the query in `getTagsForProblem` function

frontend/server/src/Controllers/User.php
```diff
@@ -2186,6 +2186,28 @@ private static function getCreatedProblems(int $identityId): array {
...
 
+    /**
+     * @return list<Problem>
+     */
+    private static function getBookmarkedProblems(int $identityId): array {
+        $bookmarkedProblemsList = \OmegaUp\DAO\ProblemBookmarks::getAllBookmarkedProblems(
+            $identityId
+        );
+
+        $relevantColumns = ['title', 'alias', 'submissions', 'accepted', 'difficulty', 'quality_seal'];
```
pabo99: Are all these fields really necessary? Preferably, only send the ones that will be displayed in the UI

frontend/server/src/Controllers/User.php
```diff
@@ -2186,6 +2186,28 @@ private static function getCreatedProblems(int $identityId): array {
...
+        /** @var list<Problem> */
+        $filteredProblems = [];
+        foreach ($bookmarkedProblemsList as $problem) {
+            if (\OmegaUp\DAO\Problems::isVisible($problem)) {
```
pabo99: Oh, I see why you’re using the entire `Problems` object: to send it to the `isVisible` function.

Maybe you could add that filter directly in the query of `getAllBookmarkedProblems` function. That way, you also ensure that you’ll get the same list both in the endpoint and here

frontend/server/src/Controllers/ProblemBookmark.php
```diff
@@ -0,0 +1,190 @@
...
+    public static function apiList(\OmegaUp\Request $r): array {
+        $r->ensureIdentity();
+
+        // Get all bookmarked problems
+        $currentUserId = $r->identity->identity_id;
+        $bookmarkedProblems = \OmegaUp\DAO\ProblemBookmarks::getAllBookmarkedProblems(
+            $currentUserId
+        );
+
+        $totalCount = count($bookmarkedProblems);
```
pabo99: Normally, in endpoints where we send both the total number of items and the list of items, it’s because the view displays them paginated. 

If here you’re considering sending all the bookmarked problems and getting the total from counting the list, you can remove the `getBookmarkedProblemsCount` function.

frontend/server/src/Controllers/ProblemBookmark.php
vijay-x-dev: Sure

frontend/server/src/Controllers/ProblemBookmark.php
vijay-x-dev: I added those anticipating the UI work, but since they’re generated automatically, I’ll drop the manual psalm-type declarations

frontend/server/src/Controllers/ProblemBookmark.php
vijay-x-dev: Thank you for the helpful suggestion. I’ll have apiToggle check the current bookmark state, call the same create/delete helpers (create if not bookmarked, delete if it is), and return ['status' => 'ok', 'bookmarked' => bool]. Please let me know if this matches what you expect

frontend/server/src/Controllers/ProblemBookmark.php
vijay-x-dev: Oh, that was my mistake .I ’ll switch to existsByPK directly and drop the extra helper.

frontend/server/src/Controllers/ProblemBookmark.php
pabo99: Sure, it sounds great!

frontend/server/src/Controllers/ProblemBookmark.php
vijay-x-dev: Thanks for the note. I’ll remove getBookmarkedProblemsCount and now derive total from the returned list in apiList

frontend/server/src/Controllers/User.php
vijay-x-dev: I’ll move the visibility filter into getAllBookmarkedProblems (added Problems.visibility > 0 in the query) and removed the extra isVisible check in User::getBookmarkedProblems, so both places now use the same filtered list.

frontend/server/src/Controllers/User.php
vijay-x-dev: Got it. I'll keep the bookmarked problem payload to just the displayed fields

frontend/server/src/Controllers/ProblemBookmark.php
```diff
@@ -0,0 +1,151 @@
...
+            'problem_alias',
+            fn (string $alias) => \OmegaUp\Validators::alias($alias)
+        );
+
+        $targetProblem = \OmegaUp\DAO\Problems::getByAlias($problemAlias);
+        if (is_null($targetProblem) || is_null($targetProblem->problem_id)) {
+            throw new \OmegaUp\Exceptions\NotFoundException('problemNotFound');
+        }
+
+        $currentUserId = $r->identity->identity_id;
```
pabo99: For consistency, could you rename this variable to `$currentIdentityId` and apply the same change everywhere else you use this name

frontend/server/src/Controllers/ProblemBookmark.php
```diff
@@ -0,0 +1,151 @@
...
+        $bookmarkedProblems = \OmegaUp\DAO\ProblemBookmarks::getAllBookmarkedProblems(
+            $currentUserId
+        );
+
+        $relevantColumns = ['alias', 'title'];
+        $problemList = [];
+        foreach ($bookmarkedProblems as $problemItem) {
+            /** @var BookmarkListItem */
+            $problemList[] = $problemItem->asFilteredArray($relevantColumns);
+        }
```
pabo99: It seems you’re using this block of code both here and in the `getBookmarkedProblems` function. Would you consider moving it into the DAO function instead, so that from that point you only filter the fields you actually need?

frontend/server/src/Controllers/User.php
```diff
@@ -2186,6 +2187,27 @@ private static function getCreatedProblems(int $identityId): array {
...
+    private static function getBookmarkedProblems(int $identityId): array {
+        $bookmarkedProblemsList = \OmegaUp\DAO\ProblemBookmarks::getAllBookmarkedProblems(
+            $identityId
+        );
+
+        $relevantColumns = ['alias', 'title'];
+        /** @var list<BookmarkProblem> */
+        $filteredProblems = [];
+        foreach ($bookmarkedProblemsList as $problem) {
+            // Problems are filtered by visibility in the DAO query.
```
pabo99: When I suggested fetching only the relevant fields, I was referring to what comes directly from the database query. 

However, looking at it now, I think it actually makes sense to retrieve all the fields so you can build the full `Problems` object and reuse the `isVisible` function. That way, you can also remove the visibility condition from the query itself, since the function will handle it. Otherwise, we’d end up maintaining the necessary conditions in two different places instead of just one

frontend/server/src/Controllers/ProblemBookmark.php
vijay-x-dev: sure

frontend/server/src/Controllers/ProblemBookmark.php
vijay-x-dev: I'm keeping the filtering in controllers to use isVisible() with full Problems objects.

Regarding duplication: As we planned to display bookmarked problems in the profile `Problems tab alongside solved/unsolved/created problems` via `ExtraProfileDetails` (similar to the others). Since bookmarks will be part of the profile data, a separate `ProblemBookmark::apiList()` endpoint may not be needed. Should I refactor it to reuse `User::getBookmarkedProblems()` and keep the endpoint for future use, or remove it. What are your thoughts?

frontend/server/src/Controllers/User.php
vijay-x-dev: Done. Updated to retrieve all fields for full Problems objects and use isVisible(). Removed the visibility condition from the SQL query; visibility is now handled by isVisible() in the controllers.

frontend/server/src/Controllers/ProblemBookmark.php
pabo99: Ah, you can also use the `isVisible()` function from the DAO; in fact, that’s where you’re already building the `Problems` object.

Regarding the duplication, I’m referring specifically to this block:
```php
        $relevantColumns = ['alias', 'title'];
        $problemList = [];
        foreach ($bookmarkedProblems as $problemItem) {
            /** @var BookmarkListItem */
            $problemList[] = $problemItem->asFilteredArray($relevantColumns);
        }
```

You’re using it both here and in the `User` controller. What do you think about adding this logic as part of the `getAllBookmarkedProblems` function so it ends up looking like this:

```php

        $relevantColumns = ['alias', 'title'];
        $bookmarkedProblems = [];
        foreach ($resultRows as $rowData) {
            $problem = new \OmegaUp\DAO\VO\Problems($rowData);

            if (\OmegaUp\DAO\Problems::isVisible($problem)) {
                /** @var BookmarkProblem */
                $bookmarkedProblems[] = $problem->asFilteredArray($relevantColumns);
            }
        }

        return $bookmarkedProblems;
```

With this change, you could even remove the `getBookmarkedProblems` function you created in the `User` controller.

frontend/server/src/DAO/ProblemBookmarks.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+namespace OmegaUp\DAO;
+
+/**
+ * ProblemBookmarks Data Access Object (DAO).
+ *
+ * {@link \OmegaUp\DAO\VO\ProblemBookmarks}.
```
pabo99: It seems that PSALM doesn’t really like this annotation. Could you remove it?
```suggestion
```

frontend/server/src/Controllers/ProblemBookmark.php
vijay-x-dev: Oh, sorry for the confusion , now I got your point, thanks for clarifying.

frontend/server/src/Controllers/ProblemBookmark.php
```diff
@@ -0,0 +1,144 @@
+<?php
+
+namespace OmegaUp\Controllers;
+
+/**
+ * ProblemBookmarkController
+ *
+ * @psalm-type BookmarkListItem=array{alias: string, title: string}
```
pabo99: You are using the same type declaration in the `User` controller, but under a different name. Is it possible to unify them?

Also, you could declare it in the DAO and reuse it in both places where it is currently defined

vijay-x-dev:
Hi @pabo99 , could you please give this a final review when you get a chance. Once it’s done, I’ll proceed with the frontend changes. Thanks!

pabo99 (commented):
I forgot to submit the review. Sorry.

One last comment and we’re done.

frontend/server/src/Controllers/ProblemBookmark.php
vijay-x-dev: Done. Unified the type declarations

vijay-x-dev:
Sorry @pabo99 , I didn't notice those test failures in the CI. Now it's fixed, I've added the `bookmarkedProblems` property to the test data in all three test files (MainInfo.test.ts, SidebarMainInfo.test.ts, and ViewProfile.test.ts).

Issue #8608, opened 2025-12-09
[FEATURE] Reduce code duplication in CI by creating a reusable 'prepare-env' action


Ankitsinghsisodya:
As a developer, I want to refactor the repeated Docker setup steps in ci.yml into a reusable Composite Action so that I can reduce code duplication and simplify workflow maintenance.

## Context

**How has this issue affected you?**
While reviewing the `CI.yml` workflow, I noticed that the `cypress`, `python`, and `teaching-assistant` jobs all contain identical blocks of code (approximately 80-100 lines each) to set up the testing environment. This includes steps for pulling Docker images, starting containers, configuring `/etc/hosts`, and bootstrapping the database.

This violation of the DRY (Don't Repeat Yourself) principle makes the workflow file difficult to read and risky to maintain. If a change is needed in the environment setup (e.g., changing a Docker flag or a database user), it must be manually applied in three different places, increasing the risk of configuration drift and errors.

**What is the goal to achieve?**
The goal is to extract these common steps into a local Composite Action (e.g., `.github/actions/prepare-env/action.yml`). This will allow the main `ci.yml` to call a single `uses:` line for the setup, significantly reducing line count and ensuring that all testing jobs use a consistent, single source of truth for their environment configuration.

I want to work on this issue.

pabo99:
Go ahead, @Ankitsinghsisodya!

Ankitsinghsisodya:
@pabo99 okey sir!!

PR #8609, opened 2025-12-10
feat: add reusable CI environment setup action

Ankitsinghsisodya:
# Description

Refactors the `ci.yml` workflow by extracting the repeated environment setup steps into a new reusable Composite Action located at `.github/actions/prepare-env`.

Key changes include:
- Created `.github/actions/prepare-env/action.yml` to handle Docker setup, host aliasing, database waiting, and default configuration generation.
- Updated the `cypress`, `python`, and `teaching-assistant` jobs to use this new action instead of repeating the same 80-100 lines of code.
- Added inputs to the action (`install_host_python_deps`, `extra_python_requirements`) to handle specific dependency needs for different jobs.

This reduces code duplication, adheres to the DRY principle, and simplifies future maintenance of the test environment.

Fixes: #8608

# Comments

This is a refactoring change. The actual logic for setting up the environment remains the same as before, just centralized. The `php` job was intentionally left untouched as it uses Service Containers rather than the Docker Compose setup used by the other jobs.

# Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,8 +35,8 @@ jobs:
           - 5672:5672
           - 15672:15672
         env:
-          RABBITMQ_DEFAULT_USER: omegaup
-          RABBITMQ_DEFAULT_PASS: omegaup
+          RABBITMQ_DEFAULT_USER: 'omegaup'
+          RABBITMQ_DEFAULT_PASS: 'omegaup'
 
     steps:
       - uses: actions/checkout@v4


.github/actions/prepare-env/action.yml
```diff
@@ -0,0 +1,114 @@
+name: 'Prepare Environment'
```
pabo99: Could you please remove the single quotes from the workflow file (action names, descriptions, input values like `'true'`/`'false'`, etc.) to maintain consistency with the rest of our workflows?

We typically avoid single quotes unless strictly necessary

.github/actions/prepare-env/action.yml
```diff
@@ -0,0 +1,114 @@
...
+        EOF
+
+        # Wait for common services
+        wait-for-it -t 60 mysql:13306
+        wait-for-it -t 60 gitserver:33861
+        wait-for-it -t 60 redis:6379
+
+        # Wait for mysql to be fully ready
+        for i in $(seq 30); do
+          if [[ "$(mysql -uomegaup -pomegaup --host=mysql --port=13306 --skip-column-names --batch -e 'SELECT 1;' || echo 0)" == "1" ]]; then
```
pabo99: Could you verify that we actually have access to the `omegaup` database instead of just running `SELECT 1`? 

```suggestion
          if [[ "$(mysql -uomegaup -pomegaup --host=mysql --port=13306 --skip-column-names --batch -e 'USE omegaup; SELECT 1;' || echo 0)" == "1" ]]; then
```

pabo99 (commented):
Thank you for this refactoring work, @Ankitsinghsisodya. The changes look great and follow best practices for DRY (Don't Repeat Yourself) principles.

To ensure this refactor doesn't introduce significant overhead, could you please provide a performance analysis report? it would be helpful to have a comparison of CI execution times before and after the changes

.github/actions/prepare-env/action.yml
Ankitsinghsisodya: ok sir

Ankitsinghsisodya: @pabo99 so should I remove 
```
RABBITMQ_DEFAULT_USER: 'omegaup'
RABBITMQ_DEFAULT_PASS: 'omegaup'
```
also which are present in ci.yml.

pabo99: If you didn’t modify those lines in this PR, you can leave them as they are for now

pabo99 (approved):
Great!

Issue #8548, opened 2025-11-24
It would be better to make a first interaction github workflow for newcomers


antedotee:
Whenever a newcomer is raising an issue, creating a pull request or commenting for the first time, it would be better that an automated comment regarding the discord channel, link for documentation and some other things which might be helpful for newcomers. ( I earlier saw someone asking link for discord channel, thats how I thought of this issue)
Also, this will make the experience of first time contributors much smoother.

antedotee:
@pabo99 Your thoughts on this?

Adez017:
@antedotee  @pabo99  , i would like to work on this issue . if you agree please assign this task to me.

pabo99:
Thanks, @antedotee! It's an excellent proposal!

Since you already have some other issues assigned, @Adez017 could take on this one, and you could help by reviewing the raised PR to ensure it meets the expected requirements.

What do you think about this??

antedotee:
Sure, assign the issue to him and I will review it and then propose if any changes are there to make.

pabo99:
The issue is now assigned to you, @Adez017! Feel free to ask if you have any questions

antedotee:
Also, @Adez017  make sure to use first-interactions github action from gha marketplace.
https://github.com/marketplace/actions/first-interaction

Adez017:
Sure @antedotee

Adez017:
hi @pabo99 , @antedotee  just to make sure we need a workflow that interact with the newcomers only . if that so is that mean that it should only appears for the first contribution only ??

antedotee:
Yes @Adez017.

Adez017:
ok

antedotee:
@Adez017 Any updates? If you are facing any issues, you can ask here :).

Adez017:
> [@Adez017](https://github.com/Adez017) Any updates? If you are facing any issues, you can ask here :).

sure @antedotee  , working on it

Issue #8520, opened 2025-11-10
[BUG] CSS in problem creator should be fixed


antedotee:
## Current Behavior

<img width="1035" height="728" alt="Image" src="https://github.com/user-attachments/assets/60328ab3-49c3-4bd3-962b-9615ff69af65" />

The preview block just looks a bit odd.

## Expected Behavior

I think it should be equally and similarly spaced just as the box on the right. Also, I believe it would be better if we can omit the need of save button(I think I will have to create seperate issue for that).

antedotee:
@pabo99 If the issue is worthy enough to be resolved, I would like to take this up.

viveksahu92:
I would like to work on this issue. Kindly assign it to me. I will ensure that the spacing is consistent with the box on the right and look into removing the need for the "Save" button as discussed.

pabo99:
Thanks for reporting this issue, @antedotee ! Since you already have issues assigned, this one will be assigned to @viveksahu92

antedotee:
Yes, I am happy that someone is willing to work upon this issue. In case  @viveksahu92  is unable to solve, just ask me as I solved this issue last night.
Thanks.

aaradhygaur22:
@viveksahu92 Bro can i work with you, i just want to learn

Irfan762:
Hi @antedotee  , I’d like to work on this issue. Please assign it to me if it’s available.

Irfan762:
Thanks for the update 👍
I understand the issue is currently assigned.
I’ll keep an eye on it and would be happy to help if it becomes available or if any assistance is needed.

PrinceLogic:
I want to work on this issue. 
@antedotee Can you check the pr of @viveksahu92 and if it failed then to assign me.

PR #8620, opened 2025-12-12
fixed the 403 error on size labeler PR

antedotee:
Currently, the size labeler workflow is giving 403 error as I saw on the latest PR. I have included the fix for that in this PR.

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,7 +1,7 @@
 name: PR Size Labeler
 
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, synchronize, reopened]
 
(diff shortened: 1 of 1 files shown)


antedotee:
@pabo99 Sorry for the workflow failing, I have included the fix in this PR.

antedotee:
Also, @pabo99 , there seems to be no update on the #8548 and #8520 issue. It's been a while. Should I pick them or wait more instead?

pabo99:
Could you hold off for a bit? 

One of the issues already has a PR under review, and the other is being clarified, so it makes sense to keep that status. In a couple of days, we’ll review again and see if you can pick them up

Issue #8275, opened 2025-05-16
[FEATURE] [SQL Queries Optimization] Refinar consulta que obtiene todos los registros de `Course_Clone_Log`


pabo99:
En algunas pruebas se manda llamar `Course_Clone_Log::getAll()`:

```sql
            SELECT
                `Course_Clone_Log`.`course_clone_log_id`,
                `Course_Clone_Log`.`ip`,
                `Course_Clone_Log`.`course_id`,
                `Course_Clone_Log`.`new_course_id`,
                `Course_Clone_Log`.`token_payload`,
                `Course_Clone_Log`.`timestamp`,
                `Course_Clone_Log`.`user_id`,
                `Course_Clone_Log`.`result`
            FROM
                `Course_Clone_Log`
``` 
Revisar si se puede omitir desde el análisis, o refinar una búsqueda con los datos que se esperan haber obtenido.

Ref:
https://github.com/omegaup/omegaup/blob/b59b3717cee2a550be32994122c7e4828c719f5b/frontend/tests/controllers/CourseCloneTest.php#L63

coderrDikshant:
Hey @pabo99! I’m interested in working on this issue — could you please assign it to me?

Ankitsinghsisodya:
Hi @pabo99, I see this issue was assigned a while ago but hasn't been closed. Is it still available for contribution?

pabo99:
Hello @coderrDikshant, are you still interested in resolving this issue? 

If not, the issue will be assigned to @Ankitsinghsisodya, who has shown interest in working on it

Ankitsinghsisodya:
Hi @pabo99, thanks for the update!
If @coderrDikshant is no longer available, I’m happy to take over the issue.
Please let me know if I should proceed.

pabo99:
Go ahead, @Ankitsinghsisodya ! The issue is now assigned to you!

Ankitsinghsisodya:
@pabo99
I have submitted a Pull Request for this issue. I look forward to your feedback.

PR #8622, opened 2025-12-12
Refactor CourseCloneLog to add countByCourseId method and update tests

Ankitsinghsisodya:
# Description

Replaced the inefficient usage of `Course_Clone_Log::getAll()` in `CourseCloneTest.php` with a direct SQL count query to optimize memory usage and performance.

Specific changes:
- Added `countByCourseId()` method to `CourseCloneLog.php`. This uses the existing `course_id` index for an efficient `COUNT` query.
- Updated `CourseCloneTest.php` (lines 62-67), replacing the `getAll()` + `assertArrayContainsWithPredicateExactlyOnce()` combination with `assertSame(1, CourseCloneLog::countByCourseId(...))`.

Fixes: #8275

# Comments

The previous implementation fetched all records from the `Course_Clone_Log` table into memory to verify a single insertion, which was inefficient as noted in the issue description. The new implementation pushes the counting logic to the database layer.

# Checklist:

### DIFF
--- a/.github/workflows/assign-issues.yml
+++ b/.github/workflows/assign-issues.yml
@@ -1,0 +1,53 @@
+name: "Assign Issues"
+
+on:
+    issue_comment:
+        types: [created]
+    schedule:
+        # Run daily to handle reminders and automatic unassignments
+        - cron: "0 3 * * *"
+    workflow_dispatch:
+
+jobs:
+    assign:
+        runs-on: ubuntu-22.04
+        permissions:
+            issues: write
+            pull-requests: write
+        steps:
+            - name: Assign or unassign issues
+              uses: takanome-dev/assign-issue-action@beta
+              with:
+                  github_token: ${{ secrets.GITHUB_TOKEN }}
+                  maintainers: ""
+                  self_assign_cmd: /assign
+                  self_unassign_cmd: /unassign
+                  # Unassign after 5 days if no PR is opened
+                  days_until_unassign: 7
+                  enable_auto_suggestion: true
+                  allow_self_assign_author: true
+                  enable_reminder: true
+                  reminder_days: auto
+                  reminder_comment: >
+                      Quick reminder: please open a pull request within 2.5 days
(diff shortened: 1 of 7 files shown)


frontend/server/src/DAO/CourseCloneLog.php
```diff
@@ -12,4 +12,22 @@
...
+            FROM
+                `Course_Clone_Log`
+            WHERE
+                `course_id` = ?;';
+        /** @var int */
+        $count = \OmegaUp\MySQLConnection::getInstance()->GetOne(
+            $sql,
+            [$courseId]
+        );
+        return intval($count);
```
pabo99: Here it’s not necessary to parse to int, since you’re already specifying on line 26 the type that the `GetOne` function will return

frontend/tests/controllers/CourseCloneTest.php
```diff
@@ -59,10 +59,11 @@ public function testCreateCourseClone() {
...
         $this->assertSame($courseAlias, $courseClonedData['alias']);
-        $this->assertArrayContainsWithPredicateExactlyOnce(
-            \OmegaUp\DAO\CourseCloneLog::getAll(),
-            fn (\OmegaUp\DAO\VO\CourseCloneLog $courseLog) =>
-                $courseLog->course_id === $courseData['course_id']
+        $this->assertSame(
+            1,
+            \OmegaUp\DAO\CourseCloneLog::countByCourseId(
+                $courseData['course_id']
+            )
```
pabo99, line 66: Could you revert the assertion to how it was? just replace the function you created. That way, we avoid leaving static data that could break the test if more elements are added

Ankitsinghsisodya:
@pabo99 I’ve implemented the changes based on your feedback. Kindly take a look when you get a moment.

Ankitsinghsisodya:
@pabo99 , could you please take a look at the PRs when you have some time?

pabo99:
> @pabo99 , could you please take a look at the PRs when you have some time?

It LGTM!

PR #8623, opened 2025-12-12
search now uses idx_schools_name

### LINKED ISSUE #8014: Create index for column `Name` in table `Schools` and use it in DAO

antedotee:
Optimized the query to use index instead of a full table scan. Added tests for partial prefix matches and adjusted tests to match prefix matching behaviour.

Fixes #8014


pabo99:
Could you resolve the conflict here?

Issue #8624, opened 2025-12-13
[BUG] Simultaneous addition and removal of size tags when opening a PR


antedotee:
Currently, when someone raises a PR, size tag gets added, then removed, and then once again gets added. I think this is particularly arising from the labeler.yml and label-pr.yml tag race condition, most likely.

<img width="551" height="39" alt="Image" src="https://github.com/user-attachments/assets/6b3a03d6-1d13-4d28-92d4-feabc659126b" />

Also, when someone makes a commit, ideally if the commit has size/M and earlier the size was size/M, the removal and the addition shouldn't happen.

antedotee:
I will look into it, once I get time. I will be glad if someone else picks this up to resolve it in the meantime.

Ankitsinghsisodya:
Can I work on this @antedotee ?

pabo99:
Assigned to @Ankitsinghsisodya !

Ankitsinghsisodya:
@pabo99
I have submitted a Pull Request for this issue. I look forward to your feedback.

antedotee:
Closing this issue as it is reduntant.

PR #8630, opened 2025-12-15
Fix race condition causing duplicate add/remove of size labels on PRs

Ankitsinghsisodya:
Fixed labeler race condition and size tag flickering

# Description

This PR fixes the race condition between the path-based labeler and the size labeler workflows. Previously, these two workflows ran in parallel, causing a race condition where size tags would be added, removed, and added again (flickering).

**Changes made:**
* Merged the separate `label-pr.yml` and `labeler.yml` workflows into a single sequential workflow file.
* Implemented a custom JavaScript step for size labeling that checks for idempotency (it only updates labels if the size category has actually changed).
* Added logic to auto-create size labels if they are missing from the repository, ensuring the workflow doesn't fail on fresh forks.
* Ensured `sync-labels: false` is set for the path labeler so it doesn't wipe out size labels.

Fixes: #8624

# Comments

The new workflow ensures that:
1. Labels are created (if missing).
2. Path labels are applied.
3. Size labels are calculated and applied **only if necessary**.

This eliminates the "add-remove-add" notification spam on PRs.

# Checklist:

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -197,9 +197,13 @@ Note: This PR might require additional review time due to its size.`;
             
             // Add new size label
             core.info(`Adding size label: ${newSizeLabel}`);
-            await github.rest.issues.addLabels({
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              issue_number: issueNumber,
-              labels: [newSizeLabel]
-            });
+            try {
+              await github.rest.issues.addLabels({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: issueNumber,
+                labels: [newSizeLabel]
+              });
+            } catch (error) {
+              core.warning(`Failed to add label ${newSizeLabel} to issue #${issueNumber}: ${error.message}`);
+            }


Ankitsinghsisodya:
Hey @pabo99,

I’ve updated the tests to resolve the failure we were seeing in ranking.test.ts.

Upon debugging, I found that scoreboard.start_time was being created at module load using the real system time, which conflicted with jest.useFakeTimers(). This mismatch caused mergeRankings() to compute the wrong currentDelta.

I fixed this by moving the start_time initialization so it relies on the mocked now inside the test logic. The tests are passing locally now!

antedotee:
I think this PR creates a lot of unnecessary changes, rather than solving the core bug which was originally there. I am having exams currently, I will try to look at it when I will get time.

.github/workflows/unified-labeler.yml
```diff
@@ -0,0 +1,172 @@
...
+      - name: Create size labels if missing
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const labels = [
+              { name: 'size/XS', color: '0e8a16', description: 'Very small PR (≤10 lines)' },
+              { name: 'size/S', color: '28a745', description: 'Small PR (11–50 lines)' },
+              { name: 'size/M', color: 'fbca04', description: 'Medium PR (51–100 lines)' },
+              { name: 'size/L', color: 'd93f0b', description: 'Large PR (101–200 lines)' },
+              { name: 'size/XL', color: 'b60205', description: 'Very large PR (>200 lines)' }
```
pabo99: Now that you’re unifying this workflow, I’d like to slightly adjust these values. Would it be possible to leave them like this:

```suggestion
              { name: 'size/M', color: 'fbca04', description: 'Medium PR (51–150 lines)' },
              { name: 'size/L', color: 'd93f0b', description: 'Large PR (151–250 lines)' },
              { name: 'size/XL', color: 'b60205', description: 'Very large PR (>250 lines)' }
```

This gives a bit more margin for PRs with extensive functionality.

We’ll also need to update this everywhere it’s used (including the PR template).

pabo99:
> I think this PR creates a lot of unnecessary changes, rather than solving the core bug which was originally there. I am having exams currently, I will try to look at it when I will get time.

Thank you, @antedotee!

Ankitsinghsisodya:
@pabo99, thank you for pointing this out—I’ll look into it and try to fix the issue.

Ankitsinghsisodya:
@pabo99, the issue is that pull_request_target runs the workflow defined in main, ignoring the changes in this PR. As a result, my fix hasn't actually executed yet, which explains why we are still seeing the old behavior in the logs.

pr-race
See the jobs which are executed are from the workflow defined in main.

pabo99:
@antedotee, could you help me review the changes in this PR? Is this the right approach?

antedotee:
@pabo99 , I saw the changes and I feel like these changes are making a custom github action workflow. Though this is great,  I believe, why to reinvent the wheel when we have something already working. This PR defeats the purpose of fixing the race condition which was supposed to be the problem initially. I explained the problem and the most probable solution here, https://github.com/omegaup/omegaup/issues/8671#issuecomment-3678026726
If we decide to implement custom github action workflow, then we would have to test this and then merge this. 
So, it is better to close this PR as we are fighting for an overoptimisation which have diminishing returns in my opinion.
At the end, the take is yours.

Ankitsinghsisodya:
Hi @antedotee, @pabo99,

Thanks for the review and the time spent analyzing this!

I'd like to respectfully clarify a few points about my approach:

**What my PR specifically solves:**

1. **Race condition elimination** — By merging the path labeler and size labeler into a single sequential workflow, we guarantee no parallel execution conflicts. The current setup with two separate workflows will always have the potential for race conditions.
2. **Idempotency** — My solution only updates labels when the size category _actually changes_, preventing the "add-remove-add" notification spam entirely.
3. **Auto-creation of labels** — Ensures the workflow doesn't fail on fresh forks that lack pre-configured labels.

**Regarding the simpler solution:**
The proposed fix in #8671 using `fail_if_xl: true` addresses the XL warning message issue, but it doesn't actually fix the **race condition** between the two workflows—which was the original problem (#8624 ). The flickering behavior (labels being added, removed, then added again) would persist because the path labeler and size labeler would still run in parallel.

That said, I'm absolutely open to simplifying my implementation if there are specific concerns about maintainability. Perhaps we could:

- Keep the unified workflow approach but simplify the JavaScript logic?
- Or test both approaches and compare the actual behavior?

I'd appreciate a second look with these points in mind. Happy to make adjustments based on concrete feedback!

antedotee:
I initially created the issue, thinking that the fix to this will be small. I didn't expected this to be so big. I believe the problem is not worth optimizing for, as this is how the codely action also works in other repos, also. Rest is upto Pabo to think whether we want to work upon this or not. If you ask me, no.

Ankitsinghsisodya:
@pabo99 sir, should I close this PR?

Issue #8627, opened 2025-12-14
[BUG] Optimize getContestPrintDetailsForTypeScript to use bulk SELECT (Fix N+1 Read)


Ankitsinghsisodya:
## Current Behavior

In `frontend/server/src/Controllers/Contest.php`, specifically inside the `getContestPrintDetailsForTypeScript` method (around line 773), the code iterates through a list of problems to fetch their details one by one.

**The problematic code:**
```php
foreach ($details['problems'] as $index => $problem) {
    // This executes a separate SELECT query for every single iteration
    $problemDetails = \OmegaUp\DAO\Problems::getByAlias($problem['alias']); 
    // ... processing ...
}
```

I want to work on this issue

pabo99:
Assigned

Ankitsinghsisodya:
@pabo99
I have submitted a Pull Request for this issue. I look forward to your feedback.

PR #8638, opened 2025-12-16
Optimize bulk fetching of contest problems to reduce N+1 query issues

Ankitsinghsisodya:
# Description

Optimized `getContestPrintDetailsForTypeScript` in `Contest.php` to fix an N+1 performance issue.

Previously, the code iterated through the problem list and executed a `SELECT` query for every single problem (`Problems::getByAlias`). I have refactored this to:
1. Added a new method `Problems::getByAliases` in the DAO to support bulk fetching via `WHERE alias IN (...)`.
2. Updated the controller to fetch all problem details in a single query before the loop.
3. Mapped the results by alias for O(1) access during the iteration.

This reduces the database load from **N** queries to **1** query for this operation.

Fixes: #8627

# Checklist:

### DIFF
--- a/frontend/server/src/Controllers/Notification.php
+++ b/frontend/server/src/Controllers/Notification.php
@@ -3,7 +3,7 @@
  namespace OmegaUp\Controllers;
 
 /**
- * BadgesController
+ * NotificationController
  *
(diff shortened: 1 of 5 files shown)


pabo99 (approved):
Great, thanks for fixing this!

PR #8655, opened 2025-12-18
Fix Docker race condition in CI by disabling volume copying

Ankitsinghsisodya:
# Description

This PR fixes the race condition causing Python and Cypress tests to fail with `mkdir ... file exists` errors (Issue #7812).

The issue arises because `frontend`, `gitserver`, and `grader` services all mount the shared volume `omegaupdata` at `/var/lib/omegaup`. When `docker compose up` runs, these services race to populate the volume with content from their respective images, specifically clashing when creating the `problems.git` directory.

Changes made:
1. **Disabled Volume Copying:** Updated `docker-compose.yml` to use the long syntax for volume mounting with `nocopy: true` for the `frontend`, `gitserver`, and `grader` services. This prevents Docker from attempting to copy content from the container image into the shared volume at startup.
2. **Pre-created Directories:** Updated the CI workflow to manually create the `omegaup_omegaupdata` volume and the `problems.git` directory using a temporary Alpine container. This ensures the necessary directory structure exists before the main services start, which is required now that `nocopy` is enabled.
3. **Startup Ordering:** Added a dependency for `gitserver` on `grader` to improve startup stability.

Fixes: #7812

# Comments

The use of `nocopy: true` is the critical part of this fix; it ensures that the containers do not interfere with the shared volume during initialization. Because auto-population is disabled, the manual step to `mkdir -p /var/lib/omegaup/problems.git` via Alpine was added to the workflow to ensure the application finds the expected paths.

# Checklist:

### DIFF
--- a/frontend/server/src/Controllers/Contest.php
+++ b/frontend/server/src/Controllers/Contest.php
@@ -1754,19 +1754,28 @@ class Contest extends \OmegaUp\Controllers\Controller {
             ],
         ];
 
-        \OmegaUp\DAO\ProblemsetIdentityRequest::create(new \OmegaUp\DAO\VO\ProblemsetIdentityRequest([
-            'identity_id' => $r->identity->identity_id,
-            'problemset_id' => $contest->problemset_id,
-            'request_time' => \OmegaUp\Time::get(),
-        ]));
+        try {
+            \OmegaUp\DAO\DAO::transBegin();
 
-        foreach ($admins as $admin) {
-            \OmegaUp\DAO\Notifications::create(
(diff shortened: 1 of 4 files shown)


pabo99 (approved):
We’ll approve this PR and monitor the behavior of the workflows. If the same error shows up again in the future, we’ll see whether it’s necessary to reopen the issue or create new ones with updated information

Ankitsinghsisodya:
Thanks for the approval, @pabo99! Agreed—monitoring the workflows is the best next step. I'll stand by in case any adjustments are needed later.

PR #8666, opened 2025-12-19
added the fix to remove stale tag when someone does an activity

### LINKED ISSUE #8662: [BUG] The instruction "Please comment or update this issue to remove the stale label" is not working as expected
## Current Behavior

A comment was made on an issue requesting the removal of the "Stale" label, but it remained.

## Expected Behavior

When commenting on an issue that has the "Stale" label, it should be removed.

The workflow should be able to detect whenever there are changes in an issue and remove that label.

antedotee:
Added event triggers so the workflow runs immediately when activity occurs:
issue_comment (type: created) — triggers when someone comments on an issue or PR
issues (types: edited, reopened) — triggers when an issue is edited or reopened
pull_request (types: edited, reopened) — triggers when a PR is edited or reopened

Fixes #8662


pabo99 (approved):
Let's approve it!

Issue #8649, opened 2025-12-17
[FEATURE] Add Max Streak, Total Active Days, and Total Submissions to User Heatmap


k-vijay-05:
To add activity statistics `(total submissions in a year, total active days, and max streak)` to the user profile heatmap, similar to how platforms like LeetCode display these metrics above their heatmap. This feature would show the total submissions count, total active days, and max streak (longest consecutive days with submissions) in a header section above the heatmap grid, making it `easy for users to track their coding activity patterns at a glance`.

## Context

The implementation can be done entirely on the client-side using the existing heatmap data , no server-side changes required. The calculation logic will iterate through the same date map that powers the heatmap visualization to compute total submissions, count active days, and track the longest consecutive streak of days with submissions. All statistics will be calculated per selected year and automatically update when users switch between years.

### Reference: LeetCode User Heatmap

<img width="1248" height="262" alt="Image" src="https://github.com/user-attachments/assets/15952ad0-4362-4bb7-a1fe-42042348e74e" />

### Current Implementation: OmegaUp User Heatmap

<img width="1114" height="327" alt="Image" src="https://github.com/user-attachments/assets/047596be-1b24-4930-8e07-c85e522efa2c" />

I’d like to work on this feature and would welcome any feedback or suggestions.

pabo99:
Go ahead, @k-vijay-05 !

PR #8667, opened 2025-12-19
Add statistics (yearly submissions, active days, max streak) to user activity heatmap

vijay-x-dev:
# Description

This change enhances the user activity heatmap by adding derived statistics `for the selected year`:

1. Total submissions 
2. Total active days
3. Max streak

Fixes: #8649 

Screenshot:
Screenshot 2025-12-20 003035

# Checklist:

### DIFF
--- a/.github/workflows/assign-issues.yml
+++ b/.github/workflows/assign-issues.yml
@@ -1,0 +1,53 @@
+name: "Assign Issues"
+
+on:
+    issue_comment:
+        types: [created]
+    schedule:
+        # Run daily to handle reminders and automatic unassignments
+        - cron: "0 3 * * *"
+    workflow_dispatch:
+
+jobs:
+    assign:
+        runs-on: ubuntu-22.04
+        permissions:
+            issues: write
+            pull-requests: write
+        steps:
+            - name: Assign or unassign issues
+              uses: takanome-dev/assign-issue-action@beta
+              with:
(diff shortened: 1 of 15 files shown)


vijay-x-dev:
> The change looks good. Thanks for adding this feature.
> 
> There’s an issue not caused by your changes, but I’d like you to address it in this same PR:
> 
> The `renderHeatmap` function is being called from different places, and it seems many of them are triggered at startup.
> 
> Could you refactor the function and add a validation so that the heatmap is rendered only once?

Sure! @pabo99

vijay-x-dev:
Added a `hasRendered boolean` flag to prevent duplicate renders within the same render cycle:

1. `Early return guard`: Added if (this.hasRendered) return; at the start of renderHeatmap() to skip rendering if already rendered
2. `Reset flag on data changes`: Reset hasRendered = false in all watchers (onAvailableYearsChange, onDataChange,  onLoadingChange) to  allow re-rendering when data actually changes
3. `Set flag after render`: Set hasRendered = true after successfully rendering heatmap

The heatmap now renders only once. Verified using `debug output`showing render count reduced from 3 times to 1 time on initial load.

**`Debugger screenshot (Initial code):`**
Screenshot 2025-12-23 174731

**`Debugger screenshot (Updated code):`**
Screenshot 2025-12-23 173959

Issue #8660, opened 2025-12-19
[FEATURE] Migrate template.tpl notifications to a reusable Vue component


Ankitsinghsisodya:
## Context

Currently, the error, warning, info, and success messages shown across the platform are detached from the Vue components. They are rendered from a static `<div>` in `template.tpl` and manipulated using jQuery in `ui.ts`.

This creates several issues:
- **Inconsistency**: The rest of the UI is built with Vue, but status messages rely on direct DOM manipulation
- **Maintainability**: jQuery-based code is harder to test and debug compared to Vue components
- **Coupling**: The status display logic is tightly coupled to specific DOM element IDs (`#status`)
- **Animation**: Uses jQuery's `slideDown()`/`slideUp()` instead of Vue transitions

The goal is to modernize this system by:
1. Creating a reusable `StatusMessage.vue` component
2. Using a reactive store to manage status state
3. Migrating existing `ui.error()`, `ui.success()`, `ui.warning()`, `ui.info()` calls to use the new component
4. Eventually removing the jQuery dependency for this feature

I want to work on this issue.

pabo99:
Assigned to you, @Ankitsinghsisodya !

PR #8670, opened 2025-12-20
feat(ui): migrate template.tpl notifications to reusable Vue component

Ankitsinghsisodya:
## Description
This PR migrates the platform's notification system (error, warning, info, success) 
from a legacy jQuery-based implementation to a modern, reusable Vue component.

### Changes made:
- Created [GlobalNotifications.vue](cci:7://file:///home/ankitsinghsisodya/Desktop/BIT-Ankit/frontend/www/js/omegaup/components/common/GlobalNotifications.vue:0:0-0:0) component to handle rendering of notification 
  messages with Vue transitions
- Implemented [notificationsStore.ts](cci:7://file:///home/ankitsinghsisodya/Desktop/BIT-Ankit/frontend/www/js/omegaup/notificationsStore.ts:0:0-0:0) Vuex store module to manage notification 
  visibility and content, replacing direct DOM manipulation
- Added [global_notifications.ts](cci:7://file:///home/ankitsinghsisodya/Desktop/BIT-Ankit/frontend/www/js/omegaup/common/global_notifications.ts:0:0-0:0) entry file to mount the component
- Refactored [ui.ts](cci:7://file:///home/ankitsinghsisodya/Desktop/BIT-Ankit/frontend/www/js/omegaup/ui.ts:0:0-0:0) methods (`ui.error`, `ui.success`, `ui.warning`, `ui.info`) 
  to dispatch updates to the new Vuex store instead of using jQuery `slideDown()`/`slideUp()`
- Replaced the static HTML `#status` div in [template.tpl](cci:7://file:///home/ankitsinghsisodya/Desktop/BIT-Ankit/frontend/templates/template.tpl:0:0-0:0) with a Vue mount point
- Cleaned up jQuery click handler in [status.dismiss.js](cci:7://file:///home/ankitsinghsisodya/Desktop/BIT-Ankit/frontend/www/js/status.dismiss.js:0:0-0:0)
- Added webpack entry point for `common_global_notifications`

Fixes: #8660

### Notes:
- The behavior of existing calls to `ui.error()` throughout the codebase remains 
  unchanged; only the underlying mechanism for rendering the message has been swapped out.
- Animations use CSS/JS transitions instead of jQuery's `slideDown`/`slideUp`

## Checklist:


frontend/www/js/omegaup/common/global_notifications.ts
```diff
@@ -0,0 +1,20 @@
+import GlobalNotifications from '../components/common/GlobalNotifications.vue';
```
pabo99: nit:
```suggestion
import common_GlobalNotifications from '../components/common/GlobalNotifications.vue';
```

frontend/www/js/omegaup/components/common/GlobalNotifications.vue
```diff
@@ -0,0 +1,130 @@
+<template>
```
pabo99, line 1: Would it be possible to add this component along with its test to another PR?

This way we can streamline the reviews.

As a bonus, you could also include the stories for the component in the new PR

frontend/www/js/omegaup/components/common/GlobalNotifications.vue
```diff
@@ -0,0 +1,130 @@
...
+
+  dismiss(): void {
+    notificationsStore.dispatch('dismissNotifications');
+  }
+
+  // Transition hooks for dynamic height animation
+  beforeEnter(el: HTMLElement): void {
+    el.style.height = '0';
+    el.style.opacity = '0';
+    el.style.overflow = 'hidden';
```
pabo99: Is there a better way to create these styles?

We’re not too fond of using inline styles due to the risks involved—such as data injection, maintainability issues, and performance drawbacks.

frontend/www/js/omegaup/notificationsStore.ts
```diff
@@ -0,0 +1,155 @@
+import Vue from 'vue';
```
pabo99, line 1: Could you move the creation of the store and its corresponding test to another PR, and include the store call in `ui.ts` within that same PR?

frontend/www/js/omegaup/ui.ts
```diff
@@ -160,29 +133,10 @@ export function ignoreError(response: { error?: string; payload?: any }): void {
...
-    .attr('data-counter', statusCounter + 1)
-    .slideUp({
-      complete: () => {
-        statusElement
-          .removeClass('animating')
-          .attr('data-counter', statusCounter + 2);
-      },
-    });
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export function dismissNotifications(_originalStatusCounter?: number): void {
```
pabo99: Looks like this function parameter is no longer used. 

It was previously for adding seconds to hide stacked messages, but since counters aren’t added in the new version, it can likely be removed—just check if any calls still include it

frontend/www/js/omegaup/components/common/GlobalNotifications.vue
Ankitsinghsisodya: I will refactor the component to use pure CSS transition classes instead of inline JavaScript styles.

Ankitsinghsisodya:
@pabo99  sir, it is done. Now, you can position notifications in any corner or as banners. I made sure the default behavior didn't change, so no refactoring is needed for existing alerts.

frontend/www/js/omegaup/notificationsStore.ts
Ankitsinghsisodya: @pabo99 Thanks for the suggestion.

I looked into splitting the store and component into separate PRs, but I realized it introduces a dependency issue that breaks the application state:

1. **The Producer (`ui.ts`):** If I merge the Store and `ui.ts` changes first, the system effectively stops showing notifications because `ui.ts` will dispatch to Vuex, but the **Consumer** (`GlobalNotifications.vue`) won't be there to render them yet.
2. **The Consequence:** This renders the platform's feedback system invisible during the transition and causes the Cypress tests (which look for alert elements) to fail immediately.

Since the logic (`ui.ts`) and the view (`GlobalNotifications.vue`) are tightly coupled in this migration, they need to be deployed together to ensure the notification system remains atomic and functional. I'll keep them in this single PR to ensure CI passes and the UI doesn't break.

Ankitsinghsisodya:
@pabo99 sir, I have made all the changes suggested by you. Kindly review the PR whenever you are free.

pabo99 (approved):
Great!!!

Ankitsinghsisodya:
@pabo99 sir, can we merge with PR since it has been approved?

pabo99:
Before merging, could you tell me where I can configure the default position of the messages?

I’d like to see how they behave in each of the possible positions

Ankitsinghsisodya:
@pabo99 The default position is configured in `notificationsStore.ts` at line 48:

```typescript
position: NotificationPosition.Top,
```

### Available Positions

| Position      | Enum Value                         | Description                                                       |
| ------------- | ---------------------------------- | ----------------------------------------------------------------- |
| Top (default) | `NotificationPosition.Top`         | Full-width banner at the top of the page                          |
| Bottom        | `NotificationPosition.Bottom`      | Fixed full-width banner at the bottom                             |
| Top-Right     | `NotificationPosition.TopRight`    | Toast-style corner notification (400px max, rounded, with shadow) |
| Bottom-Right  | `NotificationPosition.BottomRight` | Toast-style corner notification (400px max, rounded, with shadow) |

### How to Change Position

**Option 1: Change the global default**
Modify line 48 in `notificationsStore.ts` to use any of the above enum values.

**Option 2: Per-notification positioning**
Pass `position` when calling the notification methods:

```typescript
import { NotificationPosition } from './notificationsStore';

ui.success('Success message', { position: NotificationPosition.TopRight });
ui.error('Error message', { position: NotificationPosition.BottomRight });
```

The current implementation defaults to `Top` to maintain backward compatibility with the existing behavior.

pabo99:
I’ve been testing the notifications component and I still found some places where messages don’t show up when they should 

Messages

Could you take a look at this case?

Ankitsinghsisodya:
@pabo99 sir, it was just a CSS problem with positioning. The last commit has only 7-8 line changes. I have fixed it. Actually, the notifications were working; the problem was that we needed to scroll up to see them. Now I have fixed it, and we can go forward with the review.

frontend/www/js/omegaup/common/global_notifications.ts
```diff
@@ -0,0 +1,31 @@
...
+      // This handles the legacy loading/root element visibility in a Vue-native way
+      return createElement('omegaup-global-notifications', {
+        on: {
+          'ui-ready': () => {
+            const loadingEl = document.getElementById('loading');
+            const rootEl = document.getElementById('root');
+            if (loadingEl) loadingEl.style.display = 'none';
+            if (rootEl) rootEl.style.display = 'block';
+          },
+        },
```
pabo99: It looks like this logic is legacy, probably from before the arena was migrated to Vue. I can’t find those elements in the current HTML, and they don’t seem necessary with the current architecture.

Could you please confirm if they are no longer used, and if so, safely remove this code? This would help keep things cleaner

pabo99:
I hadn’t noticed the conflict. Could you please resolve it and push the change?

pabo99:
@tvanessa @heduenas 

I’d like to introduce the new component for displaying alert, info, error, and success messages. It’s built entirely in Vue — no more jQuery in this part of the system. Thanks to @Ankitsinghsisodya for developing this solution.

One improvement is that it now supports configurable positioning. By default, messages appear in the same location as before, but now we can choose where they show up. For example, in this case I used a position that makes certain messages less intrusive:

image

Looking ahead, we could even group messages, with each group having its own default position. That way, we wouldn’t need to change the default value or adjust multiple settings every time

frontend/www/js/omegaup/common/global_notifications.ts
Ankitsinghsisodya: Thanks for pointing out. I will remove it.

Ankitsinghsisodya:
@pabo99 sir, I think it would be a good idea to add stack messages.

pabo99 (approved):
Thank you, so much, @Ankitsinghsisodya!

Ankitsinghsisodya:
@pabo99, thank you. Working on this migration was a great experience, and I learned a lot during the process.

Issue #8616, opened 2025-12-11
[FEATURE] A User Comparison tool so that I can visualize my progress and stats against other users side-by-side


Ankitsinghsisodya:
## Context

_How has this issue affected you? What is the goal to achieve?_

Competitive programming is inherently social, yet currently, there is no easy way to directly compare statistics (Rank, Problems Solved, Topic Mastery) with a friend or rival without manually switching back and forth between profile tabs.

The goal is to add a social/competitive element to the platform by creating a new page (e.g., `/compare`). This page would allow a user to enter two usernames and generate a side-by-side visualization of their performance metrics, fulfilling a common request in competitive communities to "Compare with a Friend."

I want to work on this issue.

Ankitsinghsisodya:
https://github.com/user-attachments/assets/7be8f7b0-4f51-426f-ab4b-c9b35aee6e00

@pabo99 Sir, I have created a demo video showcasing this feature for better clarity. I would greatly appreciate your feedback on the current implementation, as well as any suggestions for improvements or additional enhancements you think would be valuable.

Ankitsinghsisodya:
@pabo99 Sir, combining this feature with #8656 would significantly enhance the competitive programming experience for users on the omegaUp platform.

Ankitsinghsisodya:
@pabo99, I have implemented the feature and look forward to your feedback. I attempted to split this into separate PRs, but the tests failed due to dependencies between the frontend and backend. As a result, I have kept it as a single PR to ensure everything passes correctly.

pabo99:
The feature looks great, thanks for working on this issue!

The review may take some time due to the size of the PR, so I’d like to ask for your patience.

I have a suggestion before the review:

On the _User Ranking_ page, you added a button to compare users. What do you think about adding another button labeled “Select two users to _compare”?_ This would trigger checkboxes on the left side of the table, and once two different users are selected, the “_Compare users_” button would become enabled

Ankitsinghsisodya:
@pabo99 Thank you for the positive feedback! I completely understand regarding the review time; I will be patient.

Regarding your suggestion for the User Ranking page: I think that is a great idea and would definitely improve the user flow. I will work on implementing the "Select to compare" toggle and the checkbox functionality right away and update this PR with the changes.

pabo99:
One more idea:

If the user selects only one username, it would be great to enable another button: "_Compare vs me_".

I would appreciate it if these changes could be applied to the current PR.

Thank you in advance!

[...]

PR #8678, opened 2025-12-20
Feature: Implementing Compare User

Ankitsinghsisodya:
# Description

Add a user comparison feature that allows users to compare two profiles side-by-side, showing their statistics and rankings.

**Backend Changes:**
- Added `apiCompare()` and `getCompareDetailsForTypeScript()` methods to `User.php` controller
- Added TypeScript API bindings in `api.ts` and types in `api_types.ts`
- Added nginx rewrite and PHP entry point for `/compare/` route
- Added comprehensive PHPUnit tests for the comparison API

**Frontend Changes:**
- Added `CompareUsers.vue` - Main comparison page with user inputs
- Added `CompareCard.vue` - Individual user profile card component
- Added `compare.ts` - Vue entry point for the compare page
- Updated `Rank.vue` - Added link to compare page
- Updated `NavbarItems.vue` - Added navigation link
- Added translation strings in all 4 languages (en, es, pt, pseudo)

**Features:**
- Compare any two users by username
- Display profile pictures, ranks, solved problems, and contest counts
- Visual highlighting based on comparison results
- Privacy-aware: respects private profile settings
- URL state management for shareable comparison links
- Error handling with user-friendly messages
- Accessible alt text for profile images

Fixes: #8616

# Comments

The feature is accessed via `/compare/` or through the "Compare Users" link in the ranking page.

**API Endpoint:** `GET /api/user/compare/?username1=<user1>&username2=<user2>`

# Checklist:

## Files Changed
- `frontend/server/src/Controllers/User.php`
- `frontend/www/js/omegaup/api.ts`
- `frontend/www/js/omegaup/api_types.ts`
- `frontend/tests/controllers/UserCompareTest.php`
- `frontend/server/nginx.rewrites`
- `frontend/www/compare.php`
- `webpack.config-frontend.js`
- `frontend/www/js/omegaup/components/user/CompareUsers.vue`
- `frontend/www/js/omegaup/components/user/CompareCard.vue`
- `frontend/www/js/omegaup/user/compare.ts`
- `frontend/www/js/omegaup/components/user/Rank.vue`
- `frontend/www/js/omegaup/components/common/NavbarItems.vue`
- Translation files (en, es, pt, pseudo)

## Architecture of feature
https://gist.github.com/Ankitsinghsisodya/452145efcf7e22d780cdf4ab398210ed

## Demo video for the compare feature

https://github.com/user-attachments/assets/f36079fa-a9f4-4121-9bea-2c3c370124de

### DIFF
--- a/frontend/www/js/omegaup/components/problem/creator/cases/Sidebar.vue
+++ b/frontend/www/js/omegaup/components/problem/creator/cases/Sidebar.vue
@@ -2,7 +2,7 @@
   <div>
     <div class="d-flex align-items-center justify-content-between">
       <h5 class="mb-0 d-none d-md-inline">{{ T.problemCreatorGroups }}</h5>
-      <div>
+      <div class="d-flex flex-nowrap align-items-center">
         <b-button
           data-toggle-layout-sidebar
           size="sm"
@@ -100,57 +100,59 @@
     <div>
       <b-card class="border-0">
         <b-row class="mb-1">
-          <b-button
-            data-sidebar-groups="ungrouped"
-            variant="light"
-            data-placement="top"
-            :title="T.problemCreatorUngroupedCases"
-            class="w-84"
-            @click="showUngroupedCases = !showUngroupedCases"
-          >
-            <div class="d-flex justify-content-between">
-              <div class="mr-2 text-truncate">
-                {{ T.problemCreatorUngrouped }}
+          <div class="d-flex flex-nowrap align-items-center w-100">
+            <b-button
+              data-sidebar-groups="ungrouped"
+              variant="light"
+              data-placement="top"
+              :title="T.problemCreatorUngroupedCases"
+              class="w-84"
+              @click="showUngroupedCases = !showUngroupedCases"
+            >
+              <div class="d-flex justify-content-between">
+                <div class="mr-2 text-truncate">
+                  {{ T.problemCreatorUngrouped }}
+                </div>
+                <div class="d-inline-block text-nowrap">
+                  <b-badge
+                    data-sidebar-ungrouped-cases="count"
+                    variant="primary"
+                    class="mr-1"
+                    >{{ ungroupedCases.length }}</b-badge
+                  >
+                  <b-badge data-sidebar-ungrouped-cases="points" variant="info">
+                    {{ Math.round(getTotalPointsForUngroupedCases) }}
+                    {{ T.problemCreatorPointsAbbreviation }}</b-badge
+                  >
+                </div>
               </div>
-              <div class="d-inline-block text-nowrap">
-                <b-badge
-                  data-sidebar-ungrouped-cases="count"
-                  variant="primary"
-                  class="mr-1"
-                  >{{ ungroupedCases.length }}</b-badge
-                >
-                <b-badge data-sidebar-ungrouped-cases="points" variant="info">
-                  {{ Math.round(getTotalPointsForUngroupedCases) }}
-                  {{ T.problemCreatorPointsAbbreviation }}</b-badge
-                >
-              </div>
-            </div>
-          </b-button>
-          <b-dropdown variant="light" size="sm" right no-caret>
-            <template #button-content>
-              <BIconThreeDotsVertical />
-            </template>
-            <b-dropdown-item disabled
-              ><b-row>
-                <div class="ml-6">
-                  <BIconTrash variant="danger" font-scale=".95" />
-                </div>
-                <div class="ml-8">
-                  {{ T.problemCreatorDeleteGroup }}
-                </div>
-              </b-row>
-            </b-dropdown-item>
-            <b-dropdown-item @click="deleteUngroupedCases()"
-              ><b-row>
-                <div class="ml-6">
-                  <BIconTrash variant="danger" font-scale=".95" />
-                </div>
-                <div class="ml-8">
-                  {{ T.problemCreatorDeleteCases }}
-                </div>
-              </b-row>
-            </b-dropdown-item>
-          </b-dropdown>
+            </b-button>
+            <b-dropdown variant="light" size="sm" right no-caret>
+              <template #button-content>
+                <BIconThreeDotsVertical />
+              </template>
+              <b-dropdown-item disabled
+                ><b-row>
+                  <div class="ml-6">
+                    <BIconTrash variant="danger" font-scale=".95" />
+                  </div>
+                  <div class="ml-8">
+                    {{ T.problemCreatorDeleteGroup }}
+                  </div>
+                </b-row>
+              </b-dropdown-item>
+              <b-dropdown-item @click="deleteUngroupedCases()"
+                ><b-row>
+                  <div class="ml-6">
+                    <BIconTrash variant="danger" font-scale=".95" />
+                  </div>
+                  <div class="ml-8">
+                    {{ T.problemCreatorDeleteCases }}
+                  </div>
+                </b-row>
+              </b-dropdown-item>
+            </b-dropdown>
+          </div>
           <b-collapse v-model="showUngroupedCases" class="w-100">
             <b-card class="border-0 w-100">
               <b-row


Ankitsinghsisodya:
@pabo99 I’ve updated the PR with the requested changes!

1. I added the "Select two users to compare" toggle, which triggers the checkbox visibility on the left side.
2. I also implemented the "Compare vs me" button, which becomes available when a single username is selected.

Everything should now be working as discussed. Ready for your review whenever you have a moment!


https://github.com/user-attachments/assets/64ecff1b-f3e3-443e-9720-357a48be0763

frontend/server/nginx.rewrites
```diff
@@ -100,12 +100,15 @@ rewrite ^/problem/([a-zA-Z0-9_+-]+)/stats/?$ /problems/stats.php?problem_alias=$
 rewrite ^/profile/?$ /profile/index.php last;
 rewrite ^/profile/([a-zA-Z0-9_+.-]+)/?$ /profile/index.php?username=$1 last;
 rewrite ^/rank/?$ /rank/users.php last;
+rewrite ^/rank/users/?$ /rank/users.php last;
 rewrite ^/rank/authors/?$ /rank/authors.php last;
 rewrite ^/rank/schools/?$ /rank/schools.php last;
+rewrite ^/rank/compare/?$ /compare.php last;
```
pabo99: For consistency, could you move this file to the `/rank/` directory?

frontend/server/nginx.rewrites
```diff
@@ -100,12 +100,15 @@ rewrite ^/problem/([a-zA-Z0-9_+-]+)/stats/?$ /problems/stats.php?problem_alias=$
...
 rewrite ^/privacypolicy/?$ /privacypolicy.php last;
 rewrite ^/schoolofthemonth/?$ /schools/schoolofthemonth.php last;
 rewrite ^/schools/?$ /course/ permanent;
 rewrite ^/schoolsrank/?$ /rank/schools/ permanent;
+rewrite ^/compare/?$ /compare.php last;
```
pabo99: Both entry points you’re adding point to the same file. Could you remove this one?

frontend/templates/es.lang
```diff
@@ -273,6 +274,15 @@ codersOfTheMonthRank = "Top 100 coders del mes pasado"
 codersOfTheMonthUser = "Coder"
 collecionOtherTags = "Otras etiquetas"
 collectionTitle = "Colecciones de Problemas"
+compareButton = "Comparar"
+compareDescription = "Compara los perfiles y estadísticas de dos usuarios"
```
pabo99: Can you  use 'coders' instead of 'usuarios' to make the texts more gender-free

frontend/templates/es.lang
```diff
@@ -273,6 +274,15 @@ codersOfTheMonthRank = "Top 100 coders del mes pasado"
...
+compareEnterUsername = "Ingresa el nombre de usuario"
+compareUser1Label = "Usuario 1"
```
pabo99: Same here:
```suggestion
compareUser1Label = "Coder 1"
```

frontend/templates/es.lang
```diff
@@ -273,6 +274,15 @@ codersOfTheMonthRank = "Top 100 coders del mes pasado"
...
+compareUser2Label = "Usuario 2"
```
pabo99: Same here as well:
```suggestion
compareUser2Label = "Coder 2"
```

frontend/templates/es.lang
```diff
@@ -273,6 +274,15 @@ codersOfTheMonthRank = "Top 100 coders del mes pasado"
...
+compareUserNotFound = "Usuario no encontrado"
```
pabo99:
```suggestion
compareUserNotFound = "Coder no encontrado"
```

frontend/templates/es.lang
```diff
@@ -273,6 +274,15 @@ codersOfTheMonthRank = "Top 100 coders del mes pasado"
...
+compareUsersError = "Ocurrió un error al comparar los usuarios"
```
pabo99:
```suggestion
compareUsersError = "Ocurrió un error al comparar coders"
```

frontend/templates/es.lang
```diff
@@ -273,6 +274,15 @@ codersOfTheMonthRank = "Top 100 coders del mes pasado"
...
+compareUsersTitle = "Comparar Usuarios"
```
pabo99:
```suggestion
compareUsersTitle = "Comparar Coders"
```

frontend/templates/es.lang
```diff
@@ -907,6 +917,7 @@ navAuthorRanking = "Ranking de contribuidores"
 navBlog = "Blog"
 navCoderOfTheMonth = "Coder del mes general"
 navCoderOfTheMonthFemale = "Coder del mes para ellas"
+navCompareUsers = "Comparar Usuarios"
```
pabo99:
```suggestion
navCompareUsers = "Comparar Coders"
```

frontend/templates/es.lang
```diff
@@ -978,6 +989,7 @@ omegaupTitleCollections = "Colecciones"
 omegaupTitleCollectionsByAuthor = "Problemas por autor"
 omegaupTitleCollectionsByLevel = "Problemas por nivel"
 omegaupTitleCommonIndex = "Bienvenido"
+omegaupTitleCompareUsers = "Comparar Usuarios"
```
pabo99:
```suggestion
omegaupTitleCompareUsers = "Comparar Coders"
```

frontend/templates/es.lang
```diff
@@ -1687,6 +1702,8 @@ scoreboardShowParticipantsNames = "Mostrar nombres de participantes"
 searchSchool = "Buscar escuela"
 searchUser = "Buscar coder"
 selectProblemLevelDesc = "Para agregar el problema, debes elegir su nivel correspondiente"
+selectTwoUsersToCompare = "Selecciona dos usuarios para comparar"
```
pabo99, line 1710:
```suggestion
selectTwoUsersToCompare = "Selecciona dos coders para comparar"
```

frontend/templates/es.lang
```diff
@@ -2151,6 +2168,7 @@ wordsSeeCourse = "Ver el curso"
 wordsSeeGeneralRanking = "Ver el ranking general"
 wordsSeeLatestSubmissions = "Ver los últimos envíos realizados por %(username)"
 wordsSeeSolution = "Ver Solución"
+wordsSelect = "Seleccionar"
```
pabo99, line 2178: We still haven’t found the best way to name the translation keys.

What we do want to avoid over time is using the `words` prefix, since it lacks context. 

Could you use a different prefix? Perhaps this could be related to the view where the text appears

pabo99 (commented):
Oh, this change is quite extensive.

In future PRs, could you separate the server logic from the UI logic? That can help make reviews faster.

For now, I’m leaving some minor comments, and then I’ll continue with the review of the components and the server changes.

Overall, the change looks great. Just one observation regarding the position of the buttons: could they be placed outside the header's card? That area feels overloaded once more information is added.

What do you think about placing them at the same level as the search bar? In fact, the message that appears saying filters cannot be used also doesn’t look good.

If it’s possible to fix this here, feel free to do so. If it’s too extensive, we can leave it for another issue.

Ankitsinghsisodya:
image
Thank you for the feedback!

**Regarding separating server logic from UI logic:** Absolutely, I'll keep this in mind for future PRs. I agree that splitting backend (API/controllers) and frontend (Vue components) changes into separate PRs would make reviews more manageable.

**Regarding the button placement:** Done! I've moved the compare buttons from the header to the card-body area, placing them at the same level as the search bar. The layout is now:
- **Header:** Just the title and help link (much cleaner)
- **Card body row 1:** Search input, Search button, Filter dropdown
- **Card body row 2:** Compare buttons

**Regarding the filter messages:** I've also removed the "must login to filter" and "must update basic info" badge messages that were cluttering the UI. These can be addressed more elegantly in a future iteration if needed.

Ankitsinghsisodya:
@pabo99 Sir, may we kindly proceed with reviewing this PR?

frontend/server/src/Controllers/User.php
```diff
@@ -5122,6 +5124,150 @@ public static function getDocsForTypeScript(\OmegaUp\Request $r) {
...
+    /**
+     * Get comparison data for two users
+     *
+     * @return UserCompareData|null
+     */
+    private static function getUserCompareData(
+        ?\OmegaUp\DAO\VO\Identities $loggedIdentity,
+        ?string $username
+    ): ?array {
+        if (is_null($username) || $username === '') {
```
pabo99: Could you simlpify a bit?
```suggestion
        if (empty($username)) {
```

frontend/server/src/Controllers/User.php
```diff
@@ -5122,6 +5124,150 @@ public static function getDocsForTypeScript(\OmegaUp\Request $r) {
             'entrypoint' => 'common_docs',
         ];
     }
+
+    /**
+     * Get comparison data for two users
+     *
+     * @return UserCompareData|null
+     */
+    private static function getUserCompareData(
```
pabo99, line 5148: Would it be possible to cache this content?

Since comparisons could be performed without many validations, this might result in too many round trips to the database each time a comparison is made

The user cache should be invalidated when the user solves a problem or joins a new contest

frontend/www/js/omegaup/components/user/CompareCard.vue
```diff
@@ -0,0 +1,158 @@
...
+.stats-grid {
+  display: grid;
+  grid-template-columns: repeat(2, 1fr);
+  gap: 1rem;
+}
+
+.stat-item {
+  text-align: center;
+  padding: 0.5rem;
+  background-color: #f8f9fa;
```
pabo99: Can you add this color and the others as variables?

frontend/www/js/omegaup/components/user/CompareCard.vue
```diff
@@ -0,0 +1,158 @@
...
+            v-if="profile.gravatar_92"
+            :src="profile.gravatar_92"
+            :alt="profileImageAlt"
+            class="rounded-circle mb-2"
+            width="92"
+            height="92"
+          />
+        </a>
+        <h5 class="card-title mb-1">
+          <a :href="profileUrl" :class="profile.classname">
```
pabo99: Can you check why the classname is not working?

I ran a test with ranked users:

Image

But in the comparison view I see this:

Image

frontend/www/js/omegaup/components/user/CompareCard.vue
```diff
@@ -0,0 +1,158 @@
...
+            }}</strong>
+          </div>
+        </div>
+
+        <!-- Contests -->
+        <div class="stat-item">
+          <div class="stat-label">{{ T.profileContests }}</div>
+          <div class="stat-value">
+            <strong>{{
+              contestsCount !== null ? contestsCount : 'N/A'
```
pabo99: Could you add this string: `N/A` to the localization files?

frontend/www/js/omegaup/components/user/Rank.vue
```diff
@@ -72,6 +105,14 @@
       <table class="table mb-0 table-responsive-sm">
         <thead>
           <tr>
+            <th
+              v-if="isSelectionMode"
+              scope="col"
+              class="text-center align-middle"
+              style="width: 40px"
```
pabo99: Could you add this style as a CSS class instead of using inline styles? 

You could also use a smaller font size for this class, since it will only be used temporarily while the user is selecting usernames to compare

frontend/www/js/omegaup/user/compare.ts
```diff
@@ -0,0 +1,56 @@
...
+    return;
+  }
+
+  let payload: UserComparePayload;
+  try {
+    payload = JSON.parse(payloadElement.innerText) as UserComparePayload;
+  } catch (e) {
+    console.error('Failed to parse payload JSON:', e);
+    return;
+  }
```
pabo99: Here you can use the PayloadParsers to directly obtain the payload from the types:

```suggestion
  const payload = types.payloadParsers.UserComparePayload();
```

This way, you no longer need to redeclare the interfaces you created above

frontend/www/js/omegaup/user/compare.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  const containerElement = document.getElementById('main-container');
+  if (!containerElement) {
+    console.error('Main container element not found');
+    return;
+  }
```
pabo99: This validation is kind of unnecessary, so it can be removed
```suggestion
```

frontend/www/js/omegaup/components/user/CompareUsers.vue
```diff
@@ -0,0 +1,220 @@
...
+        this.user2 = response.user2 as UserCompareData | null;
+        this.errorMessage = null;
+      })
+      .catch((error: { message?: string }) => {
+        console.error('Compare error:', error);
+        this.errorMessage = error.message || T.compareUsersError;
+      })
+      .finally(() => {
+        this.isLoading = false;
+      });
```
pabo99: Generally, all API calls are handled in controller files (`.ts`).
Could move this block to `compare.ts` and only emit the action from here.

You might need to propagate the action through multiple components.

frontend/www/js/omegaup/components/user/CompareUsers.vue
```diff
@@ -0,0 +1,220 @@
...
+    })
+      .then((response) => {
+        this.user1 = response.user1 as UserCompareData | null;
+        this.user2 = response.user2 as UserCompareData | null;
+        this.errorMessage = null;
+      })
+      .catch((error: { message?: string }) => {
+        console.error('Compare error:', error);
+        this.errorMessage = error.message || T.compareUsersError;
+      })
```
pabo99: What do you think about delegating this to the notifications component you recently modernized?
```suggestion
      .catch(ui.error)
```

This way, you no longer need to worry about adding the div where the error message is displayed.

Even if the position is not exactly the same, this shouldn’t happen very often, since the endpoint doesn’t throw many errors

Ankitsinghsisodya:
@pabo99 sir, I have done all the changes suggested you.

frontend/www/js/omegaup/components/user/CompareCard.vue
```diff
@@ -0,0 +1,170 @@
...
+    const rank = this.profile.rankinfo.rank;
+    if (rank == null) return T.profileRankUnrated;
+    return `#${rank}`;
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+$stat-item-bg: #f8f9fa;
+$stat-label-color: #6c757d;
```
pabo99: The idea of having these colors as variables is to unify them within a single file, keeping in mind the moment when dark mode is planned to be implemented in omegaUp. 

So it would be recommended to move them to `frontend/www/sass/base/_variables.scss` and, preferably, use the same naming convention for the identifiers. 

Here’s an example:
https://github.com/omegaup/omegaup/blob/71eca68f6ff39b86bd09aaef261ca2708fdec3a3/frontend/www/js/omegaup/components/arena/Clarification.vue#L195-L205

frontend/www/js/omegaup/components/user/CompareCard.vue
```diff
@@ -0,0 +1,170 @@
...
+        </div>
+
+        <!-- Problems Solved -->
+        <div class="stat-item">
+          <div class="stat-label">{{ T.profileSolvedProblems }}</div>
+          <div class="stat-value">
+            <strong>{{
+              solvedProblemsCount !== null
+                ? solvedProblemsCount
+                : T.wordsNotApplicable
```
pabo99, line 45: I can't find the definition of this key

frontend/www/js/omegaup/components/user/CompareUsers.vue
```diff
@@ -0,0 +1,179 @@
...
+  @Prop({ default: false }) isLoading!: boolean;
+
+  T = T;
+  inputUsername1: string = '';
+  inputUsername2: string = '';
+
+  created(): void {
+    this.inputUsername1 = this.initialUsername1 ?? this.inputUsername1;
+    this.inputUsername2 = this.initialUsername2 ?? this.inputUsername2;
+  }
```
pabo99: Is it possible to initialize directly with inputUsername?
```suggestion
  inputUsername1: string = this.initialUsername1 ?? '';
  inputUsername2: string = this.initialUsername2 ?? '';
```

frontend/www/js/omegaup/components/user/CompareUsers.vue
```diff
@@ -0,0 +1,179 @@
...
+@Component({
+  components: {
+    'omegaup-user-compare-card': user_CompareCard,
+  },
+})
+export default class CompareUsers extends Vue {
+  @Prop({ default: null }) user1!: UserCompareData | null;
+  @Prop({ default: null }) user2!: UserCompareData | null;
+  @Prop({ default: null }) initialUsername1!: string | null;
+  @Prop({ default: null }) initialUsername2!: string | null;
```
pabo99: These properties should be immutable. Can you remove the suffix?
```suggestion
  @Prop({ default: null }) username1!: string | null;
  @Prop({ default: null }) username2!: string | null;
```

Although in some places we still have it, we’re making it possible to remove it gradually. Instead, we use the suffix `current...` or `selected...` for data that can actually change

frontend/www/js/omegaup/components/user/CompareUsers.vue
```diff
@@ -0,0 +1,179 @@
...
+      if (solved2 < solved1) return 'compare-loser';
+    }
+    return 'compare-tie';
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+.compare-winner {
+  border-color: #28a745 !important;
```
pabo99: These colors should also be added as variables

frontend/www/js/omegaup/user/compare.ts
```diff
@@ -0,0 +1,74 @@
...
+      return createElement('omegaup-user-compare', {
+        props: {
+          user1: this.user1,
+          user2: this.user2,
+          initialUsername1: payload.username1,
+          initialUsername2: payload.username2,
+          isLoading: this.isLoading,
+        },
+        on: {
+          compare: (this as any).onCompare,
```
pabo99: Could you add the action right here?
```suggestion
          compare: ({
            username1,
            username2,
          }: {
            username1?: string;
            username2?: string;
          }) => {
            this.isLoading = true;

            // Update URL
            const url = new URL(window.location.href);
            if (username1) {
              url.searchParams.set('username1', username1);
            } else {
              url.searchParams.delete('username1');
            }
            if (username2) {
              url.searchParams.set('username2', username2);
            } else {
              url.searchParams.delete('username2');
            }
            window.history.pushState({}, '', url.toString());

            api.User.compare({
              username1: username1 || undefined,
              username2: username2 || undefined,
            })
              .then((response) => {
                this.user1 = response.user1;
                this.user2 = response.user2;
              })
              .catch(ui.apiError)
              .finally(() => {
                this.isLoading = false;
              });
          },
```
This way, you can remove the method that was created above

frontend/www/js/omegaup/components/user/CompareCard.vue
Ankitsinghsisodya: The key wordsNotApplicable is defined in all language files at line 2116 (value: "N/A"). It's an existing key that's also used in frontend/www/js/omegaup/components/course/FilteredList.vue

Ankitsinghsisodya:
@pabo99 sir, I have done all the changes suggested you.

frontend/www/js/omegaup/components/user/CompareCard.vue
pabo99: It makes sense, since the PR passes all the CI workflows jobs, thanks for clarifying!

frontend/www/js/omegaup/user/compare.ts
```diff
@@ -0,0 +1,109 @@
...
+          .then((response) => {
+            this.user1 = response.user1;
+            this.user2 = response.user2;
+          })
+          .catch(ui.apiError)
+          .finally(() => {
+            this.isLoading = false;
+          });
+      },
+    },
```
pabo99: This is no longer needed, so it can be removed
```suggestion
```

pabo99 (approved):
Thank you so much for adding this feature, @Ankitsinghsisodya! 

I think many students will really appreciate it

Ankitsinghsisodya:
Thanks @pabo99! It was a pleasure working on it. Really appreciate the kind words.

PR #8699, opened 2025-12-23
simplified the labeler workflow

### LINKED ISSUE #8624: [BUG] Simultaneous addition and removal of size tags when opening a PR

### LINKED ISSUE #8671: [BUG] The label workflow is unable to print message if the size of the PR is XL
Ideally, it should have printed the warning message as it is written in the code.

antedotee:
Now, if the size of the PR is XL, the workflow will fail.
Also, I have removed the reduntant code.

Regarding the #8624 issue, I looked into the various open source respositories using this github action. I found out that this is normal. If we want to fix the removal, we would need to create a custom github action, which I believe, will be an unnecessary optimization. Still, I am open to that as well.

Fixes #8671 
Fixes #8624


.github/workflows/labeler.yml
```diff
@@ -89,7 +46,7 @@ jobs:
           l_label: "size/L"
           l_max_size: "200"
           xl_label: "size/XL"
-          fail_if_xl: "false"
+          fail_if_xl: "true"
```
pabo99, line 54: If the workflow is going to fail, it might be a good idea to adjust the limits to be a bit more flexible.

I remember adjusting them somewhere else, but I don’t recall the exact place. The  treshold values would be:

- `xs_max_size`: 10
- `s_max_size`: 50
- `m_max_size`: 200
- `l_max_size`: 300

We’ll also need to update this number in the PR template.

pabo99 (commented):
Regarding the removal, we can leave it as it is for now since it doesn’t seem to have any impact.

I just left a suggestion, and once it’s applied we’ll approve the PR

pabo99 (approved):
Thank you!

pabo99:
Can you check the conflicts and resolve them?

antedotee:
Sorry for the delay, can you have a look at it now @pabo99 ?

PR #8715, opened 2025-12-28
 Keep sidebar and tabs fixed during scroll and conditionally show course navigation buttons

### LINKED ISSUE #8648: [FEATURE] Keep course assignment pages Problems/Lecture tab header and sidebar fixed while scrolling
On `course assignment pages`, when scrolling the Problems/Lecture tab, the `sidebar and the tab header scroll out of view`. Expected: only the problem content should scroll, with `sidebar and tabs fixed`.

## Context

video:

https://github.com/user-attachments/assets/8d9f41a5-a8bc-4cc2-95d1-2845b4bb7da9

This fix improves the user experience by keeping navigation visible while reading problem statements users can switch problems or tabs without losing their place.

I’d like to work on this issue.

vijay-x-dev:
# Description

This PR implements a fixed layout where the sidebar and tab header remain visible while only the problem content area scrolls.

**Improvements**

1.`Fixed Sidebar and Tab Header`
- Made the sidebar and tab header sticky/fixed so they remain visible during scrolling
- Only the main problem content area scrolls, maintaining navigation context
- Sidebar is scrollable independently when content exceeds viewport height

2.` Conditional Next/Prev Course Buttons`

- Next and previous course navigation buttons now only appear when there are adjacent courses available
- Prevents showing empty/non-functional navigation buttons

3.`Auto-scroll to Top on Problem Change`

- When switching between problems, the problem content automatically scrolls to the top
- Ensures users always start reading from the beginning of each problem

Fixes #8648 

**Videos:**

`The layout works correctly on both screen sizes, with the sidebar maintaining maximum visibility while remaining scrollable on smaller screens.`

- **Laptop screen**: _(Sidebar is visible to maximum extent and scrollable when needed)_

https://github.com/user-attachments/assets/769639f6-7b71-4fab-b7d7-628bee3c1cb8

- **Large monitor screen**:

https://github.com/user-attachments/assets/10ddd882-8a64-4750-9c5c-473b428c5f2b

# Checklist:


pabo99 (commented):
Thank you for addressing this issue.

I’d like to ask if it’s possible to apply an additional change:

I see that the assignment title and the timer take up too much space, which is very valuable when working on screens with limited resolution.

Would it be possible to adjust the position/size of these elements when scrolling down begins? I was thinking they could be placed next to the tabs.

If this is too complicated or would make the change too large, we can leave it for a new issue, no problem

vijay-x-dev:
Thank you for the feedback @pabo99! I completely agree that saving vertical space is crucial, especially on screens with limited resolution. I am interested in solving this in this PR itself. I also had this concern in mind about the space being taken up by the assignment title and timer.

I was exploring different options and found this repositioning approach where the title and timer appear in a compact header format above the tabs when scrolling begins. I've attached a video showing the current behavior. If you feel this is not suitable, I'm happy to switch back to moving them directly next to the tabs instead. Could you please provide feedback on which approach you prefer?


https://github.com/user-attachments/assets/a4920430-fa41-433b-8899-f86b17e6e96e

vijay-x-dev:
Hi @pabo99 , could you please have a look whenever you get a chance. Once I have your feedback, I’ll proceed accordingly. Thanks!

pabo99:
I'm not sure if I'm reviewing an older version, or if the browsers I'm testing on don't support the fixed styles you're adding. Here’s what I’m seeing:

https://github.com/user-attachments/assets/1942f67f-a274-47ba-8b38-a27cda5d1378

I can't see the header elements adjusting as expected. Could you help me review this? I tested on:

- Opera GX:
```
Update stream: Early Access
Core: 125.0.5729.58
System: macOS Version 26.2 (Build 25C56) 26.2.0 arm64
Chromium version: 141.0.7390.125
```

- Safari:
```
Version 26.2 (21623.1.14.11.9)
```

vijay-x-dev:
> I'm not sure if I'm reviewing an older version, or if the browsers I'm testing on don't support the fixed styles you're adding. Here’s what I’m seeing:
> 
>  Grabacion.de.pantalla.2026-01-24.a.la.s.1.21.15.p.m.mov 
> I can't see the header elements adjusting as expected. Could you help me review this? I tested on:
> 
> * Opera GX:
> 
> ```
> Update stream: Early Access
> Core: 125.0.5729.58
> System: macOS Version 26.2 (Build 25C56) 26.2.0 arm64
> Chromium version: 141.0.7390.125
> ```
> 
> * Safari:
> 
> ```
> Version 26.2 (21623.1.14.11.9)
> ```

Hi @pabo99 ,

In my previous message, I was asking to review the idea of compact header attached in the video because I didn't commit the changes of compact header because I was exploring different options.

Now I have committed the changes for the compact header feature where the title and timer reposition above the tabs when scrolling begins.
Sorry for any confusion caused by my earlier message.

vijay-x-dev:
Hi @pabo99 , could you please review this PR when you get a chance.

pabo99 (commented):
I see what's causing us not to get the expected result. When you scroll to the end of the problem details view, the entire content starts scrolling, including the sidebar with the list of problems.

One possible fix is to remove the footer in Arena mode. I think that functionality existed before, but at some point it was lost. To achieve the expected result, you could try removing the footer.

We already have a property called `hideFooterAndHeader`, sent from the `currentPayload` object, which is used in specific cases like showing the print view of a problem.
You could add another property, maybe called `arenaMode`. For now, its behavior would be to remove the footer. In the future, we might also modify the header to hide all menu options in the navbar and avoid distractions

vijay-x-dev:
Added arenaMode to hide the footer in Arena mode, fixing scrolling where the entire content (including the sidebar) scrolled when reaching the end of problem details.

Changes:

- Updated template.tpl to check arenaMode when rendering the footer
- Added 'arenaMode' => true in Course.php methods (getAssignmentDetails and getArenaCourseDetailsForTypeScript)
- Added arenaMode?: bool to the RenderCallbackPayload type in UITools.php

Follows the existing hideFooterAndHeader pattern

frontend/templates/template.tpl
```diff
@@ -65,7 +65,7 @@
     {% jsInclude 'common_scroll_to_top' omitRuntime %}
       <div id="scroll-to-top"></div>
     {% jsInclude 'common_footer' omitRuntime %}
-    {% if not headerPayload.inContest and not hideFooterAndHeader %}
+    {% if not headerPayload.inContest and not hideFooterAndHeader and not arenaMode %}
```
pabo99, line 68: Ohhh, I see we already had an `inContest` property that should do exactly what I suggested. And apparently, the issue was with the declaration, since the render is never initializing `$headerPayload`.

A huge favor: could you please use the existing property and fix it so it works correctly?

In the future, we can rename it to cover more options with `arenaMode`

Issue #8728, opened 2026-01-03
[FEATURE] Add a "hide-tags" toggle on the pages showing problems


Ashutoshbind15:
For someone practicing problems on the site, I'd want them to be able to select whether they want to view the problem tags or not.

## Context

Currently we don't have an option to toggle problem tags visibility, adding one would help the users to practice random questions

<img width="1370" height="617" alt="Image" src="https://github.com/user-attachments/assets/0c1c639a-8197-4ea6-b677-a1138e3f47c6" />

I'd want a button just below the tags list, to toggle tag visibility

Ashutoshbind15:
Hi @pabo99, a couple of questions:

1. Does this look like a feasible request, also saw your interest in this while searching for duplicates in https://github.com/omegaup/omegaup/issues/8166
2. How does the assigning/approvals for newly proposed features work, I have self-assigned for now, but does it need a heads up from a maintainer before anyone can use the '/assign'?

antedotee:
Hi @Ashutoshbind15 , no, this /assign workflow is designed in such a way to reduce maintainer intervention. So, yes, You can start working now.

Ashutoshbind15:
Thanks for clearing it up, @antedotee

PrasannaMishra001:
hey @pabo99 , I have opened a PR that adds a toggle to hide or show problem filters on collection pages as requested in #8728. Since there was no visible progress and this is a small UI change I implemented it to move things forward. Please let me know if any changes are needed.

Ashutoshbind15:
Hi @PrasannaMishra001, thanks for your interest in contributing! I was actually working on this issue and was about to push my changes (including tests and i18n).

Please check the project README and the bot's comment above for details on how issue assignments work. There's also information in the "options for contributors" section in the above bot message about how to be added as a co-assignee when needed.

As for this issue, I've already raised a PR (#8738), so it should be covered. I'd encourage you to check out other unassigned issues or create a new one if you'd like to contribute! Thanks for understanding.

PrasannaMishra001:
@Ashutoshbind15 i saw no progress and issue was open since many days so tried to give better alter options . Still i appreciate your clarification.

pabo99:
Hello @Ashutoshbind15, Welcome to omegaUp and sorry for the late reply.


Indeed, what @antedotee mentioned is correct: the intention of using the issue assignment bot is to reduce maintainers’ intervention. But your question about whether you can start working on these changes is very valid:

[...]

PR #8736, opened 2026-01-08
feat(problem): add toggle to hide problem filters in collections

PrasannaMishra001:
# Description

This PR adds a simple toggle control on problem collection pages to allow users to hide/show the filters sidebar (tags, difficulty, and quality).

When the filters are hidden, the problems table expands to full width, making it easier for users to practice problems without being influenced by tags or metadata.

Fixes: This directly addresses the request in #8763 to allow users to optionally hide problem tags while practicing.

### UI changes
- A minimal arrow (`‹ / ›`) toggle button is added on the right side above the problems list.
- Filters are visible by default.
- Clicking the arrow hides the filters and expands the problems list.
- Clicking again restores the filters.

A short video demonstrating the behavior has been uploaded:
- One video shows the **final arrow-based toggle** (current implementation).

https://github.com/user-attachments/assets/28ddff29-fdf4-40ad-ba71-b11c2ba8efdb

- Another video shows 

https://github.com/user-attachments/assets/241d2b23-d859-4830-81fb-b16f274d6701

an **alternative toggle-switch approach** (for reference, if you suggest).
Fixes: #8763 

cc: @pabo99 

# Comments
- The implementation avoids new dependencies and reuses existing layout logic.
- Tested locally on problem collection pages with empty and populated states.

# Checklist:

- [ x] The code follows the [coding guidelines](https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/Coding-guidelines.md) of omegaUp.
- [x ] The tests were executed and all of them passed.
- [ x] If the change is large (> 200 lines), this PR was split into various Pull Requests. It's preferred to create one PR for changes in controllers + unit tests in PHPUnit,  and then another Pull Request for UI + tests in Jest, Cypress or both.


PrasannaMishra001:
@pabo99 please review this as was quoted in issue #8738 by you: https://github.com/omegaup/omegaup/issues/8728#issuecomment-3742150056 , hence i created a separate issue #8763 with this PR , waiting for review by you .

frontend/www/js/omegaup/components/problem/CollectionList.vue
```diff
@@ -10,8 +10,17 @@
         <h1 class="title-font p-0">{{ title }}</h1>
       </div>
     </div>
+    <div class="d-flex justify-content-end mb-3">
+      <button
+        class="btn btn-link filter-toggle"
+        :aria-label="filtersVisible ? 'Hide filters' : 'Show filters'"
```
pabo99: Could you add these strings to the localization files?

pabo99 (commented):
Is there a way to place the button exactly where the sidebar is going to collapse? And for extra points, could you add a sliding animation, something like what ChatGPT has:

https://github.com/user-attachments/assets/f269218d-223b-4c70-a81b-10f4d8eef5d3

That would look more naturally integrated into the view.

PrasannaMishra001:
@pabo99 thanks for the suggestion! I’ve updated the toggle so it sits closer to the sidebar edge and added a smooth width-based transition, which makes the hide/show behavior feel more natural.
The layout is based on a fixed width sidebar and a flex growing main content area, so once actual problem data is loaded, the table automatically takes up the remaining space without overlapping the filters. This should work the same way in both empty and populated states.I’ve also added the required localization strings .

https://github.com/user-attachments/assets/700a8136-d4e7-47b7-ac00-43e2f969ef07

frontend/www/js/omegaup/components/problem/CollectionList.vue
```diff
@@ -57,12 +61,27 @@
...
+          class="btn btn-link filter-toggle mb-3"
+          :aria-label="
+            filtersVisible ? T.collectionHideFilters : T.collectionShowFilters
+          "
+          @click="filtersVisible = !filtersVisible"
+        >
+          {{ filtersVisible ? '‹' : '›' }}
+          {{
+            filtersVisible ? T.collectionHideFilters : T.collectionShowFilters
+          }}
```
pabo99: Instead of displaying the text on screen, would it be possible to add it as a tooltip?

frontend/www/js/omegaup/components/problem/CollectionList.vue
```diff
@@ -57,12 +61,27 @@
...
+
+      <div class="flex-grow-1 main-content-wrapper">
+        <button
+          class="btn btn-link filter-toggle mb-3"
+          :aria-label="
+            filtersVisible ? T.collectionHideFilters : T.collectionShowFilters
+          "
+          @click="filtersVisible = !filtersVisible"
+        >
+          {{ filtersVisible ? '‹' : '›' }}
```
pabo99: Would it be possible to add FontAwesome icons?

It seems that `bootstrap-vue` already includes them

frontend/www/js/omegaup/components/problem/CollectionList.vue
```diff
@@ -57,12 +61,27 @@
           "
         ></omegaup-problem-filter-quality>
       </div>
-      <div class="col p-0">
+
+      <div class="flex-grow-1 main-content-wrapper">
+        <button
```
pabo99: Is there a way to place this button between the two divs?

The issue with placing it here is that it takes space away from the problems table. Maybe you didn’t notice because the problem collection you tested with was empty

If you want to have a collection with problems, you can do the following:

- From the database, update all problems with the field `quality_seal = 1`
- Run the script `./stuff/cron/aggregate_feedback.py`
- Reload the "Browser problems" page. The problems should now belong to a collection

PrasannaMishra001:
@pabo99 Thanks for the review. I’ve addressed all the suggestions by moving the toggle between the sidebar and the table so it doesn’t affect the table width,replacing the on-screen text with localized tooltip strings and FontAwesome icons and adding a smooth collapse/expand animation for the filters.
I also tested the layout with populated collections by adding sample problems setting `quality_seal = 1` .

https://github.com/user-attachments/assets/15f13c59-916b-47bf-ace4-7570b27ec3a8

me learning to setup db should make future testing and UI-related changes easier as well :) which took most time.

pabo99 (approved):
Great! Thank you, @PrasannaMishra001 !

PR #8757, opened 2026-01-12
[DOCS] Improve AI generated readme docs

Prashant-thakur77:
# Description

This PR enhances the auto-generated README for controllers to show the structure of complex return types by adding a separate table that lists all fields with their types and optional status.

## Before the changes

Returns

| Name | Type |
|------|------|
| `job` | `types.AiEditorialJobDetails` |

## After Changes Apllied
Returns

| Name | Type |
|------|------|
| `job` | `types.AiEditorialJobDetails` |

**`types.AiEditorialJobDetails` fields:**

| Name | Type | Optional |
|------|------|----------|
| `created_at` | `Date` | No |
| `error_message` | `string` | Yes |
| `is_retriable` | `boolean` | No |
| `job_id` | `string` | No |
| `md_en` | `string` | Yes |
| `md_es` | `string` | Yes |
| `md_pt` | `string` | Yes |
| `problem_alias` | `string` | No |
| `status` | `string` | No |
| `validation_verdict` | `string` | Yes |

Screenshot From 2026-01-12 15-44-02

This improvement makes the API documentation much more useful for developers by showing exactly what fields are available in complex return types like types.AiEditorialJobDetails, types.UserProfile, etc.

Fixes: #8514 

## Changes made:
●  Added two helper methods to APIGenerator class in APITool.php:
●  getTypeFields(): Extracts field definitions from complex types
●  generateTypeFieldsTable(): Creates a Markdown table showing all fields with their types and optional status
●  Modified generateDocumentation() to automatically add field tables after complex return types
●  The solution handles optional fields (marked with ? in TypeScript) and displays them in a separate "Optional" column


Prashant-thakur77:
@pabo99  I have updated and resolved the issue,do tell if any changes are required :).
ANd please do tell is any other tasks are present which might need assistance,I would love to help and learn from you all. And please tell me if thereris any way to know what currently is being worked on omegaup ,so if i need to contribute anywhere i can contribute so that it can meet the goals of omegaup.

frontend/server/cmd/APITool.php
```diff
@@ -838,6 +838,69 @@ private function parseDocComment(string $docblock): \Psalm\Internal\Scanner\Pars
...
+    }
+    private function generateTypeFieldsTable(
+        string $typeName,
+        array $typeFields
+    ): void {
+        if (empty($typeFields)) {
+            return;
+        }
+        echo "\n**`{$typeName}` fields:**\n\n";
+        echo "| Name | Type | Optional |\n";
```
pabo99: What if we just focus on the required data and replace the title here?
```suggestion
        echo "| Name | Type | Required |\n";
```

frontend/server/cmd/APITool.php
```diff
@@ -838,6 +838,69 @@ private function parseDocComment(string $docblock): \Psalm\Internal\Scanner\Pars
...
+        echo "|------|------|----------|\n";
+
+        foreach ($typeFields as $field) {
+            $optionalText = $field['optional'] ? 'Yes' : 'No';
```
pabo99: Can you just mark the required fields with the `✓` symbol?

That way it’s visually clearer to understand (you'll need to change the logic to get the correct data)

One more thing: could you add this column to the Requests tables? A field is optional when the request type includes `null`

frontend/server/src/Controllers/README.md
```diff
@@ -4858,6 +4975,13 @@ Get Contests which a certain user has participated in
...
+**`types.UserProfileContests` fields:**
+
+| Name    | Type                                         | Optional |
+| ------- | -------------------------------------------- | -------- |
+| `[key`  | `string]: { data: types.ContestParticipated` | No       |
+| `place` | `number`                                     | No       |
+
 ## `/api/user/create/`
 
 ### Description
```
pabo99, line 5113: Oh, this case seems more challenging. Let me share a possible approach, and we can decide together if it’s feasible for this initial correction stage.

How complicated would it be to build something like the following:

**Data structure for `types.UserProfileContests`:**
```json
{
  "[contest-key: string]": {
    "data": {
      "alias": "string",
      "finish_time": "Date",
      "last_updated": "Date",
      "start_time": "Date",
      "title": "string"
    },
    "place": "number"
  }
}
```
?
Where this:
```json
    {
      "alias": "string",
      "finish_time": "Date",
      "last_updated": "Date",
      "start_time": "Date",
      "title": "string"
    }
```
is the value of `types.ContestParticipated`

Or a bit simpler:


**Data structure for `types.UserProfileContests`:**
```json
{
  "[contest-key: string]": {
    "data": "`types.ContestParticipated`",
    "place": "number"
  }
}
```


This would only apply to those with this type of return:
```php
array<string, array{...}>
```
It seems that only this and `types.AssignmentProgress` fall into that category at the moment

frontend/server/src/Controllers/README.md
```diff
@@ -2604,6 +2625,19 @@ Calls to /status grader
 | -------- | -------------------- |
 | `grader` | `types.GraderStatus` |
 
+**`types.GraderStatus` fields:**
+
+| Name                  | Type                         | Optional |
+| --------------------- | ---------------------------- | -------- |
+| `broadcaster_sockets` | `number`                     | No       |
+| `embedded_runner`     | `boolean`                    | No       |
+| `queue`               | `{ run_queue_length: number` | No       |
```
pabo99: It seems this is a case of nested Objects/Lists. We should check whether the value of `queue` can be converted into a `PSALM` type, which would be this: 
```php
queue: array{running: list<array{name: string, id: int}
``` 
to:
```php
queue: QueueInfo
``` 
where:
```php
QueueInfo=array{running: list<array{name: string, id: int}>
``` 
And with this, you would apply the same principle as in the previous case, where we show the `JSON` structure directly

frontend/server/src/Controllers/README.md
```diff
@@ -3509,6 +3554,15 @@ Returns the solution for a problem if conditions are satisfied.
 | ---------- | ------------------------ |
 | `solution` | `types.ProblemStatement` |
 
+**`types.ProblemStatement` fields:**
+
+| Name       | Type                      | Optional |
+| ---------- | ------------------------- | -------- |
+| `images`   | `{ [key: string]: string` | No       |
```
pabo99: For this particular case we can use something like:


| Name       | Type                      | 
|-|-|
| `images`   | Record<string, string> | 

Or even better, replace in the code:
```php
images: array<string, string>
```

with:
```php
images: ImageMap
```

where
```php
ImageMap=array<string, string>
```
And in the end, we reuse the functionality of JSON structures

frontend/server/src/Controllers/README.md
```diff
@@ -4386,7 +4440,28 @@ contestant's machine and the server.
...
 | --------- | ---------------------- |
 | `session` | `types.CurrentSession` |
-| `time`    | `number`               |
+
+**`types.CurrentSession` fields:**
+
+| Name                         | Type                         | Optional |
+| ---------------------------- | ---------------------------- | -------- |
+| `apiTokenId`                 | `number`                     | Yes      |
+| `api_tokens`                 | `types.ApiToken[]`           | No       |
```
pabo99: - For these, could you identify the `[]` pattern so that they render as:
```suggestion
| `api_tokens`                 | List[`types.ApiToken`]           | No       |
```

Prashant-thakur77:
@pabo99 I have updated and apllied the changes requested by you.please review when you have time and tell if any changes are required and (i have added custom formatting as when i was using the default i.e how it was done before the table alignment was not proper ).

frontend/server/cmd/APITool.php
Prashant-thakur77: Done - Changed "Optional" column to "Required" with ✓ symbol for required fields. Much clearer now!

frontend/server/cmd/APITool.php
Prashant-thakur77: Done - Using ✓ for required fields. Applied to both return type tables AND request parameter tables.

frontend/server/src/Controllers/README.md
Prashant-thakur77: Complex types like UserProfileContests now show as formatted JSON blocks instead of messy tables,But still you might check for any furhter changes in the current implementation

frontend/server/src/Controllers/README.md
Prashant-thakur77: Nested objects like queue show as formatted JSON structures for better readability

frontend/server/src/Controllers/README.md
Prashant-thakur77: types.ApiToken[] now shows as List[types.ApiToken] across all documentation

frontend/server/src/Controllers/README.md
pabo99: For this PR it's excellent! Thank you for applying the suggested changes!

frontend/server/src/Controllers/README.md
```diff
@@ -2219,16 +2234,22 @@ Returns details of a given course
...
 | Name          | Type                       |
 | ------------- | -------------------------- |
 | `assignments` | `types.AssignmentProgress` |
 
+**`types.AssignmentProgress` fields:**
+
+| Name   | Type                      | Required |
+| ------ | ------------------------- | -------- |
+| `[key` | `string]: types.Progress` | ✓        |
+
```
pabo99: A formatted JSON block should have been displayed here, right? Just like in `types.UserProfileContests`

frontend/server/src/Controllers/README.md
```diff
@@ -4451,16 +4517,37 @@ contestant's machine and the server.
 
 ### Parameters
 
-| Name         | Type           | Description |
-| ------------ | -------------- | ----------- |
-| `auth_token` | `null\|string` |             |
+| Name         | Type           | Description | Required |
+| ------------ | -------------- | ----------- | -------- |
+| `auth_token` | `null\|string` |             |          |
```
pabo99, line 4524: Just a suggestion: 

Since you already added the Required column, we can omit `null\` in the  Type column. However, if you think it provides more context for developers, you can leave it as is—no problem

frontend/server/src/Controllers/README.md
Prashant-thakur77: updated as needed.I missed it before in the logic 
Screenshot From 2026-01-24 11-03-00

frontend/server/src/Controllers/README.md
Prashant-thakur77: I recommend keeping the null prefix because it provides important semantic information distinct from the Required column the Type shows what values a field can hold, while Required indicates if it must be present. For example, required: true, type: null string means the field must be present but can be null, versus required: false, type: string which is optional but must be string when present. This complete type information is consistent with TypeScript conventions and helps developers understand the API without cross-referencing columns.

BUt still i would like your insight on it and will do so for the pr.

Prashant-thakur77:
> This updated version is much more detailed than the previous one. The suggestions I left in the comments, aside from the text alignment issues you had already identified, can be left for future changes.
> 
> I truly appreciate all the effort you put into implementing these improvements!

I have updated with the new changes:) Please review when you have time.

PR #8764, opened 2026-01-13
Frontend for carousel management

### LINKED ISSUE #8765: [FEATURE] Create a ui for managing the carousels.
Recently the APIs have been made for carousel management. The UI for this needs to be developed.
This new feature will only be available for the site admin.
Using this we will be able to make the carousel on the homepage dynamic, which currently serves static data from the static file.

## Context
Issue: #6916

iqbalcodes6602:
# Description
- i created the api `getCarouselManagementDetailsForTypeScript` for getting the list of items when for the new page, which is being called in this file `frontend/www/admin/carousel.php`
- configured the `nginx.rewrites` to render the component on this route `/admin/carousel.php` and in the frontend `webpack.config-frontend.js`
- i then added another entry in the navbar for the Carousal managemnt for admins
- added the `frontend/www/js/omegaup/admin/carousel.ts` file for api calls
- `frontend/www/js/omegaup/components/admin/Carousel.vue` is the main component page for carousal management.
- the exisitng carousal on the home page, have replaced the static content to dynamic `frontend/www/js/omegaup/components/homepage/Carousel.vue`.

Fixes: #8765 

# Screenshots
Screenshot 2026-01-13 at 7 42 05 PM

image

Screenshot 2026-01-13 at 7 37 10 PM

image

# Checklist:


frontend/server/src/Controllers/CarouselItems.php
```diff
@@ -173,4 +175,44 @@ public static function apiListActive(\OmegaUp\Request $r): array {
...
+                    'carouselItems' => array_map(
+                        fn(\OmegaUp\DAO\VO\CarouselItems $item): array => [
+                            'carousel_item_id' => $item->carousel_item_id ?? 0,
+                            'title' => $item->title ?? '',
+                            'excerpt' => $item->excerpt ?? '',
+                            'image_url' => $item->image_url ?? '',
+                            'link' => $item->link ?? '',
+                            'button_title' => $item->button_title ?? '',
+                            'expiration_date' => $item->expiration_date,
+                            'status' => $item->status == 'active'
```
pabo99: If this is going to return a boolean, can it be called `is_active` instead?

frontend/server/src/Controllers/CarouselItems.php
```diff
@@ -173,4 +175,44 @@ public static function apiListActive(\OmegaUp\Request $r): array {
...
+            'templateProperties' => [
+                'payload' => [
+                    'carouselItems' => array_map(
+                        fn(\OmegaUp\DAO\VO\CarouselItems $item): array => [
+                            'carousel_item_id' => $item->carousel_item_id ?? 0,
+                            'title' => $item->title ?? '',
+                            'excerpt' => $item->excerpt ?? '',
+                            'image_url' => $item->image_url ?? '',
+                            'link' => $item->link ?? '',
+                            'button_title' => $item->button_title ?? '',
```
pabo99: Could you keep the value obtained from the query? It's preferable to get null values rather than returning magic ones

frontend/templates/en.lang
```diff
@@ -2005,6 +2025,7 @@ wordsCopyToClipboard = "Copy to clipboard"
 wordsCountry = "Country"
 wordsCountryRegion = "Country / Region"
 wordsCourseContent = "Course content"
+wordsCreate = "Create"
```
pabo99: Can you add the prefix `carousel` instead of `word`? The same applies to the rest of the text strings with that prefix

frontend/templates/es.lang
```diff
@@ -177,7 +177,23 @@ cannotAddProb = "No puedes agregar problemas porque no eres el director del conc
 cannotEditUngroupedGroup = "No es posible editar un grupo sin agrupar"
 cannotRemoveProblem = "No puedes eliminar problemas porque no eres el director del concurso."
 cannotRemoveProblemWithSubmissions = "No puedes eliminar un problema que ya tiene envíos"
+carouselButtonTitle = "Título del botón"
+carouselCreateNew = "Crear nuevo elemento del carrusel"
+carouselDeleteConfirm = "Item do carrossel de arquivos"
```
pabo99:
```suggestion
carouselDeleteConfirm = "Archivar elemento del carrusel"
```

frontend/templates/es.lang
```diff
@@ -1974,6 +1993,7 @@ wordsAlias = "Alias"
 wordsAll = "Todos"
 wordsAnswer = "Respuesta"
 wordsAnyLanguage = "Cualquier Lenguaje"
+wordsArchive = "Archivo"
```
pabo99:
```suggestion
wordsArchive = "Archivar"
```

frontend/templates/es.lang
```diff
@@ -2033,6 +2054,7 @@ wordsEnglish = "Inglés"
 wordsExam = "Examen"
 wordsExecution = "Ejecución"
 wordsExperiments = "Experimentos"
+wordsExpire = "Expirar"
```
pabo99:
```suggestion
wordsExpire = "Expirado"
```

frontend/templates/en.lang
```diff
@@ -2033,6 +2054,7 @@ wordsEnglish = "English"
 wordsExam = "Exam"
 wordsExecution = "Execution"
 wordsExperiments = "Experiments"
+wordsExpire = "Expire"
```
pabo99:
```suggestion
wordsExpire = "Expired"
```

frontend/www/js/omegaup/components/common/Navbar.vue
```diff
@@ -244,6 +244,14 @@
...
                       T.navMyQualityNomination
                     }}</a>
+                    <div v-if="isAdmin" class="dropdown-divider"></div>
+                    <a
+                      v-if="isAdmin"
+                      class="dropdown-item"
+                      href="/admin/carousel/"
+                      data-nav-carousel-management
+                      >{{ T.navCarouselManagement }}</a
+                    >
```
pabo99: Could you remove this option from the menu? It's already too cluttered. The best approach will be to plan how to include the administrative activities in a new menu option, but for now, let’s leave it out

frontend/server/src/Controllers/CarouselItems.php
```diff
@@ -173,4 +175,44 @@ public static function apiListActive(\OmegaUp\Request $r): array {
...
+    ): array {
+        $r->ensureMainUserIdentity();
+        if (!\OmegaUp\Authorization::isSystemAdmin($r->identity)) {
+            throw new \OmegaUp\Exceptions\ForbiddenAccessException();
+        }
+
+        return [
+            'templateProperties' => [
+                'payload' => [
+                    'carouselItems' => array_map(
```
pabo99: I see that this array mapping is now being used in at least three different places. How about adding a helper function that handles retrieving the carousel items?

It might need a parameter to indicate whether to return the full list or only the active items.

With this change, `apiListActive` would only return this list. What do you think about merging it with `apiList`, and having this accept an optional `active_only` parameter?

frontend/www/js/omegaup/components/homepage/Carousel.vue
```diff
@@ -60,8 +84,109 @@ import carouselConfig from '../../carousel.config';
...
+        // Fallback to empty array if no slides
+        if (this.slides.length === 0) {
+          this.slides = [];
+        }
+      })
+      .catch((error) => {
+        console.error('Error loading carousel items:', error);
+        // On error, set empty array (carousel won't display)
+        this.slides = [];
+      });
```
pabo99: Normally, we try to make all API calls from `.ts` files, which helps keep the data layer isolated from the components. 

How about moving this call to `carousel.ts` and executing it at the start of the script, so that when the Vue instance is created, the list of slides is already populated?

frontend/www/js/omegaup/components/homepage/Carousel.vue
```diff
@@ -60,8 +84,109 @@ import carouselConfig from '../../carousel.config';
...
   T = T;
-  // Reverse the entries so that newer ones appear first.
-  slides = carouselConfig.reverse();
+  slides: SlideData[] = [];
+
+  mounted(): void {
+    this.loadCarouselItems();
+  }
+
+  parseJsonField(field: string | undefined): { [key: string]: string } {
```
pabo99: Is there a way to avoid using `undefined`?

frontend/www/js/omegaup/components/admin/Carousel.vue
```diff
@@ -0,0 +1,597 @@
...
+    <div class="card-header">
+      <h2 class="card-title">{{ T.omegaupTitleCarouselManagement }}</h2>
+    </div>
+    <div class="card-body">
+      <div class="mb-2">
+        <label class="mr-2">{{ T.wordsLanguage }}:</label>
+        <select
+          v-model="currentLanguage"
+          class="form-control d-inline-block"
+          style="width: auto"
```
pabo99: Could you avoid using inline styles? 

Instead, you can apply a [Bootstrap 4](https://getbootstrap.com/docs/4.1/utilities/sizing/) class, it looks like `w-auto` has the same effect

frontend/www/js/omegaup/components/admin/Carousel.vue
```diff
@@ -0,0 +1,597 @@
...
+        >
+          <option value="en">English</option>
+          <option value="es">Español</option>
+          <option value="pt">Português</option>
```
pabo99: Can you use localized strings here as well?

frontend/www/js/omegaup/components/admin/Carousel.vue
```diff
@@ -0,0 +1,597 @@
...
+  truncateText(text: string, maxLength: number): string {
+    if (!text || text.length <= maxLength) {
+      return text || '';
+    }
+    return text.substring(0, maxLength) + '...';
+  }
+
+  formatDate(date: Date): string {
+    return new Date(date).toLocaleString();
+  }
```
pabo99: Could you use the existing `time.formatDateTime()` function here instead of creating a new one?

frontend/www/js/omegaup/components/admin/Carousel.vue
```diff
@@ -0,0 +1,597 @@
...
+
+  formatDateForInput(date: Date): string {
+    const d = new Date(date);
+    const year = d.getFullYear();
+    const month = String(d.getMonth() + 1).padStart(2, '0');
+    const day = String(d.getDate()).padStart(2, '0');
+    const hours = String(d.getHours()).padStart(2, '0');
+    const minutes = String(d.getMinutes()).padStart(2, '0');
+    return `${year}-${month}-${day}T${hours}:${minutes}`;
+  }
```
pabo99: Could you check if there is a function in the `time` library that retrieves this string? If not, you could move this function into that file.

frontend/www/js/omegaup/components/admin/Carousel.vue
```diff
@@ -0,0 +1,597 @@
...
+          <tr v-if="filteredCarouselItems.length === 0">
+            <td colspan="9" class="text-center">
+              {{ T.carouselNoItems }}
+            </td>
+          </tr>
+        </tbody>
+      </table>
+    </div>
+
+    <!-- Create/Edit Modal -->
```
pabo99, line 118: We're not big fans of using modals, but in this case we can keep it since it's an administrative task. 

However, could you try implementing the [Bootstrap-Vue modal plugin](https://bootstrap-vue.org/docs/components/modal/)?
This way, the code stays cleaner because many actions are delegated to the plugin

frontend/templates/en.lang
iqbalcodes6602: okay done.

frontend/templates/es.lang
iqbalcodes6602: done

frontend/templates/es.lang
iqbalcodes6602: done

frontend/templates/es.lang
iqbalcodes6602: done

frontend/templates/en.lang
iqbalcodes6602: done

frontend/www/js/omegaup/components/common/Navbar.vue
iqbalcodes6602: sure we can remove it for now, and plan later on how to put it.

frontend/www/js/omegaup/components/admin/Carousel.vue
iqbalcodes6602: sure, will do that.

frontend/www/js/omegaup/components/admin/Carousel.vue
iqbalcodes6602: sure done.

frontend/www/js/omegaup/components/admin/Carousel.vue
iqbalcodes6602: yes, thanks.

frontend/www/js/omegaup/components/admin/Carousel.vue
iqbalcodes6602: thank you for pointing out, we already have a similar function. I will use that.

frontend/server/src/Controllers/CarouselItems.php
iqbalcodes6602: sure done.

frontend/server/src/Controllers/CarouselItems.php
iqbalcodes6602: yes we can implement this entire thing.

frontend/www/js/omegaup/components/admin/Carousel.vue
iqbalcodes6602: yes we can do that

frontend/www/js/omegaup/components/homepage/Carousel.vue
iqbalcodes6602: sure done

frontend/server/src/Controllers/CarouselItems.php
iqbalcodes6602: done

frontend/server/src/Controllers/CarouselItems.php
```diff
@@ -132,45 +180,47 @@ public static function apiList(\OmegaUp\Request $r): array {
...
-                    'title' => $item->title ?? '',
-                    'excerpt' => $item->excerpt ?? '',
-                    'image_url' => $item->image_url ?? '',
-                    'link' => $item->link ?? '',
-                    'button_title' => $item->button_title ?? '',
-                    'expiration_date' => $item->expiration_date,
-                    'status' => $item->status == 'active'
-                ],
-                \OmegaUp\DAO\Base\CarouselItems::getAll()
+            'carouselItems' => self::transformCarouselItems(
```
pabo99, line 189: I see that transformCarouselItems filters active elements when needed, but $items also handles filtering. Is there a way to improve the logic to avoid duplication of responsibilities?

frontend/www/js/omegaup/admin/carousel.ts
```diff
@@ -0,0 +1,89 @@
...
+              .catch(ui.apiError);
+          },
+          'update-item': (item: types.CarouselItem): void => {
+            api.CarouselItems.update({
+              carousel_item_id: item.carousel_item_id,
+              title: item.title,
+              excerpt: item.excerpt,
+              image_url: item.image_url,
+              link: item.link,
+              buttonTitle: item.button_title,
```
pabo99, line 59: nit:
```suggestion
              button_title: item.button_title,
```
for consistency with the rest of the parameters

frontend/www/js/omegaup/admin/carousel.ts
```diff
@@ -0,0 +1,89 @@
...
+              expiration_date: item.expiration_date
+                ? item.expiration_date.toISOString()
```
pabo99, line 61: Could you use `formatDateLocal()` instead of `toISOString()`. I noticed there’s a note in that function suggesting a different format

frontend/www/js/omegaup/components/homepage/Carousel.vue
```diff
@@ -60,8 +84,111 @@ import carouselConfig from '../../carousel.config';
...
+        if (this.slides.length === 0) {
+          this.slides = [];
+        }
+      })
+      .catch((error) => {
+        console.error('Error loading carousel items:', error);
+        // On error, set empty array (carousel won't display)
+        this.slides = [];
+      });
+  }
```
pabo99, line 191: If this is always loaded from the start, could we fetch it directly from the `payload` instead? That would let us avoid using `mounted` here and also prevent making API calls from within the component

frontend/www/js/omegaup/components/homepage/Carousel.vue
```diff
@@ -51,7 +51,31 @@
 import { Vue, Component } from 'vue-property-decorator';
 import T from '../../lang';
 import homepageSlide from './Slide.vue';
-import carouselConfig from '../../carousel.config';
```
pabo99, line 54: This file will no longer be used. It would be worth removing it

frontend/www/js/omegaup/components/admin/Carousel.vue
```diff
@@ -0,0 +1,496 @@
...
+        >
+          <option value="en">{{ T.wordsEnglish }}</option>
+          <option value="es">{{ T.wordsSpanish }}</option>
+          <option value="pt">{{ T.wordsPortuguese }}</option>
+        </select>
+        <label class="mr-2 ml-3">{{ T.wordsStatus }}:</label>
+        <select
+          v-model="currentStatusFilter"
+          class="form-control d-inline-block"
+          style="width: auto"
```
pabo99, line 21: This needs to be added as a Bootstrap 4 class

frontend/www/js/omegaup/components/admin/Carousel.vue
```diff
@@ -0,0 +1,496 @@
...
+                  50,
+                )
+              }}
+            </td>
+            <td>
+              <img
+                v-if="item.image_url"
+                :src="item.image_url"
+                :alt="getMultilingualText(item.title, currentLanguage)"
+                style="max-width: 100px; max-height: 60px"
```
pabo99, line 68: This should also be added as a class; if there’s something in Bootstrap that achieves a similar result, even better

frontend/www/js/omegaup/components/admin/Carousel.vue
```diff
@@ -0,0 +1,496 @@
...
+        cancelTitle: T.wordsCancel,
+        okVariant: 'danger',
+        hideHeaderClose: false,
+        centered: true,
+      },
+    );
+    if (confirmed) {
+      this.$emit('delete-item', item.carousel_item_id);
+    }
+  }
```
pabo99, line 388: Would it make sense to avoid using `(this as any).$bvModal` here? A declarative confirmation modal controlled by local state might be easier to type and would align better with other confirmation flows in the frontend. I mean, you could use a new b-modal right behind the modal with the carousel form:

Would it make sense to avoid using `(this as any).$bvModal` here? A more declarative confirmation modal controlled via local state could be easier to type and would align better with other confirmation flows in the frontend.

For example, you could define a `b-modal` alongside the carousel form:

```vue
    <b-modal
      v-model="showDeleteConfirmationModal"
      :title="T.carouselDeleteConfirm"
      :ok-title="T.carouselArchive"
      :cancel-title="T.wordsCancel"
      ok-variant="danger"
      :hide-header-close="false"
      centered
      @ok="deleteSelectedItem"
      @hidden="onDeleteConfirmationHidden"
    >
      <p>{{ T.carouselDeleteMessage }}</p>
    </b-modal>
```

Then manage its visibility with local state:
```ts 
  showDeleteConfirmationModal = false;
```
The `deleteSelectedItem` method could simply emit the action:
 ```ts
  deleteSelectedItem(): void {
    if (!this.selectedItemToDelete) {
      return;
    }
    this.$emit('delete-item', this.selectedItemToDelete.carousel_item_id);
    this.selectedItemToDelete = null;
  }
```
You could also introduce an intermediate state to track the item being deleted, setting it when the modal opens and clearing it on confirm or cancel:
```ts
  selectedItemToDelete: types.CarouselItem | null = null;
```

And the `confirmDelete` function could be renamed to better reflect this intermediate step, since it would only select the item and display the confirmation modal, something like:
```ts
  confirmDelete(item: types.CarouselItem): void {
    this.selectedItemToDelete = item;
    this.showDeleteConfirmationModal = true;
  }
```
And finally, `onDeleteConfirmationHidden` would be used to clear the selected item state, regardless of how the modal is closed—whether via keyboard shortcuts, clicking outside the modal, or any other allowed interaction.
```ts
  onDeleteConfirmationHidden(): void {
    this.selectedItemToDelete = null;
  }
```

frontend/www/js/omegaup/components/admin/Carousel.vue
```diff
@@ -0,0 +1,496 @@
...
+    } else {
+      this.currentItem.expiration_date = undefined;
+    }
+
+    if (this.isEditing) {
+      this.$emit('update-item', this.currentItem);
+    } else {
+      this.$emit('create-item', this.currentItem);
+    }
+    (this.$refs.carouselItemModal as any)?.hide();
```
pabo99, line 413: To avoid casting to `any`, we could define a proper interface for the modal. We can also improve this by using the `@Ref` decorator from `vue-property-decorator`.

For example, you could declare a reference to the modal like this:
```ts
  @Ref('carouselItemModal') readonly carouselItemModal!: ModalRef;
```
Then define a `ModalRef` interface with the methods you actually use:
```ts
interface ModalRef {
  show(): void;
  hide(): void;
}
``` 
With this in place, you can call the modal methods directly without casting:
```suggestion
    this.carouselItemModal.hide();
```
This removes the need for `(this as any)` and keeps the code properly typed and easier to maintain

frontend/www/js/omegaup/components/admin/Carousel.vue
```diff
@@ -0,0 +1,496 @@
...
+      image_url: '',
+      link: '',
+      button_title: '',
+      expiration_date: undefined,
+      is_active: true,
+    };
+  }
+
+  showItemModal(): void {
+    (this.$refs.carouselItemModal as any)?.show();
```
pabo99, line 286: Same case here:
```suggestion
    this.carouselItemModal.show();
```

frontend/www/js/omegaup/components/admin/Carousel.vue
```diff
@@ -0,0 +1,496 @@
...
+
+  onItemModalOk(evt: any): void {
+    const form = this.$refs.carouselForm as HTMLFormElement | undefined;
+    if (form && !form.checkValidity()) {
+      evt.preventDefault();
+      form.reportValidity();
+      return;
+    }
+    this.saveItem();
+  }
```
pabo99, line 435: To stay consistent with avoiding the use of `any`, we can use the proper event type (`Event`). We can also take advantage of the `@Ref` decorator, assuming it has already been exported, to reference the form element.

For example, it could be declared like this:
```ts
  @Ref('carouselForm') readonly carouselForm!: HTMLFormElement;
```

Then we can use it directly like this:

```suggestion
  onItemModalOk(evt: Event): void {
    if (!this.carouselForm.checkValidity()) {
      evt.preventDefault();
      this.carouselForm.reportValidity();
      return;
    }
    this.saveItem();
  }
```
And we can also get rid of that annoying `undefined`. It would be good to consider avoiding it throughout the rest of the component as well, if possible.

frontend/server/src/Controllers/CarouselItems.php
```diff
@@ -5,22 +5,58 @@
...
+        /** @var \OmegaUp\DAO\VO\CarouselItems[] $filteredItems */
+        $filteredItems = array_values($filteredItems);
+
+        /** @var list<CarouselItem> */
+        return array_map(
+            fn(\OmegaUp\DAO\VO\CarouselItems $item): array => [
+                'carousel_item_id' => $item->carousel_item_id,
+                'title' => $item->title,
+                'excerpt' => $item->excerpt,
+                'image_url' => $item->image_url,
```
pabo99, line 37: For practical purposes, this works as a temporary solution. However, we should consider a better way to manage image uploads to have more control over them.

I’m thinking about using an `Amazon S3 bucket` for this, but we can leave that for a separate issue so we don’t block this PR. Does that sound good?

frontend/templates/es.lang
```diff
@@ -183,7 +183,29 @@ cannotAddProb = "No puedes agregar problemas porque no eres el director del conc
...
+carouselDeleteMessage = "¿Está seguro de que desea archivar este elemento del carrusel?"
+carouselEditItem = "Editar elemento del carrusel"
+carouselExcerpt = "Extracto"
+carouselExpirationDate = "Fecha de expiración"
+carouselExpirationDateHint = "Dejar vacío para que no expire"
+carouselExpire = "Expirado"
+carouselImageUrl = "URL de la imagen"
+carouselInactive = "Inactivo"
+carouselItemCreated = "Elemento del carrusel creado correctamente"
+carouselItemDeleted = "Elemento del carrusel eliminado correctamente"
```
pabo99, line 201: For consistency with the rest of the wording:
```suggestion
carouselItemArchived = "Elemento del carrusel archivado correctamente"
```

frontend/www/js/omegaup/components/admin/Carousel.vue
```diff
@@ -0,0 +1,496 @@
...
+        </select>
+        <label class="mr-2 ml-3">{{ T.wordsStatus }}:</label>
+        <select
+          v-model="currentStatusFilter"
+          class="form-control d-inline-block"
+          style="width: auto"
+        >
+          <option value="active">{{ T.carouselActive }}</option>
+          <option value="archived">{{ T.carouselArchive }}</option>
+          <option value="all">{{ T.wordsAll }}</option>
```
pabo99, line 25: Would it be possible to use dedicated strings for the statuses instead? It looks like some strings are being reused across different contexts, which can lead to incorrect translations in all supported languages.

We should define specific keys for these statuses to ensure proper translations per context. For example:
```suggestion
          <option value="active">{{ T.carouselStatusActive }}</option>
          <option value="archived">{{ T.carouselStatusArchived }}</option>
          <option value="all">{{ T.carouselStatusAll }}</option>
```
And then provide the appropriate translations for each language:
```ts
en: "Active" | "Archived" | "All"
es: "Activo" | "Archivado" | "Todos"
pt: "Ativo" | "Arquivado" | "Todos"
```

frontend/www/js/omegaup/components/admin/Carousel.vue
```diff
@@ -0,0 +1,496 @@
...
+                <font-awesome-icon :icon="['fas', 'edit']" />
+              </button>
+            </td>
+            <td>
+              <button
+                class="btn btn-sm btn-danger"
+                @click.prevent="confirmDelete(item)"
+              >
+                <font-awesome-icon :icon="['fas', 'archive']" />
+              </button>
```
pabo99, line 106: This button should be hidden when the user is viewing archived items. Could you add that condition?

pabo99 (commented):
Overall, everything looks good and is working as expected.

For future PRs, it would be great to avoid changes of this size, as they make reviews more time-consuming and harder to follow. Ideally, changes could be split into smaller PRs, for example, separating backend and frontend work, or even breaking things down further by components or responsibilities.

With that in mind, and once the suggested changes are addressed, this should be good to merge.

pabo99 (approved):

Issue #8773, opened 2026-01-14
[FEATURE] Add browser back button support for contest filter


Ankitsinghsisodya:
## Title

As a **user**, I want **the browser back button to restore my previous contest filter, sort order, and tab selections** so that **I can easily navigate back to my previous view without manually reselecting all my options**.

## Context

When browsing the contests page (`/arena/`), users can apply various filters, sort orders , and switch between tabs (Current, Future, Past). However, when a user changes these options and then presses the browser's back button, the page doesn't restore the previous state properly.

**Current behavior:**

- Selecting a filter/sort option updates the URL query parameters
- Pressing browser back button navigates to the previous URL
- But the Vue component's internal state doesn't sync with the restored URL parameters
- The UI shows stale or inconsistent state compared to the URL

**Expected behavior:**

- When the user clicks the browser back button, the page should restore the exact filter, sort order, tab, and search query from the previous URL
- The contest list should refresh to reflect the restored state
- This provides a seamless navigation experience consistent with standard web behavior

**Impact:**

- Users lose their previous filter/sort selections when navigating back
- This forces users to manually reselect their options, leading to a poor UX
- This is especially frustrating when exploring different filter combinations

**Technical notes:**

- The component (`ContestListv2.vue`) already has watchers to sync internal state when props change
- The solution requires listening to `popstate` events and updating the Vue app's state accordingly
- The parent entry point (`contest_listv2.ts`) needs to handle the `popstate` event and pass updated props to the component

Ankitsinghsisodya:
https://github.com/user-attachments/assets/c8b87185-c375-42fd-ba4b-1816cdd514b4

Ankitsinghsisodya:
@pabo99 sir, I have fixed this locally. Sir should I raise a PR to fix this?

pabo99:
Sure!

PR #8776, opened 2026-01-15
[FEATURE] Add browser back button support for contest filter

Ankitsinghsisodya:
# Description

Add browser back/forward button support for contest filter, sort order, and tab selections on the `/arena/` page.

**Changes made:**

- Extract URL parsing logic into reusable `parseUrlState()` function
- Make `tab`, `page`, `sortOrder`, `filter` reactive Vue data properties
- Add `popstate` event listener to handle browser back/forward navigation
- Add prop watchers in `ContestListv2.vue` to sync internal state when props change

**Problem:** When users changed filters/sort options and pressed the browser's back button, the URL would change but the Vue component's internal state wouldn't sync with the restored URL parameters, resulting in stale or inconsistent UI state.

**Solution:** Added a `popstate` event listener that parses URL parameters and updates the Vue instance's reactive data properties. The component's existing watchers automatically detect prop changes and refresh the contest list.

Fixes: #8773

# Checklist:


Ankitsinghsisodya:
I added ESLint disable comment for vue/no-deprecated-destroyed-lifecycle rule since the codebase uses Vue 2, where beforeDestroy is the correct lifecycle hook (beforeUnmount is Vue 3 only).

Ankitsinghsisodya:
@pabo99 sir, could you please review this PR also whenever you are free.

pabo99 (commented):
The change looks good. There's only one case that doesn't work as expected:

https://github.com/user-attachments/assets/39fa4574-0d2b-4aa3-8587-b24f5f8c0394

It's not correctly detecting the hash change.

We should check whether the `popstate` event is triggered correctly when the hash changes.

Ankitsinghsisodya:
@pabo99 I have fixed that bug.

PR #8788, opened 2026-01-17
Add book link to Help dropdown: Estructura de Datos y sus aplicaciones en Programación Competitiva

### LINKED ISSUE #8786: Agregar link de descarga de libro "Estructura de Datos y sus aplicaciones en Programación Competitiva" en el menú de ayuda
Recientemente nos compartieron un libro muy completo el cual sería interesante compartir.

El libro se puede descargar de forma gratuita desde la página de la Editorial UTP en el enlace: https://hdl.handle.net/11059/16567

El autor del libro hizo énfasis en compartir el link de descarga en lugar del archivo en PDF. Entonces ese link es el que se deberá agregar.

El autor es Hugo Humberto Morales Peña

anuragdubey137:
### Description
Added a new link to the Help dropdown menu:

- "Estructura de Datos y sus aplicaciones en Programación Competitiva"
- Opens in a new tab
- Styling matches existing items

### Screenshot
Help dropdown with new book link
Help dropdown with new book link 2

### Fixes
Closes #8786


frontend/www/js/omegaup/components/common/NavbarItems.vue
```diff
@@ -224,6 +224,13 @@
             target="_blank"
             >{{ T.navAlgorithmsBook }}</a
           >
+          <a
+            class="dropdown-item text-wrap"
+            href="https://hdl.handle.net/11059/16567"
+            target="_blank"
+          >
+            Estructura de Datos y sus aplicaciones en Programación Competitiva
```
pabo99, line 232: For consistency, could you add this string to the localization files? Similar to line 225

frontend/www/js/omegaup/components/common/NavbarItems.vue
```diff
@@ -224,6 +224,13 @@
...
+          </a>
```
pabo99, line 233: The author’s name will also need to be added.

Since this will be a lot of text, is there a way to extend the width of this menu? Maybe it could behave like the notifications dropdown, which expands based on the content being displayed.

pabo99 (commented):
Just a couple of suggestions

frontend/www/js/omegaup/components/common/NavbarItems.vue
anuragdubey137: Thanks for the suggestion! 
I’ll move this string to the localization files for consistency, similar to line 225, and update the component accordingly.

frontend/www/js/omegaup/components/common/NavbarItems.vue
anuragdubey137: Yes, I was thinking along the same lines .

Since the author’s name adds more text, I agree that the Help dropdown should be able to expand based on its content. I’ll follow a similar approach to the notifications dropdown by allowing the menu width to grow naturally (instead of being constrained to the default Bootstrap width), while still keeping reasonable max-width limits and proper text wrapping.

I’ll add the author’s name to the localized string as well and adjust the dropdown styling accordingly.

pabo99: In a follow-up change, we can modernize our navbar so that the options include:

- Icon
- Title
- Subtitle

Here's an example:

Captura de pantalla 2026-01-19 a la(s) 1 41 48 p m


This way we could optimize space a bit and the menu would provide more context. For now, we can apply what I suggested before

anuragdubey137: Thanks for the suggestion! I agree that this structure would provide better context and optimize space. For now, I’m working on the earlier suggestion we discussed and will raise a PR for that shortly. We can revisit this improved dropdown structure in a follow-up if it still makes sense.

anuragdubey137: Hi @pabo99, I updated NavbarItems.vue to use the new translation key instead of a hardcoded string. I’ve added it to all lang.*.json files, but after rebuilding, it’s not showing on the local frontend. Could you advise what I might be missing in the generation process?

pabo99: I still don't see the changes you mentioned. Did you forget to push your changes?

pabo99:
Closed in favor of #8895

Issue #8529, opened 2025-11-15
[FEATURE] login and sign up present on same page


impakshal:
Login and Signup lie on same page.. For signing up a user needs to scroll down to signup div
<img width="1158" height="806" alt="Image" src="https://github.com/user-attachments/assets/f9d8096e-c5a2-4cbc-a026-54598a5818ed" />

Rather login and sign up pages can be seperated as below image shows. Making things convienient for user. It is also a good development practice
<img width="637" height="430" alt="Image" src="https://github.com/user-attachments/assets/9600aecc-afea-45a9-a570-51487b0b2f92" />

impakshal:
Hello @pabo99  The above issue was created as i thought login and sign up structure can be improved. If you like the suggestion i will be more than happy to work upon it

shubham220420:
hey @impakshal If you dont mind can I work on this , I am willing to contribute on this, thanks :)
cc: @pabo99

aryan7071:
Hi! I would like to work on this issue. I have previously worked on similar projects, and I’m confident I can contribute effectively here. I’d love to take this up — please assign it to me if it’s available.

Ptmishra69:
hey can i be assigned for this issue

pabo99:
> Hello [@pabo99](https://github.com/pabo99) The above issue was created as i thought login and sign up structure can be improved. If you like the suggestion i will be more than happy to work upon it

Hello, @impakshal!  Thanks so much for your interest in contributing to omegaUp!

I really like the idea of modernizing this view! I just have a couple of suggestions to keep in mind:
- Try to stick to the official color palette
- Consider including space for the Google login option — with this new layout, you could even keep it visible across both components

impakshal:
Sure.. Will soon raise a PR for the same.. Thanks

amanpandey2587:
Hi @pabo99 , since there has no pull request created for this issue and stale marking has been added as well , could you please assign this issue to me.
Thanks

pabo99:
Sure, thanks for your interest in working on this, @amanpandey2587!
  
I’ll go ahead and assign the issue to you so you can proceed.  

If you run into any difficulties, just let me know and we can reassign it if needed.

PR #8811, opened 2026-01-19
chore: implementing login signup tab

amanpandey2587:
# Description
Merged the loginIn and signUp functionality into a toggle state block for clean UI.

Earlier View: 
image

Current View:
image
image

Fixes: #8529

# Comments
Manual Testing of the login and signup functionality was done thoroughly.

# Checklist:


frontend/www/js/omegaup/components/login/Signin.vue
```diff
@@ -1,18 +1,62 @@
...
-    <omegaup-signup
-      :has-visited-section="hasVisitedSection"
-      :validate-recaptcha="validateRecaptcha"
-      :use-signup-form-with-birth-date="useSignupFormWithBirthDate"
-      @register-and-login="(request) => $emit('register-and-login', request)"
-    >
-    </omegaup-signup>
+    <ul class="nav nav-tabs mb-3" role="tablist">
+      <li class="nav-item" role="presentation">
+        <button
```
pabo99: Could you add fragment (hash) navigation here? That way a user could access the "Sign up" or "Log in" view directly (for example, via URLs like `#signup` or `#login`). Also, when reloading the page the user would remain on the same form/tab. 

Here's an example:
https://github.com/omegaup/omegaup/blob/af35573e6d0a4141830fa730484aff21ffec434f/frontend/www/js/omegaup/components/group/Edit.vue#L9-L28

frontend/www/js/omegaup/components/login/Signin.vue
```diff
@@ -1,18 +1,62 @@
...
+          class="nav-link"
+          :class="{ active: activeTab === 'login' }"
```
pabo99: Can you use an `enum` for the different types of tabs that will be added? This way we reduce the risk of making typos

frontend/www/js/omegaup/components/login/Signin.vue
```diff
@@ -38,5 +82,55 @@ export default class Signin extends Vue {
   @Prop({ default: false }) useSignupFormWithBirthDate!: boolean;
 
   T = T;
+  activeTab: string = 'login';
 }
 </script>
+
+<style scoped>
+.nav-tabs {
+  border-bottom: 1px solid #dee2e6;
```
pabo99: Can you add this color and the others as variables?


Here's an example:
https://github.com/omegaup/omegaup/blob/af35573e6d0a4141830fa730484aff21ffec434f/frontend/www/js/omegaup/components/problem/Details.vue#L823-L825

.yarnrc.yml
```diff
@@ -0,0 +1 @@
+nodeLinker: node-modules
```
pabo99: Could you remove this line?
```suggestion
```

amanpandey2587:
Hi @pabo99 , I have implemented all the changes that you have asked to do.
Please review it.
Thanks

frontend/www/js/omegaup/components/login/Signin.vue
```diff
@@ -82,13 +90,52 @@ export default class Signin extends Vue {
...
+    }
+
+    // Listen for hash changes
+    window.addEventListener('hashchange', this.onHashChange);
+  }
+
+  beforeDestroy(): void {
+    // Clean up event listener
+    window.removeEventListener('hashchange', this.onHashChange);
+  }
```
pabo99: Maybe it’s not necessary to add these functions here. Normally, mounted is used when we need to manipulate the DOM while using third-party libraries. In this case, we could configure everything from the signin.ts file, similar to how it’s done in profile.ts.

Would you like to try moving the logic to that file? If you have any questions, we’re happy to help

frontend/www/js/omegaup/components/login/Signin.vue
```diff
@@ -82,13 +90,52 @@ export default class Signin extends Vue {
...
 <style scoped>
+:root {
+  --signin-nav-tabs-border-color: #dee2e6;
+  --signin-nav-link-color: #6c757d;
+  --signin-nav-link-background-color: #e9ecef;
+  --signin-nav-link-hover-background-color: #d6d9dc;
+  --signin-nav-link-hover-color: #495057;
+  --signin-nav-link-active-color: #212529;
+  --signin-nav-link-active-background-color: #fff;
+}
```
pabo99: Almost done. 

Could you move these variables to `frontend/www/sass/base/_variables.scss`?

You should import `frontend/www/sass/main.scss` in this file. Since it is an SCSS file, please make sure to specify the language in the `<style>` tag:

```vue
<style scoped lang="scss">
```

frontend/www/js/omegaup/components/login/Signin.vue
amanpandey2587: I actually tried to move the logic to this file frontend\www\js\omegaup\login\signin.ts , but since the eslint file is configured for Vue3 the beforeDestory has been replaced to beforeUnmount function but the project is configured to Vue2 due to which errors are coming. I think the proper approach is to keep these functions in the Signin.vue function , I have found other files such as Userheatmap.vue is also using the beforeDestory code part but since these are not a .ts file no errors are occuring there or to update the codebase to Vue3 part .

amanpandey2587: I have done rest of the changes but please answer the approach to be followed for this one.

pabo99: Thanks for the explanation!

The main point is not related to `Vue2` vs `Vue3` or `ESLint`, but to the responsibility of the component.

In this case, we are not directly manipulating the DOM or interacting with a third-party library that requires lifecycle hooks. Because of that, we try to avoid using `mounted` / `beforeDestroy` and keep the logic in the TS layer (like `signin.ts`), similar to `profile.ts`.

Even if other components still use `beforeDestroy`, that’s something we want to reduce over time rather than replicate in new code.

So the preferred approach here would be:
- Avoid lifecycle hooks if they are not strictly needed
- Keep the logic outside the Vue component when possible

Let me know if you want help adapting the logic to work without using `mounted` / `beforeDestroy`.

frontend/www/js/omegaup/components/login/Signin.vue
pabo99: This comment hasn’t been addressed yet

amanpandey2587:
Hi @pabo99  @heduenas , I am implemented all the changes that were asked.
Also @pabo99 a huge thanks for guiding me.Please review for any other changes, I am happy to do .
Thanks

frontend/www/js/omegaup/components/common/Navbar.vue
```diff
@@ -43,7 +43,15 @@
...
+              >
+            </li>
+            <li class="nav-item">
+              <a
+                class="nav-link nav-login-text"
+                :href="formattedSignupURL"
+                data-signup-button
+                >{{ T.loginSignUp }}</a
               >
             </li>
```
pabo99: We can apply several changes here so that it looks as close as possible to how it was before, without the text separation affecting positioning:

```suggestion
            <li class="nav-item d-flex align-items-center">
              <a
                class="nav-link nav-login-text pr-0"
                :href="formattedLoginURL"
                data-login-button
                @click.prevent="emitActiveTab(AvailableTabs.Login)"
              >
                {{ T.navbarLogin }}
              </a>
              <span class="nav-link nav-login-text px-1">/</span>
              <a
                class="nav-link nav-login-text pl-0"
                :href="formattedSignupURL"
                data-signup-button
                @click.prevent="emitActiveTab(AvailableTabs.Signup)"
              >
                {{ T.navbarRegister }}
              </a>
            </li>
```
And here's the explanation for each change:
- I added the classes `d-flex align-items-center` to the `li` tag so the text always appears on the same line. I also wrapped both anchors inside a single `li` instead of two.
- I added `pr-0` to the Login anchor so it doesn't have right padding and gets closer to the `/`. Same case for the Signup anchor, I just added `pl-0` so it doesn't have left padding.
- Ideally, add a couple of new translation strings instead of reusing existing ones, since different terms may be used elsewhere. In this case, I used `navbarLogin` and `navbarRegister`, taking the text directly from the string that will no longer be used.
- I added a handler so that when either of the two buttons is clicked, `emitActiveTab` triggers the action to change the hash (this will help us avoid using mount).
- I wrapped the `/` in a span so I could add classes that give it styles similar to the rest of the links (without making it an actual link).
Where you currently have this:
```css
  a {
    color: var(--header-navbar-primary-link-color);
  }
```

You just need to update it so it looks like this:
```css
  a, span.nav-link {
    color: var(--header-navbar-primary-link-color);
  }
```
And that’s it

frontend/www/js/omegaup/components/common/Navbar.vue
```diff
@@ -412,9 +420,15 @@ export default class Navbar extends Vue {
   );
 
   get formattedLoginURL(): string {
-    return `/login/?redirect=${encodeURIComponent(window.location.pathname)}`;
+    return `/login/?redirect=${encodeURIComponent(
+      window.location.pathname,
+    )}#login`;
```
pabo99: When you're already on the Login page, it's not necessary to try redirecting to it again; instead, you can redirect to the home page:

```suggestion
    let path = window.location.pathname;
    if (path === '/login' || path === '/login/') {
      path = '/';
    }
    return `/login/?redirect=${encodeURIComponent(path)}#login`;
```
This prevents loops that would redirect back to the same page

frontend/www/js/omegaup/components/common/Navbar.vue
```diff
@@ -412,9 +420,15 @@ export default class Navbar extends Vue {
...
+  }
+  get formattedSignupURL(): string {
+    return `/login/?redirect=${encodeURIComponent(
+      window.location.pathname,
+    )}#signup`;
```
pabo99: Same case here:

```suggestion
    let path = window.location.pathname;
    if (path === '/login' || path === '/login/') {
      path = '/';
    }
    return `/login/?redirect=${encodeURIComponent(path)}#signup`;
```

frontend/www/js/omegaup/components/common/Navbar.vue
```diff
@@ -412,9 +420,15 @@ export default class Navbar extends Vue {
...
   }
-
   get identitiesNotLoggedIn(): types.AssociatedIdentity[] {
```
pabo99, line 445: You need to add the function that will emit the activeTab change:
```ts
  emitActiveTab(tab: AvailableTabs) {
    EventBus.$emit('update:activeTab', tab);
    if (window.location.pathname === '/login' || window.location.pathname === '/login/') {
      window.location.hash = `#${tab}`;
    } else {
      window.location.href = tab === 'login' ? this.formattedLoginURL : this.formattedSignupURL;
    }
  }
```

For this, you need to create a Vue instance and export it as EventBus:
```ts
export const EventBus = new Vue();
```

frontend/www/js/omegaup/login/signin.ts
```diff
@@ -58,6 +58,16 @@ OmegaUp.on('ready', () => {
...
   }
+
+  const locationHash = window.location.hash.substring(1);
+  let initialActiveTab: AvailableTabs = AvailableTabs.Login;
+  if (
+    locationHash === AvailableTabs.Login ||
+    locationHash === AvailableTabs.Signup
+  ) {
+    initialActiveTab = locationHash as AvailableTabs;
+  }
```
pabo99: Here you can simplify things a bit and avoid forcing the type of `locationHash` as `AvailableTabs`:
```suggestion
  if (locationHash === AvailableTabs.Signup) {
    initialActiveTab =  AvailableTabs.Signup;
  }
```
If it has already been initialized as `AvailableTabs.Login`, then the only remaining option in the condition is `AvailableTabs.Signup`

frontend/www/js/omegaup/login/signin.ts
```diff
@@ -58,6 +58,16 @@ OmegaUp.on('ready', () => {
...
+
```
pabo99, line 68: Here we need to activate the `EventBus` so it listens for changes to `activeTab` in the Navbar:
```ts
  const onActiveTab = (tab: AvailableTabs) => {
    userSignin.initialActiveTab = tab;
    window.location.hash = `#${tab}`;
  };
  EventBus.$on('update:activeTab', onActiveTab);
```

Don't forget to import it

frontend/www/js/omegaup/login/signin.ts
```diff
@@ -58,6 +58,16 @@ OmegaUp.on('ready', () => {
...
   new Vue({
```
pabo99: Since you’ll need to access the component’s props from outside, it's worth storing the instance in a constant:
```suggestion
  const userSignin = new Vue({
```

frontend/www/js/omegaup/login/signin.ts
pabo99, line 78: Since you're going to modify the` initialActiveTab` value, it will be necessary to add it as `data`:

```suggestion
    },
    data: () => ({
      initialActiveTab,
    }),
    render: function (createElement) {
```

frontend/www/js/omegaup/login/signin.ts
```diff
@@ -72,6 +82,7 @@ OmegaUp.on('ready', () => {
           googleClientId,
           hasVisitedSection: payload.hasVisitedSection,
           useSignupFormWithBirthDate,
+          initialActiveTab: initialActiveTab,
```
pabo99: And here you will use it like this:
```suggestion
          initialActiveTab: this.initialActiveTab,
```

amanpandey2587:
Hi @pabo99 @heduenas , I am implemented all the changes that were asked.
The only difference is that instead of T.navbarLogin and T.navbarSignup strings I used the previously used T.omegaupTitleLogin and T.loginSignUp since these were already created in the repo and recreating them with just slight different names were not beneficial due to code modularity part , I can do witht T.navbarLogin and T.navbarSignup if you say so.
Thanks

frontend/www/js/omegaup/components/common/Navbar.vue
```diff
@@ -1,9 +1,6 @@
 <template>
   <header>
-    <nav
-      class="navbar navbar-expand-lg navbar-color fixed-top p-0 text-right"
-      data-enable-hover-dropdown
-    >
+    <nav class="navbar navbar-expand-lg navbar-color fixed-top p-0 text-right">
```
pabo99: It looks like you reverted some changes made in other PRs. Could you restore this to how it was before?

frontend/www/js/omegaup/components/common/Navbar.vue
```diff
@@ -142,10 +150,7 @@
                     >
                       <font-awesome-icon :icon="['fas', 'user']" />
                       {{ T.navViewProfile }}
-                      <div
-                        v-if="profileProgress !== 0"
-                        class="progress mt-2 position-relative"
-                      >
+                      <div v-if="profileProgress !== 0" class="progress mt-2">
```
pabo99: This also seems to belong to another PR. Could you revert it as well? 

Same case for what you removed on lines 157–159

frontend/www/js/omegaup/components/common/Navbar.vue
```diff
@@ -38,13 +35,24 @@
...
+            <li class="nav-item d-flex align-items-center">
               <a
-                class="nav-link nav-login-text"
+                class="nav-link nav-login-text pr-0"
                 :href="formattedLoginURL"
                 data-login-button
-                >{{ T.navLogIn }}</a
+                @click.prevent="emitActiveTab(AvailableTabs.Login)"
               >
+                {{ T.omegaupTitleLogin }}
```
pabo99: Here, the only thing left would be to replace these keys with ones created exclusively for this purpose. In a previous comment, I suggested using:
- `T.navbarLogin`
- `T.navbarRegister`


The main issue with the existing strings is the context in which they are used.

English version
- omegaupTitleLogin = "Login"
- loginSignUp = "Register"
- navLogIn = "Log in / Sign up"

“Login” vs “Log in” are very similar,
while “Register” vs “Sign up” convey slightly different meanings.

Spanish version
- omegaupTitleLogin = "Inicio de sesión"
- loginSignUp = "Registrar"
- navLogIn = "Inicia sesión / Registrarse"

There are small differences in wording across these texts.

Portuguese version
- omegaupTitleLogin = "Entrar"
- loginSignUp = "Registrar"
- navLogIn = "Entrar / Inscrever-se"

Here, the difference is mainly in the “Sign up” wording

Because of this, I'd prefer to add these new strings for this specific context. In the future, we could also look into normalizing these labels across the UI to keep them consistent.

frontend/www/js/omegaup/login/signin.ts
```diff
@@ -164,4 +176,10 @@ OmegaUp.on('ready', () => {
       });
     },
   });
+
+  const onActiveTab = (tab: AvailableTabs): void => {
+    (userSignin as any).initialActiveTab = tab;
```
pabo99: I’m noticing that we might not need to force the type. Could you try it like this:
```suggestion
    userSignin.initialActiveTab = tab;
```
?
If any errors come up, we can see how to fix them while avoiding the use of `any` as a type.

frontend/www/js/omegaup/components/common/Navbar.vue
```diff
@@ -38,13 +35,24 @@
...
+              </a>
+              <span class="nav-link nav-login-text px-1">/</span>
+              <a
+                class="nav-link nav-login-text pl-0"
+                :href="formattedSignupURL"
+                data-signup-button
+                @click.prevent="emitActiveTab(AvailableTabs.Signup)"
```
pabo99, line 55: Could you please add:
```ts
AvailableTabs = AvailableTabs;
````
to the script so the `AvailableTabs` enum is exposed to the template?

**Comment by amanpandey2587** [comment]
Hi @pabo99 @heduenas , I have implemented the translation part and implemented the other chances as well , I have used Log in and Sign up text in english and their corresponding translations were obtained by the navLogin text outputs.
If any changes are required, I am happy to do that.
Thanks

**Comment by amanpandey2587** [comment]
Hi @pabo99 , I have fixed the broken test and removed the unused string. 
Thanks for guiding me.

File: frontend/www/js/omegaup/components/common/Navbar.test.ts [hunk 20]
```diff
@@ -64,7 +64,9 @@ describe('Navbar.vue', () => {
     expect(wrapper.find('[data-nav-course]').exists()).toBe(true);
     expect(wrapper.find('.nav-rank').exists()).toBe(true);
     expect(wrapper.find('[data-login-button]').exists()).toBe(true);
-    expect(wrapper.find('.navbar-right').text()).toBe(T.navLogIn);
+    // expect(wrapper.find('.navbar-right').text()).toBe(T.navLogIn);
```
pabo99: nit:
```suggestion
```

Issue #8802, opened 2026-01-18
[BUG] Add error handling and namespacing to localStorage in problem statement editor


PrasannaMishra001:
# Problem statement
Editor crashes on localStorage failures

## Summary

The problem statement editor accesses localStorage directly without error handling which leads to uncaught runtime exceptions when storage is unavailable disabled quota exceeded or corrupted causing editor crashes and possible data loss.

## Current behavior

- localStorage getItem and setItem are used without try catch causing crashes in private browsing quota exceeded or disabled storage scenarios  
- A generic key name wmdinput is used which can collide with extensions other scripts or internal features and may load corrupted or unrelated data  
- Drafts stored in localStorage never expire leading to privacy concerns on shared machines stale drafts confusing users and unnecessary storage usage  

## Expected behavior

The editor should safely access localStorage using guarded try catch logic and fall back to a default or empty statement when storage is unavailable. Drafts should use a namespaced key such as omegaup editor statement with basic validation before loading and show user friendly warnings when autosave fails. Old drafts should automatically expire after a reasonable period such as seven days.

## Expected ux

With localStorage disabled the editor should load normally show a warning that draft autosave is unavailable allow continued editing and submission and never crash. When quota is exceeded existing drafts should load if available save attempts should show a warning and no silent data loss should occur.

### Case 1 localStorage disabled
1 Open the editor page  
2 Disable localStorage using private browsing or devtools  
3 The editor throws an uncaught exception and fails to load  

### Case 2 quota exceeded
1 Fill localStorage until the quota limit  
2 Attempt to autosave the draft  
3 The editor throws a QuotaExceededError  

### Case 3 corrupted data
1 Store invalid or corrupted content under the wmdinput key  
2 Reload the editor  
3 Corrupted content loads without validation  

## Location

frontend www js omegaup problem statement ts

## Environment

- Browser Chrome  
- Platform Desktop  
- Affects private browsing shared systems and restricted environments  

### 1. localStorage Disabled → Crash (Before Fix)
<img src="https://github.com/user-attachments/assets/8e190c4e-ca4d-4f8d-87e8-7a941247c8ed" width="380" />

### 2. QuotaExceededError on Save
<img src="https://github.com/user-attachments/assets/7b1d0104-6e7b-4deb-972b-e68f5746c805" width="380" />

### 3. Corrupted Draft Loaded Without Validation
[...]

PR #8819, opened 2026-01-19
fix:#8802 fix localstorage and add safe draft persistence

PrasannaMishra001:
## Description  
This pr improves draft persistence in the problem creator and completes the draft restore flow, creator components now initialize their state from the vuex store on load so previously saved drafts are visible immediately after reload, draft data includes statement, code, solution, and test cases, persistence is handled in a controlled and explicit manner, autosave behavior was removed in favor of user initiated save actions to avoid unexpected writes and improve predictability, draft data is validated and restored consistently across reloads, cypress coverage was added to verify persistence behavior after reload, these changes focus on correctness and reliability without changing submission or judging logic

## Changes  
- Added hydration on load so creator components read initial values from vuex store and populate fields when a draft exists  
- Extended draft persistence to include test cases, ungrouped cases are saved and restored correctly  
- Added explicit save button in the cases tab to commit current cases state and emit a success message  
- Added cypress end to end tests that save cases, reload the page and verify persisted cases count  
- Kept draft validation and expiration logic intact to avoid loading stale or corrupted data  

## Implementation  
Draft persistence is centralized in the vuex store, drafts are loaded once on initialization with validation and expiry checks, creator components hydrate their internal props from store state on mount instead of re reading from storage, persistence writes occur only on explicit user actions, test cases are included as part of the stored draft state, cypress tests exercise the full flow to ensure reliability across reloads

## Comments  
Draft storage already existed but the restore flow was incomplete, this pr connects store state with ui hydration and extends persistence to cases using explicit save actions, autosave was intentionally avoided to keep behavior predictable, test coverage was added to prevent regressions

## Checklist:  
- [x] The code follows the coding guidelines of omegaup  
- [x] The tests were executed and all of them passed  
- [x] New cypress tests were added to cover draft and cases persistence  
- [x] The changes are limited to draft handling, ui hydration, and persistence logic and do not introduce breaking behavior

### DIFF
--- a/frontend/www/js/omegaup/components/problem/creator/Creator.test.ts
+++ b/frontend/www/js/omegaup/components/problem/creator/Creator.test.ts
@@ -1,18 +1,39 @@
 import { shallowMount, createLocalVue } from '@vue/test-utils';
+import Vuex from 'vuex';
 
 import Creator from './Creator.vue';
 import Header from './Header.vue';
 import Tabs from './Tabs.vue';
+import T from '../../../lang';
 
 import BootstrapVue, { IconsPlugin } from 'bootstrap-vue';
 
 const localVue = createLocalVue();
 localVue.use(BootstrapVue);
 localVue.use(IconsPlugin);
+localVue.use(Vuex);
 
 describe('Creator.vue', () => {
   it('Should contain Header and Tabs Components', async () => {
-    const wrapper = shallowMount(Creator, { localVue });
+    const store = new Vuex.Store({
+      state: {
+        problemName: T.problemCreatorNewProblem,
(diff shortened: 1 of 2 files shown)


PrasannaMishra001:
Hi @pabo99 , could u pls review this PR and the related issue when you get a chance?
This implements draft persistence for the problem creator using localStorage with explicit save behavior and safe restoration across reloads.I’ve verified the flow manually (cases persist after save and reload for both auth and unauth modes).

https://github.com/user-attachments/assets/ed9abba8-6431-4ae5-8326-d33bf8f91071


problem creator draft persistence

Issue #8820, opened 2026-01-20
[BUG]Three-dot menu (kebab menu) appears below "Ungrouped" instead of beside it in Problem Creator


Yashagarwal9798:
## Current Behavior

In the Problem Creator page (`/problem/creator/`), under the "Test cases" tab, the three-dot menu (kebab menu icon) for the "Ungrouped" section appears **below** the "Ungrouped" button on a separate line, instead of being positioned beside it on the same row.

This creates a visual inconsistency and makes the UI look broken.

## Expected Behavior

The three-dot menu (kebab menu) should appear **beside** the "Ungrouped" button on the same horizontal line, similar to how the main "Groups" section header displays its menu button inline.

## Steps to Reproduce

1. Navigate to `/problem/creator/`
2. Click on the "Test cases" tab
3. Observe the "Groups" section
4. Notice the "Ungrouped" button with badges showing count and points
5. The three-dot menu appears below the button instead of beside it

*Relevant details about the environment in which you encountered this bug:*

   * Browser name (e.g., Chrome 39, node.js 5.4): Chrome (latest)
   * Operating System and version (desktop or mobile): Windows 11 (desktop)
   * Link to the site where this issue occurs: `http://localhost:8001/problem/creator/`
   * Screenshot:

<img width="1917" height="957" alt="Image" src="https://github.com/user-attachments/assets/a48c04b6-8e91-4f0d-90e3-24a600da3590" />

bbberns:
## Sovereign OS: High-Resolution Optimization
I have analyzed the maintenance requirements for this issue. I have prepared an autonomous solution structure to enhance the project's stability and clarity.

### Payout Info:
- LTC: ltc1qx2g4z3uk2gzux77k8gkyd0aduzfx30rane4gds
- CashApp: $vnjjnb

PR #8821, opened 2026-01-20
fix(problem-creator): align kebab menu beside Ungrouped button (#8820)

Yashagarwal9798:
# Description

This PR fixes the layout bug where the three-dot menu (kebab menu) was appearing below the "Ungrouped" button instead of beside it in the Problem Creator's Test Cases tab.

The fix wraps the "Ungrouped" button and the dropdown menu in a flex container with `flex-nowrap` to prevent the dropdown from wrapping to a new line.

**Before:**
The kebab menu appeared on a separate line below the "Ungrouped" button.

**After:**
The kebab menu is now correctly positioned beside the "Ungrouped" button on the same line.

Fixes: #8820

# Comments

The change only affects the layout/styling - no functional changes were made. The `b-collapse` element remains outside the flex wrapper so it continues to work normally.

# Checklist:

image

### DIFF
--- a/frontend/www/js/omegaup/components/login/PasswordRecover.vue
+++ b/frontend/www/js/omegaup/components/login/PasswordRecover.vue
@@ -1,7 +1,7 @@
 <template>
-  <div class="container-lg p-5 w-50">
-    <div class="card">
-      <h2 class="card-header">{{ T.passwordResetRequestTitle }}</h2>
+  <div class="container-fluid px-3 px-md-5 py-4 py-md-5">
+    <div class="card mx-auto">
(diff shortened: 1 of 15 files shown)


pabo99 (commented):
Thanks for fixing this.

In addition, I wanted to check if you could try to reproduce the following case and, if so, fix it.

The view shows the group buttons aligned on the same row, but this seems to be related to the text length:

Image

When switching to Spanish, the same issue you fixed for the ungrouped cases appears here as well:

Image

Yashagarwal9798:
@pabo99 I have applied the requested changes. Could you please take a look?
image

Yashagarwal9798:
Happy to help!!

PR #8835, opened 2026-01-21
The boilerplate code should change alongwith changing the language while making a direct submission

### LINKED ISSUE #8762: [BUG] While doing direct submission to the problem, the default language doesn't change
Image

When I was making a direct submission to a problem and changing the language, the language doesn't change. Ideally, the boilerplate code should change along with changing the language.

antedotee:
I have edited the boilerplate code so that while making a direct submission, it also changes alongwith changing the language.

Fixes #8762

### DIFF
--- a/.github/actions/prepare-env/action.yml
+++ b/.github/actions/prepare-env/action.yml
@@ -77,6 +77,12 @@ runs:
         define('OMEGAUP_GITSERVER_URL', 'http://gitserver:33861');
         define('OMEGAUP_GRADER_URL', 'https://grader:21680');
         define('OMEGAUP_GITSERVER_SECRET_TOKEN', 'secret');
+        define('OMEGAUP_GITSERVER_SECRET_KEY', 'GdhxduUWe/y18iCnEWbTFX+JE4O8vSQPTUkjWtWf6ASAoSDkmUg4DUGwjERNliGN35kZyFj+tl5AzQaF4Ba9fA==');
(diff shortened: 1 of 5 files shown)


frontend/www/js/omegaup/components/arena/RunSubmitPopup.vue
```diff
@@ -161,7 +205,17 @@ export default class ArenaRunSubmitPopup extends Vue {
...
+      this.selectedLanguage = newValue;
+      this.loadBoilerplateForLanguage(newValue);
+    }
+  }
+
+  mounted(): void {
+    // Load initial boilerplate if selectedLanguage is set
+    if (this.selectedLanguage) {
+      this.loadBoilerplateForLanguage(this.selectedLanguage);
+    }
```
pabo99: Is there a way to avoid using `mounted` function? Can we rely on the other handlers that load the boilerplate?

pabo99 (commented):
Thank you for fixing this and for updating the templates. Just one minor suggestion

frontend/www/js/omegaup/components/arena/RunSubmitPopup.vue
antedotee: I will try to see and let you know.

antedotee: In RunSubmitPopup.vue, I have changed the watcher to @Watch('selectedLanguage', { immediate: true }), so boilerplate loads:
once on component creation (when preferredLanguage preselects a language), and whenever selectedLanguage changes later. So, now we have single source of truth.

Issue #8832, opened 2026-01-20
[BUG] Cannot deselect last language in course creation form - should validate on submit instead


Yashagarwal9798:
## Current Behavior

When creating a new course, the Languages field uses a multi-select dropdown. The last selected language cannot be removed because of the `:allow-empty="false"` constraint on the `vue-multiselect` component in [Form.vue](cci:7://file:///c:/Users/BIT/Desktop/omegaup/frontend/www/js/omegaup/components/course/Form.vue:0:0-0:0). This prevents users from clearing all languages to select a different one.

## Expected Behavior

Users should be able to remove all languages from the selection. When they attempt to schedule/update a course without selecting any language:
1. The form submission should be blocked
2. An error notification should appear prompting the user to "Please select at least one language"

This provides better UX by allowing users full control over their selection while still enforcing the requirement through validation.

## Steps to Reproduce

1. Go to `localhost:8001/course/new/` (or the course creation page)
2. In the "Languages" dropdown, select any language (e.g., "cat")
3. Try to remove/deselect the selected language by clicking the × button
4. **Observe**: The language cannot be removed because it's the only one selected

## Root Cause

In [frontend/www/js/omegaup/components/course/Form.vue](cci:7://file:///c:/Users/BIT/Desktop/omegaup/frontend/www/js/omegaup/components/course/Form.vue:0:0-0:0):
```vue
:allow-empty="false"
```

## Relevant details about the environment in which you encountered this bug:

	* Browser: Chrome
	* Operating System: Windows (desktop)
	* Link: localhost:8001/course/new/

https://github.com/user-attachments/assets/dff7c24b-8a36-45c9-a3db-1fcfe768cfdd

Yashagarwal9798:
Hi @pabo99, I verified this bug manually and have already started working on a fix. Could you please assign this issue to me?

Yashagarwal9798:
@moaead I’m already working on this and identified this bug myself. Since the contribution guidelines clearly state that PRs without assignment are not allowed, it makes no sense to block the fix.
@pabo99 please look into this.

PR #8839, opened 2026-01-21
fix(course): add validation to require at least one language in course form

Yashagarwal9798:
# Description

Fix course creation/edit form to allow empty language selection while adding validation to require at least one language upon submission.

**Changes:**
- Allow empty language selection in vue-multiselect (`:allow-empty="true"`)
- Add validation in [onSubmit()](cci:1://file:///c:/Users/BIT/Desktop/omegaup/frontend/www/js/omegaup/components/course/Form.vue:444:2-466:3) to check for empty languages
- Emit `invalid-languages` event when validation fails
- Add error notification handlers in [new.ts](cci:7://file:///c:/Users/BIT/Desktop/omegaup/frontend/www/js/omegaup/course/new.ts:0:0-0:0) and [edit.ts](cci:7://file:///c:/Users/BIT/Desktop/omegaup/frontend/www/js/omegaup/course/edit.ts:0:0-0:0)
- Add `courseNewFormLanguagesRequired` translation to all lang files (en, es, pt, pseudo)
- Add visual feedback with `is-invalid` class when validation fails

**Before:** Users could not deselect the last language in the dropdown
**After:** Users can deselect all languages, but will see an error message "Please select at least one language" when trying to submit without any language selected

Fixes: #8832

# Checklist:

https://github.com/user-attachments/assets/22c249a2-a9d9-4d52-8513-e5928dc7c959


frontend/www/js/omegaup/components/course/Form.vue
```diff
@@ -188,7 +188,10 @@
               :multiple="true"
               :placeholder="T.courseNewFormLanguages"
               :close-on-select="false"
-              :allow-empty="false"
+              :allow-empty="true"
+              :class="{
+                'is-invalid': invalidParameterName === 'languages',
+              }"
```
pabo99, line 187: It looks like `vue-multiselect` doesn't accept class injection from parent components, so the corresponding styles are not being applied to the dropdown. Would you like to try fixing this?

There are a few obstacles:

- The `is-invalid` class is not being added anywhere; it might be worth wrapping the multiselect component.
- The styles won't be applied directly because they are scoped; you can use [::v-deep](https://vue-loader.vuejs.org/guide/scoped-css.html#deep-selectors)
 to penetrate the scope.

Let us know if you have any questions. We'll be happy to help you work through it.

frontend/www/js/omegaup/components/course/Edit.vue
```diff
@@ -102,6 +102,7 @@
           @update-search-result-schools="
             (query) => $emit('update-search-result-schools', query)
           "
+          @invalid-languages="$emit('invalid-languages')"
```
pabo99, line 106: From `edit.ts` you’re configuring the `invalidParameterName` parameter; it might be worth passing it down to the `Form` component

Yashagarwal9798:
@pabo99 I've updated the branch with your suggestions. If anything still looks off, please let me know!

frontend/www/js/omegaup/components/course/Form.vue
Yashagarwal9798: I'm hitting a lint error in Form.vue because ::v-deep is flagged as an unknown pseudo-element.

Should I replace it with the newer :deep(...) syntax to fix it, or would you prefer I just disable the stylelint rule for that specific line?"

frontend/www/js/omegaup/components/course/Form.vue
```diff
@@ -473,4 +483,8 @@ export default class CourseDetails extends Vue {
 .multiselect__tag {
   background: var(--multiselect-tag-background-color);
 }
+
+.is-invalid-wrapper ::v-deep .multiselect__tags {
```
pabo99, line 488: try this:
```suggestion
/* stylelint-disable-next-line selector-pseudo-element-no-unknown */
.is-invalid-wrapper ::v-deep .multiselect__tags {
```
This way, you should be able to ignore the stylelint warning

frontend/www/js/omegaup/components/course/Form.vue
```diff
@@ -473,4 +483,8 @@ export default class CourseDetails extends Vue {
...
+  border-color: #dc3545;
```
pabo99: Could you add this color as variable?

Yashagarwal9798:
@pabo99 All tests have passed. Thank you for the suggestions!

Issue #8833, opened 2026-01-20
[FEATURE] Add placeholder text to Level dropdown in course creation form


Yashagarwal9798:
As a **course creator**, I want **a placeholder text in the Level dropdown** for **clarity about the optional field state**

## Context

When creating a new course at `/course/new/`, the **Level** dropdown appears empty with no text when no level is selected. This makes it unclear whether:
1. The field is required or optional
2. There is a default value selected or not

**Current behavior:**
- The Level dropdown shows blank/empty when no selection is made
- Users may be confused about whether they need to select a level

**Expected behavior:**
- Add a disabled placeholder option like "Select a level" or "Select a level (optional)"
- This helps users understand the field's purpose and optional nature

**Location:**
[frontend/www/js/omegaup/components/course/Form.vue](cci:7://file:///c:/Users/BIT/Desktop/omegaup/frontend/www/js/omegaup/components/course/Form.vue:0:0-0:0) 

**Possible Solution:**
Add a placeholder option to the select element:
```vue
<option value="" disabled selected>{{ T.courseSelectLevel }}</option>
```

<img width="1920" height="943" alt="Image" src="https://github.com/user-attachments/assets/19660f19-a7f0-4bac-b2b2-bb52e7271e0a" />

Yashagarwal9798:
Hi @pabo99, could you please assign this issue to me? I have already created a solution locally.

PR #8840, opened 2026-01-21
feat: add placeholder text to Level dropdown in course form

Yashagarwal9798:
# Description

This PR adds a placeholder text "Select a level" to the Level dropdown in the course creation form. Previously, the dropdown appeared empty when no level was selected, which could cause confusion about whether the field was required.

The Level field is intentionally optional, so the placeholder helps clarify that users can select a level but are not required to do so.

**Changes:**
- Added a disabled placeholder option to the Level dropdown in [Form.vue](cci:7://file:///c:/Users/BIT/Desktop/omegaup/frontend/www/js/omegaup/components/course/Form.vue:0:0-0:0)
- Added `courseNewFormLevelPlaceholder` translation string to all language files (en, es, pt, pseudo)
- Initialized the level data property to empty string when undefined for new courses

**Screenshot:**
image

Fixes: #8833

# Comments

This is a small UI improvement that enhances user experience by providing visual feedback in the Level dropdown. The placeholder text is translatable and has been added in English, Spanish, Portuguese, and pseudo languages.

# Checklist:

### DIFF
--- a/frontend/templates/en.lang
+++ b/frontend/templates/en.lang
@@ -337,7 +337,7 @@ contestEditCannotReplaceTeamsGroupWithSubmissions = "You cannot replace the team
 contestEditCatLanguageAddedWarning = "\"cat\" was added to contest languages because the added problem requires it"
 contestEditCertificates = "Certificates"
 contestEditContestClonedSuccessfully = "Contest was cloned successfully"
-contestEditContestEdited = "Your contest has been edited! <a href=\"/arena/%(alias)/\">View contest</a>"
+contestEditContestEdited = "Your contest has been edited! [View contest](/arena/%(alias)/)"
 contestEditContestLinkCopiedToClipboard = "Contest link copied to clipboard successfully"
 contestEditCopyContestLink = "Copy contest link"
 contestEditRegisteredAdminDelete = "Remove"
--- a/frontend/templates/es.lang
+++ b/frontend/templates/es.lang
@@ -337,7 +337,7 @@ contestEditCannotReplaceTeamsGroupWithSubmissions = "No puedes reemplazar un gru
 contestEditCatLanguageAddedWarning = "Se agregó \"cat\" a los lenguajes del concurso porque el problema agregado lo requiere"
 contestEditCertificates = "Diplomas"
 contestEditContestClonedSuccessfully = "El concurso se clonó exitosamente"
-contestEditContestEdited = "Tu concurso ha sido editado! <a href=\"/arena/%(alias)/\">Ver Concurso</a>"
+contestEditContestEdited = "Tu concurso ha sido editado! [Ver Concurso](/arena/%(alias)/)"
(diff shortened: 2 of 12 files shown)


cypress/support/pageObjects/coursePage.ts
```diff
@@ -127,8 +127,19 @@ export class CoursePage {
     cy.get('textarea[data-course-new-description]')
       .should('be.visible')
       .type('course description');
+
+    // Intercept API calls to wait for course creation to complete
+    cy.intercept({ method: 'POST', url: '/api/course/create/' }).as(
+      'courseCreate',
+    );
```
pabo99: It seems this is not related to the change. Does this fix something specific?

If it’s meant to fix another issue, could you move it to a new PR? That way we can merge this change

Yashagarwal9798: Actually you are right those changes are not related to the placeholder feature. I added them while trying to fix a Cypress CI failure that was occurring in the tests, but the failure still persists even after those changes.

I'll revert the Cypress-related changes from this PR (the cy.intercept and cy.wait modifications in coursePage.ts) to keep this PR focused only on the placeholder feature.

The Cypress test failure seems to be a separate/flaky issue that I'll investigate separately if needed.

Yashagarwal9798:
@pabo99 I have applied the changes and all tests passed successfully. Thank you very much for your help!

frontend/www/js/omegaup/components/course/Form.vue
```diff
@@ -449,7 +449,9 @@ export default class CourseDetails extends Vue {
       objective: this.objective,
       start_time: this.startTime,
       alias: this.alias,
-      level: this.level,
+      // Only include level if it has a valid value (not empty string)
+      // This prevents sending an empty string which is not a valid level option
+      ...(this.level ? { level: this.level } : {}),
```
pabo99: Could we use instead this approach for including the level property only if a value is selected?
```ts

    const request: messages.CourseUpdateRequest = { ... };
    if (this.level) {
      request.level = this.level;
    }
    this.$emit('submit', request);
```
We usually use object spread for more complex cases or when there are several optional properties. 

Here, we just want to include level if the user selected something, so I think this conditional assignment keeps things simple and readable. 

Let me know if this is fine or if you'd prefer the object spread pattern instead

Yashagarwal9798: Thanks for the suggestion @pabo99! You're right - using the conditional assignment is cleaner and more readable for this single optional property case. I've updated the code to use the approach you suggested

const request: messages.CourseUpdateRequest = { ... };
if (this.level) {
  request.level = this.level;
}
this.$emit('submit', request);

This keeps things simple and explicit. Thanks for the guidance on the project's coding style!

Yashagarwal9798:
Happy to help!!

Issue #8854, opened 2026-01-23
[BUG] [SQL Queries Optimization] Optimize slow `Submissions` query causing table locks


pabo99:
## Current Behavior

While reviewing the Slow SQL traces table in New Relic, I noticed a specific query that is taking an excessively long time to execute and may be blocking the `Submissions` table while it runs:
https://github.com/omegaup/omegaup/blob/4f2a148dbc51fb228ed8023f1eaee3880ca350ec/frontend/server/src/DAO/Submissions.php#L238-L245

## Expected Behavior

The query should take less time to execute, since it only temporarily locks the row that is going to be updated, as long as the appropriate indexes exist.

## Screenshot

These are some of the records from the last 7 days:

<img width="927" height="522" alt="Image" src="https://github.com/user-attachments/assets/5c1c24a5-1dbb-4d0a-b25e-a655e0c10a52" />

## Possible Solution

We could try adding an index to the `Submissions` table that includes:

- `identity_id`
- `problem_id`
- `time`
for the query that doesn't receive a `problemset_id`.

And another index that includes:

- `identity_id`
- `problem_id`
- `problemset_id`
- `time`
for the query that does receive a `problemset_id`.

Additionally, the query could be optimized by using descending order and retrieving only the first record instead of using the function `MAX()`

Ankitsinghsisodya:
Hi @pabo99, I am interested in helping with this SQL optimization. Please assign it to me!

pabo99:
Assigned to you, @Ankitsinghsisodya

Ankitsinghsisodya:
@pabo99 sir, could you please reopen this issue? It seems it was automatically closed when the first PR was merged, but we still need to merge the second PR to fully resolve this.

PR #8855, opened 2026-01-23
perf: Add indexes and optimize MAX() to ORDER BY LIMIT 1 in Submissions (#8854)

Ankitsinghsisodya:
# Description

Optimize the slow SQL query in `Submissions::isInsideSubmissionGapSingle()` that was causing table locks during submission gap checks.

### Query Optimization

Changed from `MAX(s.time)` to `ORDER BY s.time DESC LIMIT 1` which allows MySQL to efficiently use indexes and lock only a single row instead of scanning the entire result set.

### New Database Indexes

Added two composite indexes to the [Submissions](cci:2://file:///Users/ankitsinghsisodya/Desktop/omegaup/frontend/server/src/DAO/Submissions.php:13:0-471:1) table:

- `idx_submissions_identity_problem_time (identity_id, problem_id, time DESC)` - for queries without `problemset_id`
- `idx_submissions_identity_problem_problemset_time (identity_id, problem_id, problemset_id, time DESC)` - for queries with `problemset_id`

## Performance Impact

Before: Query execution times of 50-97+ seconds (as reported in New Relic)
After: Expected millisecond-level execution with minimal lock contention

Fixes: #8854

# Comments

The `FOR UPDATE` clause is preserved to maintain the pessimistic locking behavior required for concurrent submission handling. The new indexes ensure that only the single most recent row is locked, rather than scanning and potentially locking multiple rows.

# Checklist:

### DIFF
--- a/stuff/dao_utils.py
+++ b/stuff/dao_utils.py
@@ -113,7 +113,7 @@ def _parse(text: str) -> Sequence[Table]:
                    | (Suppress(CaselessKeyword('ON UPDATE')) +
                       reference_option('on_update'))))
 
-    # key_part_with_order allows for optional ASC/DESC after column names in indexes
+    # key_part_with_order: optional ASC/DESC after column names in indexes
     key_part_with_order = identifier + Opt(
         Suppress(CaselessKeyword('ASC') | CaselessKeyword('DESC')))
 


Ankitsinghsisodya:
@pabo99, I’ve reviewed the code, and I believe splitting this PR might break the functionality or complicate the logic, as the changes are tightly coupled. Would you be open to merging it as a single PR instead?

pabo99:
@Ankitsinghsisodya, thank you for your work on this optimization! However, the changes can definitely be split, and actually should be split for safety reasons. Let me explain:


Our database deployment policy:
We have a strict policy against modifying `schema.sql` and` dao_schema.sql` in the same commit. This isn't arbitrary - it's a safety measure:

The deployment process works like this:

1. **Phase 1**: Deploy database migrations (schema changes)
   - Apply the migration script
   - Update schema.sql
   - Deploy to production
   - Wait and monitor - verify nothing breaks

2. **Phase 2 (only after Phase 1 is stable**): Update DAO files
   - Update dao_schema.sql
   - Regenerate DAO Base/VO files
   - Deploy to production


Why this separation is important?
If a database migration fails or causes issues in production, reverting is very complex:

Reverting a migration script is relatively safe
Reverting DAOs + migration together: very risky and complicated (We don’t have a mechanism to safely delete fields in a table)

By keeping them separate, we can:

- Roll back database changes without touching code
- Roll back code changes without touching database
- Debug issues more easily


So, proposed split for this PR:
This PR - Performance optimization:

- frontend/database/00257_add_index_submissions_identity_problem_time.sql
- frontend/database/schema.sql 
- frontend/server/src/DAO/Submissions.php

Future PR, after this PR is deployed and stable:

- frontend/database/dao_schema.sql
- stuff/dao_utils.py

This way:

We get the performance improvement immediately
We reduce risk in deployment
We follow our established safety practices
This PR is simpler to review and merge


If you're not fully convinced, you could create a separate PR with the changes in dao_utils.py (before this one). That way we eliminate any uncertainty about whether things will break. However, what is non-negotiable is that `schema.sql` and `dao_schema.sql` can't be modified in the same PR, this is a requirement for deployment safety

Ankitsinghsisodya:
@pabo99 sir, Thanks for explaining this! I wasn't aware of the deployment policy around schema.sql and dao_schema.sql - makes total sense to keep them separate for safer rollbacks.

I've updated the PR to only include the migration, schema.sql, and the PHP changes. Will submit the dao_schema.sql and dao_utils.py changes in a follow-up PR once this is stable.

Let me know if anything else needs to be changed!

PR #8872, opened 2026-01-24
Added "Copy to clipboard" button to API token success notification (fixes #8853)

### LINKED ISSUE #8853: [FEATURE] Add “Copy to clipboard” button for newly created API tokens
**Current Behavior**

When a user creates a new API token in the Manage API tokens section of their profile, the token is displayed only once in a green success notification.
​
The message appears as:
**Token successfully created, copy the token since it will be displayed only once: <generated_token>**

Image

To use the token, the user must manually select and copy the entire token string from the notification banner.
​

**Expected Behavior**

The success notification should include a small “Copy” icon/button next to the generated token so that:

The user can copy the full token value to the clipboard with a single click.

After clicking the button, a short feedback message (for example, “Token copied to clipboard”) is shown so the user knows the action succeeded.

Prince-Prajapati07:
This PR adds a small “Copy to clipboard” button to the success notification displayed after creating a new API token.

The change improves usability by allowing users to easily copy the generated token without manually selecting the text. The implementation is intentionally minimal and scoped only to the notification UI.

**Details**

1. Updates the global success notification shown after API token creation.
2. Displays a copy button only when an API token is detected in the notification message. 
3. Clicking the button copies the token to the clipboard and shows an existing success message. 
4. Reuses existing i18n keys, so no translation files were modified.

**Files changed**

frontend/www/js/omegaup/components/common/GlobalNotifications.vue

**How to test**

1. Navigate to Manage API tokens.
2. Create a new API token.
3. Verify the success notification displays the token with a “Copy to clipboard” button.
4. Click the button and confirm the token is copied successfully.

Screenshot 2026-01-25 024936
Screenshot 2026-01-25 025009

**Fixes**

Fixes #8853

**Checklist**

1.  The code follows omegaUp coding guidelines
2.  Existing tests were executed and passed
3.  No new i18n keys were added
4.  Change is small and self-contained


Prince-Prajapati07:
Hi @pabo99 sir, the PR is ready for review. I rebased onto main, added tests, and ran lint/tests locally. The gh actions run needs a maintainer to approve the workflow. could one of you approve it? Thanks.

frontend/www/js/omegaup/components/common/GlobalNotifications.vue
pabo99: I’m not a fan of adding more complexity to the global notifications component. This should only focus on displaying the visual element where the messages will appear.

Let me explain how I think we could solve this issue:

As you can see, the message that’s displayed is already wrapped in a `Markdown` component. This is a lite version designed specifically to render only basic elements.

We have another, more complete Markdown component that we use to render problems with everything needed. In that component, there’s already a similar action we need to implement where the input/output table is displayed:

Image

Would it be possible for you to try reapplying this in the lite `Markdown` component? This is what’s done in the other component:
https://github.com/omegaup/omegaup/blob/c0a2b0ef1e9fed324671fe16fbd30aa326400e91/frontend/www/js/omegaup/components/problem/ProblemMarkdown.vue#L78-L103

- Where it first selects the element using a selector,
- Then creates the button,
- And adds a click listener.

In that case, it shows a message that the text was copied successfully. That won’t work exactly for you because it would overwrite the message. But since this message will show a token that won’t be displayed again, I’d like the message to remain visible and only show a small note next to the button saying that the token was copied to the clipboard

Prince-Prajapati07:
Thanks @pabo99 sir for the feeback! I refactored this so the copy behavior lives in the markdown component only and cleaned up the globalnotifications changes. I’ve also updated the branch

Let me know if you’d like anything adjusted.

.gitignore
```diff
@@ -128,3 +128,4 @@ cypress/downloads/*
 # to check for inefficient queries.
 # Once the inefficient queries are fixed, this file should be deleted.
 stuff/inefficient_queries.csv
+start.sh
```
pabo99: Is it necessary to add this? If so, could you add a comment explaining why it would be useful to include it?

That way, you can also resolve the conflict that appeared

frontend/www/js/omegaup/components/Markdown.vue
```diff
@@ -42,11 +44,80 @@ export default class Markdown extends Vue {
 
   mounted(): void {
     this.root.innerHTML = this.html;
+    this.injectApiTokenCopyButton();
```
pabo99: Ok, the idea of moving all the logic to this component, besides keeping the `GlobalNotifications` component clean, is to make it reusable in all cases where something needs to be copied, regardless of whether it's a token, a code block, etc.

So, can you rename it to something more generic? For example:
```suggestion
    this.injectCopyButton();
```

frontend/www/js/omegaup/components/Markdown.vue
```diff
@@ -42,11 +44,80 @@ export default class Markdown extends Vue {
...
+    }
+
+    const tokenValue = match[1].trim();
+    const [beforeText, afterText] = template.split(tokenPlaceholder);
+
+    while (paragraph.firstChild) {
+      paragraph.removeChild(paragraph.firstChild);
+    }
+
+    paragraph.appendChild(document.createTextNode(beforeText));
```
pabo99: Continuing with the idea of making this function more generic and reapplying some changes from `ProblemMarkdown`, we would need to find a way to access the token code. Looking through the localization files, it seems everything is plain text.

We can modify the string to include labels that allow us to access it via a selector, and that aren’t easily repeated in other text strings. For example, we can wrap the token in a span > code:
```ts
apiTokenSuccessfullyCreated = "Token successfully created, copy the token since it will be displayed only once: <span><code>%(token)</code></span>"
```
and do the same for other languages.

Then here we could access these elements via `querySelectorAll`:
```suggestion
    this.root.querySelectorAll('span > code').forEach((code) => {
```

So that inside the `forEach` you access the `span` element:
```ts
      const parent = code.parentElement; // span
      // If there are no matching elements, just exit
      if (!parent) return;
      // Prevent injecting multiple times
      if (parent.querySelector('button.copy-btn')) return;
```

frontend/www/js/omegaup/components/Markdown.vue
```diff
@@ -42,11 +44,80 @@ export default class Markdown extends Vue {
...
+
+    const tokenSpan = document.createElement('span');
+    tokenSpan.textContent = tokenValue;
+    paragraph.appendChild(tokenSpan);
```
pabo99: You no longer need to create the span element since we already added it in the source text, so it can be removed
```suggestion
```

frontend/www/js/omegaup/components/Markdown.vue
```diff
@@ -42,11 +44,80 @@ export default class Markdown extends Vue {
...
+
+    const copyButton = document.createElement('button');
+    copyButton.type = 'button';
+    copyButton.className = 'btn btn-light btn-sm ml-2';
```
pabo99: What do you think about replacing the `btn-light `class with `btn-link` instead? That way it won't have borders and only the icon will be shown. Additionally, you can add an extra class in case you need to apply custom styles.
```suggestion
    copyButton.className = 'btn btn-link btn-sm ml-2 copy-btn';
```

frontend/www/js/omegaup/components/Markdown.vue
```diff
@@ -42,11 +44,80 @@ export default class Markdown extends Vue {
...
+    copyButton.appendChild(document.createTextNode('📋'));
+    copyButton.title = (T as any).wordsCopyToClipboard || '';
```
pabo99: nit:
```suggestion
    copyButton.title = T.wordsCopyToClipboard;
```
This is the right way to access localization texts, since there is a linter to prevent using any nonexistent string

frontend/www/js/omegaup/components/Markdown.vue
```diff
@@ -42,11 +44,80 @@ export default class Markdown extends Vue {
...
+    copyButton.setAttribute('aria-label', copyButton.title);
+
+    const successHint = document.createElement('span');
+    successHint.className =
+      'api-token-copy-hint text-success small ml-2 d-none';
+    successHint.textContent = '✓';
```
pabo99: Could you use this emoji, which is more eye-catching ✅? This way, you no longer need to add the `text-success` class, and the `api-token-copy-hint` class can be more general.
```suggestion
      successHint.className = 'copy-hint small ml-2 d-none';
      successHint.textContent = '✅';
```

frontend/www/js/omegaup/components/Markdown.vue
```diff
@@ -42,11 +44,80 @@ export default class Markdown extends Vue {
...
+
+    copyButton.addEventListener('click', (event: Event) => {
+      event.preventDefault();
+      event.stopPropagation();
+      ui.copyToClipboard(tokenValue);
```
pabo99: Now you can access the content to copy as follows:
```suggestion
      ui.copyToClipboard(code.textContent || '');
```

frontend/www/js/omegaup/components/Markdown.vue
```diff
@@ -42,11 +44,80 @@ export default class Markdown extends Vue {
...
+      successHint.classList.remove('d-none');
```
pabo99: If you want to add an animation so that the hint element fades in and fades out you can add something like this:
```ts
        // Force reflow to restart the animation if clicked multiple times
        void successHint.offsetWidth;
        successHint.classList.add('fade-in');
        successHint.classList.remove('fade-out');
        setTimeout(() => {
          successHint.classList.remove('fade-in');
          successHint.classList.add('fade-out');
        }, 2000); // Starts fading out after 2s
        setTimeout(() => {
          successHint.classList.add('d-none');
          successHint.classList.remove('fade-out');
        }, 2500); // Fades out after 2.5s
```

frontend/www/js/omegaup/components/Markdown.vue
```diff
@@ -42,11 +44,80 @@ export default class Markdown extends Vue {
...
+    });
+
+    paragraph.appendChild(copyButton);
+    paragraph.appendChild(successHint);
+    paragraph.appendChild(document.createTextNode(afterText));
```
pabo99: And now you just need to append the elements you created to the span:
```suggestion
      parent.appendChild(copyButton);
      parent.appendChild(successHint);
```

frontend/www/js/omegaup/components/Markdown.test.ts
```diff
@@ -13,4 +28,37 @@ describe('Markdown.vue', () => {
...
   });
+
+  it('Does not inject copy button for non-token messages', () => {
+    const wrapper = shallowMount(omegaup_Markdown, {
+      propsData: {
+        markdown: 'Just a regular notification message.',
+      },
+    });
+
+    expect(wrapper.find('button').exists()).toBe(false);
```
pabo99: To simplify a bit:
```suggestion
    expect(wrapper.find('button').exists()).toBeFalsy();
```

frontend/www/js/omegaup/components/Markdown.test.ts
```diff
@@ -13,4 +28,37 @@ describe('Markdown.vue', () => {
...
+  it('Injects copy button and copies token for API token notification', async () => {
+    const template = (T as any).apiTokenSuccessfullyCreated as string;
+    const token = 'abc123token';
+    const message = template.replace('%(token)', token);
+
+    const wrapper = shallowMount(omegaup_Markdown, {
+      propsData: {
+        markdown: message,
+      },
+    });
```
pabo99: We have a function to format strings with dynamic data, and we can use it here:

```suggestion
    const token = 'abc123token';
    const wrapper = shallowMount(omegaup_Markdown, {
      propsData: {
        markdown: ui.formatString(T.apiTokenSuccessfullyCreated, { token }),
      },
    });
```

frontend/www/js/omegaup/components/Markdown.test.ts
```diff
@@ -13,4 +28,37 @@ describe('Markdown.vue', () => {
...
+
+    const button = wrapper.find('button');
+    expect(button.exists()).toBe(true);
```
pabo99: nit:
```suggestion
    expect(button.exists()).toBeTruthy();
```

frontend/www/js/omegaup/components/Markdown.test.ts
```diff
@@ -13,4 +28,37 @@ describe('Markdown.vue', () => {
...
+
+    await button.trigger('click');
+
+    expect(ui.copyToClipboard).toHaveBeenCalledWith(token);
+
+    const hint = wrapper.find('.api-token-copy-hint');
+    expect(hint.exists()).toBe(true);
```
pabo99: nit:
```suggestion
    expect(button.exists()).toBeTruthy();
```

frontend/www/js/omegaup/components/Markdown.test.ts
```diff
@@ -13,4 +28,37 @@ describe('Markdown.vue', () => {
...
+    });
+
+    const button = wrapper.find('button');
+    expect(button.exists()).toBe(true);
+
+    await button.trigger('click');
+
+    expect(ui.copyToClipboard).toHaveBeenCalledWith(token);
+
+    const hint = wrapper.find('.api-token-copy-hint');
```
pabo99: As suggested in the comments:
```suggestion
    const hint = wrapper.find('.copy-hint');
```

Prince-Prajapati07:
@pabo99 sir! Thanks for the feedback, I’ve refactored the logic to be generic in **markdown.vue**, updated styles and tests, and also removed the accidental **start.sh** entry from .gitignore. Conflicts are resolved now. Please take another look when you get a chance. thanks!

frontend/www/js/omegaup/components/Markdown.vue
```diff
@@ -42,11 +44,61 @@ export default class Markdown extends Vue {
...
+      }
+
+      const copyButton = document.createElement('button');
+      copyButton.type = 'button';
+      copyButton.className = 'btn btn-link btn-sm ml-2 copy-btn';
+      copyButton.textContent = T.wordsCopyToClipboard;
+
+      const copyHint = document.createElement('span');
+      copyHint.className = 'copy-hint';
+      copyHint.textContent = T.wordsCopiedToClipboard || 'Copied';
```
pabo99: nit:
```suggestion
      copyHint.textContent = T.wordsCopiedToClipboard;
```

pabo99 (commented):
One last suggestion

Prince-Prajapati07:
> One last suggestion

Thanks @pabo99 sir! Applied the change.

PR #8881, opened 2026-01-25
Fix: Reduce gap between 'Or upload the file' label and 'Choose file' button (#8829)

### LINKED ISSUE #8829: [BUG] Large gap between "Or upload the file" label and "Choose file" button in Problem Creator
## Current Behavior

In the Problem Creator page (`/problem/creator/`), there is a large horizontal gap between the "Or upload the file" label and the "Choose file" button. This creates an inconsistent visual appearance compared to the "Language" row above it.

## Expected Behavior

The "Or upload the file" label should be positioned closer to the "Choose file" button, similar to how the "Language" label is positioned relative to its dropdown. The layout should be visually consistent across all form rows.

## Steps to Reproduce

1. Navigate to `localhost:8001/problem/creator/` (or the production equivalent)
2. Click on the "Code" tab
3. Observe the spacing between "Or upload the file" label and the "Choose file" button
4. Compare it with the spacing between "Language" label and its dropdown

* Browser name (e.g., Chrome 39, node.js 5.4): Chrome (latest)
* Operating System and version (desktop or mobile): Windows 11 Desktop
* Link to the site where this issue occurs: `/problem/creator/` (Code tab)
* Screenshot:

Image

Yashagarwal9798:
### Description
Fixes #8829

This PR reduces the large horizontal gap between the "Or upload the file" label and the "Choose file" button in the Problem Creator Code tab.

### Changes
- Updated the CSS classes in `CodeTab.vue` to use proper Bootstrap grid alignment
- Label now uses `col-sm-2 col-form-label mb-0 pr-2` to maintain left margin consistency with the "Language" row while reducing right padding
- File input wrapper uses `col-auto pl-0 d-flex align-items-center` to minimize the gap between label and button

### Before
Large gap between "Or upload the file" label and "Choose file" button, inconsistent with the "Language" row layout.

### After
The label and button are now closer together while maintaining proper left alignment consistent with other form rows.

# Checklist:

image

### DIFF
--- a/.github/workflows/assign-issues.yml
+++ b/.github/workflows/assign-issues.yml
@@ -8,6 +8,11 @@ on:
         - cron: "0 3 * * *"
     workflow_dispatch:
 
+# Prevent race condition when multiple users comment /assign simultaneously
+concurrency:
+    group: assign-issue-${{ github.event.issue.number || 'scheduled' }}
+    cancel-in-progress: false
+
 jobs:
     assign:
         runs-on: ubuntu-22.04


pabo99 (commented):
Since you're already modifying the styles of that form, could you also review each breakpoint to make sure it looks as well‑adjusted as possible? Preferably, the texts should not split into more than one line. 

Ok:
Image

Fix:
Image

Fix:
Image

Fix:
Image

It seems to be ok:
Image

Yashagarwal9798:
@pabo99 I have pushed the requested changes. Could you please take a look when you have a moment?

image
image

frontend/www/js/omegaup/components/problem/creator/code/CodeTab.vue
```diff
@@ -40,13 +40,16 @@
...
         </label>
-        <div class="col-sm-7">
+        <div
+          class="col-12 col-sm-auto pl-sm-0 d-flex align-items-center overflow-hidden"
+        >
           <input
             data-problem-creator-code-input
-            class="w-100"
+            class="text-truncate"
+            style="max-width: 100%"
```
pabo99: We usually don’t use inline styles. Instead, could you use a Bootstrap class? This [one](https://getbootstrap.com/docs/4.0/utilities/sizing/) does exactly what you need:
```suggestion
            class="text-truncate mw-100"
```

pabo99 (commented):
Could you fix the styles for the language field as well?

Image

frontend/www/js/omegaup/components/problem/creator/code/CodeTab.vue
Yashagarwal9798: Hi @pabo99, thanks for the review!

I've made both changes:

1. **Applied your suggestion** - Replaced the inline `style="max-width: 100%"` with Bootstrap's `mw-100` class on the file input.

2. **Fixed the Language field styling** - Updated the Language row to use the same responsive layout classes (`col-12 col-sm-auto`, `form-group row align-items-center`, etc.) as the file upload row, which should fix the broken appearance.


image

PR #8902, opened 2026-01-28
fix: Align description text inline on Clone course page

### LINKED ISSUE #8875: [BUG] Description text not aligned properly on Clone course page
## Current Behavior

On the "Clone course" page (`/course/{alias}/clone/`), the description text is displayed below the "Description:" label instead of being inline beside it. This creates an inconsistent layout compared to other fields like "Name:" and "Created by:" which display their values on the same line as the labels.

## Expected Behavior

The description text should be displayed beside/inline with the "Description:" label, maintaining consistency with the other fields on the page. The layout should follow the same pattern as:
- **Name:** value (inline)
- **Created by:** value (inline)
- **Description:** value (should also be inline)

## Steps to Reproduce:

1. Navigate to any course edit page
2. Click on "Clone" to open the clone course page
3. Click on Generate clone link and then use that link to open a clone course page.
4. Observe that the description text appears below the "Description:" label instead of beside it

* Browser name: Chrome
* Operating System and version: Windows
* Screenshot:

Image

Yashagarwal9798:
Replaced omegaup-markdown component with a simple span element to display the course description inline with its label, maintaining consistency with other fields like Name and Created by.

Fixes #8875

# Description

Fixed the description text alignment issue on the Clone course page (`/course/{alias}/clone/`). The description was displaying below the "Description:" label instead of inline beside it.

**Problem:** The `omegaup-markdown` component renders as a `<div>` element containing a `<p>` tag with default margins, causing the description text to appear on a new line with extra spacing.

**Solution:** Replaced the `omegaup-markdown` component with a simple `<span>` element to display the description text inline, matching the behavior of the "Name:" and "Created by:" fields.

image

# Checklist:


frontend/www/js/omegaup/components/course/CloneWithToken.vue
```diff
@@ -17,7 +17,7 @@
         </p>
         <p>
           <span class="font-weight-bold">{{ T.wordsDescription }}: </span>
-          <omegaup-markdown :markdown="course.description"></omegaup-markdown>
+          <span>{{ course.description }}</span>
```
pabo99: What happens if the description contains rich elements, such as a link? I think we should keep the Markdown component and instead adjust the way the information is displayed. Maybe you can adapt the rest of the displayed data so that it uses the same styles as the description.

You could also try adjusting the margins and padding of the Markdown for this case. It seems that the component already has an attribute that removes them, try using `full-width`

Yashagarwal9798: @pabo99 I've pushed the requested changes. Could you please review them whenever you have time?

Yashagarwal9798:
Happy to help!!

PR #8905, opened 2026-01-29
Add System_Settings table and Migration

PrasannaMishra001:
# Description

This PR adds the database migration and schema updates required to introduce a System_Settings table, which will support site-wide configuration options in a safe and deployable way.

## Changes
- adds migration script `00260_system_settings.sql` to create the System_Settings table
- updates `schema.sql` to include the new table definition
- introduces initial setting `ephemeral_grader_enabled` with default value enabled

## Table structure
- `setting_key`: unique identifier for each setting
- `setting_value`: value stored as text
- `setting_description`: human-readable documentation
- created_at and updated_at timestamps for audit purposes

This is part 1 of the admin-configurable EphemeralGrader feature. Follow-up PRs will handle:
- backend helpers and api endpoints
- admin ui for managing settings
- integration with relevant controllers

Part of: #8319

# Comments

This PR intentionally contains only the migration and schema changes, following the database deployment policy. DAO and application-level changes will be submitted in subsequent PRs after this is stable.

# Checklist:
- [x] The code follows the [coding guidelines](https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/Coding-guidelines.md) of omegaUp.
- [x] The tests were executed and all of them passed.
- [x] If you are creating a feature, the new tests were added.
- [x] If the change is large (> 200 lines), this PR was split into various Pull Requests. It's preferred to create one PR for changes in controllers + unit tests in PHPUnit, and then another Pull Request for UI + tests in Jest, Cypress or both.

PR #8906, opened 2026-01-29
Add backend API endpoints for system settings management

### LINKED ISSUE #8319: [FEATURE] Permitir que se deshabilite el EphemeralGrader desde la configuración del sitio

## Context

PrasannaMishra001:
# Description

adds backend support for managing site wide system settings via admin apis

introduces a systemsettings helper layer to abstract database access and provide typed helpers for reading and updating boolean settings
adds admin api endpoints to fetch and update system settings securely with proper validation and authorization
registers api routes for the new admin endpoints following existing api conventions
adds translation strings for all supported languages to ensure consistent api error and response messages

this pr builds on the system_settings migration and schema changes and focuses only on backend apis and helpers without any ui changes

fixes: #8319

# Comments

this is part 3 of the ephemeralgrader admin toggle feature and depends on the system settings migration pr
the apis and helper methods are designed to be reusable for future site wide configuration flags

Depends on**
- Waiting for **PR #8905**
- Waiting for **PR #8906** 

# Checklist:


frontend/server/src/SystemSettings.php
```diff
@@ -0,0 +1,54 @@
...
+     * @return bool
+     */
+    public static function getBooleanSetting(
+        string $settingKey,
+        bool $default
+    ): bool {
+        try {
+            /** @var null|string $value */
+            $value = \OmegaUp\MySQLConnection::getInstance()->GetOne(
+                'SELECT `setting_value` FROM `System_Settings` WHERE `setting_key` = ?',
```
pabo99: All database-related changes must live in the DAOs, so this PR will need to remain on hold until the PR that introduces the DAO creation is approved

PrasannaMishra001:
@pabo99 this backend for the system settings feature updated , adds admin get/update endpoints, gating the grader IDE page on the setting and exposes the flag on the problem details so the frontend can hide it. Review whenever you have time.

frontend/server/src/Controllers/Admin.php
```diff
@@ -321,4 +321,76 @@ public static function getMaintenanceMessage(?string $lang = null): ?array {
...
+     *
+     * @return array{status: string}
+     */
+    public static function apiUpdateSystemSettings(\OmegaUp\Request $r): array {
+        $r->ensureMainUserIdentity();
+        if (!\OmegaUp\Authorization::isSystemAdmin($r->identity)) {
+            throw new \OmegaUp\Exceptions\ForbiddenAccessException();
+        }
+
+        $ephemeralGraderEnabled = $r->ensureOptionalBool(
```
pabo99, line 360: This configuration should always exist. Could we use `ensureBool` instead and let it fail when this configuration is missing?

frontend/server/src/Controllers/Admin.php
```diff
@@ -321,4 +321,76 @@ public static function getMaintenanceMessage(?string $lang = null): ?array {
...
+            'ephemeral_grader_enabled'
+        );
+        if (!is_null($ephemeralGraderEnabled)) {
+            \OmegaUp\DAO\SystemSettings::setBooleanSetting(
```
pabo99, line 364: Since this value is likely to be read frequently and updated rarely, it might be worth caching this setting and future ones. If we do that, we could invalidate the cache here after an update

frontend/server/src/Controllers/Admin.php
```diff
@@ -321,4 +321,76 @@ public static function getMaintenanceMessage(?string $lang = null): ?array {
...
+            throw new \OmegaUp\Exceptions\ForbiddenAccessException();
+        }
+
+        return [
+            'entrypoint' => 'admin_settings',
+            'templateProperties' => [
+                'title' => new \OmegaUp\TranslationString(
+                    'omegaupTitleAdminSettings'
+                ),
+                'payload' => [],
```
pabo99, line 392: Could we list all available settings here?

frontend/server/src/Controllers/Problem.php
```diff
@@ -3042,6 +3042,13 @@ public static function getProblemDetails(
                 intval($loggedIdentity->identity_id)
             );
         }
+
+        // Lets the problem page hide the embedded ephemeral grader when an
+        // admin has disabled it from the site settings.
+        $response['ephemeralGraderEnabled'] = \OmegaUp\DAO\SystemSettings::getBooleanSetting(
```
pabo99: Here, you could read the settings from the cache. You can look for other examples in the codebase to see how to access them.

Thinking about it a bit more, maybe we should avoid adding these queries to every route where the settings are needed, since it could become harder to control where they are being loaded.

Could we move this logic to `commonPayload` instead? Since it's included in all views, it would be easier to expose the settings from there and access them once they are cached

pabo99 (commented):
Just a few suggestions and it's okay to me

frontend/server/src/Controllers/Admin.php
PrasannaMishra001: added caching now in DAO, invalidation point also covered. I am only caching when row exists and storing bool as an int as `Cache::get()` treats stored false as a miss. added cache clear as well as it is not truncated unlike DB.

frontend/server/src/Controllers/Problem.php
PrasannaMishra001: ok so i found `UITools::getCommonPayload()` which is included in every view. So moved this logic there and will expose now cached DAO, remaining i am doing in frontend one where it reads value from `commonPayload`

frontend/server/src/Controllers/Admin.php
PrasannaMishra001: kept `ensureBool` optional so an admin can update one of several settings later without sending all of them.

frontend/server/src/Controllers/Admin.php
PrasannaMishra001: It already returns a settings object, so each new setting is just another key there and i kept the endpoint as the general read API now that the value is also in the common payload.

PrasannaMishra001:
@pabo99 please review this if okay, will take system_settings work forward

PrasannaMishra001:
@pabo99 closing this one, its backend already landed in main when #8938 (which was stacked on top of this) got squash merged, so apiGetSystemSettings / apiUpdateSystemSettings and the cached SystemSettings reads are all in main now. nothing left here to merge on its own.

PR #8908, opened 2026-01-29
Fixes #8808  Form validation focus fixed

### LINKED ISSUE #8808: [Form validation does not focus the missing field when creating a contest]
## Current Behavior

While creating a contest, if some required fields are left blank and I click the "Create Contest" button, the page only shows a generic message like “Please fill the details”.  

The form does NOT automatically point to or focus the specific field that was missed, so the user has to manually search which input is empty.

## Expected Behavior

When the user submits the form with missing required fields, the platform should:

- Highlight the exact field that is empty/invalid  
- Scroll or focus to that field automatically  
- Show a clear message near that input  

This would make the process smoother and less confusing for users.

## Steps to Reproduce

1. Go to the Create Contest page  
2. Leave one or more required fields empty  
3. Click the "Create Contest" button  
4. Observe that only a generic message appears and the form does not point to the missing field

## Environment

- Browser: Chrome (latest)
- Operating System: Desktop – Windows/Linux
- Site: OmegaUp contest creation page
- Screenshot: (attached above)

Image

ParvTyagi:
## Summary
This PR addresses Issue #8808 by improving the validation of the contest creation form.
Instead of showing only a generic error message, the form now guides users
directly to the missing or invalid required fields.

## Problem (Issue #8808)
When a user submitted the "Create Contest" form with required fields left empty,
The platform only displayed a generic message like "Please fill in the details".
The form did not highlight, scroll to, or focus on the specific missing input,
forcing users to search for the error manually.

## Solution
This change enhances the validation behaviour by:
- Highlighting the exact required field that is missing or invalid
- Automatically scrolling to and focusing on that field
- Displaying a clear, contextual error message near the input

This results in a smoother and more user-friendly contest creation experience.

## How to Test

1. Navigate to the Create Contest page
2. Leave one or more required fields empty
3. Click the "Create Contest" button
4. Verify that:
   - The page scrolls to the missing field
   - The field is visually highlighted
   - An error message appears near the field

### Before

https://github.com/user-attachments/assets/ac003b79-b4ca-47a8-9b66-0fff9396cc45

### After

https://github.com/user-attachments/assets/6d2c8447-98d0-47ec-b743-8459aa5d5ffb

## Related Issue
Fixes #8808

### DIFF
--- a/frontend/www/js/omegaup/contest/new.ts
+++ b/frontend/www/js/omegaup/contest/new.ts
@@ -1,7 +1,6 @@
 import { OmegaUp } from '../omegaup';
 import { types } from '../api_types';
 import Vue from 'vue';
-import T from '../lang';
 import contest_NewForm from '../components/contest/Form.vue';
 import * as ui from '../ui';
 import * as api from '../api';


ParvTyagi:
@pabo99 
I’ve addressed the feedback by updating the contest form and fixing the related lint issues.

This PR focuses on:
Cleaning up the Vue template to comply with project lint rules
Improving form validation handling
Adding the required translation strings

I’d really appreciate it if you could take a look and let me know if there’s anything I should adjust or improve further.

Thanks for your time!

frontend/www/js/omegaup/components/contest/Form.vue
```diff
@@ -699,7 +1051,181 @@ export default class Form extends Vue {
...
+          this.localErrors[fieldName] = T.contestNewFormTitleRequired;
+          return false;
+        }
+        break;
+      case FieldName.Alias:
+        if (!this.alias || this.alias.trim().length === 0) {
+          this.localErrors[fieldName] = T.contestNewFormShortTitleRequired;
+          return false;
+        }
+        if (!/^[a-z0-9-]+$/.test(this.alias)) {
```
pabo99: Oh no, where did you get these regular expressions from?

For the regular expression for the alias, you can base it on this:
https://github.com/omegaup/omegaup/blob/bd1b09ceb1e8b819034c968b0cb6d439088342e3/frontend/server/src/Validators.php#L227

You need to change the error message as well

frontend/www/js/omegaup/components/contest/Form.vue
```diff
@@ -4,30 +4,70 @@
...
+        </ul>
+        <button
+          type="button"
+          class="close"
+          aria-label="Close"
+          @click="localErrors = {}"
+        >
+          <span aria-hidden="true">&times;</span>
+        </button>
+      </div>
```
pabo99: We already have a global notifications system. Is there a way to use it instead of the one you’re creating?

The notifications support Markdown, so they can perfectly render bold text as well as list bullets

ParvTyagi: Hi @pabo99, thanks for the detailed review!

I’ve updated the alias validation to match the server-side regex in Validators.php (/^[a-zA-Z0-9_-]+$/) and adjusted the corresponding error message.

I’ve also removed the custom validation summary and switched to using the existing global notifications system so we can take advantage of Markdown rendering for bold text and lists.

Please let me know if everything looks good now, or if you’d like any further changes.

frontend/templates/en.lang
```diff
@@ -384,38 +384,48 @@ contestNew = "New contest"
 contestNewForm = "IOI Style"
 contestNewFormAdmissionMode = "Admission mode"
 contestNewFormAdmissionModeDescription = "Select contest admission mode. There are three different modalities:<br /><ul><li>Private: Contest is only visible to coders who have been explicitly invited.</li><li>Registration/open: Anyone can register, only people admitted by admins have access.</li><li>Public: Anyone can participate.</li></ul>"
+contestNewFormAliasHelp = "Only lowercase letters, numbers, and hyphens. Cannot be changed later."
```
pabo99: Uppercase letters are also valid, could you update this string?

frontend/templates/en.lang
```diff
@@ -384,38 +384,48 @@ contestNew = "New contest"
...
+contestNewFormAliasInvalid = "Alias can only contain lowercase letters, numbers, and hyphens"
```
pabo99: Same here

frontend/www/js/omegaup/components/contest/Form.vue
```diff
@@ -227,93 +388,123 @@
...
+                  >
+                    {{ localErrors[FieldName.Scoreboard] }}
+                  </div>
+                  <small class="form-text text-muted">{{
+                    T.contestNewFormScoreboardPercentRange
+                  }}</small>
                 </div>
               </div>
-              <div class="row">
+              <div v-if="canSetRecommended" class="row">
```
pabo99: This recommended section should be kept as is, since it's where the user is informed about how their contest can appear in the recommended list

frontend/www/js/omegaup/components/contest/Form.vue
```diff
@@ -706,77 +1135,192 @@ export default class Form extends Vue {
...
   onSubmit() {
+    if (!this.validateForm()) {
+      const errorsList = Object.values(this.localErrors)
+        .filter(Boolean)
+        .map((error) => `- ${error}`)
+        .join('\n');
+      displayStatus({
+        message: `**${T.formValidationSummaryTitle}**\n${errorsList}`,
+        type: MessageType.Danger,
+      });
```
pabo99: Ok, let’s adjust this a bit.

`displayStatus` is the general function for showing messages, but there is another function that already contains the specific type we need. In this case, you can use the function `ui.error()`. This one only takes a single string parameter:
```suggestion
      ui.error(`**${T.formValidationSummaryTitle}**<ul>${errorsList}</ul>`);
```
Notice that I wrapped the error list in a `<ul>` tag because I found that Markdown does not render the information correctly if we use line breaks with `\n`.

frontend/www/js/omegaup/components/contest/Form.vue
```diff
@@ -706,77 +1135,192 @@ export default class Form extends Vue {
...
+    this.currentScoreMode = preset.currentScoreMode;
+    this.localErrors = {};
   }
 
   onSubmit() {
+    if (!this.validateForm()) {
+      const errorsList = Object.values(this.localErrors)
+        .filter(Boolean)
+        .map((error) => `- ${error}`)
+        .join('\n');
```
pabo99: It also fails to render the line break here, so you can use:
```suggestion
      const errorsList = Object.values(this.localErrors)
        .filter(Boolean)
        .map((error) => `<li>${error}</li>`)
        .join('');
```

pabo99 (commented):
The test failures look legit. Apply these suggestions and then we’ll check the cause.

frontend/www/js/omegaup/components/contest/Form.vue
ParvTyagi: “I addressed your feedback.”

frontend/www/js/omegaup/components/contest/Form.vue
```diff
@@ -706,79 +1143,196 @@ export default class Form extends Vue {
...
+    this.penaltyType = preset.penaltyType;
+    this.showScoreboardAfter = preset.showScoreboardAfter;
+    this.currentScoreMode = preset.currentScoreMode;
+    this.localErrors = {};
   }
 
   onSubmit() {
+    if (this.update) {
+      this.$emit('update-contest');
+    }
```
pabo99: This emit should normally be placed after all the validations. It actually seems to be already there, so could you remove it from here?
```suggestion
```

frontend/www/js/omegaup/components/contest/Form.vue
```diff
@@ -706,79 +1143,196 @@ export default class Form extends Vue {
...
+        }
+      });
+      return;
+    }
+
+    this.isSubmitting = true;
+
     if (!this.languages || this.languages.length === 0) {
       this.$emit('invalid-languages');
+      this.isSubmitting = false;
```
pabo99: This condition was added in a recent PR, but with your changes, it seems it may no longer be necessary.

Could you help me validate? If it’s not needed, you can remove all the logic related to `invalid-languages` for contests. 

Perhaps the `clear-language-error` logic could also be removed

frontend/www/js/omegaup/components/contest/Form.vue
```diff
@@ -143,17 +276,27 @@
                       :multiple="true"
                       :placeholder="T.contestNewFormLanguages"
                       :close-on-select="false"
-                      :allow-empty="true"
+                      :allow-empty="false"
```
pabo99: In the PR I mentioned earlier, the value of this attribute was also changed to allow removing all elements. Could you revert it to how it was?
```suggestion
                      :allow-empty="true"
```

ParvTyagi: Done.

frontend/www/js/omegaup/components/contest/Form.vue
ParvTyagi: A few tests were failing due to this, so I had introduced that logic. I’ve removed it now as per your suggestion.

frontend/www/js/omegaup/components/contest/Form.vue
ParvTyagi: You’re right — after the validation refactor, Languages is now handled entirely through localErrors like the rest of the fields.
The old invalid-languages and clear-language-error logic is no longer necessary.
I’ve removed that condition and the related event/listener so the validation is fully self-contained within the form.

ParvTyagi:
@pabo99 
These is a issue with the test data.
The test is failing because both `initialStartTime` and `initialFinishTime` are created with `new Date()`, resulting in identical timestamps. The form's validation requires `finishTime > startTime`, so when they're equal, validation fails and the `update-contest` event is never emitted.

ParvTyagi:
I’ve made some changes in `Form.test.ts` to address this. Could you please review them and let me know if anything else needs improvement?

pabo99 (approved):
Thank you, so much!

Issue #8901, opened 2026-01-28
[BUG] Race condition allows multiple users to self-assign the same issue simultaneously


Yashagarwal9798:
## Current Behavior

When two or more users comment `/assign` on an issue at approximately the same time, all of them get assigned to the issue instead of only the first one. This happens because the GitHub Action workflow runs in parallel for each comment, and each run sees the issue as unassigned before any assignment is completed.

## Expected Behavior

Only the first user who comments `/assign` should be assigned to the issue. Subsequent `/assign` requests on an already-assigned issue should be rejected with an appropriate message (e.g., "This issue is already assigned to @username").

## Steps to Reproduce

1. Open any unassigned issue
2. Have two different users comment `/assign` within a few seconds of each other
3. Observe that both users get assigned to the issue

## Root Cause

This is a race condition in the workflow. The `assign-issue-action` runs in parallel when multiple `issue_comment` events are triggered simultaneously. Each workflow run checks if the issue is assigned before the other run completes its assignment, resulting in both users being assigned.

## Suggested Fix

Add `concurrency` control to the workflow to ensure only one assignment workflow runs per issue at a time

Yashagarwal9798:
@pabo99 This specific error also occurred in issue #8900. I'd love to work on resolving it. Could you please assign me?

waygeance:
Hi @Yashagarwal9798, I think we can add priority here: the person who opened the issue should get priority in assignment. We can set a cooling period of 5 minutes during which, if anyone tries to assign themselves, they won’t be able to. That time will be given only to the person who raised the issue.

As for the race condition, we can make the time more precise ,if it’s currently measured in minutes, we can go for seconds and so on. What do you think? Any points to add on this @pabo99 ? Also @Yashagarwal9798 if you want we can co-author this PR?

Ankitsinghsisodya:
Hi @waygeance, thanks for the suggestions!

regarding the cooling period and priority logic, I think that might add a bit too much complexity to the workflow script. We generally want to keep the bot logic as stateless and simple as possible.

On the technical side, making the timestamps more precise unfortunately won't solve the race condition. The issue here is that the GitHub Action runners are executing in parallel. Even with nanosecond precision, if two runners check the "Is Assigned?" state at the same time (before either has written the "Assigned" state), they will both proceed.

[...]

PR #8917, opened 2026-01-30
fix: add concurrency control to prevent race condition in issue assignment #8901

Yashagarwal9798:
# Description

Fixes a race condition where multiple users commenting `/assign` simultaneously on the same issue would all get assigned, instead of only the first user.

**Root Cause:** When multiple `issue_comment` events trigger the workflow in parallel, each run checks if the issue is assigned before any other run completes its assignment.

**Solution:** Added GitHub Actions `concurrency` control with the issue number as the group key. This ensures only one assignment workflow runs per issue at a time, with subsequent workflows queued instead of running in parallel.

Fixes: #8901

# Comments

This is a minimal, non-breaking change that uses GitHub Actions' built-in concurrency feature - no external dependencies or complex locking mechanisms needed.

# Checklist


Yashagarwal9798:
Happy to help!!

PR #8891, opened 2026-01-26
Allow EphemeralGrader to be enabled/disabled in admin settings

PrasannaMishra001:
# Description

This pr adds a site level admin setting to enable or disable the ephemeralgrader ide dynamically without redeploying the platform

- introduced a system_settings based configuration with dao helpers and admin apis to persist and retrieve the ephemeralgrader enabled state
- enforced the setting across backend flows including grader ide rendering and problem details so users cannot access the ide when disabled
- added a new admin settings ui with a toggle that auto saves changes shows success or error feedback and remains consistent after reload
- integrated translations webpack entrypoint cache prevention and loading guards to ensure correct rendering and no flicker or stale state
- added an admin only navigation entry so the feature is discoverable and usable by site administrators

Fixes: #8319

Screenshot 2026-01-26 205019

https://github.com/user-attachments/assets/9418f1d6-a230-4bdd-b0ef-e0eec14a87ed

# Comments

this change touches both backend and frontend to ensure the setting is reliable ,discoverable across the platform  
please review the system_settings usage and admin api naming as these may be reused for future site wide configurations

# Checklist

- [x] The code follows the [coding guidelines](https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/Coding-guidelines.md) of omegaUp.
- [x] The tests were executed and all of them passed.
- [x] If you are creating a feature, the new tests were added.
- [ ] If the change is large (> 200 lines), this PR was split into various Pull Requests.

PR #8918, opened 2026-01-30
Added DAO layer for System_Settings table

### LINKED ISSUE #8319: [FEATURE] Permitir que se deshabilite el EphemeralGrader desde la configuración del sitio

PrasannaMishra001:
## Description

This PR introduces the **DAO/VO layer** for the `System_Settings` table and refactors existing code to use it,keeping omegaup’s standard database access patterns.

All database access related to system settings is now centralized in the DAO layer instead of using direct SQL queries.

This PR is a **dependency for PR #8906**, which implements the backend logic for system settings.

Fixes: #8319 

# Comments
- This PR only adds the DAO/VO layer and does not introduce user-facing changes.
- It should be **merged before PR #8906** to keep the feature split clean and reviewable.

 Depends on
- Waiting for **PR #8905**
# Checklist:

### DIFF
--- a/frontend/server/src/DAO/SystemSettings.php
+++ b/frontend/server/src/DAO/SystemSettings.php
@@ -1,0 +1,118 @@
+<?php
+
+namespace OmegaUp\DAO;
+
+/**
+ * SystemSettings Data Access Object (DAO).
+ *
+ * Esta clase contiene toda la manipulacion de bases de datos que se necesita
+ * para almacenar de forma permanente y recuperar instancias de objetos
+ * {@link \OmegaUp\DAO\VO\SystemSettings}.
+ */
+class SystemSettings extends \OmegaUp\DAO\Base\SystemSettings {
+    /**
+     * Get a system setting by key.
+     *
+     * @param string $key The setting key
+     * @return \OmegaUp\DAO\VO\SystemSettings|null The setting object or null if not found
+     */
+    public static function getByKey(string $key): ?\OmegaUp\DAO\VO\SystemSettings {
+        $fields = join(
+            ', ',
+            array_keys(
+                \OmegaUp\DAO\VO\SystemSettings::FIELD_NAMES
+            )
+        );
+        $sql = "SELECT {$fields} FROM System_Settings WHERE setting_key = ? LIMIT 1;";
+        /** @var array{created_at: \OmegaUp\Timestamp, setting_description: null|string, setting_id: int, setting_key: string, setting_value: null|string, updated_at: \OmegaUp\Timestamp}|null */
+        $row = \OmegaUp\MySQLConnection::getInstance()->GetRow($sql, [$key]);
+        if (empty($row)) {
+            return null;
+        }
+        return new \OmegaUp\DAO\VO\SystemSettings($row);
+    }
+
+    /**
+     * Get a boolean system setting value.
+     *
+     * @param string $key The setting key
+     * @param bool $default Default value if setting not found
+     * @return bool The boolean setting value
+     */
+    public static function getBooleanSetting(
+        string $key,
+        bool $default = true
+    ): bool {
+        $setting = self::getByKey($key);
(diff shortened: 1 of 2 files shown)


vijay-x-dev:
Hi @PrasannaMishra001 
Once PR #8905 is merged and stabilized, please update `dao_schema.sql` in this PR with the corresponding changes.

PrasannaMishra001:
@pabo99 these are DAO changes. I simply cherry-picked them from https://github.com/omegaup/omegaup/pull/8891 and took this DAO part as one of its sub-PRs. Since PR 1 is already merged, I’ll update the remaining backend changes right after this one is merged. Please review these and feel free to suggest anything from your side.

PrasannaMishra001:
@pabo99 pls review this part 2 of #8319  whenever you get chance to continue part3 which will close the issue

frontend/server/src/DAO/SystemSettings.php
```diff
@@ -0,0 +1,112 @@
...
+ */
+class SystemSettings extends \OmegaUp\DAO\Base\SystemSettings {
+    /**
+     * Get a system setting by key.
+     *
+     * @param string $key The setting key
+     * @return \OmegaUp\DAO\VO\SystemSettings|null The setting object or null if not found
+     */
+    public static function getByKey(string $key): ?\OmegaUp\DAO\VO\SystemSettings {
+        $sql = 'SELECT * FROM System_Settings WHERE setting_key = ? LIMIT 1;';
```
pabo99: This (`SELECT * FROM ...`) has caused issues in the past when new fields are added to a table, since there can be a deploy window where the DAOs and the DB schema are out of sync. To mitigate this, we introduced `FIELD_NAMES`, which contains a static list of fields and helps keep these changes synchronized without affecting the current behavior of the DAOs:
```suggestion
        $fields = join(
            ', ',
            array_keys(
                \OmegaUp\DAO\VO\SystemSettings::FIELD_NAMES
            )
        );
        $sql = "SELECT {$fields} FROM System_Settings WHERE setting_key = ? LIMIT 1;";
```

frontend/server/src/DAO/SystemSettings.php
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * Get a boolean system setting value.
+     *
+     * @param string $key The setting key
+     * @param bool $default Default value if setting not found
+     * @return bool The boolean setting value
+     */
+    public static function getBooleanSetting(
```
pabo99, line 42: I see that you have getters/setters for boolean values and string values. Would it also be necessary to add support for integer values?

This raises another question for me: how do we know what data type each configuration returns? Would it make sense to include an additional field in the table to indicate the setting's type?

pabo99 (commented):
Just a couple of suggestions. Would it also be possible to add tests covering the functions you're introducing?

frontend/server/src/DAO/SystemSettings.php
PrasannaMishra001: done, now as per #6713 .

frontend/server/src/DAO/SystemSettings.php
PrasannaMishra001: for this parent issue, only setting in use is boolean ephemeral_grader_enabled one and string access ones cover the rest, so kept it to bool and string here. 
And for int values and setting_type column so a type column we can add after a basic editor with frontend is created later. 
I think it is cleaner to add them after these merged. I will raise separate issue later to take this as well.

PrasannaMishra001:
@pabo99 in frontend PR of admin settings page also because before rendering we have to know the datatype in either case while calling the checkbox for bool ,also the names name getBooleanSetting() self explain. This is also how maintenance and experiments system already work, neither storing the type. So i think settings type column not required at the moment, or am i missing something ?

pabo99 (approved):
Great!

Issue #8920, opened 2026-01-30
[FEATURE]  Implement autocomplete search functionality on the compare coders page


vijay-x-dev:
Implement autocomplete search on the compare coders page using the existing `omegaup-common-typeahead` component.  Users can search and select coders from a dropdown.

PR #8922, opened 2026-01-30
Add autocomplete search to compare coders page

vijay-x-dev:
# Description

Added autocomplete search functionality to the compare coders page, similar to the search coder feature. Users can now search and select coders from a dropdown with matching suggestions. The Compare button is only enabled when both users are entered.

Fixes: #8920 

https://github.com/user-attachments/assets/a8c0d71f-fbcb-4c4d-8438-18e88c7ae58b

# Checklist:


frontend/www/js/omegaup/user/compare.ts
```diff
@@ -64,6 +68,31 @@ OmegaUp.on('ready', () => {
...
+          'update-search-result-users': ({
+            query,
+            field,
+          }: {
+            query: string;
+            field: 'user1' | 'user2';
+          }) => {
+            api.User.list({ query })
+              .then(({ results }) => {
+                const formattedResults = results.map(
```
pabo99, line 120: Could you exclude from the list the user that is being compared against, if it is already selected?

This way, we avoid comparing a user against themselves.

Here is an example of how problems are excluded; maybe it can help you:
https://github.com/omegaup/omegaup/blob/58a6409eb51e5cd7574435c4af85af7e3baf3b49/frontend/www/js/omegaup/contest/edit.ts#L191-L206

vijay-x-dev:
Hi @pabo99 

Thanks for the feedback! I've implemented both requested features:

**`1. Preload usernames from URL parameters:`**
When users navigate from the ranking list view the usernames are now preloaded and displayed as chips with the same formatting as if they were manually searched (showing "username (Name)" format). This is handled in the `mounted()` hook.


**`2. Exclude already selected user from search results:`**
I've added filtering logic in the `update-search-result-users` handler. When searching in one field, the user already selected in the other field is excluded from the results. This prevents users from comparing themselves against themselves.


https://github.com/user-attachments/assets/26964b0a-d1e0-4229-82e2-53daac35c76e

Issue #8919, opened 2026-01-30
[REFACTOR] Server Controller layer


xx-Tasneem-Ahmed-xx:
## Content

The controller layer contains some issues:

- Many unused function parameters and local variables
- Cognitive complex functions
- Missing default case statement
- More than 3 return statements in a single function

## Impact of the Problem

As a new contributor, I need the code to be readable and easier to trace.
As a developer, I need to be able to test individual methods in isolation.

## Proposed Change

- Remove all unused variables
- Separate complex functions to smaller helper ones while preserving the function behaviour
- Add default case statement to fall back on. (I intend to add a default value when appropriate but there are other options for when it isn't: throw an error, simply log the error or fall back silently). I would like to hear your opinion on the preferred strategy @pabo99 .
- Have only one return statement while preserving the exact logic and outputs.

### Note
I plan to address this issue incrementally by submitting small prs, each refactoring two or three controllers for easy review.

pabo99:
Hi, @xx-Tasneem-Ahmed-xx, it sounds great!

Is it possible to start with controllers that are less complex to better estimate the scope of this activity? For example, you could begin with the following controllers:

- `ACL.php`
- `Admin.php`
- `Authorization.php`
- `Badge.php`
- `CarouselItems.php`

This approach will help evaluate how extensive the refactoring might become before moving on to more complex controllers

xx-Tasneem-Ahmed-xx:
That’s a fair point, @pabo99 , thank you. I agree that starting with less complex controllers is a good way to estimate the scope. I’ll open a small demo PR refactoring one or at most two of the less complex controllers.

For clarity, the controllers you referenced are already in good shape at this stage and don’t require refactoring yet.

pabo99:
Hi @xx-Tasneem-Ahmed-xx, can you continue with another change similar to #8929? This way you can complete the 3 merged PRs with the GSoC label to register for Phase 3, which corresponds to the review of proposal drafts.

xx-Tasneem-Ahmed-xx:
Thank you so much for the heads up, @pabo99. I will follow up with a pr ASAP.

PR #8929, opened 2026-01-31
refactor: certificate and controller php files

xx-Tasneem-Ahmed-xx:
# Description
## Here is a summary of the changes:
1. **Certificate.php**

- **getMonthName()** - Replaced switch statement with match expression
Before: Used a 12-case switch statement (30 lines)
After: PHP 8.0 `match` expression with literal translation strings (~16 lines)

- **getPlaceSuffix()** 
Before: Multiple `if-elseif` statements assigning to intermediate variable
After: Direct `match` expression for modulo cases
  
- **getCertificatePdf()** - Consolidated multiple early returns
Before: Multiple if statements with early returns
After: Uses if-elseif-else chain with a single return at the end

2. **Controller.php**
**updateValueProperties()** Method now uses 3 helper methods instead of a complex loop:
- **parsePropertyConfiguration()** - Separates the logic for parsing property configurations (handles both simple string properties and complex array configurations with transforms/importance flags)
- **getTransformedValue()** - Applies transformation functions to property values if provided
- **isImportantChange()** - Determines whether a property change should be flagged as important

Part of: #8919

# Comments
No functional or behavioral changes were introduced. All existing logic and outputs remain the same.

# Checklist:


frontend/server/src/Controllers/Certificate.php
```diff
@@ -107,32 +107,21 @@ public static function getValidationForTypeScript(\OmegaUp\Request $r) {
...
-                return $translator->get('certificatePdfMonth9');
-            case 10:
-                return $translator->get('certificatePdfMonth10');
-            case 11:
-                return $translator->get('certificatePdfMonth11');
-            default:
-                return $translator->get('certificatePdfMonth12');
-        }
+
+        return match ($month) {
```
pabo99, line 111: Oh, I had never used this instruction before, thank you for introducing it!

PR #8938, opened 2026-02-01
feat: add admin ui, client gating for system settings

### LINKED ISSUE #8319: [FEATURE] Permitir que se deshabilite el EphemeralGrader desde la configuración del sitio

PrasannaMishra001:
## Description
Adds the admin settings UI for the ephemeral_grader_enabled toggle, wires client-side gating in the grader IDE and problem details and updates translations and the webpack entry for the new admin page. The disabled state now renders the generic ViewUnavailable component from #9919, so both the embedded grader and /grader/ephemeral/ show the same consistent view.

https://github.com/user-attachments/assets/547c53e9-2e74-46a0-88c0-040b59d13b40

Fixes: #8319

# Comments
This PR contains only frontend/UI changes split from the original EphemeralGrader feature. Backend APIs and DAO were moved to earlier PRs.

Depends on**
- Waiting for **PR #8905**
- Waiting for **PR #8906** 
- Waiting for **PR #8918** 

# Checklist:


frontend/www/js/omegaup/components/admin/Settings.vue
```diff
@@ -0,0 +1,96 @@
...
+  <div class="card">
+    <div class="text-white bg-primary card-header">
+      <div class="card-title h4">
+        {{ T.omegaupTitleAdminSettings }}
+      </div>
+    </div>
+    <div class="card-body">
+      <div v-if="loading" class="text-center">
+        <div class="spinner-border" role="status">
+          <span class="sr-only">Loading...</span>
```
pabo99: Could you add this text to the localization files?

frontend/www/js/omegaup/components/admin/Settings.vue
```diff
@@ -0,0 +1,96 @@
...
+        this.ephemeralGraderEnabled = Boolean(
+          response.settings.ephemeralGraderEnabled,
+        );
+      }
+    } catch (error: any) {
+      this.ui.apiError(error);
+    } finally {
+      this.loading = false;
+    }
+  }
```
pabo99: In another PR, I suggested exposing these settings through the payload. Could you revert this block?

This also lets us avoid using `mounted()`, since the data is already available when the component is rendered

frontend/www/js/omegaup/components/admin/Settings.vue
```diff
@@ -0,0 +1,96 @@
...
+        ephemeral_grader_enabled: newValue,
+      });
+      this.ui.success(T.ephemeralGraderUpdated);
+    } catch (error: any) {
+      this.ui.apiError(error);
+      // Rollback on failure
+      this.ephemeralGraderEnabled = oldValue;
+    } finally {
+      this.updating = false;
+    }
```
pabo99: Could you move this to `settings.ts`?

This would help keep the data layer separated from the presentation layer. The component would only need to dispatch the action, while `settings.ts` would be responsible for handling the update

frontend/www/js/omegaup/components/common/Navbar.vue
```diff
@@ -273,6 +273,17 @@
...
+                    <div class="dropdown-divider"></div>
+                    <a
+                      class="dropdown-item"
+                      href="/admin/settings.php"
+                      data-nav-admin-settings
+                    >
+                      <font-awesome-icon :icon="['fas', 'cog']" />
+                      {{ T.omegaupTitleAdminSettings }}
+                    </a>
+                  </template>
```
pabo99: For now, we haven't planned to expose the Admin menu. Could we leave that for a follow-up change where we include all the actions and sections available to admins? If so, could you revert this part for now?

On a related note, could you add `settings.php` to the `nginx.rewrites` mapping? That would give us a more user-friendly URL

frontend/www/js/omegaup/grader/ide.ts
```diff
@@ -14,6 +15,17 @@ OmegaUp.on('ready', () => {
...
+
+  if (!payload.ephemeralGraderEnabled) {
+    const alert = document.createElement('div');
+    alert.className = 'alert alert-danger';
+    alert.setAttribute('role', 'alert');
+    alert.textContent = T.ephemeralGraderDisabled;
+    main.innerHTML = '';
+    main.appendChild(alert);
+    return;
+  }
```
pabo99: To avoid introducing additional DOM elements for this case, could we add a simple condition in `EphemeralGrader.vue` instead?

Longer term, I'd like us to have a generic "View unavailable" component that can be controlled from the server and rendered consistently across the application

frontend/www/js/omegaup/components/problem/Detailsv2.vue
```diff
@@ -79,7 +79,7 @@
         </omegaup-overlay>
         <template v-if="problem.accepts_submissions">
           <omegaup-arena-ephemeral-grader
-            v-if="!problem.karel_problem"
+            v-if="!problem.karel_problem && problem.ephemeralGraderEnabled"
```
pabo99: There may be more scenarios in the future where this component should be hidden. Could you use a computed property instead? That would give us a single place to add any future conditions as they come up

frontend/www/js/omegaup/components/admin/Settings.vue
PrasannaMishra001: removed this entirely, as now the values come from the payload.

frontend/www/js/omegaup/components/admin/Settings.vue
PrasannaMishra001: done, it is a prop from commonPayload and so removed.

frontend/www/js/omegaup/components/common/Navbar.vue
PrasannaMishra001: reverted this and added admin/settings link for now. Will be happy to take up broader admin menu work when that's planned.

frontend/www/js/omegaup/components/problem/Detailsv2.vue
PrasannaMishra001: done, added `shouldShowEphemeralGrader`

frontend/www/js/omegaup/grader/ide.ts
PrasannaMishra001: done, replaced with simple `v-if` block in the component.

for view unavailable compoent, will raise and issue to work on platform wise such reusable compoent driver by serverflag (as in commonPayload).

frontend/www/js/omegaup/components/admin/Settings.vue
PrasannaMishra001: done so that component only emits the change now. `settings.ts` handles `updateSytemSettings` and the toast separately now.

PrasannaMishra001:
@pabo99 updated the disabled state which now renders as per #9919 instead , this will show both grader and grader/ide similar view. Also added test for it.

updated the description of this PR and added working video.

frontend/www/js/omegaup/grader/ide.ts
pabo99: I just updated the branch in #9919, so it should be merged soon

I have one small suggestion: now that you have a dedicated component for the unavailable view, could it be rendered directly from the `getGraderForTypeScript` entrypoint?

That way, we can keep the IDE component focused on the IDE itself and avoid including any unavailable-state logic there

pabo99 (commented):
The change looks good! I just have one question: why aren't the header and footer displayed when navigating to the unavailable view?

Image

I'd prefer for this card to be rendered within those wrappers so the page remains consistent with the rest of the application.

PrasannaMishra001:
@pabo99 actually the graderpage was rendered as full screen app, 'getGraderForTypeScript ' was setting hide flag for header/footer as true, now i changed that when enabled then hide otherwise render inside these wrappers. 
image
image

frontend/www/js/omegaup/grader/ide.ts
PrasannaMishra001: done.

frontend/www/js/omegaup/grader/unavailable.ts
```diff
@@ -0,0 +1,16 @@
+import Vue from 'vue';
```
pabo99, line 1: We might end up disabling more non-grader views in the future, which would require duplicating files like this. Maybe in a follow-up PR we can decouple this file from the grader and pass the description through the payload instead

pabo99 (approved):
Oh, I see why this PR ended up being so large, we're tackling two separate issues here:

* Enabling the system settings for administrators
* Displaying the "View Unavailable" component for the IDE

For future changes, let's try to keep each PR focused on a single issue. That will make the review process much easier

PrasannaMishra001:
thanks for merging this @pabo99. just noting for the record, this one was actually stacked on top of #8906 (the backend), which i had pushed first and meant to go in before this. since this got squash merged, #8906's backend (apiGetSystemSettings / apiUpdateSystemSettings and the cached reads) came along into main with it, so #8906 is effectively redundant now and i will close it. my mistake for not marking this one as draft while its dependency was still open. going forward i will draft stacked PRs and call out the dependency at the top of the description so the order stays clear and you wont get confused now onwards, i actually did mentioned dependency which might was overlooked

Issue #8941, opened 2026-02-01
[FEATURE] Display remaining self-assign count in /assign bot response


Yashagarwal9798:
_As a **contributor**, I want **the /assign bot response to show how many self-assigns I have remaining** so that **I can better manage my issue assignments and know my limits**._

## Context

When using the `/assign` command, a user can self-assign up to 5 issues. Currently, the bot response only confirms the assignment with messages like:
- "Assigned! Please open a pull request within 7 days..."
- Welcome message with contributor guidelines

However, there is no indication of how many self-assigns the user has used or how many remain. This can lead to confusion when a user tries to assign themselves to another issue and gets rejected due to reaching the limit.

### Proposed Enhancement

Add the remaining self-assign count to the bot's response message. For example:
- "Assigned! You have **4 self-assigns remaining** (out of 5). Please open a pull request within 7 days..."

Or as a separate line:
- " Self-assign usage: 1/5 used (4 remaining)"

### Benefits
- Helps contributors plan which issues they want to work on
- Reduces surprise when hitting the self-assign limit
- Improves transparency of the contribution process

<img width="1046" height="195" alt="Image" src="https://github.com/user-attachments/assets/0d642d03-6ded-478e-b188-2fe7a0ad80ff" />

Yashagarwal9798:
@pabo99 I would love to work on this issue. Could you please assign it to me?

PR #8949, opened 2026-02-02
feat(workflow): Display remaining self-assign count in /assign bot response

Yashagarwal9798:
# Description

This PR adds functionality to display the remaining self-assign count when a user uses the `/assign` command on an issue.

**Changes:**
- Add a new step to count user's historical self-assignments by searching issues where the user commented `/assign` and was successfully assigned
- Show remaining self-assign slots (out of 5) in the assignment message
- Use timeline API to verify self-assignments and detect voluntary unassigns

**Behavior:**
- Only `/unassign` (voluntary self-unassignment) frees a slot
- Auto-unassignment after 7 days does NOT free a slot
- Issue closure does NOT free a slot

**Example messages:**
- ` You have **3** self-assigns remaining out of 5.`
- ` You have reached your maximum of 5 assigned issues.`

Fixes: #8941

# Comments

The implementation uses GitHub's Search API and Timeline API to accurately track self-assignments across all issues, even those that were later closed or unassigned.

# Checklist

### DIFF
--- a/.github/workflows/assign-issues.yml
+++ b/.github/workflows/assign-issues.yml
@@ -19,6 +19,8 @@ jobs:
         permissions:
             issues: write
             pull-requests: write
+        env:
+            MAX_ASSIGNMENTS: 5
         steps:
             - name: Check for experienced issue creator exception
               id: pre_check
@@ -108,7 +110,7 @@ jobs:
                                   issue_number: issue.number,
                                   body: `Assigned to @${commenter}!\n\n` +
                                         `As an experienced contributor (${approvedPRCount} merged PRs) working on your own issue, ` +
-                                        `you're exempt from the standard 5-assignment limit. Please open a pull request within 7 days. ` +
+                                        `you're exempt from the standard ${process.env.MAX_ASSIGNMENTS}-assignment limit. Please open a pull request within 7 days. ` +
                                         `If no PR is opened, this issue will be automatically unassigned to keep the queue moving.`
                               });
                               
@@ -135,7 +137,7 @@ jobs:
                       }
                       
                       const commenter = context.payload.comment.user.login;
-                      const MAX_ASSIGNMENTS = 5;
+                      const MAX_ASSIGNMENTS = parseInt(process.env.MAX_ASSIGNMENTS || '5');
                       
                       // Count all open issues currently assigned to this user
                       const { data: assignedIssues } = await github.rest.issues.listForRepo({
@@ -174,9 +176,9 @@ jobs:
                       or this issue will be unassigned so others can pick it up.
                   assigned_label: ""
                   pin_label: ""
-                  max_assignments: 5
+                  max_assignments: ${{ env.MAX_ASSIGNMENTS }}
                   max_assignments_message: >
-                      You already have the maximum number of assigned issues (5).
+                      You already have the maximum number of assigned issues (${{ env.MAX_ASSIGNMENTS }}).
                       Please wrap up or unassign one before taking another.
                   block_assignment: true
                   block_assignment_comment: >


Yashagarwal9798:
@pabo99  Could you please take a look and let me know if any logical changes are needed?

pabo99 (commented):
Wouldn’t it be more relevant to simply count the number of issues currently assigned to the user, rather than tracking the issues they’ve self-assigned?

This is the relevant information, and it would simplify the script, independent of issue assignment changes. Thoughts?

Yashagarwal9798:
Hi @pabo99 sir, thanks for the feedback!

You raise a good point. I initially tracked self-assignments specifically because:

1. The `max_assignments` limit in the action is designed for **self-assignment** limits
2. Maintainer-assigned issues (via `/assign @username`) don't count against a user's self-assign quota
3. This way, maintainers can assign additional work to trusted contributors without affecting their self-assign limit

However, I understand your concern about simplicity. If the intended behavior is to show "total currently assigned issues" regardless of how they were assigned, I can simplify the script to just count all open issues assigned to the user.

Could you clarify the expected behavior:
- **Option A:** Show remaining self-assign slots (current implementation) - useful if maintainer assignments shouldn't count against the limit
- **Option B:** Show total currently assigned issues (simpler) - useful if users just need to know their current workload

can you please tell me which approach aligns better with omegaUp's workflow, and I'll update accordingly!

pabo99:
> Hi @pabo99 sir, thanks for the feedback!
> 
> You raise a good point. I initially tracked self-assignments specifically because:
> 
> 1. The `max_assignments` limit in the action is designed for **self-assignment** limits
> 2. Maintainer-assigned issues (via `/assign @username`) don't count against a user's self-assign quota
> 3. This way, maintainers can assign additional work to trusted contributors without affecting their self-assign limit
> 
> However, I understand your concern about simplicity. If the intended behavior is to show "total currently assigned issues" regardless of how they were assigned, I can simplify the script to just count all open issues assigned to the user.
> 
> Could you clarify the expected behavior:
> 
> * **Option A:** Show remaining self-assign slots (current implementation) - useful if maintainer assignments shouldn't count against the limit
> * **Option B:** Show total currently assigned issues (simpler) - useful if users just need to know their current workload
> 
> can you please tell me which approach aligns better with omegaUp's workflow, and I'll update accordingly!

Thanks for laying out both approaches so clearly! For omegaUp's workflow, Option B fits better, we'd like to show the total number of currently assigned issues, regardless of how they were assigned. That way contributors always have a clear view of their workload

Yashagarwal9798:
@pabo99 sir, Thanks for the clarification! I'll simplify to count total current assignments.

One edge case to consider: Since the action's `max_assignments` limit only blocks *self-assignment* but not *maintainer-assignment* (via UI or `/assign @user`), users could theoretically have more than 5 issues if maintainers assign them directly.

**Example:** If a maintainer assigns 6 issues to a trusted contributor, the "remaining count" would technically be negative.

**Proposed handling:** Since my message step only runs after a successful assignment (when `assigned == 'yes'`), the over-limit case is already handled by the action's built-in `max_assignments_message`. My step will simply show remaining count after successful assignments:

- **Under limit:** "You have 3 issues assigned. You can self-assign 2 more."
- **At limit (after this assignment):** "You have reached the maximum of 5 concurrent assignments."

For the edge case where users have >5 maintainer-assigned issues and try `/assign`, the action blocks them and shows its own message - so no negative count issue occurs.

Does this approach work for you?

pabo99:
Yes, that makes sense. Thanks for laying out the edge case.

I agree your approach avoids the negative count issue and keeps the messages consistent. Let’s go with this plan

Yashagarwal9798:
@pabo99 sir I have applied the requested changes. Could you please take another look?

.github/workflows/assign-issues.yml
```diff
@@ -121,7 +121,41 @@ jobs:
...
+                  github-token: ${{ secrets.GITHUB_TOKEN }}
+                  script: |
+                      const comment = context.payload.comment.body.trim();
+                      if (comment !== '/assign') {
+                          core.setOutput('current_count', '0');
+                          return;
+                      }
+                      
+                      const commenter = context.payload.comment.user.login;
+                      const MAX_ASSIGNMENTS = 5;
```
pabo99: This value is stored in two different places. If we were to change that value, ideally we should only have to modify it in one place. Could we add an environment variable to store this value?

Yashagarwal9798: @pabo99 sir I've updated the PR based on your feedback. Please let me know if there is anything else I should adjust.

pabo99 (approved):
Great!

Yashagarwal9798:
happy to help!!

PR #8954, opened 2026-02-02
[FEATURE] Display warning/ban reasons to problem authors

Ankitsinghsisodya:
# Description

Display warning/ban reasons to problem authors in the problem view.

When a problem receives a warning or is banned by a Quality Reviewer, the author can now see the reason(s) directly in the problem interface. This helps authors understand what needs to be fixed, especially since email notifications often end up in spam folders.

**Changes:**

- **Backend**: Added logic in `getProblemDetailsForTypeScript()` to fetch and include `warningReasons` for problem owners when the problem has warning/banned visibility
- **Frontend**: Added Bootstrap alert banner in `SettingsSummary.vue` that displays warning (yellow) or ban (red) reasons
- **Translations**: Added `problemWarningBecause` and `problemBannedBecause` keys in all language files
- **Tests**: Added `testDemotionRationalesDisplayedToOwner()` test in `QualityNominationTest.php`

**UI Preview:**

Warning state (visibility = -1):

```
┌─────────────────────────────────────────────────────┐
│  ⚠️ This problem has a warning because:            │
│  • The problem has inappropriate content           │
└─────────────────────────────────────────────────────┘
```

Banned state (visibility ≤ -2):

```
┌─────────────────────────────────────────────────────┐
│  🚫 This problem has been banned because:          │
│  • The problem violates community guidelines       │
└─────────────────────────────────────────────────────┘
```

Fixes: #4135

# Comments

The warning reasons are only visible to the problem owner (controlled by `showVisibilityIndicators` prop). The banner appears below the problem title and above the settings table.

# Checklist:


Ankitsinghsisodya:
***Problem Creator View:***
image
***Random User View:***
image

frontend/www/js/omegaup/components/problem/SettingsSummary.vue
```diff
@@ -50,6 +50,29 @@
...
+    <!-- Warning/Ban Reasons Banner -->
+    <div
+      v-if="showWarningReasons && warningReasons.length"
+      :class="[
+        'alert',
+        'mx-auto',
+        'w-75',
+        'mb-3',
+        isBanned ? 'alert-danger' : 'alert-warning',
+      ]"
```
pabo99: If only the `alert-danger`/`alert-warning` classes are reactive, can we add the others as regular classes? Something like:
```suggestion
      class="alert mx-auto w-75 mb-3"
      :class="{ isBanned ? 'alert-danger' : 'alert-warning' }"
```

frontend/server/src/Controllers/Problem.php
```diff
@@ -2824,6 +2824,29 @@ public static function getProblemDetails(
...
+
+            // If problem has warning/banned status and user is the owner,
+            // show the warning reasons so they can fix the issue
+            $isWarningOrBanned = in_array(
+                $problem->visibility,
+                [
+                    \OmegaUp\ProblemParams::VISIBILITY_PRIVATE_WARNING,
+                    \OmegaUp\ProblemParams::VISIBILITY_PUBLIC_WARNING,
+                    \OmegaUp\ProblemParams::VISIBILITY_PRIVATE_BANNED,
+                    \OmegaUp\ProblemParams::VISIBILITY_PUBLIC_BANNED,
```
pabo99, line 2836: Now that you are more familiar with the different types of problem visibilities, could you help me assess the magnitude of this change #4601?

In the PR https://github.com/omegaup/omegaup/pull/4602 there is more relevant information, such as replacing the numeric field for visibility with an enum type instead.  

What are your thoughts on this?

pabo99 (commented):
Just minor suggestions, and that's ok!

frontend/server/src/Controllers/Problem.php
Ankitsinghsisodya: The idea makes sense - enforcing valid values at the DB level is definitely better than just relying on application validation. Right now someone could potentially insert weird visibility values directly in the DB.

That said, I think the complexity here is being underestimated. Changing from int to ENUM would touch a lot of the codebase. A few things to watch out for:

- There are numeric comparisons like `visibility > 0` and `visibility >= 2` scattered throughout the code for access control. All of these would need to be rewritten.
- The TypeScript types in [api_types.ts](cci:7://file:///Users/ankitsinghsisodya/Desktop/omegaup/frontend/www/js/omegaup/api_types.ts:0:0-0:0) would need updating too
- Migration needs to handle converting existing data correctly

Also noticed PR #4602 has been marked stale, so it probably needs a fresh look to work with the current codebase.

One alternative worth considering: if the main goal is just to prevent invalid values, we could add a CHECK constraint or trigger on the existing int column. Less disruptive than a full ENUM migration, though ENUM is arguably cleaner long-term.

What do you all think?

PR #8957, opened 2026-02-02
Fix: Initialize Help page earlier to prevent blank /help screen

### LINKED ISSUE #8944: [BUG] Missing /help/ page - returns generic 404 error
## Description

When navigating to `https://omegaup.com/help/`, the page returns a generic HTTP 404 error instead of providing helpful support resources or guidance.

## Current Behavior

The `/help/` route displays:
- A generic Omega (Ω) symbol icon
- "HTTP/1.1 404 - Not Found" message
- A single link to "omegaup.com" with no other guidance

## Expected Behavior

The `/help/` route should either:
1. Display a proper Help/Support page with FAQ, documentation, or support contact information
2. Redirect to a valid help resource (e.g., blog.omegaup.com or support documentation)
3. Show a user-friendly error page with suggestions to navigate to other sections

## Why This Is an Issue

- Users expecting a help section via the navigation menu find a confusing 404 page
- The error page provides no guidance on where to find help resources
- It diminishes user experience for first-time visitors seeking assistance
- The generic 404 page doesn't match the professional OmegaUp branding and UI

## Steps to Reproduce

1. Visit https://omegaup.com/
2. Click on the "Help" dropdown menu in the navigation bar
3. Or navigate directly to https://omegaup.com/help/
4. Observe the generic 404 error page

Image

viveksahu92:
This pull request adds a dedicated Help & Support page at /help/ and makes sure it always loads correctly.
Loaded the Help page bundle via a new common_help entry in webpack.config-frontend.js.

Initialized the Help page on DOMContentLoaded without waiting for OmegaUp.ready(), so the page does not stay blank if other modules fail.
​Added translations for the Help page in English, Spanish, Portuguese, and pseudo languages .
​Updated nginx rewrites to map /help/ to help.php.

Fixes: #8944
​
Screenshot 2026-02-02 214148


viveksahu92:
Hi, @pabo99 please could you review this PR?

pabo99 (commented):
Could you resolve the conflicts? Otherwise, it will be very difficult to complete the review.

frontend/templates/en.lang
```diff
@@ -546,10 +546,10 @@ courseCardCourseResume = "Resume"
 courseCardImpartedBy = "Imparted by: **%(school_name)**"
 courseCardMustLogIn = "You must login to see these courses."
 courseCardProgress = "Progress:"
-courseCardPublicLevelAdvanced = "★★★ Advanced"
-courseCardPublicLevelIntermediate = "★★ Intermediate"
-courseCardPublicLevelIntroductory = "★ Introductory"
-courseCardsListSearch = "🔍 Course name"
+courseCardPublicLevelAdvanced = "â˜…â˜…â˜… Advanced"
```
pabo99: It looks like the special characters were saved with the wrong encoding. Could you make sure your editor is set to save files in UTF‑8 without BOM before committing? That way the special characters will display correctly and we’ll avoid mojibake in the diffs.

frontend/server/src/Controllers/User.php
```diff
@@ -5201,6 +5259,18 @@ function () use ($loggedIdentity, $identity): array {
...
+    /**
+     * Returns the help resources for TypeScript
+     *
+     * @return array{templateProperties: array{payload: UserHelpPayload, title: \OmegaUp\TranslationString}, entrypoint: string}
+     *
+     * @psalm-type UserHelpPayload = array{helpResources: list<array{name: string, url: string, external: bool, icon: string}>}
+     */
+    public static function getHelpForTypeScript(\OmegaUp\Request $r): array
+    {
+        return self::apiHelp($r);
```
pabo99: Normally we don’t call endpoints directly from PHP; instead, we create bridge functions to get what we need. The main reason is that it's hard to trace what is being sent in the requests.

From what I can see, the function doesn't require any data from the request. Could you change it to a regular function instead?

frontend/server/src/Controllers/User.php
```diff
@@ -5140,10 +5142,66 @@ public static function getDocsForTypeScript(\OmegaUp\Request $r) {
...
+            ],
+        ];
+
+        return [
+            'templateProperties' => [
+                'payload' => ['helpResources' => $helpResources],
+                'title' => new \OmegaUp\TranslationString('omegaupTitleHelp'),
+            ],
+            'entrypoint' => 'common_help',
+        ];
```
pabo99: Could you move this to `getHelpForTypescript`? That way, if at some point we want to add more things, we can do it from there without having to move it into this function

frontend/www/js/omegaup/shim-tsx.d.ts
```diff
@@ -0,0 +1,14 @@
...
+
+declare global {
+    namespace JSX {
+        interface Element extends import('vue').VNode { }
+        interface ElementClass extends import('vue').default { }
+        interface IntrinsicElements {
+            [elem: string]: any;
+        }
+    }
+}
```
pabo99: Could you revert this change?

Here’s an explanation of why:
https://github.com/omegaup/omegaup/blob/6f1e07bca656151bea72fdf57a8d0026fd54a1dd/frontend/www/docs/Coding-guidelines.md?plain=1#L119-L127

diag_linter.py
```diff
@@ -0,0 +1,29 @@
+
+import re
```
pabo99: What are all these changes useful for?

If you found errors in the linters that need to be fixed, it's better to open an issue and address them separately. Doing it here only slows down the review process

pabo99 (commented):
I can’t really gauge the size of this PR because of the lines modified in the localization files.
It would be best to split this change into two:

- The first focusing on all the server‑side changes
- The second with all the UI‑related changes

That way we can move faster with the reviews.

viveksahu92:
​"Hi @pabo99 , I have fixed all the checks. Checks are now passing. Could you please review it again? If the PR is perfect, please merge it."

viveksahu92:
Screenshot 2026-02-09 003326

viveksahu92:
Hi @pabo99, just checking in on this PR whenever you have a moment. I've addressed the previous checks, and it's ready for your final sign-off. Thanks!"

frontend/server/src/Controllers/Admin.php
```diff
@@ -6,6 +6,7 @@ class Admin extends \OmegaUp\Controllers\Controller {
     /**
      * Get stats for an overall platform report.
      *
+     * @param \OmegaUp\Request $r
```
pabo99: Here it's not necessary to redeclare the type, since the parameter implicitly indicates it will be of that type.
Also, this change isn't related to this controller at all. Could you revert this line?
```suggestion
```

frontend/server/src/Controllers/Admin.php
```diff
@@ -39,7 +40,7 @@ public static function apiPlatformReportStats(\OmegaUp\Request $r): array {
                      * @param array{gender: string, users: int} $row
                      * @return array<string, int>
                      */
-                    fn (array $row) => [$row['gender'] => $row['users']],
+                    fn (array $row): array => [$row['gender'] => $row['users']],
```
pabo99: Same case as the previous comment: on line 41 the more detailed type is already defined, so it’s not necessary to repeat it here:
```suggestion
                    fn (array $row) => [$row['gender'] => $row['users']],
```

frontend/server/src/Controllers/User.php
```diff
@@ -522,7 +524,7 @@ public static function apiLogin(\OmegaUp\Request $r): array {
      *
      * @omegaup-request-param string $old_password
      * @omegaup-request-param null|string $password
-     * @omegaup-request-param mixed $permission_key
+     * @omegaup-request-param string $permission_key
```
pabo99: This change should be accompanied by correcting the parameter in the code, which means avoiding the use of:
```php
$r['param`]
```
and instead using:
```php
$r->ensureString('param')
// or
$r->ensureOptionalString('param')
// for strings
```
This adjustment enforces declaring the correct type.

But since this change isn't related to what I just mentioned, you could revert this line and perhaps open a new issue to fix that case specifically

frontend/www/js/omegaup/common/help.ts
```diff
@@ -0,0 +1,20 @@
+import common_Help from '../components/common/Help.vue';
+import Vue from 'vue';
+import { types } from '../api_types';
+
+document.addEventListener('DOMContentLoaded', () => {
```
pabo99: nit:
```suggestion
OmegaUp.on('ready', () => {
```
This is the convention we usually follow to indicate that the document is ready, meaning it has finished loading and can now be interacted with

frontend/www/js/omegaup/components/common/Help.vue
```diff
@@ -0,0 +1,104 @@
...
+
+    <h4>{{ T.helpAdditionalResources }}</h4>
+    <ul class="list-unstyled">
+      <li class="mb-2">
+        <a href="/privacypolicy/">{{ T.wordsPrivacyPolicy }}</a>
+      </li>
+      <li class="mb-2">
+        <a href="/docs/">{{ T.helpDocumentation }}</a>
+      </li>
+    </ul>
```
pabo99: That's not the correct path for the privacy policy, and it's already included in the footer.

Also, the link to the documentation is already included in the cards above, so we can remove this for now.

Perhaps in the future, if we want to display additional resources, we could include them in the payload using the same format as the other resources, along with a flag to indicate that only minimal information is shown—this way the information doesn't end up scattered.

frontend/www/js/omegaup/components/common/Help.vue
```diff
@@ -0,0 +1,104 @@
...
+      </div>
+    </div>
+
+    <div class="row">
+      <div
+        v-for="resource in helpResources"
+        :key="resource.name"
+        class="col-md-4 mb-4"
+      >
+        <div class="card h-100">
```
pabo99: The content looks good. Could we also add an icon in the payload for each of the cards you’re adding?

pabo99:
BTW,  I like this design better:

Screenshot 2026-02-02 214148

Could you display it that way? I mean having the main card with a title, and inside it the rest of the cards.

viveksahu92:
> BTW, I like this design better:
> 
> Screenshot 2026-02-02 214148
> Could you display it that way? I mean having the main card with a title, and inside it the rest of the cards.


"Hi @pabo99, I have updated the layout to include a main container card as suggested. Please let me know if this looks correct or if any further changes are needed. If everything looks good, please consider merging this pull request."

Screenshot 2026-02-13 162930

pabo99:
It seems that some of the suggestions I left earlier haven't been addressed yet. Could you please review them again and apply the required changes?

Also, in the last screenshot you shared, I'd like the title to use the card title styles — just like in the first screenshot you included in this PR

pabo99:
Could you please run:
```bash
./stuff/lint.sh fix -- frontend/www/js/omegaup/api_types.ts
```
?

frontend/server/src/Controllers/User.php
```diff
@@ -5349,6 +5351,58 @@ public static function getCompareDetailsForTypeScript(
         ];
     }
 
+    /**
+     * @return array{entrypoint: string, templateProperties: array{payload: UserHelpPayload, title: \OmegaUp\TranslationString}}
+     */
+    public static function getHelpForTypeScript() {
```
pabo99: Since all this content is static and we don’t have plans to incorporate it into the DB in the near future, can you remove all of this entrypoint and instead do something like:

https://github.com/omegaup/omegaup/blob/801bfa27acb743d0f93db4587cc2da1aec4781bf/frontend/www/group/new.php#L1-L13

Where the payload is sent empty from that file.

And for the content, can you create a file similar to:

https://github.com/omegaup/omegaup/blob/801bfa27acb743d0f93db4587cc2da1aec4781bf/frontend/www/js/omegaup/testimonials.config.ts#L1-L72

So you can use it in the component?

pabo99 (commented):
Could you please run:
```bash
./stuff/lint.sh fix -- frontend/www/js/omegaup/api_types.ts
```
?
Also, in previous versions, an excerpt was shown for each item (at least in the screenshot you shared), and in this latest revision it’s no longer there. Could you add it back?

frontend/www/js/omegaup/components/common/Help.vue
```diff
@@ -0,0 +1,60 @@
...
+        <div
+          v-for="resource in helpResources"
+          :key="resource.name"
+          class="col-md-6 col-lg-4 mb-4"
+        >
+          <div class="card h-100 shadow-sm">
+            <div class="card-body">
+              <h5 class="card-title">
+                <span v-if="resource.icon" class="mr-2">{{
+                  resource.icon
```
pabo99: Is it possible to use Font Awesome icons instead of the ones you're currently using?

frontend/www/js/omegaup/components/common/Help.vue
```diff
@@ -0,0 +1,60 @@
...
+                {{ resource.description[currentLocale] }}
+              </p>
+              <a
+                :href="resource.url"
+                :target="resource.external ? '_blank' : '_self'"
+                :rel="resource.external ? 'noopener noreferrer' : ''"
+                class="btn btn-primary btn-sm"
+              >
+                {{ T.helpVisit }}
+                <span v-if="resource.external"> ↗</span>
```
pabo99: Here you can use a Font Awesome icon as well; it looks like this one works fine: `external-link-alt`

frontend/www/js/omegaup/components/common/Help.vue
```diff
@@ -0,0 +1,60 @@
...
+          :key="resource.name"
+          class="col-md-6 col-lg-4 mb-4"
+        >
+          <div class="card h-100 shadow-sm">
+            <div class="card-body">
+              <h5 class="card-title">
+                <span v-if="resource.icon" class="mr-2">{{
+                  resource.icon
+                }}</span>
+                {{ resource.title[currentLocale] }}
```
pabo99: Can you use `T.locale` directly? That way you don’t need the `currentLocale` computed property

frontend/templates/pt.lang
```diff
@@ -856,6 +856,21 @@ groupsIdentitiesSuccessfullyCreated = "As identidades foram criadas e ligadas ao
 groupsIdentityWarning = "Salve as senhas em um local seguro, porque depois de criar as identidades, não haverá maneira de recuperar as senhas."
 groupsInvalidCsv = "Faça o upload de um arquivo CSV válido."
 groupsUploadCsvFile = "Subir CSV arquivo"
+helpAdditionalResources = "Recursos Adicionais"
```
pabo99: Some of these strings are already obtained from `help.config.ts`. Could you review and remove the ones that are no longer needed?

frontend/server/src/Controllers/User.php
```diff
@@ -5349,6 +5349,7 @@ public static function getCompareDetailsForTypeScript(
         ];
     }
 
+
```
pabo99: nit:
```suggestion
```

pabo99 (commented):
Much better! Just a few suggestions.

viveksahu92:
Hi @pabo99 , now cl pipeline was fixed can you review again?

frontend/server/src/Controllers/User.php
pabo99: This suggestion is still pending

frontend/www/js/omegaup/components/common/Help.vue
```diff
@@ -0,0 +1,59 @@
...
+          :key="resource.name"
+          class="col-md-6 col-lg-4 mb-4"
+        >
+          <div class="card h-100 shadow-sm">
+            <div class="card-body">
+              <h5 class="card-title">
+                <span v-if="resource.icon" class="mr-2">{{
+                  resource.icon
+                }}</span>
+                {{ T[`helpResource${capitalizeFirst(resource.name)}`] }}
```
pabo99: These text strings do not exist in the localization files. Additionally, the ideal approach would be to retrieve them directly from the `help.config.ts` file, just as it is done in `carousel.config.ts`

frontend/www/js/omegaup/components/common/Help.vue
```diff
@@ -0,0 +1,59 @@
...
+        <div
+          v-for="resource in helpResources"
+          :key="resource.name"
+          class="col-md-6 col-lg-4 mb-4"
+        >
+          <div class="card h-100 shadow-sm">
+            <div class="card-body">
+              <h5 class="card-title">
+                <span v-if="resource.icon" class="mr-2">{{
+                  resource.icon
```
pabo99: This only renders the name of the item, not the icon. Also, I had suggested using Font Awesome icons, but I don't see them included.

frontend/www/js/omegaup/components/common/Help.vue
```diff
@@ -0,0 +1,59 @@
...
+                }}
+              </p>
+              <a
+                :href="resource.url"
+                :target="resource.external ? '_blank' : '_self'"
+                :rel="resource.external ? 'noopener noreferrer' : ''"
+                class="btn btn-primary btn-sm"
+              >
+                {{ T.helpVisit }}
+                <span v-if="resource.external"> ↗</span>
```
pabo99: This also needs to be included using Font Awesome

frontend/www/media/omegaup.png
pabo99, line 1: Coul you revert this change? It looks like this image is still being used in the `README`

frontend/www/js/omegaup/components/common/Help.vue
```diff
@@ -0,0 +1,64 @@
...
+        <div
+          v-for="resource in helpConfig"
+          :key="resource.title[T.locale]"
+          class="col-md-6 col-lg-4 mb-4"
+        >
+          <div class="card h-100 shadow-sm">
+            <div class="card-body">
+              <h5 class="card-title">
+                <span v-if="resource.icon" class="mr-2">
+                  <i :class="getIconClass(resource.icon)" aria-hidden="true"></i>
```
pabo99: The icons are still not visible. Can you try adding them using the `FontAwesomeIcon` component, as it is done in other places?

frontend/www/js/omegaup/components/common/Help.vue
```diff
@@ -0,0 +1,64 @@
...
+                {{ resource.description[T.locale] }}
+              </p>
+              <a
+                :href="resource.url"
+                :target="resource.external ? '_blank' : '_self'"
+                :rel="resource.external ? 'noopener noreferrer' : ''"
+                class="btn btn-primary btn-sm"
+              >
+                {{ T.helpVisit }}
+                <i v-if="resource.external" class="fas fa-external-link-alt ml-2" aria-hidden="true"></i>
```
pabo99: This one should also be added using the `FontAwesomeIcon` component

pabo99 (commented):
Almost done!

viveksahu92:
HI @pabo99 , could you please see my new changes .
viedo:

https://github.com/user-attachments/assets/b2198927-f3a0-4a1c-9a95-a111e400848e

frontend/server/src/Controllers/User.php
```diff
@@ -57,6 +57,8 @@
  * @psalm-type VerificationParentalTokenDetailsPayload=array{hasParentalVerificationToken: bool, message: string}
  * @psalm-type UserDocument=array{name: string, url: string}
  * @psalm-type UserDocsPayload=array{docs: array<string, list<UserDocument>>}
+ * @psalm-type HelpResource=array{name: string, url: string, icon: string, external: bool}
+ * @psalm-type UserHelpPayload=array{helpResources: list<HelpResource>}
```
pabo99: These types seem to no longer be used. Could you remove them?

frontend/www/js/omegaup/help.config.ts
```diff
@@ -0,0 +1,128 @@
...
+    title: {
+      en: 'GitHub Repository',
+      es: 'Repositorio en GitHub',
+      pt: 'Repositório no GitHub',
+    },
+    description: {
+      en:
+        'Contribute to omegaUp by exploring the source code on GitHub. Report issues, submit pull requests, and help improve the platform.',
+      es:
+        'Contribuya a omegaUp explorando el código fuente en GitHub. Reporte problemas, envíe solicitudes de extracción y ayude a mejorar la plataforma.',
```
pabo99: Nit:
```suggestion
        'Contribuye a omegaUp explorando el código fuente en GitHub. Reporta problemas, envía Pull Requests y ayuda a mejorar la plataforma.',
```

For consistency with the other descriptions' translations

frontend/www/js/omegaup/components/common/Help.vue
```diff
@@ -0,0 +1,102 @@
...
+  components: {
+    FontAwesomeIcon,
+  },
+})
+export default class CommonHelp extends Vue {
+  T = T;
+  helpConfig = helpConfig;
+  mounted() {
+    document.title = `${this.T.omegaupTitleHelp} \u2013 omegaUp`;
+  }
```
pabo99: This is no longer necessary since the title is obtained from the `render` function in `PHP`. So it can be removed:



```suggestion
```


If you found any rendering issues, it would be better to report them in a new issue so they can be fixed globally

stuff/teaching_assistant/teaching_assistant.py
```diff
@@ -137,7 +137,7 @@ def get_runs_from_course_endpoint(
     returns the list of run_ids and corresponding_users from last 30 days.
     """
     endpoint = (
-        f"/api/course/runs?"
+        f"api/course/runs?"
```
pabo99, line 140: This change seems unrelated to this PR. Could you revert it?

frontend/www/help.php
```diff
@@ -0,0 +1,13 @@
...
+namespace OmegaUp;
+require_once(dirname(__DIR__, 1) . '/server/bootstrap.php');
+
+\OmegaUp\UITools::render(
+    fn (\OmegaUp\Request $r) => [
+        'templateProperties' => [
+            'payload' => [],
+            'title' => new \OmegaUp\TranslationString('omegaupTitleHelp'),
+        ],
+        'entrypoint' => 'common_help',
```
pabo99, line 11: This entry is still missing in the `webpack.config-frontend.js` file. The PHP file also still needs to be added to `nginx.rewrites`

pabo99 (commented):
This looks much better overall — just a few minor suggestions left.

One thing that would help a lot is avoiding force pushes when possible, since they make it harder to track whether previous suggestions/comments were already addressed

Issue #8896, opened 2026-01-27
[FEATURE]Improve custom test case execution for problems


PrasannaMishra001:
As a problem solver, I want custom test case execution to be clearly visible and easy to use so that I can debug my solution effectively before submitting

## Context

omegaUp already provides an ephemeral grader that allows running code on custom input without submitting
However this functionality is hard to discover and not intuitive for many users
The custom test option is placed at the bottom of the problem view and requires understanding multiple case tabs which makes quick debugging difficult .

## Proposed Solution
- Make custom input execution more visible in the problem view
- Improve labels and layout to make debugging faster and simpler.

PrasannaMishra001:
@pabo99 i would like to work on this,pls /assign

pabo99:
Thanks for the suggestion! This functionality is already available in the problem view through the ephemeral grader. You can enter custom input and run your code without submitting, and the system will show stdout, stderr.

<img width="1027" height="678" alt="Image" src="https://github.com/user-attachments/assets/d35385e5-1900-4b34-920d-eeb8760def9d" />

If you notice any limitations or have ideas to improve the current integration, please share them here so we can track them

PrasannaMishra001:
@pabo99 you’re right😅,I just checked and saw the custom cases add option at the bottom right.
so the discoverability is still an issue. the custom test case option is placed at the bottom and is easy to miss, especially for new users
I think this can be improved mainly from a UX perspective,by making custom input and run more visible and easier to use, ig no backend changes will be needed.

PR #8959, opened 2026-02-02
[FEATURE] Improve discoverability of custom test case execution

PrasannaMishra001:
## Description

This PR improves the discoverability of custom test case execution in the problem view.

### What changed
image

- Moved the **“Add Case” form** from the bottom of the case list to the **top**.
- Added a **visible “CASES” header** to clearly label the section.
- Replaced by **“+ Case”** to make the action explicit.
- Added **input placeholders** to guide first-time users.
- Applied **minimal, scoped styling** using existing omegaUp design variables.
- Removed obsolete styling related to the old bottom-placed form.

Fixes: #8896

---

## Comments

- Changes are intentionally limited to `CaseSelector.vue` to keep the PR focused.
- Verified locally using the ephemeral grader with custom input execution.
- this is a pure UX change.

---

## Checklist

- [x] The change is small and self-contained; no PR split required.


frontend/www/js/omegaup/grader/CaseSelector.vue
```diff
@@ -260,8 +283,8 @@ div.filenames {
   flex: 1;
 }
 
-input[type='number'].case-weight {
-  width: 3em;
+.case-weight {
+  width: 3.5em;
```
pabo99: Wouldn't it be better here to have the width adjust automatically?

frontend/www/js/omegaup/grader/CaseSelector.vue
```diff
@@ -228,15 +233,33 @@ export default class CaseSelector extends Vue {
...
+
+.case-form {
+  padding: 0.5em 0.75em 0.25em;
+  border-bottom: 1px solid $omegaup-grey--lighter;
+}
+
+.btn-add-case {
+  padding: 0.375rem 0.75rem;
+  white-space: nowrap;
+}
```
pabo99: t seems that in Bootstrap these are the default padding values for buttons, so it wouldn't be necessary to override them.

Also, `white-space: nowrap;` can be replaced with the `text-nowrap `class.

This way, this block is no longer needed and it can be removed
```suggestion
```

frontend/www/js/omegaup/grader/CaseSelector.vue
```diff
@@ -1,8 +1,43 @@
 <template>
   <div class="root d-flex flex-column h-100" :class="theme">
+    <div class="case-header">
+      <h6 class="case-title mb-0">{{ T.wordsCases }}</h6>
+    </div>
     <div class="summary">
       {{ summary }}
     </div>
+    <form class="case-form" @submit.prevent="createCase()">
+      <div class="input-group">
```
pabo99, line 10: Now that you’re improving the positioning of these elements, it would also be worth making some adjustments.

Can you make the `input-group` wrap only the two inputs? The button can go on a separate line, and it can keep its 100% width so the full text is always visible.

frontend/www/js/omegaup/grader/CaseSelector.vue
```diff
@@ -1,8 +1,43 @@
...
+        <input
+          v-model="newCaseWeight"
+          class="form-control case-weight"
+          type="text"
```
pabo99: Can this be changed to type number?
```suggestion
          type="number"
```

frontend/www/js/omegaup/grader/CaseSelector.vue
```diff
@@ -1,8 +1,43 @@
...
+          :placeholder="'1'"
```
pabo99: Can this be replaced with a text that better describes the input? For example, "Case weight".

Remember to add it to the translation strings.

frontend/www/js/omegaup/grader/CaseSelector.vue
```diff
@@ -1,8 +1,43 @@
...
+        />
+        <input
+          v-model="newCaseName"
+          class="form-control"
+          type="text"
+          data-case-name
+          :placeholder="T.wordsCase + ' name'"
```
pabo99: Can you add that composite string to the translation strings?

frontend/www/js/omegaup/grader/CaseSelector.vue
```diff
@@ -1,8 +1,43 @@
...
+            :class="{
+              'btn-primary': theme == 'vs',
+              'btn-secondary': theme == 'vs-dark',
+            }"
+            type="submit"
+            :disabled="!newCaseName.length"
+            data-add-button
+            @click="createCase()"
+          >
+            + {{ T.wordsCase }}
```
pabo99: Now that the button will have more space, can you add the translation string "Add case" and use it here?

frontend/www/js/omegaup/grader/CaseSelector.vue
PrasannaMishra001: Changed .case-weight from fixed `width: 3.5em` to flex: 1 1 auto for automatic sizing

frontend/www/js/omegaup/grader/CaseSelector.vue
PrasannaMishra001: replaced this block with bootstrap `text-nowrap`

frontend/www/js/omegaup/grader/CaseSelector.vue
PrasannaMishra001: moved btn to separate line with w-100 mt-2 classes

frontend/www/js/omegaup/grader/CaseSelector.vue
PrasannaMishra001: changed this placeholder and added translation also

frontend/www/js/omegaup/grader/CaseSelector.vue
PrasannaMishra001: ok replacd and added translation

PrasannaMishra001:
@pabo99 sir pls review the chances.

image

frontend/templates/es.lang
```diff
@@ -2013,6 +2013,7 @@ wizardTitle = "Buscador de Problemas"
 wordAccepted = "Aceptado"
 wordsActions = "Acciones"
 wordsAddAdmin = "Agregar admin"
+wordsAddCase = "Agregar caso"
```
pabo99: Can you use a different suffix instead of `words`? It lacks context. The same applies to the other added strings

frontend/www/js/omegaup/grader/CaseSelector.vue
```diff
@@ -108,7 +110,7 @@ import T from '../lang';
 
 @Component
 export default class CaseSelector extends Vue {
-  newCaseWeight: number = 1;
+  newCaseWeight: any = '';
```
pabo99: Can this be of type null | number? That way we can avoid using magic values
```suggestion
  newCaseWeight: null | number = null;
```

frontend/www/js/omegaup/grader/CaseSelector.vue
```diff
@@ -212,10 +214,10 @@ export default class CaseSelector extends Vue {
 
     store.dispatch('createCase', {
       name: this.newCaseName,
-      weight: parseFloat(this.newCaseWeight.toString()),
+      weight: parseFloat(this.newCaseWeight.toString()) || 1,
```
pabo99: There shouldn’t be a need to do parsing here anymore since `newCaseWeight` has the right type.

```suggestion
      weight: this.newCaseWeight ?? 1,
```

frontend/www/js/omegaup/grader/CaseSelector.vue
```diff
@@ -212,10 +214,10 @@ export default class CaseSelector extends Vue {
...
     });
 
-    this.newCaseWeight = 1;
+    this.newCaseWeight = '';
```
pabo99: nit:
```suggestion
    this.newCaseWeight = null;
```

frontend/templates/es.lang
PrasannaMishra001: changing to `caseName` and `caseWeight`

frontend/www/js/omegaup/grader/CaseSelector.vue
PrasannaMishra001: replaced this.

frontend/templates/en.lang
```diff
@@ -22,6 +22,7 @@ activityReportReport = "Report"
 activityReportSummaryContest = "This is a report of the activity that happened within this contest. In order to protect contestants' privacy, their IP addresses have been obfuscated using a perfect hash. That means that there is a one-to-one correspondence between IP addresses and Origin IDs."
 activityReportSummaryCourse = "This is a report of the activity that happened within this course. In order to protect participants' privacy, their IP addresses have been obfuscated using a perfect hash. That means that there is a one-to-one correspondence between IP addresses and Origin IDs."
 activityReportUsers = "Account name"
+addCase = "Add case"
```
pabo99: Now they have less context ='(

Could they be renamed to:

- **_caseSelectorAddCase_**
- **_caseSelectorCaseName_**
- **_caseSelectorCaseWeight_**
?

PR #8980, opened 2026-02-04
Revamp self assign workflow

### LINKED ISSUE #8807: [FEATURE] Creating a custom workflow for assigning PRs
As many of you have noticed, the assign workflow is causing some issues, especially when someone unassigns the issue and then someone tries to assign it. Also, with the growing requirements, I believe it will be best to create a custom workflow catering to the requirements of our organisation. 
I will try to list down all the objectives we want to achieve through the worklow in the comments. Once we have listed down all the objectives, we can start working on it.

antedotee:
## Updated Self-Assign Workflow Overview

This repository now uses a **single GitHub Actions workflow** (`assign-issues.yml`) to handle self-assignment via issue comments. All previous JS-based assign-issues scripts and the stateful auto-unassign logic have been removed.

Key properties:

- **Trigger**: On `issue_comment` (type: `created`) for issues (not PRs).
- **Scope**: Enforces limits **within this repository only** 

---

### `/assign`

- Must be **exactly** `/assign`:
  - No usernames (e.g., `/assign @user` is rejected).
  - No extra text (e.g., `/assign me` is rejected).
- Always treated as **self-assign**:
  - The commenter is the only user added as an assignee, if allowed.

### `/unassign`

- Must be **exactly** `/unassign`.
- Only unassigns the **commenting user**.
- If the commenter is not currently assigned, the command is ignored (no-op, just logs).

Any other comment content is ignored by this workflow.

---

### 1. Bot comments

- **All bot comments are ignored**:
  - If `comment.user.type === "Bot"`, the workflow logs and returns.
  - There is no special handling any bot.

### 2. Issue creator association (where the issue came from)

To use `/assign` on an issue:

- The issue must be created by a user whose `author_association` is one of ALLOWED_ISSUE_CREATORS: "heduenas pabo99 carlosabcs tvanessa":

If the issue creator’s association is **not** in this set (e.g., `NONE`, `CONTRIBUTOR`, `FIRST_TIME_CONTRIBUTOR`, etc.), then:

- `/assign` is **rejected**, with a comment explaining that the feature only applies to issues created by owner/member/collaborator and naming the actual association of the creator.
- For your own issues, ask maintainer to assign the issue manually.

### 3. Commenting user association (new contributors)

When a user runs `/assign`, the workflow checks their `author_association`:

- If the association is **`FIRST_TIME_CONTRIBUTOR`**, **`FIRST_TIMER`**, or **`NONE`**:
  - The workflow checks whether they have **at least one merged PR in this repo**.
  - If they **do not** have any merged PRs here:
    - `/assign` is rejected with a message:
      - First-time contributors cannot use auto-assignment yet.
      - They are asked to request assignment from maintainers.
  - If they **do** have a merged PR:
    - They are allowed to proceed like any other contributor.

- If the association is `CONTRIBUTOR`, `MEMBER`, `COLLABORATOR`, `OWNER`, etc.:
  - No extra merged-PR check is required; they can proceed (subject to other rules).

Fixes #8807


antedotee:
CC @pabo99(I haven't yet tested it in my fork), since it will require quite a bit of time along with multiple users commenting. I  am thinking of rolling this out in prod, and if we encounter any errors, fix it on the fly and modifying it according to user requirements on the go.

Ankitsinghsisodya:
@pabo99 sir and @antedotee , the maintenance required for the current self-assign workflow has become overly complicated and cumbersome. We are encountering bugs on a daily basis, which is shifting focus away from critical feature work. 

I propose we switch to a much simpler flow: allow issue authors to assign and unassign themselves, while maintainers retain full authority to assign or unassign anyone. This would significantly reduce the maintenance burden while ensuring the maintainer's decision remains final.

pabo99:
Thank you both for your input and effort on this workflow.

@antedotee, I appreciate the modular approach and the time invested in implementing and documenting all these features. Given the complexity and the amount of work put into this, I think it's fair to give your changes a chance by enabling the current modular workflow in the repository, as you suggested.

At the same time, @Ankitsinghsisodya raises a valid point about long-term maintainability and the need to keep our workflows simple and reliable.

How about we proceed with enabling the current modular workflow, monitor its stability and user experience for a reasonable period, and if we encounter persistent issues or maintenance overhead, we can revisit Ankit’s proposal for a simpler flow? This way, we recognize the effort and innovation, but also remain open to simplifying if needed.

Let's keep communication open and prioritize what works best for the team and the project

antedotee:
@pabo99 I have used search query only, so I don't think we will face such inconsistencies which was raised in the PR you mentioned.
Also, regarding the problems we are facing with the workflow, yes, I think there are efforts getting invested. But if we can handle this in the short term, I think in long term, this will save significant time.
Let's work together and make it work @Ankitsinghsisodya @pabo99 🙌. If we face problems for a prolonged period of time, we can switch to what Ankit has proposed.

pabo99:
There’s one more conflict. Could you resolve it? Once it’s resolved, I’ll review the PR

pabo99:
Can you check the conflicts and resolve them?

.github/workflows/assign-issues.yml
```diff
@@ -1,222 +1,382 @@
...
-    schedule:
-        # Run daily to handle reminders and automatic unassignments
-        - cron: "0 3 * * *"
-    workflow_dispatch:
+
+env:
+    # Maximum number of open issues a user can have assigned in THIS repository.
+    MAX_OPEN_ASSIGNMENTS: 2
+    # Space or comma-separated list of usernames allowed to open issues that support /assign.
+    ALLOWED_ISSUE_CREATORS: "heduenas pabo99 carlosabcs tvanessa"
```
pabo99: Is it possible that we use the creator associations (`OWNER`, `MEMBER`, `COLLABORATOR`) here instead of that hardcoded list? That would help ensure that any issue reported by any collaborator is eligible to be assigned

.github/workflows/assign-issues.yml
```diff
@@ -1,222 +1,382 @@
...
     issue_comment:
         types: [created]
-    schedule:
-        # Run daily to handle reminders and automatic unassignments
-        - cron: "0 3 * * *"
-    workflow_dispatch:
+
+env:
+    # Maximum number of open issues a user can have assigned in THIS repository.
+    MAX_OPEN_ASSIGNMENTS: 2
```
pabo99, line 9: With these changes, it will be necessary to update the `CONTRIBUTING.md` and `frontend/www/docs/Issue-Assignment-Workflow.md` documentation. Could you help me with that?

pabo99:
Could you check this error as well?

```bash
2026-03-04T13:30:38+00:00;INFO;verifying reversemap usage in all workflows
2026-03-04T13:30:38+00:00;INFO;./.github/workflows/assign-issues.yml uses actions/github-script@v7, whose version v7 is not a commit hash
2026-03-04T13:30:38+00:00;ERROR;Action reference problems found; see https://github.com/omegaup/omegaup/blob/main/CONTRIBUTING.md#github-action-reference-discipline
Error: Process completed with exit code 90.
```

Thank you in advance!

antedotee:
> Could you check this error as well?
> 
> ```shell
> 2026-03-04T13:30:38+00:00;INFO;verifying reversemap usage in all workflows
> 2026-03-04T13:30:38+00:00;INFO;./.github/workflows/assign-issues.yml uses actions/github-script@v7, whose version v7 is not a commit hash
> 2026-03-04T13:30:38+00:00;ERROR;Action reference problems found; see https://github.com/omegaup/omegaup/blob/main/CONTRIBUTING.md#github-action-reference-discipline
> Error: Process completed with exit code 90.
> ```
> 
> Thank you in advance!

Yes, it is probably failing because we are using tag instead of a commit hash. Sorry, I didn't look at it. Fixing it!

antedotee:
PTAL @pabo99.

pabo99 (commented):
The change looks good, but the documentation in `CONTRIBUTING.md` still needs to be updated.

In addition, all references to `takanome-dev/assign-issue-action`, which is no longer used, should be removed

Issue #8961, opened 2026-02-03
[BUG] /assign command fails silently for experienced issue creators due to invalid GraphQL query


Yashagarwal9798:
## Description

The "experienced issue creator exception" feature in the `/assign` command is broken. When a user with 10+ merged PRs tries to self-assign an issue they created using `/assign`, the workflow crashes silently with no response.

## Current Behavior

When commenting `/assign` on an issue I created (and I have 12+ merged PRs), I receive **no response at all** - no success message, no error message, nothing. The issue remains unassigned.

## Expected Behavior

Users with 10+ merged PRs who comment `/assign` on their own issues should:
1. Be automatically assigned to the issue
2. Receive a success message indicating they bypassed the standard 5-assignment limit

## Steps to Reproduce

1. Have 10+ merged PRs in the repository
2. Create a new issue
3. Comment /assign on that issue
4. Observe: No response, issue remains unassigned
5. Check Actions tab: The *Assign Issues* workflow crashes with the GraphQL error

Yashagarwal9798:
@pabo99 I have already implemented a fix locally. Could you please assign this issue to me so I can open a Pull Request?

PR #8981, opened 2026-02-04
fix: use search query instead of invalid pullRequests author argument

Yashagarwal9798:
# Description

The GraphQL query in the experienced contributor check was using `pullRequests(author: $author)` which is not a valid GitHub GraphQL API argument. This caused the workflow to crash silently, preventing users with 10+ merged PRs from self-assigning their own issues.

Changed to use the `search` query with `is:pr is:merged author:X` filter which is the correct way to query PRs by author in GitHub's GraphQL API.

## Error that was occurring:
GraphqlResponseError: Request failed due to following response errors:

- Field 'pullRequests' doesn't accept argument 'author'
- Variable $author is declared by anonymous query but not used

## Changes Made:
- Replaced `repository.pullRequests(author: $author)` with `search(query: "is:pr is:merged author:X repo:owner/repo")`
- Updated result parsing from `result.repository.pullRequests` to `result.search`
- Added `... on PullRequest { }` fragment since search returns a union type

Fixes: #8961

# Comments

This is a minimal fix that only changes the GraphQL query mechanism. The business logic (counting merged PRs, pagination, checking if >= 10 PRs) remains unchanged.

# Checklist:


Yashagarwal9798:
@pabo99 I have traced the specific flow that causes the crash. It only occurs when a user tries to self-assign an issue they created using the exact command /assign.

The Crash Flow (Fails):
 1. Trigger: User comments /assign on their own issue.
 2. Pre-check: System confirms the user is the Issue Creator.
 3. Logic: It attempts to count merged PRs to verify "Experienced User" status.
 4. Query: Executes GraphQL query pullRequests with argument author.
 5. Error: 💥 CRASH – "Field 'pullRequests' doesn't accept argument 'author'."
 6. Result: Silent failure (no response).

The Standard Flow (Works): The crash is avoided in other scenarios because the "Experienced User" check is skipped:
- If the comment is not exactly /assign (e.g., "please /assign").
- If the user is not the creator of the issue.

In these cases, the buggy GraphQL query is never executed, so the standard assignment flow succeeds.

Issue #8988, opened 2026-02-05
[BUG] Weak randomness for password-reset tokens in `ApiUtils::getRandomString()`


shiavm006:
## Why This Is a Problem
- rand() and uniqid() are not cryptographically secure and are documented as unsuitable for security-sensitive tokens.
- md5() is a deterministic hash and does not add entropy or make weak randomness secure.

Since these tokens are used in password-reset links, an attacker who can approximate the generation time and send many guesses has a better chance of guessing a valid token than they should.
This contrasts with Session::registerSession(), which already uses random_bytes() + sha256, following modern best practices.

shiavm006:
cc @pabo99 what r u thoughts on it

pabo99:
Thanks for the clarification. I agree this is a valid concern, and switching to a cryptographically secure approach here is the right call

PR #8991, opened 2026-02-06
Use CSPRNG and sha256 for password reset tokens

shiavm006:
### Summary

This PR strengthens the security of password-reset tokens by switching from weak randomness (`rand()` + `uniqid()` + `md5`) to a cryptographically secure random source, and by using `sha256` instead of `sha1` when storing and verifying reset digests. This addresses [#8988] (Weak randomness for password-reset tokens in `ApiUtils::getRandomString()`).

Fixes: #8988 

### Changes

- `frontend/server/src/ApiUtils.php`
  - Change `ApiUtils::getRandomString()` from:
    - `md5(uniqid(strval(rand()), true))`
    - **to** `bin2hex(random_bytes(32))` (256-bit CSPRNG, hex encoded).

- `frontend/server/src/Controllers/Reset.php`
  - In `apiCreate()` and `apiGenerateToken()`:
    - Store the reset digest using `hash('sha256', $token)` instead of `hash('sha1', $token)`.
  - In `apiUpdate()`:
    - Verify the provided `reset_token` by comparing `hash('sha256', $resetToken)` with the stored `reset_digest`.

### DIFF
--- a/.github/actions/prepare-env/action.yml
+++ b/.github/actions/prepare-env/action.yml
@@ -77,6 +77,12 @@ runs:
         define('OMEGAUP_GITSERVER_URL', 'http://gitserver:33861');
         define('OMEGAUP_GRADER_URL', 'https://grader:21680');
         define('OMEGAUP_GITSERVER_SECRET_TOKEN', 'secret');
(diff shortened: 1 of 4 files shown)


shiavm006:
cc @pabo99 i think it should just be  fine did what we discussed

pabo99 (approved):
Thanks!

Issue #8993, opened 2026-02-06
[BUG] Some workflow uses latest version of ubuntu


antedotee:
## Current Behavior

Same as title

## Expected Behavior
Ideally, all workflows should use 22.04 as the default version, as latest is an ever moving target and can sometimes cause problems.

PR #8994, opened 2026-02-06
fixed the ubuntu version in workflows

AgrapujyaLashkari:
# Description

Fixed the workflow to use ubuntu-22.04 rather than the latest one.

Fixes: #8993


AgrapujyaLashkari:
CC @pabo99 . Please have a look.

PR #8995, opened 2026-02-06
added the first interaction workflow for newcomers

AgrapujyaLashkari:
# Description

I have added the first-interactions workflow and added the discord link and the wiki link for newcomers. The workflow will run only for the first time, when the newcomer has raised its first issue or PR.

Fixes: #8548


AgrapujyaLashkari:
CC @pabo99 . I have raised PR as the issue was stale for a prolonged period of time. Also, if you want me to add any other information for newcomer, I will be happy to implement that too.
Thanks and regards.

pabo99:
Thank you, @AgrapujyaLashkari.

Could you request that the issue be assigned to you? That way we'll avoid the error in the workflows

.github/workflows/greetings.yml
```diff
@@ -0,0 +1,37 @@
...
+      - uses: actions/first-interaction@v3.1.0
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          issue-message: |
+            Hi @${{ github.actor }} — thanks for opening your first issue!
+
+            1. Discord Link: https://discord.com/invite/K3JFd9d3wk
+            2. Documentation (GitHub Wiki): https://github.com/omegaup/omegaup/wiki
+
+            Most contributor discussion happens in the `dev-training` channel.
```
pabo99: Could you add this as a variable that can be used for all the messages created here? That way, if we need to update it, we only do it in one place instead of in multiple places.

pabo99:
Hello, @AgrapujyaLashkari! The issue linked to this PR is now assigned to you. When you push your changes with the suggestions I shared, the `check‑issue‑link` job should pass smoothly

pabo99 (approved):
Thanks, @AgrapujyaLashkari!

Issue #8992, opened 2026-02-06
Security: Default config contains real-looking secrets (operational security risk)


shiavm006:
`config.default.php` contains high-entropy secrets that look like production values. They’re committed to the repo, so anyone can use them.

**Problem**

If a deployment relies on defaults (staging, new prod region, forgot to set `config.php`), it uses these public secrets. No error, no warning.

**Affected keys (lines 37–77):**

- `OMEGAUP_GITSERVER_SECRET_KEY` / `OMEGAUP_GITSERVER_PUBLIC_KEY` — can forge gitserver auth
- `OMEGAUP_GRADER_SECRET` — can forge grader requests
- `OMEGAUP_COURSE_CLONE_SECRET_KEY` — can forge course clone tokens

**Fix**

Replace these with obvious placeholders like `CHANGE_ME` so misconfigured deployments fail loudly instead of silently using public secrets. Optionally add startup checks in production to reject placeholder values.

shiavm006:
cc @pabo99 what u think about it

Yashagarwal9798:
Hello @shiavm006, just a heads-up that the self-assign feature is working again! You should be able to assign this issue to yourself.

shiavm006:
@Yashagarwal9798 ya just waiting for the @pabo and others thoughts on this

pabo:
> @Yashagarwal9798 ya just waiting for the @pabo and others thoughts on this

LGTM!

...but you meant to tag @pabo99

PR #8997, opened 2026-02-06
fix: replace real-looking secrets in config.default.php with placeholder

shiavm006:
## Description

Replaces real-looking secrets in `config.default.php` with placeholders to avoid insecure defaults. Environments that relied on these defaults could silently use public, known secrets.

## Problem

`config.default.php` contained high-entropy secrets that looked like production values:
- `OMEGAUP_GITSERVER_SECRET_KEY` / `OMEGAUP_GITSERVER_PUBLIC_KEY`
- `OMEGAUP_GRADER_SECRET`
- `OMEGAUP_COURSE_CLONE_SECRET_KEY`
- `OMEGAUP_GOOGLE_SECRET` / `OMEGAUP_GOOGLE_CLIENTID`
- `OMEGAUP_EXPERIMENT_SECRET`

Any deployment (staging, new prod region, developer env) that did not override these in `config.php` would use these public secrets, allowing:
- Gitserver auth forgery
- Grader request forgery
- Course clone token forgery

## Solution

- **config.default.php:** Replace these with `CHANGE_ME` / `xxxxx` placeholders.
- **test_config.default.php:** Add test-only secret definitions so tests keep working.
- **prepare-env action & developer-environment.sh:** Add explicit secret overrides for CI and dev environments.

Misconfigured deployments now fail when these values are used instead of silently using public secrets.

fixes : #8992


shiavm006:
cc @pabo99

frontend/server/config.default.php
```diff
@@ -120,11 +111,8 @@
 # ####################################
 # GOOGLE LOGIN CONFIG
 # ####################################
-try_define('OMEGAUP_GOOGLE_SECRET', 'acmtr0Y37vnTVJV4BwmdhOsK');
-try_define(
-    'OMEGAUP_GOOGLE_CLIENTID',
-    '982542692060-lf9htvij4ba13fiufpqeldic0qqqvird.apps.googleusercontent.com'
-);
+try_define('OMEGAUP_GOOGLE_SECRET', 'xxxxx');
+try_define('OMEGAUP_GOOGLE_CLIENTID', 'xxxxx');
```
pabo99, line 115: This configuration points to localhost and it helped me log in from my local environment. Could you also add it in `prepare-env` and in `environment-development`, or do you have a better strategy?

shiavm006: u r right  those values were useful for local Google login. I will add them to prepare-env and developer-environment, using the same pattern as the gitserver/grader/course-clone secrets. That should restore local login while keeping config.default.php safe.

shiavm006: ig now it should work fine

Issue #9001, opened 2026-02-06
[BUG] State-changing API endpoints can be called via GET query parameters


shiavm006:
## Current Behavior

API request construction uses `$_REQUEST` and does not enforce HTTP method for mutating APIs.

- `frontend/server/src/ApiCaller.php:230`

As a result, endpoints like `User::apiUpdate` can be invoked via GET query params instead of POST.

## Expected Behavior

Mutating APIs should reject GET and enforce POST/PUT/PATCH (and CSRF protection should align with that method policy).

shiavm006:
cc @pabo99  like whats ur point on this would love to know

pabo99:
Thanks for bringing this up! I really appreciate you highlighting it. 

This is an important issue for the platform’s security and reliability

PR #9005, opened 2026-02-07
Enforce POST for mutating API endpoints (reject GET with 405)

shiavm006:
## Problem
State-changing API endpoints (e.g. `User::apiUpdate`) could be called via GET query parameters because the server used `$_REQUEST` and did not check HTTP method.

## Solution
- Reject GET for mutating endpoints and return **405 Method Not Allowed** with a clear error message.
- Read-only endpoints (list, details, profile, etc.) still allow GET.
- Two read-only endpoints whose names contain a mutating verb (`listAssociatedIdentities`, `statusVerified`) are allowlisted so GET remains allowed for them.

Fixes: #9001


shiavm006:
cc @pabo99 just check it once

frontend/tests/controllers/ApiCallerGetRejectionTest.php
```diff
@@ -0,0 +1,74 @@
...
+class ApiCallerGetRejectionTest extends \OmegaUp\Test\ControllerTestCase {
+    private function cleanupRequestMethod(): void {
+        unset($_SERVER['REQUEST_METHOD']);
+        unset($_SERVER['REQUEST_URI']);
+    }
+
+    public function testGetToMutatingEndpointReturns405() {
+        $_SERVER['REQUEST_METHOD'] = 'GET';
+        $_SERVER['REQUEST_URI'] = '/api/user/update';
+        $_REQUEST = [];
```
pabo99: How about adding this initialization in the `setUp` function? You can find implementation examples in many of the test files.

frontend/tests/controllers/ApiCallerGetRejectionTest.php
```diff
@@ -0,0 +1,74 @@
+<?php
+// phpcs:disable VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
+
+/**
+ * GET method rejection for mutating API endpoints.
+ */
+class ApiCallerGetRejectionTest extends \OmegaUp\Test\ControllerTestCase {
+    private function cleanupRequestMethod(): void {
+        unset($_SERVER['REQUEST_METHOD']);
+        unset($_SERVER['REQUEST_URI']);
```
pabo99, line 16: This could very well go in the `tearDown` function. That way you avoid creating this function, and you also no longer need to wrap all the tests in `try/catch` statement.

frontend/tests/controllers/ApiCallerGetRejectionTest.php
```diff
@@ -0,0 +1,74 @@
+<?php
+// phpcs:disable VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
```
pabo99: Can you remove this line?
It looks like the linter isn't complaining about `$_REQUEST` not being used.

frontend/tests/controllers/ApiCallerGetRejectionTest.php
shiavm006: Done .. the common initialization is now in setUp()

frontend/tests/controllers/ApiCallerGetRejectionTest.php
shiavm006: done that u that make sense

Issue #8962, opened 2026-02-03
[BUG]Redis getOrSet causes cache stampede when OMEGAUP_CACHE_IMPLEMENTATION is redis


shiavm006:
## Current Behavior

When `OMEGAUP_CACHE_IMPLEMENTATION` is `redis`, `RedisCacheAdapter::getOrSet()` in `frontend/server/src/Cache.php` does not acquire any distributed lock before calling the expensive `$setFunc()` callback on a cache miss. Every concurrent request that sees a miss runs `$setFunc()` and then races to set the key. Only one request wins the `SET NX`, but all of them have already executed the callback. This is a cache stampede: N concurrent misses cause N redundant recomputations (e.g. N heavy DB/aggregation calls), which can spike load and latency when cache is cold or after invalidation.

The base `CacheAdapter::getOrSet()` (used by APCCacheAdapter) prevents this with a file lock per `lockGroup` (`/tmp/omegaup-cache-*.lock`). `RedisCacheAdapter` overrides `getOrSet()` and does not use that lock (and file locks don't work across multiple frontend servers). The TODO at lines 336–338 in `Cache.php` documents this: "This does not acquire a global lock to avoid this from being called multiple times. Maybe we should try a Redlock here."

`Cache::getFromCacheOrSet()` is used in many places (ranks, contest/problem lists, session, user profile, problem statement/settings, run counts, course progress, etc.), so the impact is high when Redis is used with multiple frontend processes.

## Expected Behavior

On a cache miss, only one process should run `$setFunc()` per key (or per lock group); others should wait or retry and then read the value written by the first. Behavior should match the file-lock path used by the base adapter (single flight / no stampede).

shiavm006:
cc @pabo99 @antedotee

PR #9014, opened 2026-02-07
Fix Redis getOrSet cache stampede with distributed lock

shiavm006:
## Problem

When `OMEGAUP_CACHE_IMPLEMENTATION` is `redis`, `RedisCacheAdapter::getOrSet()` does not take a lock before calling the expensive `$setFunc()` on a cache miss. Every concurrent request that sees a miss runs `$setFunc()` and then races to set the key; only one write wins, but all have already run the callback. That causes a cache stampede (N redundant recomputations) and can spike load and latency when the cache is cold or after invalidation.

The base `CacheAdapter::getOrSet()` (used by APCu/InProcess) avoids this with a file lock per `lockGroup`. The Redis adapter overrides `getOrSet()` and did not use any lock (and file locks don’t work across multiple frontend servers). The TODO at lines 336–338 in `Cache.php` noted this and suggested Redlock.

## Solution

- **Acquire a Redis lock** per `lockGroup` before calling `$setFunc()`: `SET lockKey token NX PX 30000` (30s TTL).
- **Waiters** that fail to acquire the lock poll `fetch($key)` until the key is set (or the lock expires), then return from cache.
- **Release** the lock only if we still hold it using a Lua script that deletes the key only when its value equals our token.
- **Double-check** after acquiring the lock: if the key was already set (e.g. by a previous holder), return it without running `$setFunc()`.

Behavior now matches the base adapter: only one process runs `$setFunc()` per key/lockGroup on a miss; others wait and then read from cache (single flight, no stampede).

Fixes: #8962


shiavm006:
cc @pabo99 sir could u check this out

Issue #9022, opened 2026-02-08
[BUG]  Scoreboard Merge Button Alignment Issue


ParvTyagi:
# Scoreboard Merge Button Alignment Issue

## Title

Scoreboard Merge button is not aligned with the contest dropdown

---

## Description

On the **Scoreboard Merge** page, the **"Show Total Scoreboard"** button is not vertically aligned with the multiselect contest dropdown. The button appears slightly higher than the dropdown, causing a visual misalignment in the layout.

---

## Steps to Reproduce

1. Navigate to the Scoreboard Merge page:

   ```
   /scoreboardmerge/
   ```
2. Observe the layout of the contest multiselect dropdown and the **Show Total Scoreboard** button.

---

## Expected Behavior

The **Show Total Scoreboard** button should be vertically aligned on the same baseline as the multiselect contest dropdown.

---

## Actual Behavior

The button is vertically offset from the dropdown.

### Root Cause

The parent row uses `align-items-center`, which vertically centers all child elements relative to the tallest column. Since the dropdown column includes a `<label>` above the multiselect component, the button is centered against the combined height of the **label + dropdown**, causing it to appear higher than the dropdown itself.

---

## Fix

Update the row container alignment in **ScoreboardMerge.vue**:

**File:** `ScoreboardMerge.vue`

**Change:**

```diff
- align-items-center
+ align-items-end
```

This ensures the button aligns with the bottom of the multiselect dropdown instead of being centered relative to the entire column height.

---

### Current Behavior (Before Fix)
**Placeholder:**

https://github.com/user-attachments/assets/3191f9d8-b39f-43c5-a4e4-c9244ede5357

---

### Expected Behavior (After Fix)
**Placeholder:**

https://github.com/user-attachments/assets/4f2a3746-ec26-495c-93a7-458563c25ed4

---

## Notes

* This change affects layout only and has no impact on functionality.
* The fix is safe and scoped to the Scoreboard Merge page UI.

PR #9024, opened 2026-02-08
Fix Scoreboard Merge Button Alignment (Issue #9022)

ParvTyagi:
# Fix Scoreboard Merge Button Alignment (Issue #9022)

## Summary

Fixes issue **#9022**: a UI alignment issue on the **Scoreboard Merge** page where the **"Show Total Scoreboard"** button was not aligned with the contest multiselect dropdown.

---

## Fix

Updated the row alignment from `align-items-center` to `align-items-end` so the button aligns with the bottom of the dropdown instead of being centered relative to the label + dropdown height.

---

## Result

The button and the contest multiselect dropdown are now properly aligned on the same line, improving visual consistency.

---

## Notes

* UI-only change
* No functional impact

Fixes : #9022

### DIFF
--- a/.github/workflows/assign-issues.yml
+++ b/.github/workflows/assign-issues.yml
@@ -174,8 +174,9 @@ jobs:
                   reminder_comment: >
                       Quick reminder: please open a pull request within 3.5 days
                       or this issue will be unassigned so others can pick it up.
-                  assigned_label: ""
-                  pin_label: ""
+                  # Disabled: empty strings cause GitHub API validation error (see #9033)
+                  # assigned_label: ""
+                  # pin_label: ""
                   max_assignments: ${{ env.MAX_ASSIGNMENTS }}
                   max_assignments_message: >
                       You already have the maximum number of assigned issues (${{ env.MAX_ASSIGNMENTS }}).
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
 
     services:
       mysql:
-        image: mysql:8.0.23
+        image: mysql:8.0.34
         ports:
           - 13306:13306
         env:
--- a/frontend/www/docs/Architecture.md
+++ b/frontend/www/docs/Architecture.md
@@ -51,7 +51,7 @@ The list of technologies we will use to build the application are as follows:
 
 | Technology          | Purpose                | Version  |
 | ------------------- | ---------------------- | -------- |
-| MySql            | Database             | [8.0.39] |
+| MySql            | Database             | [8.0.34] |
 | PHP              | Controllers        | [8.1.2]   |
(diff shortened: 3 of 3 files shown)


frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
```diff
@@ -4,7 +4,7 @@
       <h3 class="card-title mb-0">{{ T.omegaupTitleScoreboardmerge }}</h3>
     </div>
     <div class="card-body">
-      <div class="row align-items-center">
+      <div class="row align-items-end">
         <div class="form-group col-md-9" data-merge-contest-name>
```
pabo99: How about assigning 8 columns here, so the button's div can take 4 columns? This way it won't wrap into two lines and the layout will look cleaner.

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
```diff
@@ -4,7 +4,7 @@
...
           <label>{{ T.wordsContest }}:</label>
```
pabo99: What do you think if we remove this label?

IMO, since the multiselect already has a descriptive placeholder, this label is no longer necessary

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
pabo99: Could you add the `w-100` class here so it always takes up the maximum possible width?

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
ParvTyagi: Done!

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
ParvTyagi: That's a good idea.

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
ParvTyagi: Done!

frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
```diff
@@ -18,7 +17,7 @@
             @select="onSelect"
           ></multiselect>
         </div>
-        <div class="form-group col-md-3 text-right">
+        <div class="form-group col-md-4 text-right w-100">
```
pabo99: The `w-100` class should be applied to the button instead, so it spans the full width at all screen resolutions

pabo99 (commented):
One minor suggestion!

pabo99 (approved):
It looks much better! Thank you, @ParvTyagi!

Issue #9027, opened 2026-02-08
[FEATURE] Add reset button to ephemeral grader code editor to restore default code template


vijay-x-dev:
Adding a reset button  in the ephemeral grader that restores the default code template for the current language.

PR #9028, opened 2026-02-08
Add reset button to ephemeral grader editor to restore default code template

vijay-x-dev:
# Description

Adds a reset button in the ephemeral grader that restores the default code template for the current language. `This is useful when users want to start a new approach, allowing them to quickly reset to the default template without manually clearing or rewriting code.`

Fixes: #9027 

**Video:**

https://github.com/user-attachments/assets/e7cbea62-9c0b-4cd5-a9df-a86fe4dd77d2

# Checklist:


frontend/www/js/omegaup/grader/MonacoEditor.vue
```diff
@@ -148,6 +169,25 @@ export default class MonacoEditor extends Vue {
...
+
+    if (this.isInteractive) {
+      // For interactive problems, use original interactive templates
+      defaultTemplate = templates.originalInteractiveTemplates[extension];
+    } else {
+      // For regular problems, use source templates
+      defaultTemplate = templates.sourceTemplates[extension];
+    }
+
+    this.contents = defaultTemplate;
```
pabo99: Could you simplify this a bit?


```suggestion
    if (this.isInteractive) {
      // For interactive problems, use original interactive templates
      this.contents = templates.originalInteractiveTemplates[extension];
      return;
    }
    // For regular problems, use source templates
    this.contents = templates.sourceTemplates[extension];
```

This way you don’t need to create a new variable.

vijay-x-dev:
> I just have one question — do you know what default data is loaded or which template is being used?
> 
> The thing is, I don't recall having changed the code for some languages, and yet when I clicked on reset button, the code was replaced. Let me show you an example:
> 
>  20260228-0010-48.2827788.mp4 
> Other than that, it's only the minor suggestion I left you.

The default data is the template in `GraderTemplates.ts`. 

In your C# example, the code had been changed ( // your code here )part, so reset restored the original template.

https://github.com/user-attachments/assets/c172572a-be66-491f-bf65-8424e01a5752

PR #9032, opened 2026-02-09
Fix #9031 : validate auth token format in Session:: isAuthTokenValid()

### LINKED ISSUE #9031: [BUG] Fix isAuthTokenValid no-op in Session.php
### Summary

`Session::isAuthTokenValid()` was a no-op that always returned `true`, bypassing any client-side token format validation. This PR implements actual validation against the known token format.

### Problem

```php
private static function isAuthTokenValid(string $authToken): bool {
    //do some other basic testing on authToken
    return true;
}
```

Any arbitrary string passed as an auth token would be accepted as structurally valid, deferring all rejection to the database lookup. This is a low-cost check that should happen before hitting the DB.

### Changes

The method now validates the token against its expected format: `{entropy}-{identity_id}-{sha256_hash}`

| Segment       | Expected                                      |
|---------------|-----------------------------------------------|
| `entropy`     | Exactly 30 lowercase hex chars (`2 × AUTH_TOKEN_ENTROPY_SIZE`) |
| `identity_id` | Non-empty numeric string                      |
| `hash`        | Exactly 64 lowercase hex chars (SHA-256)      |

Validation checks:
- Token splits into exactly 3 dash-separated parts
- Entropy length and charset (`[0-9a-f]`)
- Identity ID is a non-empty digit-only string
- Hash length and charset (`[0-9a-f]`)

### Impact

- **Security**: Rejects malformed tokens early, before any database or cache lookup
- **Performance**: Avoids unnecessary DB queries for obviously invalid tokens
- **Risk**: Low — only rejects tokens that could never match a legitimately generated one

### Files Changed

- `frontend/server/src/Controllers/Session.php`

ParvTyagi:
### Problem

`Session::isAuthTokenValid()` always returned `true` with no actual validation:

```php
private static function isAuthTokenValid(string $authToken): bool {
    //do some other basic testing on authToken
    return true;
}
```
Fixes #9031

### DIFF
--- a/frontend/server/src/Controllers/Session.php
+++ b/frontend/server/src/Controllers/Session.php
@@ -69,10 +69,10 @@ class Session extends \OmegaUp\Controllers\Controller {
 
     private static function isAuthTokenValid(string $authToken): bool {
         // Token format: {entropy}-{identity_id}-{sha256_hash}
-        // entropy: 30 hex chars (15 bytes via bin2hex)
+        // entropy: 2 * AUTH_TOKEN_ENTROPY_SIZE hex chars via bin2hex
         // identity_id: numeric string
         // hash: 64 hex chars (SHA-256)
-        $parts = explode('-', $authToken);
+        $parts = explode('-', $authToken, 3);
         if (count($parts) !== 3) {
             return false;
         }
--- a/frontend/tests/controllers/SessionAuthTokenTest.php
+++ b/frontend/tests/controllers/SessionAuthTokenTest.php
@@ -1,0 +1,191 @@
+<?php
+// phpcs:disable VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
+
+/**
+ * Tests for auth token format validation in Session controller.
+ */
+class SessionAuthTokenTest extends \OmegaUp\Test\ControllerTestCase {
+    /**
+     * A valid auth token should produce a valid session.
+     */
+    public function testValidAuthTokenProducesSession(): void {
+        ['identity' => $identity] = \OmegaUp\Test\Factories\User::createUser();
+        $login = self::login($identity);
+
+        \OmegaUp\Controllers\Session::invalidateLocalCache();
+        $session = \OmegaUp\Controllers\Session::getCurrentSession(
+            new \OmegaUp\Request([
+                'auth_token' => $login->auth_token,
+            ])
+        );
+        $this->assertTrue($session['valid']);
+        $this->assertNotNull($session['identity']);
+    }
+
+    /**
+     * Tokens with wrong number of parts should be rejected.
+     */
+    public function testTokenWithWrongPartCountIsRejected(): void {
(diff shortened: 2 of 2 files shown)


frontend/tests/controllers/SessionAuthTokenTest.php
```diff
@@ -0,0 +1,191 @@
...
+        \OmegaUp\Controllers\Session::invalidateLocalCache();
+        $session = \OmegaUp\Controllers\Session::getCurrentSession(
+            new \OmegaUp\Request([
+                'auth_token' => $token,
+            ])
+        );
+        $this->assertFalse(
+            $session['valid'],
+            'Token with non-hex entropy should be rejected'
+        );
```
pabo99, line 67: Is it enough to just know that the token is invalid? If we needed to know the reason, we would have to change the `isAuthTokenValid` function so that it throws different exceptions depending on the specific error.

From what I see, you only need to know whether it’s valid or not, but please help me confirm this information

frontend/tests/controllers/SessionAuthTokenTest.php
```diff
@@ -0,0 +1,191 @@
...
+                new \OmegaUp\Request([
+                    'auth_token' => $token,
+                ])
+            );
+            $this->assertFalse(
+                $session['valid'],
+                "Token '{$token}' should have been rejected (wrong part count)"
+            );
+        }
+    }
```
pabo99, line 48: Could you change this to a parameterized test? That way, we can see if more than one of the tested cases fails, instead of only getting information about the first failure. We have some examples in the code. If you need help, just let us know

frontend/server/src/Controllers/Session.php
```diff
@@ -68,7 +68,37 @@ public static function getFacebookLoginUrl(): string {
...
+
+        // Validate identity_id: non-empty numeric string
+        if ($identityId === '' || !ctype_digit($identityId)) {
+            return false;
+        }
+
+        // Validate hash: exactly 64 hex characters (SHA-256)
+        if (
+            strlen($hash) !== 64
+            || preg_match('/^[0-9a-f]+$/', $hash) !== 1
```
pabo99, line 97: Could you also add the string length to the regex? That way, we don’t need to include two separate validations

frontend/server/src/Controllers/Session.php
```diff
@@ -68,7 +68,37 @@ public static function getFacebookLoginUrl(): string {
...
+        $parts = explode('-', $authToken, 3);
+        if (count($parts) !== 3) {
+            return false;
+        }
+        [$entropy, $identityId, $hash] = $parts;
+
+        // Validate entropy: exactly 30 hex characters
+        if (
+            strlen($entropy) !== 2 * self::AUTH_TOKEN_ENTROPY_SIZE
+            || preg_match('/^[0-9a-f]+$/', $entropy) !== 1
```
pabo99, line 84: Same case here

Issue #9039, opened 2026-02-10
Block production startup when critical secrets use placeholder or empty values


shiavm006:
## Summary

Refuse to start the application in **production** if any critical secret or credential is still using a placeholder value (`CHANGE_ME`, `xxxxx`) or is empty. This is a defense-in-depth measure to prevent accidental prod deployments with example/test secrets.

Running in production with placeholder or empty secrets (DB credentials, grader/gitserver secrets, OAuth secrets, etc.) is unsafe and often unintended. We want misconfigured prod deployments to **fail fast** instead of silently running with invalid secrets.

shiavm006:
cc @pabo99  sir what r ur thoughts in it

shiavm006:
hey @AiGentsy  can go for the issue just get it assign urself i would be happy to guide u on that if u have doubts

pabo99:
Hello @shiavm006, @AiGentsy. Thank you both for your interest in collaborating on this issue.

To make sure the expectations are clear, I recommend following this checklist when implementing the validation:

- [ ] Fail startup if any secret contains a placeholder value (`CHANGE_ME`, `xxxxx`).
- [ ] Fail startup if any secret is empty.
- [ ] Fail startup if any critical secret is missing from the configuration.
- [ ] Ensure validation is applied in the sandbox environment before deploying to production .
- [ ] Provide clear error messages indicating which secret caused the failure.

Following this checklist will ensure sandbox and production remain consistent and secure, while making missing or misconfigured secrets visible early

shiavm006:
hey @pabo99 i feel like @AiGentsy is not a person its automated script

PR #9043, opened 2026-02-10
Validate critical secrets in sandbox and production

shiavm006:
## Summary

Add a startup-time validation step that refuses to run the application in `sandbox` and `production` when critical secrets are missing, empty, or still using placeholder values like `CHANGE_ME` or `xxxxx`. This makes misconfigured secrets fail fast instead of silently running with unsafe defaults.

Fixes: #9039


shiavm006:
cc @pabo99 sir just check it out once

shiavm006:
CC @pabo99 just look it ones  if it  seems fine

pabo99 (commented):
Thanks for raising this PR. The overall idea is solid, but returning a 500 error at this early stage of secrets validation feels a bit drastic, especially since some secrets might be misconfigured but not actually in use yet.  

Would it be possible to explore triggering an alert in New Relic instead? Once we're confident everything is properly configured, we could revisit this stricter approach.

shiavm006:
> Would it be possible to explore triggering an alert in New Relic instead? Once we're confident everything is properly configured, we could revisit this stricter approach.

ya can do that happy to tighten it again later once config is solid.

frontend/server/bootstrap.php
```diff
@@ -2,6 +2,110 @@
...
+        implode(PHP_EOL, array_map(
+            /**
+             * @param string $failure
+             */
+            fn (string $failure): string => "  - {$failure}",
+            $failures
+        ))
+    );
+
+    error_log($message);
```
pabo99: Can `$rootLogger` be used to throw this error? If so, could you use it instead of `error_log`?

frontend/server/bootstrap.php
```diff
@@ -2,6 +2,110 @@
...
+
+    if (class_exists(\OmegaUp\NewRelicHelper::class)) {
+        \OmegaUp\NewRelicHelper::noticeError($message);
```
pabo99, line 105: Thanks, this will allow us to be notified at the time of deploy about misconfigured secrets without blocking the UI.

frontend/server/bootstrap.php
shiavm006: ya done that u could check it once

shiavm006:
i think now it is correct

Issue #9049, opened 2026-02-11
[BUG] Some of the workflows still fail because of using tags


antedotee:
## Current Behavior

Same as title

## Expected Behavior
The action-hashes check should pass

PR #9051, opened 2026-02-11
replaced the remaining tags with pinned commit hash

antedotee:
# Description
Replaced the remaining tags with the commit hash, as well as formatted the `.gha-reversemap.yml` for better strcuture.

Fixes: #9049


antedotee:
@pabo99 Earlier the workflow was running twice, one as a part of CI and one standalone. I have removed the unnecesarry one to save runner costs and minutes. PTAL.

PR #9060, opened 2026-02-12
Add Bookmarked Problems shortcut to Problems dropdown

### LINKED ISSUE #9040: [FEATURE] Add centralized Bookmarked Problems list
## Title

Add centralized Bookmarked Problems list

---

## Description

Currently, users can bookmark problems, but there is no dedicated place to view all bookmarked items.

Once a problem is bookmarked, the only way to access it again is by manually navigating back to the same problem page. This makes bookmarks difficult to use for organizing problems to practice later.

---

## Current Behavior

- Problems can be bookmarked successfully.
- There is no page or section to list all bookmarked problems.
- Users cannot manage bookmarks from a single place.

---

## Expected Behavior

Provide a centralized **Bookmarked Problems** view where users can:

- See all bookmarked problems in one place.
- Navigate directly to any bookmarked problem.
- Remove bookmarks easily.

---

## Benefits

- Makes the bookmark feature practically useful.
- Improves user workflow for practicing problems.
- Provides better organization of saved content.

---

I’d be happy to implement this feature if maintainers agree.

Prince-Prajapati07:
## Description

Bookmarked problems currently exist under Profile → Problems but are difficult to discover since they are located at the bottom of the page.

This PR adds a **Bookmarked Problems** entry to the existing **Problems** dropdown menu, providing direct access to `/profile/#problems` without introducing new routes or backend changes.

## Comments

- Reuses the existing bookmarked problems section
- Improves discoverability with a single-click shortcut
- No API, routing, or layout changes
- Minimal frontend-only change

## Checklist

- [x] Code follows omegaUp guidelines  
- [x] No backend changes  
- [x] Existing functionality reused  
- [x] Manual testing completed  

Fixes #9040


frontend/www/js/omegaup/components/common/NavbarItems.vue
```diff
@@ -112,6 +112,9 @@
             <a class="dropdown-item" href="/problem/" data-nav-problems-list>{{
               T.navViewProblemsAll
             }}</a>
+            <a class="dropdown-item" href="/profile/#problems">
+              Bookmarked Problems
```
pabo99: Can you add this string to the localization files?

All text that will be displayed in the UI must be translated into all the languages we support.

pabo99:
One more thing: could you please link this PR to its corresponding issue using: Fixes: #(issueID) in the description ?

frontend/www/js/omegaup/components/common/NavbarItems.vue
Prince-Prajapati07: Done. Added the label to the localization files and updated the navbar to use the i18n key. Thanks!

Prince-Prajapati07:
Thanks @pabo99 sir!. I like that idea, it feels more discoverable and consistent with the existing collections UX.

To keep this pr focused, I’d prefer to land the shortcut + localization first. After that, I’d be happy to open a follow-up pr exploring the collections-based approach if that works for you.

pabo99:
> Thanks @pabo99 sir!. I like that idea, it feels more discoverable and consistent with the existing collections UX.
> 
> To keep this pr focused, I’d prefer to land the shortcut + localization first. After that, I’d be happy to open a follow-up pr exploring the collections-based approach if that works for you.

That sounds good.

As a final suggestion, you could fix the linter errors by running `./stuff/lint.sh` outside the container, but within the WSL instance if you're using Windows

Prince-Prajapati07:
> > Thanks @pabo99 sir!. I like that idea, it feels more discoverable and consistent with the existing collections UX.
> > To keep this pr focused, I’d prefer to land the shortcut + localization first. After that, I’d be happy to open a follow-up pr exploring the collections-based approach if that works for you.
> 
> That sounds good.
> 
> As a final suggestion, you could fix the linter errors by running `./stuff/lint.sh` outside the container, but within the WSL instance if you're using Windows

@pabo99 sir! I’ve fixed the linter errors and pushed the changes.
Please let me know if everything looks good now or if you’d like any further tweaks.

Issue #8921, opened 2026-01-30
[FEATURE] Show school enrollment history in user profile


pabo99:
As a student, I want to see my school enrollment history to keep a record of my past and current schools

## Description

In the user profile, under the **_Manage schools_** section, it should be possible to view the user’s school enrollment history.

This information can be displayed in a table including:

- School name
- School grade
- Graduation date

The database already supports this information.

## Current behavior

Currently, the UI only allows users to add a new school, but historical information is not displayed anywhere

anuragdubey137:
hey @pabo99 
Currently, the scholar_degree is not included in the generated VO (IdentitiesSchools.php) or the schema. Because of this, the frontend cannot read or save the school grade yet ,any attempt to do so would fail or be ignored.

For now, should I raise a PR that only displays the school name and graduation date, so we can move forward with the feature while leaving scholar_degree support for a future update when the VO/schema is updated.

pabo99:
Thanks for clarifying! That approach makes sense.

Moving forward with a PR that displays only the school name and graduation date will allow us to unblock the feature for now. Please go ahead with the PR.

pabo99:
I’m also trying to find a solution for the issue you identified. Since the `IdentitiesSchools` table doesn't currently include that field, I had to review the DB schema and found the following problem:
- `Identities` table stores the information about the current school (`current_identity_school_id`).
- `Users` table stores the `scholar_degree`.

The `scholar_degree` should be moved to the `IdentitiesSchools` table, and the way to obtain the `current_scholar_degree` should be based on the `current_identity_school_id` in the `Identities` table, so we don’t need to add an extra field.

The real issue is what to do with the existing `scholar_degree` field in `Users`, since we don't yet have functionality to remove fields from the DAOs.

The approach will be to leave this field unused and inactive until we implement the feature mentioned above.

anuragdubey137:
hey @pabo99 

Thanks for the detailed explanation! I understand the situation now:

---> The IdentitiesSchools table currently doesn’t include scholar_degree.
---> The Identities table stores the current_identity_school_id, and the Users table currently stores scholar_degree.
--->The long-term approach is to move scholar_degree to IdentitiesSchools and obtain the current value via current_identity_school_id.
For now, the existing scholar_degree field in Users will remain unused and inactive.

I’m moving forward with a PR that displays only the school name and graduation date, so we can unblock this feature while leaving school grade support for a future update.

PR #9061, opened 2026-02-12
Show school enrollment history in user profile (#8921)

anuragdubey137:
# Description

This PR adds a school enrollment history table under the **"Manage Schools"** section in the user profile.

The table displays the following information for each enrollment:

- School name  
- Graduation date  

The backend was updated to return the enrollment history, and the frontend was modified to render the information in a structured table format.

# Comments

This PR only includes the minimal required changes to display:

- School name  
- Graduation date  

No additional fields were added to keep the implementation focused and aligned with the issue scope.

Screenshot from 2026-02-13 00-09-41
Fixes: #8921


frontend/server/src/DAO/IdentitiesSchools.php
```diff
@@ -74,4 +74,29 @@ public static function getByIdentityAndSchoolId(
...
+            FROM
+                Identities_Schools iss
+            INNER JOIN
+                Schools s ON s.school_id = iss.school_id
+            WHERE
+                iss.identity_id = ?
+            ORDER BY
+                iss.creation_time DESC';
+
+        return \OmegaUp\MySQLConnection::getInstance()->GetAll(
```
pabo99, line 97: It looks like Psalm is complaining because this function isn’t declaring its type correctly (in fact, it isn’t declaring it at all). Could you add it?
```suggestion
        /** @var list<array{graduation_date: null|string, identity_school_id: int, school_name: string}> */
        return \OmegaUp\MySQLConnection::getInstance()->GetAll(
```

frontend/server/src/DAO/IdentitiesSchools.php
```diff
@@ -74,4 +74,29 @@ public static function getByIdentityAndSchoolId(
 
         return new \OmegaUp\DAO\VO\IdentitiesSchools($rs);
     }
+    public static function getByIdentity(
+        \OmegaUp\DAO\VO\Identities $identity
+    ): array {
```
pabo99, line 79: Here we also prefer to declare the return type as explicitly as possible, instead of simply using an array, which is very ambiguous:

```suggestion

    /**
     * @return list<array{graduation_date: null|string, identity_school_id: int, school_name: string}>
     */
    public static function getByIdentity(
        \OmegaUp\DAO\VO\Identities $identity
    ) {
```

frontend/server/src/Controllers/User.php
```diff
@@ -1660,13 +1660,17 @@ public static function getUserProfile(
...
+        );
         return array_merge(
             $response,
             [
                 'classname' => \OmegaUp\DAO\Users::getRankingClassName(
                     $identity->user_id
                 ),
-                'programming_languages' => \OmegaUp\Controllers\User::getSupportedProgrammingLanguages()
+                'programming_languages' => \OmegaUp\Controllers\User::getSupportedProgrammingLanguages(),
+                'school_history' => $schoolHistory,
```
pabo99, line 1677: With this change, you’ll need to add `school_history` to the `UserProfileInfo` type declaration and then run` ./stuff/lint.sh` so that this propagates to the TypeScript types and the Cypress test errors stop appearing

frontend/www/js/omegaup/components/user/ManageSchools.vue
```diff
@@ -1,65 +1,84 @@
...
+          :required="false"
+          :enabled="!isCurrentlyEnrolled"
+        ></omegaup-datepicker>
+      </div>
+      <div class="mt-3">
+        <button
+          type="submit"
+          class="btn btn-primary mr-2"
+          data-save-school-changes
+        >
```
pabo99, line 58: Now that the table with the school history is being added, could you make it so that when submitting, the table gets updated?

frontend/www/js/omegaup/components/user/ManageSchools.vue
```diff
@@ -1,65 +1,84 @@
...
+          <option value="master">{{ T.userEditMaster }}</option>
+          <option value="doctorate">{{ T.userEditDoctorate }}</option>
+        </select>
+      </div>
+      <div class="form-group">
+        <label>{{ T.userEditManageSchoolsUserCurrentlyEnrolled }}</label>
+        <omegaup-radio-switch
+          :value.sync="isCurrentlyEnrolled"
+          :selected-value="isCurrentlyEnrolled"
+        ></omegaup-radio-switch>
```
pabo99, line 43: I’m thinking it might look more appealing to use the `ToggleSwitch` component here, where the active switch indicates that the user is currently enrolled. Could you try this change?

Also, another thing I noticed: if you already have a school registered with a graduation date, and you go to add a new one where you’re currently enrolled, when you change the value in the current enrolled field, the graduation date doesn’t reset, and the record is created with the previously selected graduation date. Could you make it reset when this value is changed?

frontend/www/js/omegaup/components/user/ManageSchools.vue
```diff
@@ -1,65 +1,84 @@
...
+          class="btn btn-primary mr-2"
+          data-save-school-changes
+        >
+          {{ T.wordsSaveChanges }}
+        </button>
+        <a href="/profile/" class="btn btn-cancel">{{ T.wordsCancel }}</a>
+      </div>
+    </form>
+    <div v-if="schoolHistory.length" class="card-body mt-4">
+      <h5>School Enrollment History</h5>
```
pabo99, line 65: Could you add this string to the localization files? That way, it will be displayed in the preferred language of the logged-in user

frontend/www/js/omegaup/components/user/ManageSchools.vue
```diff
@@ -1,65 +1,84 @@
...
+      <table class="table table-striped">
+        <thead>
+          <tr>
+            <th>School</th>
+            <th>Graduation Date</th>
```
pabo99, line 70: These strings should also be added to the localization files

frontend/www/js/omegaup/components/user/ManageSchools.vue
```diff
@@ -1,65 +1,84 @@
...
+          </tr>
+        </thead>
+        <tbody>
+          <tr v-for="school in schoolHistory" :key="school.identity_school_id">
+            <td>{{ school.school_name }}</td>
+            <td>{{ school.graduation_date || '—' }}</td>
```
pabo99, line 76: Could you also add the field for scholar degree? It would also be good to indicate somewhere, or in some way, which school is the current one

frontend/www/js/omegaup/components/user/ManageSchools.vue
```diff
@@ -81,7 +100,9 @@ import OmegaupRadioSwitch from '../RadioSwitch.vue';
 export default class UserManageSchools extends Vue {
   @Prop() profile!: types.UserProfileInfo;
   @Prop() searchResultSchools!: types.SchoolListItem[];
-
+  get schoolHistory() {
+    return this.profile.school_history ?? [];
+  }
```
pabo99, line 105: To keep things organized, could you move this computed property after `isCurrentlyEnrolled`? It would also be good to leave a blank space at the beginning and at the end of it

anuragdubey137:
Hey @pabo99, thank you very much for your detailed feedback! I really appreciate all the suggestions and will go ahead and implement the fixes you mentioned.
Regarding the scholar_degree field, it’s currently not displayed because, as discussed earlier, the IdentitiesSchools VO/schema does not yet include this field. Since the frontend cannot read or save it until the backend is updated, it is not visible at the moment.

pabo99:
I understand the situation with the `scholar_degree` field. Perhaps we could add the column to the table and only store the information in the current school with the degree that is already saved in the DB. Does that sound good to you?

PR #9067, opened 2026-02-12
feat: Add rate limiting to content creation endpoints (#8975)

### LINKED ISSUE #8975: [FEATURE] Add rate limiting to content creation endpoints (courses, contests, problems)
As a **platform user**, I want **rate limiting on content creation endpoints** so that **the system is protected from resource exhaustion attacks caused by users rapidly creating large volumes of content**.

---
## Context

While reviewing the codebase, I identified that authenticated users can create **unlimited courses, contests, and problems** without any rate limiting or throttling mechanism.
**Current State:**
-  Authentication required (`ensureIdentity()`)
-  Age verification (`ensureMainUserIdentityIsOver13()`)
-  Submissions have rate limiting (`submissions_gap` = 60 seconds)
-  **No rate limiting on content creation**

**Potential Attack Scenario:**
1. Attacker creates one legitimate account (passes CAPTCHA once)
2. Runs a script to create thousands of courses/contests/problems
3. Uses unique aliases to bypass UNIQUE constraints
4. Results in database bloat, performance degradation, increased storage costs
**Affected Endpoints:**

| Endpoint | File | Has Rate Limit? |
|----------|------|-----------------|
| `Course::apiCreate` | `Course.php:836` |  No |
| `Contest::apiCreate` | `Contest.php:2650` |  No |
| `Problem::apiCreate` | `Problem.php:429` |  No |
---

## Proposed Solution

Add rate limiting to content creation endpoints:
**Suggested Limits:**
- Courses: **5 per hour** per user
- Contests: **10 per hour** per user
- Problems: **20 per hour** per user

**Implementation Options:**
1. **Database tracking table** - Log creations and query before allowing new ones
2. **Redis/Memcached** - Fast in-memory rate limiting with auto-expiry
---

## Additional Notes
- This follows industry best practices (GitHub, Twitter, AWS all have rate limits)
- Minimal impact on legitimate users (most don't create >5 courses/hour)
- I'm willing to contribute a PR to implement this if approved

Yashagarwal9798:
# Description

Implements rate limiting for content creation API endpoints to prevent abuse and resource exhaustion. A new reusable `RateLimiter` class uses Redis-backed counters with fixed-window hourly bucketing to enforce per-user limits on the following endpoints:

| Endpoint | Limit |
|---|---|
| `Course::apiCreate` | 5 per hour |
| `Contest::apiCreate` | 10 per hour |
| `Problem::apiCreate` | 20 per hour |
| `Group::apiCreate` | 5 per hour |
| `TeamsGroup::apiCreate` | 5 per hour |
| `School::apiCreate` | 5 per hour |

### Key Design Decisions

- **Fixed-window bucketing**: Uses `intdiv(Time::get(), 3600)` to create hourly time buckets. Simple and effective.
- **Per-user tracking**: Rate limits are keyed by `identity_id`, so each authenticated user has independent counters per endpoint.
- **Fail-open**: If the cache backend is unavailable, requests are allowed through rather than blocked.
- **429 response**: Returns `RateLimitExceededException` (HTTP 429) with a translated error message when the limit is exceeded.
- **Cache TTL**: Counters expire after 7200 seconds (2 hours) to ensure cleanup even if a window boundary is crossed.

### Files Changed

- **`frontend/server/src/RateLimiter.php`** (new): Reusable rate limiter class with `assertWithinLimit()` static method.
- **`frontend/server/src/Controllers/{Course,Contest,Problem,Group,TeamsGroup,School}.php`**: Added rate limit checks in `apiCreate` methods, placed after authentication but before request processing.
- **`frontend/templates/{en,es,pt,pseudo}.lang`**: Added `contentCreationRateLimitExceeded` translation string.
- **`frontend/tests/controllers/RateLimiterTest.php`** (new): Unit tests covering all 6 endpoints, window reset behavior, and independent counter isolation.

Fixes #8975

# Comments

- Rate limit checks are placed after `ensureIdentity()` / `ensureMainUserIdentityIsOver13()` and `ensureNotInLockdown()` so that unauthenticated or lockdown requests are rejected before hitting rate limit logic.
- The `RateLimiter` uses `\OmegaUp\Cache` (not `CacheAdapter` directly) to ensure proper PSR-4 autoloading compatibility.
- Limits are conservative and can be adjusted per endpoint by changing the integer passed to `assertWithinLimit()`.

# Checklist:


Yashagarwal9798:
@pabo99 sir,
### CI Test Failures — Rate Limiting vs. Existing Tests

Several existing tests create more items than the new rate limits allow within a single test method, causing `RateLimitExceededException`:

| Test | Endpoint | Items Created | Limit |
|---|---|---|---|
| `ContestAddProblemTest::testAddTooManyProblemsToContest` | `Problem::apiCreate` | >20 | 20/hr |
| `ContestUpdateTest::testUpdateWhenTooManyProblemsInContest` | `Problem::apiCreate` | >20 | 20/hr |
| `SchoolCreateTest::testSchoolsList` (8 data sets) | `School::apiCreate` | >5 | 5/hr |
| `TeamGroupsTest::testTeamsGroupsListForTypeahead` (7 data sets) | `TeamsGroup::apiCreate` | >5 | 5/hr |

### Solution Applied

I added a static toggle `RateLimiter::setForTesting(bool $enabled)` that:
1. **Disables** rate limiting by default in `ControllerTestCase::setUp()` — so all existing tests continue to pass without modification.
2. **Re-enables** rate limiting only in `RateLimiterTest::setUp()` — where rate limiting behavior is explicitly tested.

This way existing tests are unaffected, and the rate limiter is still fully tested in isolation.

### Question for maintainers

Is this approach acceptable, or would you prefer that I also modify the existing tests (e.g., by having them use multiple users or reset the rate limit counter between calls) so that rate limiting stays globally enabled during all tests?

pabo99:
Before the code review, I want to ask what we can do regarding the execution of `./stuff/bootstrap-environment.py --purge`, since I'm noticing that it creates 29 consecutive contests by the same user (single token).

If you find a solution, please document the steps you took

Yashagarwal9798:
@pabo99 sir! I checked `stuff/bootstrap.json` and confirmed it creates the following rate-limited resources under a single session:

| Endpoint | Count | Rate Limit |
|---|---|---|
| `/contest/create/` | 29 | 10/hr |
| `/problem/create/` | 12 | 20/hr |
| `/course/create/` | 2 | 5/hr |
| `/school/create/` | 1 | 5/hr |

This would indeed exceed the rate limits. However, this is already handled — **system administrators are exempt from rate limiting**.

The bootstrap script authenticates via an API token (`92d8c5a...`) which is tied to `user_id=1` (the `omegaup` system admin), as defined in [`frontend/database/00178_test_apitokens.sql`](https://github.com/omegaup/omegaup/blob/main/frontend/database/00178_test_apitokens.sql#L4). Since this user is a system admin, the bypass kicks in at [`frontend/server/src/RateLimiter.php` lines 62-64](https://github.com/Yashagarwal9798/omegaup/blob/feature/rate-limit-content-creation-8975/frontend/server/src/RateLimiter.php#L62-L64):

```php
// System administrators are exempt from rate limiting
if (\OmegaUp\Authorization::isSystemAdmin($identity)) {
    return;
}
```
This check runs inside `assertWithinLimit()` before any counter logic, so the admin's requests are never counted and never blocked. The `bootstrap-environment.py --purge` flow will continue to work without hitting any rate limits.

frontend/server/src/Controllers/Contest.php
```diff
@@ -2943,6 +2943,20 @@ public static function apiCreate(\OmegaUp\Request $r) {
         // Authenticate user
         $r->ensureMainUserIdentityIsOver13();
 
+        // Rate limit: 10 contest creations per hour per user.
+        // System admins are exempt.
+        if (
+            !\OmegaUp\Authorization::isSystemAdmin(
+                $r->identity
+            )
```
pabo99: This validation is also performed inside the `assertWithinLimit` function. Could we remove this one in favor of the validation already handled there?

The same would apply to the other places where `assertWithinLimit` is called.

frontend/server/src/Controllers/Contest.php
```diff
@@ -2943,6 +2943,20 @@ public static function apiCreate(\OmegaUp\Request $r) {
...
+        ) {
+            \OmegaUp\RateLimiter::assertWithinLimit(
+                'Contest::apiCreate',
```
pabo99: It looks like the first parameter always matches the name of the endpoint that invokes the function. How about deriving it dynamically inside `assertWithinLimit` instead?

Something like:

```php
[
    'class' => $callerClass,
    'function' => $callerFunction,
    'type' => $callerType,
] = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 2)[1];

$fullClass = $callerClass ?? 'Unknown';
$shortClass = substr($fullClass, strrpos($fullClass, '\\') + 1);
$endpoint = "{$shortClass}{$callerType}{$callerFunction}";
```

This would help avoid potential typos and keep the endpoint name automatically in sync with the caller.

frontend/server/src/Controllers/Contest.php
```diff
@@ -2943,6 +2943,20 @@ public static function apiCreate(\OmegaUp\Request $r) {
...
+                $r->identity,
+                10
```
pabo99: You could also make the `$limit` parameter optional by centralizing the default values instead. For example, you could create a constant in `RateLimiter` containing the limits you're currently passing explicitly:

```php
private const DEFAULT_LIMITS = [
    'Course::apiCreate'     => 5,
    'Contest::apiCreate'    => 10,
    'Problem::apiCreate'    => 20,
    'Group::apiCreate'      => 5,
    'TeamsGroup::apiCreate' => 5,
    'School::apiCreate'     => 5,
];
```

Then inside `assertWithinLimit` you could simply do:

```php
$limit = $limit ?? (self::DEFAULT_LIMITS[$endpoint] ?? 10);
```

This would reduce duplication and make it easier to manage the default limits in a single place.

pabo99 (commented):
The change looks good. I left a few suggestions that could make the code slightly more readable

pabo99 (approved):

PR #9069, opened 2026-02-13
Fix pre-push hook false positives when merging from main

### LINKED ISSUE #9053: [BUG] Pre-push validates files from external merges
## Current Behavior

When you do `git fetch` + `git merge` to bring changes from other PRs that have already been approved and are in the main branch, the pre-push hook fails with the error:

```bash
'frontend/database/00258_add_social_media_fields_to_users.sql' already exists in the remote repository and cannot be modified.

husky - pre-push script failed (code 1)
```

Even if you did not modify that file.

## Root Cause

The pre-push hook (`stuff/git-hooks/pre-push` and `stuff/database_schema.py`) calculates the modified files using:
https://github.com/omegaup/omegaup/blob/e1947b4919bb973f739fe9ce81709439ae26706d/stuff/database_schema.py#L64-L91

This executes: `git diff --name-only origin/main`

Which includes ALL files changed since `origin/main`, including:
- Files you modified in your commits
- Files that came from merges of other PRs

Then `database_schema.py` applies this validation:
https://github.com/omegaup/omegaup/blob/e1947b4919bb973f739fe9ce81709439ae26706d/stuff/database_schema.py#L148-L163

Which forbids modifying migration scripts that already exist in the remote repository.

## Steps to reproduce

1. You merge main, which includes commits with `00258_add_social_media_fields_to_users.sql`
2. You try to push your branch
3. Pre-push calculates: `git diff --name-only origin/main`
4. It includes `00258_add_social_media_fields_to_users.sql` (even though it came from the merge, not from you)
5. `check_extra_validations()` sees that the file exists in remote
6. Push fails even though you did not touch that file

viveksahu92:
Description
This PR fixes Issue #9053, where the pre-push hook was incorrectly flagging files during validation after a developer merged changes from the main branch.

The Problem
Previously, the pre-push hook would fail with validation errors on files that the developer hadn't even touched. This happened because the script was comparing the local branch directly against origin/main.

If main contained migration files or changes from other merged PRs, the hook mistakenly treated those as "newly modified" by the current developer, leading to confusing "false positive" errors during a push.

The Fix
I have updated the logic in stuff/database_schema.py to use a Merge Base approach.

Instead of a direct comparison with the remote branch, the script now:

Finds the common ancestor (the point where your branch originally split from main).

Runs the git diff against that specific ancestor point.

Key Changes:

Added a step to calculate the merge-base between HEAD and the upstream_branch.

Updated the validation to only scan files actually modified within the current branch.

Impact
Developers can now fetch and merge main into their local branches without triggering unnecessary validation errors. The pre-push hook remains strict for your actual changes but ignores files that were already part of the upstream repository.

How to Test
Checkout a feature branch.

Fetch and merge the latest changes: git fetch origin main && git merge origin/main.

Attempt to push your branch.

Result: The hook should only validate your specific changes and ignore the merged files from main.

Fixes: #9053


viveksahu92:
Hi @pabo99 , could you review this pull request ?

stuff/database_schema.py
```diff
@@ -62,8 +62,9 @@ def _check_mutually_exclusive_schema_modifications(
 
 
 def get_modified_files(root: str) -> Set[str]:
-    '''Get the list of modified files in the current commit.'''
-    merge_base = subprocess.run(
+    '''Get the list of modified files in the current branch relative to the
+    upstream merge base.'''
+    upstream_branch = subprocess.run(
```
pabo99, line 67: Since omegaUp's development workflow allows contributors to use either `origin` or `upstream` for the official repository, it is necessary to automatically detect which remote points to `omegaup/omegaup`. This ensures the script works regardless of the user's git remote configuration. For example:
```py
base_branch = None
    for remote_name in ['upstream', 'origin']:
        remote_url_process = subprocess.run(
            [
                '/usr/bin/git',
                'remote',
                'get-url',
                remote_name,
            ],
            check=False,
            universal_newlines=True,
            stdout=subprocess.PIPE,
            stderr=subprocess.PIPE,
            cwd=root,
        )
        if remote_url_process.returncode == 0:
            remote_url = remote_url_process.stdout.strip()
            # Check if this remote points to omegaup/omegaup
            if 'omegaup/omegaup' in remote_url:
                base_branch = f'{remote_name}/main'
                break
    
    # If we couldn't find the official repo, fall back to origin/main
    if not base_branch:
        base_branch = 'origin/main'
```
The fallback to `origin/main` remains in case the official repository cannot be detected

stuff/database_schema.py
```diff
@@ -76,6 +77,54 @@ def get_modified_files(root: str) -> Set[str]:
...
+            )
+            sys.exit(1)
+        print(
+            (f'{git_tools.COLORS.HEADER}Warning: could not determine '
+             f'merge-base with upstream {upstream_branch!r}; falling back '
+             f'to direct diff against {upstream_branch!r}.'
+             f'{git_tools.COLORS.NORMAL}'),
+            file=sys.stderr,
+        )
+        merge_base = upstream_branch
```
pabo99: Here, it might only be necessary to run the verification process once we already have the `base_branch` defined:
```suggestion
    verify_process = subprocess.run(
        [
            '/usr/bin/git',
            'rev-parse',
            '--verify',
            base_branch,
        ],
        check=False,
        universal_newlines=True,
        stdout=subprocess.PIPE,
        stderr=subprocess.PIPE,
        cwd=root,
    )
    
    if verify_process.returncode != 0:
        print(
            (f'{git_tools.COLORS.FAIL}Error: could not resolve '
             f'base branch {base_branch!r}. Please run `git fetch` to '
             f'update remotes, then retry.{git_tools.COLORS.NORMAL}'),
            file=sys.stderr,
        )
        sys.exit(1)
```

stuff/database_schema.py
pabo99: Here, it will be necessary to use the triple-dot syntax (`upstream/main...HEAD`), since it automatically compares only the changes introduced by the current branch, excluding any modifications that came from merges with the base branch. This is exactly what we need: to validate only what this PR adds, not what was merged from elsewhere:
```py
                f'{base_branch}...HEAD',
```

pabo99, line 261: At this point, we can also filter the files present in the `modified_files` object

pabo99 (commented):
This PR addresses false positives in validations when working with long-lived branches updated via merges from main.

The issue arises because merged files are incorrectly validated as if they were introduced by the feature branch.

The proposed solution compares only against the official main branch, so validations detect real branch changes and ignore merge noise

viveksahu92:
Hey @pabo99, really sorry for the delay! I was caught up with some unexpected family stuff.

I just pushed the changes. I applied the git merge-base fix exactly as you suggested to 
stuff/database_schema.py

Everything should be working perfectly now.
Let me know if anything else is needed.

pabo99:
Hi @viveksahu92 , thanks for the update! I hope everything is going well.

As of today, the GSoC selection process has progressed, and we’re prioritizing PR reviews for users who have completed Phase 1. If you’re interested in participating, it’ll be important to complete the [omegaUp test](https://omegaup.com/arena/gsoc2026/) so your PR reviews can proceed more quickly.

viveksahu92:
> Hi @viveksahu92 , thanks for the update! I hope everything is going well.
> 
> As of today, the GSoC selection process has progressed, and we’re prioritizing PR reviews for users who have completed Phase 1. If you’re interested in participating, it’ll be important to complete the [omegaUp test](https://omegaup.com/arena/gsoc2026/) so your PR reviews can proceed more quickly.

Thankyou for giving this information because i want to participate in GSoc .

viveksahu92:
Hi @pabo99 ,The omegaUp test is done (3/3 problems solved) by user name is viveksahu92! Ready for a PR review when you have a time."
image

stuff/database_schema.py
pabo99: This suggestion has not been addressed yet

stuff/database_schema.py
pabo99: This hasn’t been addressed either

viveksahu92:
"Hi @pabo99 , could you review this PR again for merging" ?

stuff/database_schema.py
```diff
@@ -84,11 +133,12 @@ def get_modified_files(root: str) -> Set[str]:
                 'diff',
                 '-z',
                 '--name-only',
-                merge_base,
+                '--diff-filter=d',
+                f'{merge_base}...HEAD',
```
pabo99: Nice improvement overall. 

To make the fallback a true direct diff, when `merge-base` fails we should use `..` instead of `...`

A clean way to do it is:

```py
diff_target = f'{merge_base}...HEAD'
```

Then, in the fallback path, instead of::

```py
merge_base = upstream_branch
```

update `diff_target` to:
```py
        diff_target = f'{upstream_branch}..HEAD'
```

And reuse it here:

```suggestion
                diff_target,
```

stuff/database_schema.py
```diff
@@ -76,6 +77,54 @@ def get_modified_files(root: str) -> Set[str]:
...
+                 f'upstream {upstream_branch!r}. Please run `git fetch` to '
+                 f'update remotes, then retry.{git_tools.COLORS.NORMAL}'),
+                file=sys.stderr,
+            )
+            sys.exit(1)
+        print(
+            (f'{git_tools.COLORS.HEADER}Warning: could not determine '
+             f'merge-base with upstream {upstream_branch!r}; falling back '
+             f'to direct diff against {upstream_branch!r}.'
+             f'{git_tools.COLORS.NORMAL}'),
```
pabo99, line 124: Nit: Since this warning says “falling back to direct diff”, it’s great to keep it aligned with the actual fallback behavior (`<upstream_branch>..HEAD`). Maybe you can also include the exact ref in the message for easier debugging.
```suggestion
            (f'{git_tools.COLORS.HEADER}Warning: could not determine '
             f'merge-base with upstream {upstream_branch!r}; falling back '
             f'to direct diff target {diff_target!r}.'
             f'{git_tools.COLORS.NORMAL}'),
```

pabo99 (approved):
Great!

PR #9073, opened 2026-02-13
fix: add ensureNotInLockdown() to Problem::apiCreate (#9062)

### LINKED ISSUE #9062: [BUG] Problem::apiCreate bypasses lockdown restrictions (missing ensureNotInLockdown)
## Current Behavior
The `Problem::apiCreate` endpoint allows authenticated users to create new problems regardless of whether the controller is in "Lockdown" mode.

Upon code analysis of `frontend/server/src/Controllers/Problem.php`, the function `apiCreate` only validates that the user is over 13:

```php
public static function apiCreate(\OmegaUp\Request $r): array {
    $r->ensureMainUserIdentityIsOver13();
    // ...
}
```
It is missing the \OmegaUp\Controllers\Controller::ensureNotInLockdown(); check.
In contrast, other creation endpoints correctly enforce this:
Course::apiCreate calls ensureNotInLockdown()
Contest::apiCreate calls ensureNotInLockdown()

## Expected Behavior
Problem::apiCreate should be consistent with other core entity creation endpoints. It should enforce ensureNotInLockdown() immediately upon execution to prevent problem creation during lockdown periods (e.g., during high-load Olympiads or maintenance).

## Steps to Reproduce
1. Inspect frontend/server/src/Controllers/Problem.php.
2. Observe the apiCreate method.
3. Compare with frontend/server/src/Controllers/Course.php's apiCreate method.
4. Notice the missing static call to ensureNotInLockdown().

## Relevant details
Component: Controllers / API
Affected File: frontend/server/src/Controllers/Problem.php

Yashagarwal9798:
# Description

Add the missing `\OmegaUp\Controllers\Controller::ensureNotInLockdown()` check to `Problem::apiCreate` to enforce lockdown restrictions during problem creation.

Currently, `Problem::apiCreate` allows authenticated users to create new problems even when the system is in lockdown mode. Other core entity creation endpoints (`Course::apiCreate`, `Contest::apiCreate`) already enforce this check. This PR makes `Problem::apiCreate` consistent with them by adding the lockdown guard as the very first validation, matching the established pattern.

Fixes #9062

# Comments

This is a one-line fix. The `ensureNotInLockdown()` call is placed before `ensureMainUserIdentityIsOver13()`, matching the order used in `Course::apiCreate` and `Contest::apiCreate`.

# Checklist:


pabo99 (approved):
In the future, I’d like us to have all our entrypoints mapped—those that should be accessible in lockdown mode and those that shouldn’t—so we can call this function directly within the `render` function, rather than having to add it everywhere.

We could also extend the mapping to cover scenarios such as when the user is not logged in, or when the user is not an admin, among others.

Yashagarwal9798:
Thanks for the review and the approval, @pabo99 sir!

I completely agree having a centralized mapping for these checks (lockdown, auth, admin status) within render would definitely be cleaner and less error-prone than manually adding guards to every controller. It would prevent missing checks like this one in the future.

PR #9092, opened 2026-02-15
Optimize Problems::byIdentityType slow query

### LINKED ISSUE #9091: [BUG] Optimize slow Problems::byIdentityType query
## Summary

The SQL query tracker flagged a slow query originating from `DAO/Problems.php` (~line 291, `byIdentityType`). The EXPLAIN output showed:
- **type: ALL** (full table scan on Problems)
- **Extra: Using where; Using temporary; Using filesort**
- Complex visibility checks, tag filters, FIND_IN_SET for languages, ORDER BY submissions DESC

## Root Cause

1. **Missing index for ORDER BY**: No index supports `ORDER BY submissions DESC` combined with visibility filtering, forcing MySQL to materialize and sort.
2. **Tag filter subquery**: The `ptp` INNER JOIN used `WHERE pt.tag_id IN (SELECT t.tag_id FROM Tags t WHERE t.name IN (...))` — a correlated subquery that can prevent optimal join ordering.

## Proposed Optimizations

1. **Add composite index** `idx_problems_visibility_submissions (visibility, submissions DESC)` — Enables index-driven ordering when filtering by visibility.
2. **Simplify tag filter** — Replace `IN (subselect)` with a direct JOIN predicate `Tags t ON pt.tag_id = t.tag_id AND t.name IN (?)` for better plan choices.

## Limitations / Cannot Optimize Further (Without Schema Change)

- **FIND_IN_SET(?, p.languages)** — Non-sargable; no index can help. Normalizing to a `Problems_Programming_Languages` junction table would require a larger schema migration.
- **LIKE CONCAT('%', ?, '%')** — Prefix wildcards prevent index use. Full-text search could help but would change semantics.

antedotee:
- Add idx_problems_visibility_submissions index to support ORDER BY submissions with visibility filter, reducing full table scan and filesort
- Replace tag filter IN (subselect) with direct JOIN on Tags for better query plan optimization

### 1. Add index `idx_problems_visibility_submissions`

Composite index on `(visibility, submissions DESC)` to support:
- Visibility range filter (`p.visibility >= ?`)
- `ORDER BY submissions DESC`
- Reduces full table scan and filesort when both are used.

**EXPLAIN before (without index):**
```sql
EXPLAIN SELECT p.problem_id, p.submissions FROM Problems p
WHERE p.visibility >= 1 ORDER BY p.submissions DESC LIMIT 20;
```
Expected: `type=ALL`, `key=NULL`, `Extra: Using where; Using temporary; Using filesort`

**EXPLAIN after (with index):**
```sql
EXPLAIN SELECT p.problem_id, p.submissions FROM Problems p
WHERE p.visibility >= 1 ORDER BY p.submissions DESC LIMIT 20;
```
Expected: `type=range`, `key=idx_problems_visibility_submissions`, no filesort in Extra

### 2. Optimize tag filter in `addTagFilter`

**File:** `frontend/server/src/DAO/Problems.php`

Replaced:
```sql
WHERE pt.tag_id IN (SELECT t.tag_id FROM Tags t WHERE t.name IN (?))
```
with:
```sql
INNER JOIN Tags t ON pt.tag_id = t.tag_id AND t.name IN (?)
```

This removes the subquery and lets the optimizer choose a better join order (Tags has `UNIQUE KEY tag_name`, so lookup by name is indexed).

## What was not optimized (and why)

- **FIND_IN_SET(?, p.languages)** — Non-sargable; would require a new `Problems_Programming_Languages` table to fix.
- **LIKE CONCAT('%', ?, '%')** on title/alias — Leading wildcards prevent index use; full-text would change behavior.

Fixes #9091

### DIFF
--- a/frontend/database/00262_add_index_problems_visibility_submissions.sql
+++ b/frontend/database/00262_add_index_problems_visibility_submissions.sql
@@ -1,4 +1,4 @@
 -- Add composite index to optimize Problems::byIdentityType query
 -- Supports ORDER BY submissions DESC with visibility filter, reducing filesort and full scan
--- Resolves slow query in DAO/Problems.php byIdentityType (EXPLAIN: Using temporary; Using filesort; type: ALL)
+-- Resolves slow query in DAO/Problems.php byIdentityType
 CREATE INDEX idx_problems_visibility_submissions ON Problems (visibility, submissions DESC);


frontend/database/00262_add_index_problems_visibility_submissions.sql
```diff
@@ -0,0 +1,4 @@
+-- Add composite index to optimize Problems::byIdentityType query
```
pabo99, line 1: It looks like script `261` has already been used, and that’s causing some of the jobs in the CI workflow to fail. Could you update your branch and rename this file to avoid the errors?

antedotee: Yes, I got this in upcoming PRs and fixed it in one. Will fix in this one too.

PR #9094, opened 2026-02-15
Optimizes the problem list query when filtering by tags by removing redundant subquery

### LINKED ISSUE #9093: [BUG] Optimize query (`Problems::byIdentityType`)
## Problem
The problem list query (`Problems::byIdentityType`) shows slow performance with tag filters:
- EXPLAIN reports: `Using where; Using temporary; Using filesort`
- Full table scan on `Problems` (`type: ALL`)

## Root cause
In `addTagFilter`, the tag filter uses a redundant subquery:
WHERE pt.tag_id IN (SELECT t.tag_id FROM Tags t WHERE t.name in (?))

antedotee:
## Changes
- **`frontend/server/src/DAO/Problems.php`**: Replaced `WHERE pt.tag_id IN (SELECT t.tag_id FROM Tags t WHERE t.name in (...))` with `AND t.name IN (...)` in the existing `Tags` join.

## Rationale
- The `IN (subquery)` form was redundant because we already join `Tags t` on `pt.tag_id = t.tag_id`.
- Filtering by `t.name IN (...)` uses the `Tags.tag_name` (UNIQUE) index directly.
- Reduces work done for the tag filter in `Problems::byIdentityType`.

Fixes #9093

### DIFF
--- a/frontend/server/src/DAO/Problems.php
+++ b/frontend/server/src/DAO/Problems.php
@@ -28,6 +28,8 @@ class Problems extends \OmegaUp\DAO\Base\Problems {
         $havingClause = $requireAllTags ? 'HAVING (COUNT(pt.tag_id) = ?)' : '';
         $placeholders = array_fill(0, count($tags), '?');
         $placeholders = join(',', $placeholders);
+        // Use direct JOIN on Tags by name to avoid redundant subquery;
+        // Tags.name has UNIQUE index so the lookup is efficient.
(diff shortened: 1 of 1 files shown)


antedotee:
Closing it in the favor of #9092 and #9096 as the changes overlap with them and merging them will make it irrelevant. Will add a test in those PRs. 
Sorry, I will try to keep this in mind from next time.

Issue #9104, opened 2026-02-15
[BUG] External links use http:// instead of https:// in multiple Vue components


PrasannaMishra001:
## Current Behavior

Several Vue components use external links with `http://` instead of `https://`. This affects:

- World clock timestamp links (`timeanddate.com`)
- Blog link (`blog.omegaup.com`)
- Gravatar profile/avatar link (`gravatar.com`)

Although these endpoints correctly redirect to HTTPS, using `http://` causes unnecessary redirects and may lead to mixed-content warnings in secure contexts.

## Expected Behavior

All external links should use `https://` directly to ensure secure connections, avoid redirects, and follow best security practices.

## If necessary, add the following information (otherwise delete this section):

*Provide a link to an example, or a detailed list of steps to reproduce this bug.*

	1. Open the affected Vue components in the repository.
	2. Locate external links using `http://`.
	3. Observe that they redirect to HTTPS when accessed.
	4. Replace `http://` with `https://` to remove the redirect.

*Relevant details about the environment in which you encountered this bug:*

	* Browser name (e.g., Chrome 39, node.js 5.4): Chrome (latest)
	* Operating System and version (desktop or mobile): Ubuntu 22.04 (desktop)
	* Link to the site where this issue occurs: omegaUp frontend (local setup)
	* Screenshot: Not required (code-level issue)

PR #8115, opened 2025-03-28
Centralised URL update mechanism for all blog links on omeagup

SalmanSi:
# Description

Centralized blog URLs into blog-links-config.json and created a getBlogUrl helper.
Refactored components and translation strings (using ui.formatString) to dynamically load URLs, preventing broken links.

Fixes: #8063

# Comments

@pabo99 please check the changes and let me know if anything is required. Like you suggested, I used the ui.formatString fucntion to dynamically load the links into the lang files. I have tested the functionality manually and its working. Also, some links were broken like for user ranking methodoly, school ranking method, user categories etc.  I have updated them with the correct ones and put them all in the blog-links-config.json

# Checklist:

- [X] The code follows the [coding guidelines](https://github.com/omegaup/omegaup/wiki/Coding-guidelines) of omegaUp.
- [X] The tests were executed and all of them passed.
- [X] If you are creating a feature, the new tests were added.
- [X] If the change is large (> 200 lines), this PR was split into various Pull Requests. It's preferred to create one PR for changes in controllers + unit tests in PHPUnit,  and then another Pull Request for UI + tests in Jest, Cypress or both.

PR #9105, opened 2026-02-15
Use HTTPS for external links

PrasannaMishra001:
# Description

Replaces 6 external links using insecure `http://` with secure `https://` across multiple Vue components. These links point to external resources (`timeanddate.com`, `blog.omegaup.com`, and `www.gravatar.com`) and should use HTTPS to ensure encrypted connections and avoid mixed-content warnings in modern browsers.

All updated HTTPS endpoints were verified to respond correctly (HTTP 200). This change improves security and aligns with best practices for external resource linking.

Fixes: #9104

# Comments

This is a small security-focused cleanup. No functional or UI behavior changes were introduced. Only protocol updates were made, and all endpoints were validated before updating.

# Checklist:

### DIFF
--- a/.github/workflows/build-containers.yml
+++ b/.github/workflows/build-containers.yml
@@ -41,11 +41,11 @@ jobs:
       - name: Push containers to GitHub registry
         run: |
           echo "${{ github.token }}" | \
-            docker login https://docker.pkg.github.com "--username=${{ github.actor }}" --password-stdin
+            docker login ghcr.io "--username=${{ github.actor }}" --password-stdin
 
           for container in frontend frontend-sidecar nginx php ai-editorial-worker; do
             for tag in "latest-${{ steps.extract_branch.outputs.branch }}" "${{ github.sha }}"; do
-              container_name="docker.pkg.github.com/${{ github.repository }}/${container}:${tag}"
+              container_name="ghcr.io/${{ github.repository_owner }}/${container}:${tag}"
               docker tag "omegaup/${container}:${{ github.sha }}" "${container_name}"
               docker push "${container_name}"
             done
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -418,9 +418,12 @@ jobs:
           pytest stuff/teaching_assistant/test_ta_course_mode_pre.py -s
 
       - name: Run the teaching assistant for course mode
+        env:
+          OMEGAUP_PASSWORD: teacher123
+          OMEGAUP_LLM_KEY: omegaup
         run: |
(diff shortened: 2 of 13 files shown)


frontend/www/js/omegaup/components/common/NavbarItems.vue
```diff
@@ -217,7 +217,7 @@
           >
           <a
             class="dropdown-item"
-            href="http://blog.omegaup.com/"
+            href="https://blog.omegaup.com/"
```
pabo99: Ok, in issue #8115 a configuration file for links pointing to the blog was added precisely to avoid adding repeated URLs and reduce the chances of typos. 

You can use one of the already created indexes to place it here.

And now this pending comment makes even more sense:
https://github.com/omegaup/omegaup/pull/8115#discussion_r2019675236

Where we could also add the rule that all links must include the right protocol

frontend/www/js/omegaup/components/contest/FilteredList.vue
```diff
@@ -156,7 +156,7 @@ export default class FilteredList extends Vue {
   }
 
   getTimeLink(time: string): string {
-    return `http://timeanddate.com/worldclock/fixedtime.html?iso=${time}`;
+    return `https://timeanddate.com/worldclock/fixedtime.html?iso=${time}`;
```
pabo99: I’m thinking about whether we should extend that blog URL configuration file to all URLs used across the platform. For the purposes of this change, we could add this one and the Gravatar URL.
Let me know if you see this as viable, and if not, we can leave it as is

PrasannaMishra001: Thanks for the suggestion.
After reviewing #8115 and your comment there, I agree that extending the config to cover all external URLs is better and will make future updates easier too.

PrasannaMishra001:
@pabo99 Please confirm if I should continue implementing it in this PR and update the issue accordingly (works for me) or if you’d prefer a separate PR for this change.

PrasannaMishra001:
@pabo99 made the changes in config file to include all URLs .

pabo99:
> @pabo99 Please confirm if I should continue implementing it in this PR and update the issue accordingly (works for me) or if you’d prefer a separate PR for this change.

Oh, I didn’t get a chance to reply before you made the changes. We could have left that for another issue.

But that’s fine, I’d just prefer to handle it with grouped URLs, so it’s easier to identify the ones that belong to the blog. The intention is to keep the URLs updated, and whenever there’s a change in the blog, it can be reflected here. All of that we can leave for another change.

For now, this PR can already be approved.

Issue #9110, opened 2026-02-15
[BUG] Arena contest list creates extra browser history entries on tab/filter/sort changes


Yashagarwal9798:
## Current Behavior

When navigating the `/arena/` contest list page, every interaction that changes the view (switching between Current/Past/Future tabs, changing sort order, or applying filters) calls `window.history.pushState()`, which creates a **new entry in the browser history stack**.

This means:
- Clicking "Current" → "Past" → "Future" tabs creates 3 separate history entries, even though the user is on the **same page** just toggling views.
- All URL parameters (`tab_name`, `page`, `sort_order`, [filter](cci:1://file:///c:/Users/BIT/Desktop/omegaup/frontend/www/js/omegaup/components/arena/ContestListv2.vue:638:2-640:3)) are appended to the URL with each change.
- When the user presses the browser **Back button**, instead of navigating to the **previous page** (e.g., course list or homepage), they cycle through each of those in-page state changes.

As a result, the current view remains visible when going back from past courses as well as from current courses (even though it's the same view).

## Expected Behavior

In-page view changes (tab switches, filter changes, sort order changes) should use `history.replaceState()` instead of `history.pushState()` so that:
1. They **update the current URL** without adding new history entries.
2. The browser Back button takes the user to the **actual previous page**, not through each in-page state variation.

## Steps to Reproduce

1. Navigate to `/arena/` (the contest list page).
2. Click on the **"Past"** tab.
3. Click on the **"Current"** tab.
4. Click on the **"Future"** tab.
5. Press the browser **Back button**.
6. **Observed**: Instead of going back to the previous page, the browser cycles through Future → Current → Past → original `/arena/` load.

## Root Cause

In [contest_listv2.ts](cci:7://file:///c:/Users/BIT/Desktop/omegaup/frontend/www/js/omegaup/arena/contest_listv2.ts:0:0-0:0), the `fetch-page` event handler uses `window.history.pushState()` for user-initiated interactions (line 179):

```typescript
window.history.pushState({}, '', urlObj);
```

While replaceState is correctly used for browser back/forward navigation (via the isFromBrowserNavigation flag), it is not applied for the initial forward user interactions like tab switches, filter changes, or sort order changes.

## Suggested Fix
Use history.replaceState() for tab, filter, and sort order changes. Reserve history.pushState() only for actions that represent meaningful new navigations (e.g., search queries or pagination).

## Context
[...]

PR #9111, opened 2026-02-15
[BUG] Fix arena contest list creating extra browser history entries on tab/filter/sort changes

Yashagarwal9798:
# Description

Fix arena contest list creating extra browser history entries when switching tabs, changing sort order, or applying filters. Previously, every in-page state change used `history.pushState()`, which added a new entry to the browser history stack. This caused the browser Back button to cycle through each tab/filter/sort state instead of navigating to the actual previous page.

The fix replaces `pushState()` with `replaceState()` for all in-page state changes, and removes the partial `isFromBrowserNavigation` workaround that was only applied during browser back/forward navigation.

### Changes made:

**[ContestListv2.vue](cci:7://file:///c:/Users/BIT/Desktop/omegaup/frontend/www/js/omegaup/components/arena/ContestListv2.vue:0:0-0:0):**
- Removed `replaceState` optional field from [UrlParams](cci:2://file:///c:/Users/BIT/Desktop/omegaup/frontend/www/js/omegaup/components/arena/ContestListv2.vue:455:0-461:1) interface (no longer needed)
- Removed `isFromBrowserNavigation` flag and its related logic
- Simplified [fetchInitialContests()](cci:1://file:///c:/Users/BIT/Desktop/omegaup/frontend/www/js/omegaup/components/arena/ContestListv2.vue:525:2-539:3) by removing conditional `replaceState` usage
- Cleaned up prop watchers by removing `isFromBrowserNavigation` flag assignments

**[contest_listv2.ts](cci:7://file:///c:/Users/BIT/Desktop/omegaup/frontend/www/js/omegaup/arena/contest_listv2.ts:0:0-0:0):**
- Simplified `fetch-page` event handler to always use `history.replaceState()`
- Removed conditional `pushState`/`replaceState` branching
- Removed the `replaceState` key skip in URL parameter iteration

Fixes: #9110

# Comments

This issue was identified by @pabo99 during the review of PR #8776. The original PR added browser back/forward button support for contest filters, but introduced a side effect where every tab/filter/sort change created a new history entry. This PR addresses that side effect with a minimal, focused fix.

# Checklist:

https://github.com/user-attachments/assets/2d88a84c-dc0c-4748-a84d-4d69856cf06d

### DIFF
--- a/frontend/www/js/omegaup/arena/contest_listv2.ts
+++ b/frontend/www/js/omegaup/arena/contest_listv2.ts
@@ -165,19 +165,15 @@ OmegaUp.on('ready', () => {
             urlObj: URL;
           }) => {
             for (const [key, value] of Object.entries(params)) {
-              if (key === 'replaceState') continue; // Don't add flag to URL
               if (value) {
                 urlObj.searchParams.set(key, value.toString());
               } else {
                 urlObj.searchParams.delete(key);
               }
             }
-            // Use replaceState for browser navigation to avoid corrupting history
-            if (params.replaceState) {
-              window.history.replaceState({}, '', urlObj);
-            } else {
-              window.history.pushState({}, '', urlObj);
-            }
+            // Always use replaceState to avoid polluting browser history
+            // with in-page state changes (tab switches, filter, sort order)
+            window.history.replaceState({}, '', urlObj);
             await contestStore.dispatch('fetchContestList', {
               requestParams: params,
               name: params.tab_name,
--- a/frontend/www/js/omegaup/components/arena/ContestListv2.vue
+++ b/frontend/www/js/omegaup/components/arena/ContestListv2.vue
@@ -459,7 +459,6 @@ export interface UrlParams {
   query: string;
   sort_order: ContestOrder;
   filter: ContestFilter;
-  replaceState?: boolean; // When true, use replaceState instead of pushState (for browser navigation)
 }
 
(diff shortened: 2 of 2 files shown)


pabo99:
Oh, I think my idea in the comment you mentioned wasn’t clear enough, @Yashagarwal9798.

The expected behavior is indeed for the back button to work after each meaningful user interaction, since every view shows different information. Because of that, those transitions should use `pushState`, allowing users to navigate back through the UI states.

What I was referring to in the comment is a small issue on the initial load: The main Arena view still appears after two back clicks. This seems to happen because when entering the view for the first time, all URL parameters are added at once and a new history entry is pushed.

In that case, the initial normalization of the URL should probably use `replaceState` instead of `pushState`, so we don’t create an extra history entry.

If this is clear now, feel free to choose one of the following options:

- Close this issue and its corresponding PR, then open a new one describing the behavior mentioned above

- Update the title and description of this issue and adjust the PR accordingly

Yashagarwal9798:
Thanks for the clarification, @pabo99 sir!

I see now that the issue is specifically about the initial URL normalization creating an extra history entry, and that we do want to keep pushState for user interactions (tabs/filters) to preserve navigation history.

Since this PR changes the behavior for all interactions (which is incorrect), I will close this PR and open a fresh one that targets only the initial load issue as per your suggestion (Option 1).

Issue #9119, opened 2026-02-15
[BUG] No spacing between checkbox and label text on registration page


Yashagarwal9798:
## Current Behavior

On the registration page (`/login/?redirect=%2F#signup`), the checkbox for accepting the site's terms has little to no gap between the checkbox input and the label text _"I have read and accept the site's Use and Privacy Policy, as well as the Code of Conduct."_ This makes the text feel cramped and reduces readability.

## Expected Behavior

There should be a small but visible gap (e.g., `4px–8px` margin or padding) between the checkbox and its associated label text, following standard UI spacing conventions for form elements.

## Steps to Reproduce
1. Navigate to [https://omegaup.com/login/?redirect=%2F#signup](https://omegaup.com/login/?redirect=%2F#signup)
2. Scroll down to the checkbox at the bottom of the registration form
3. Notice the lack of spacing between the checkbox and the label text

*Relevant details about the environment in which you encountered this bug:*

  * Browser name (e.g., Chrome 39, node.js 5.4): Any browser
  * Operating System and version (desktop or mobile): Any OS
  * Link to the site where this issue occurs: https://omegaup.com/login/?redirect=%2F#signup
  * Screenshot:

<img width="1915" height="1022" alt="Image" src="https://github.com/user-attachments/assets/f3ec9288-3a9e-433e-a3dc-b4b1f8ff4052" />

PR #9120, opened 2026-02-15
fix: add spacing between checkbox and label text on signup page

Yashagarwal9798:
# Description

Added `class="pl-1"` to the checkbox label in the first signup form (`!useSignupFormWithBirthDate`) in [Signup.vue](cci:7://file:///c:/Users/BIT/Desktop/omegaup/frontend/www/js/omegaup/components/login/Signup.vue:0:0-0:0). The checkbox for "I have read and accept the site's Use and Privacy Policy, as well as the Code of Conduct." had no gap between the checkbox and the label text, making it hard to read.

The second signup form (`useSignupFormWithBirthDate`) already had `class="pl-1"` on its label (line 209), so this change makes both forms consistent.

Fixes: #9119 

# Comments

This is a one-line CSS class addition to match the existing pattern already used in the second form variant of the same component.

# Checklist

image


Yashagarwal9798:
@pabo99 I have applied the requested changes. Ready for another review whenever you have a moment!
image
image
image
image

frontend/www/js/omegaup/components/login/Signup.vue
```diff
@@ -59,17 +59,32 @@
...
-              ></omegaup-markdown>
-            </label>
+            <div class="checkbox-wrapper">
+              <input
+                v-model="privacyPolicyAccepted"
+                data-signup-accept-policies
+                type="checkbox"
+                required
+              />
+              <label for="checkbox" class="pl-1">
```
pabo99: Can you make the checkbox toggle when the label is clicked as well?

Normally, we achieve this by wrapping the input and the text inside the label, but since this text comes from the `Markdown` component, it might not work.

So feel free to use the `id` attribute to reference it instead.

The linter might complain; in that case, you can disable the rule as we do in other places, for example:

https://github.com/omegaup/omegaup/blob/da01b938c541ad6eb3a92deee67ff046a6146187/frontend/www/js/omegaup/components/login/Login.vue#L12-L28

Yashagarwal9798: @pabo99 sir, Thanks for the suggestion! I've made the following changes:

- Added unique `id` attributes to each checkbox input and matching [for](cci:1://file:///c:/Users/BIT/Desktop/omegaup/frontend/www/js/omegaup/components/login/Signup.vue:379:2-389:3) attributes on the labels, so clicking the label text now toggles the checkbox.
- Used `<!-- id-lint off -->` to suppress the linter, following the same pattern as in `Login.vue`.

Regarding CI: all Cypress tests pass except for one flaky failure in `problem_creator.cy.ts` (`"Error: The operation was canceled."`), which i think is unrelated to this PR it appears to be a timeout/memory issue in the problem creator test suite.


https://github.com/user-attachments/assets/7c2e37cd-9ddd-43ae-9390-1c83d5592c75

frontend/www/js/omegaup/components/login/Signup.vue
```diff
@@ -58,18 +58,36 @@
         </div>
 
         <div class="row justify-content-md-center">
+          <!-- id-lint off -->
```
pabo99, line 61: Just for consistency, could you re‑enable the id‑lint right after where the id is used? The same applies to the other checkbox

pabo99 (commented):
Just a minor suggestion.

frontend/www/js/omegaup/components/login/Signup.vue
Yashagarwal9798: @pabo99 sir, I have made the requested changes. Please take a look whenever you have a moment.

Issue #9128, opened 2026-02-16
[FEATURE] Drop 9 redundant indexes across 8 tables


Yashagarwal9798:
# As a database administrator, I want redundant indexes removed so that write performance improves and storage is not wasted

## Context

Several tables in [`schema.sql`](../frontend/database/schema.sql) have indexes that are exact prefixes of — or duplicates of — existing indexes (including the primary key). These redundant indexes waste storage on disk and in the InnoDB buffer pool, and slow down every `INSERT`, `UPDATE`, and `DELETE` because MySQL must maintain all indexes on write.

### Redundant indexes identified (9 instances across 8 tables)

| Table (line) | Redundant Index | Already Covered By |
|---|---|---|
| `Favorites` (384) | `KEY user_id (user_id)` | `PRIMARY KEY (user_id, problem_id)` |
| `Problemset_Problems` (885) | `KEY problemset_id (problemset_id)` | `PRIMARY KEY (problemset_id, problem_id)` |
| `Problems_Tags` (790) | `KEY problem_id (problem_id)` | `PRIMARY KEY (problem_id, tag_id)` |
| `Problem_Viewed` (714) | `KEY problem_id (problem_id)` | `PRIMARY KEY (problem_id, identity_id)` |
| `Problems_Forfeited` (765) | `KEY user_id (user_id)` | `PRIMARY KEY (user_id, problem_id)` |
| `Problems_Languages` (778) | `KEY problem_id (problem_id)` | `PRIMARY KEY (problem_id, language_id)` |
| `Coder_Of_The_Month` (183) | `KEY coder_of_the_month_id (coder_of_the_month_id)` | `PRIMARY KEY (coder_of_the_month_id)` |
| `School_Of_The_Month` (1060) | `KEY school_of_the_month_id (school_of_the_month_id)` | `PRIMARY KEY (school_of_the_month_id)` |
| `Auth_Tokens` (99) | `KEY acting_identity_id (identity_id)` | Exact duplicate of `KEY identity_id (identity_id)` — same column, different name |

**Note:** In InnoDB, a `KEY (col_a)` when a `PRIMARY KEY (col_a, col_b)` exists is redundant because the primary key's B-tree is already sorted by `col_a` first and can serve any query filtering on `col_a` alone.

### Suggested improvement

Drop the redundant indexes via a migration:

```sql
ALTER TABLE `Favorites` DROP INDEX `user_id`;
[...]

PR #9165, opened 2026-02-18
fix: drop 9 redundant indexes across 8 tables

Yashagarwal9798:
# Description

Drop 9 redundant indexes across 8 tables in [schema.sql](cci:7://file:///c:/Users/BIT/Desktop/omegaup/frontend/database/schema.sql:0:0-0:0) and [dao_schema.sql](cci:7://file:///c:/Users/BIT/Desktop/omegaup/frontend/database/dao_schema.sql:0:0-0:0) to improve write performance and reduce storage overhead.

In InnoDB, a `KEY (col_a)` is redundant when a `PRIMARY KEY (col_a, col_b)` already exists, because the PK's B-tree is sorted by `col_a` first and can serve any query filtering on `col_a` alone.

### Indexes dropped

| Table | Dropped Index | Reason |
|---|---|---|
| `Favorites` | `KEY user_id` | Prefix of PK `(user_id, problem_id)` |
| `Problemset_Problems` | `KEY problemset_id` | Prefix of PK `(problemset_id, problem_id)` |
| `Problems_Tags` | `KEY problem_id` | Prefix of PK `(problem_id, tag_id)` |
| `Problem_Viewed` | `KEY problem_id` | Prefix of PK `(problem_id, identity_id)` |
| `Problems_Forfeited` | `KEY user_id` | Prefix of PK `(user_id, problem_id)` |
| `Problems_Languages` | `KEY problem_id` | Prefix of PK `(problem_id, language_id)` |
| `Coder_Of_The_Month` | `KEY coder_of_the_month_id` | Exact duplicate of single-column PK |
| `School_Of_The_Month` | `KEY school_of_the_month_id` | Exact duplicate of single-column PK |
| `Auth_Tokens` | `KEY acting_identity_id` | Duplicate index on `identity_id` (same column as existing `KEY identity_id`) |

Fixes: #9128

# Comments

Schema-only change. No application logic modified. The dropped indexes are purely redundant — all queries previously served by them are still covered by the primary key or an existing identical index.

# Checklist


frontend/database/schema.sql
```diff
@@ -891,7 +884,6 @@ CREATE TABLE `Problemset_Problems` (
   `order` int NOT NULL DEFAULT '1' COMMENT 'Define el orden de aparición de los problemas en una lista de problemas',
   `is_extra_problem` tinyint(1) NOT NULL DEFAULT '0',
   PRIMARY KEY (`problemset_id`,`problem_id`),
-  KEY `problemset_id` (`problemset_id`),
```
pabo99, line 894: In this context, is the index `idx_problemset_problems_ids` necessary? Since it is similar to the primary key, or does the ordering influence whether we should keep it?

Yashagarwal9798: @pabo99 sir,
The `idx_problemset_problems_ids` index is **not redundant** and should be kept. Here's why:

### Column Order Matters in B-tree Indexes

In InnoDB, the column order in a composite index determines which queries can efficiently use that index. The index's B-tree is sorted by the **first column**, then by the second column within each first-column value.

### Current Indexes on `Problemset_Problems`:

| Index | Column Order | Optimizes Queries On |
|-------|-------------|----------------------|
| `PRIMARY KEY` | `(problemset_id, problem_id)` | `problemset_id` first |
| `idx_problemset_problems_ids` | `(problem_id, problemset_id)` | `problem_id` first |

### Query Patterns Served:

**PRIMARY KEY (`problemset_id`, `problem_id`)** efficiently serves:

-- "Get all problems in this contest/course"
SELECT * FROM Problemset_Problems WHERE problemset_id = ?;

**`idx_problemset_problems_ids` (`problem_id`, `problemset_id`)** efficiently serves:

-- "Get all contests/courses that contain this problem"
SELECT * FROM Problemset_Problems WHERE problem_id = ?;

### Why Both Are Needed:

Without `idx_problemset_problems_ids`, a query like `WHERE problem_id = ?` would require a **full table scan** because the PRIMARY KEY is sorted by `problemset_id` first MySQL cannot efficiently locate rows by `problem_id` using only the PK.

This is different from the indexes I dropped (like `KEY problemset_id`), which was a **prefix** of the PK. A prefix index is redundant because the PK already provides the same lookup capability. However, `idx_problemset_problems_ids` has the columns in **reversed order**, making it a distinct and necessary index for opposite-direction lookups.



Both sort orders serve different use cases  same concept applies here.

pabo99: Thank you for the detailed explanation!

pabo99 (approved):

pabo99 (approved):

Issue #9201, opened 2026-02-20
[BUG] daily-release.yml runs weekly (Sundays) not daily — name and schedule are mismatched


PrasannaMishra001:
## Current Behavior

`.github/workflows/daily-release.yml` is named `daily-release` but its cron schedule runs only once a week on **Sundays**:
```yaml
# TODO(#1624): Make this daily once we have better coverage of the frontend.
schedule:
  - cron: '0 3 * * 0'
```
`weekly-release.yml` runs on Mondays (`0 3 * * 1`) — only 24 hours apart. The skip conditions also use `exit 1` which shows as a red failure in CI instead of a neutral skip.

## Expected Behavior

The workflow name should match its actual schedule. Either rename to reflect weekly-on-Sunday behavior, or update the cron to `'0 3 * * *'` if coverage now warrants daily runs. Skip conditions should use `exit 0` + `::notice::` instead of `exit 1` to avoid false failure signals.

## If necessary, add the following information

Steps to reproduce:
1. Open `.github/workflows/daily-release.yml`
2. Check cron value: `'0 3 * * 0'` — this is Sunday only, not daily
3. Compare to `weekly-release.yml`: `'0 3 * * 1'` — Monday, just 1 day apart

*Relevant details:*
- File: `.github/workflows/daily-release.yml`
- OS: GitHub Actions (ubuntu-22.04)

PR #9217, opened 2026-02-20
fix: correct daily-release schedule comment and use notice annotations (#9201)

PrasannaMishra001:
# Description

two small fixes in `.github/workflows/daily-release.yml`:

1. the header comment said "Run every day" but the cron `'0 3 * * 0'` runs only on Sundays - updated to "Run every Sunday"
2. skip conditions used `::warning::` annotations, which show as orange warnings even for expected intentional skips - changed to `::notice::` which is the correct level for informational messages

Fixes: #9201

# Checklist:


pabo99 (approved):
Thanks!

Issue #8956, opened 2026-02-02
[FEATURE]  Enhance Code Runner Layout for Improved Usability and Engagement


ParvTyagi:
# ✨ Feature Request: Enhance Code Runner Layout for Improved Usability and Engagement

## Summary

The omegaUp code runner is functionally strong, but the current layout makes inefficient use of screen space and feels visually cramped. Large portions of the page remain unused while the editor, problem statement, and output panels are constrained to small areas.

This affects usability, readability, and the overall solving experience, especially on larger screens.

---

## Current Problems

- The code editor does not use the full available viewport height
- Significant empty whitespace on the right and bottom of the page
- Problem statement appears in a small side panel instead of a readable area
- Output/console panel is too small to comfortably view logs and results
- Controls (language, font size, Run, Submit) are scattered instead of grouped
- Right sidebar consumes width but provides little value
- Run/Submit buttons move out of view while scrolling
- Overall layout feels outdated compared to modern coding platforms

---

### 1. Full-height responsive editor

Allow the editor to expand and occupy the available vertical space using a flex layout.

### 2. Resizable split layout

Introduce a split-pane design:

- **Left:** Problem statement  
- **Center:** Code editor  
- **Bottom:** Output/console (resizable)

### 3. Larger and resizable output panel

Make compiler output/logs easier to read by allowing users to resize the panel.

### 4. Unified top toolbar

Group controls into a single clean toolbar:

### 5. Collapsible or reduced right sidebar

Either make the sidebar collapsible or move its contents into tabs to free horizontal space.

### 6. Sticky Run/Submit buttons

Keep primary actions visible while scrolling.

### 7. Visual and accessibility improvements

- Better spacing and padding in the editor
- Improved contrast and font readability
- Modernized appearance

---

## Expected Benefits

- Better use of available screen space
- Improved readability of code and problem statement
- Reduced scrolling and clutter
- More modern and professional solving experience
- Accessibility improvements
- Better experience on large monitors and laptops

---

## Scope

This change primarily involves layout/CSS improvements and does **not** affect core judging logic.

---

## Screenshot

![Code Runner Layout Issue](https://github.com/user-attachments/assets/132b650e-aaa5-4056-b693-4bef763aea10)

ParvTyagi:
@pabo99 
I’d appreciate feedback on the feasibility of these layout improvements and any technical constraints to consider before working on a PR.

JaYRaNa213:
@ParvTyagi 

[...]

PR #9221, opened 2026-02-20
UI: Add core utilities and constants for grader components (Phase 1)

ParvTyagi:
## Summary
This is the first of a series of PRs aimed at improving the standalone omegaUp code runner layout as discussed in #8956. Following reviewer feedback, I am splitting the original large PR into smaller, manageable chunks. 

This specific PR introduces the shared foundation: **Constants** and **Utility functions** required by the new UI components.

## Motivation
The current code runner requires excessive scrolling and has inefficient screen utilization. To implement a more ergonomic, flexbox-based layout, shared logic for theming, throttling, and duration parsing needed to be modernized and centralized to prevent code duplication across the new Vue components.

## Changes
- **`constants.ts`**: Centralizes editor font sizes, timing constants for feedback/debouncing, and keyboard shortcut definitions.
- **`util.ts`**: 
    - Added `announceToScreenReader` for ARIA live region support.
    - Updated `vuexGet` and `vuexSet` with better type safety.
    - Implemented `throttle` for efficient editor resize handling.
    - Exported `MonacoThemes` enum for consistent theme switching.

## Technical Notes
- Strictly compatible with **Vue 2** and **ES2015**.
- Fixed previous CI failures by removing explicit `any` types where possible.
- No functional changes to existing judging logic.

## Related PR Plan
To help reviewers, here is the order of the subsequent PRs that will follow this one:
1. **Foundation (This PR):** Utils and Constants.
2. **Leaf Components:** Isolated UI components (TextEditor, ZipViewer, etc.).
3. **Control Components:** Logic components (CaseSelector, IDESettings).
4. **Integration:** The main `Ephemeral.vue` layout update.

Fixes: #8956


pabo99:
Hi Parv, just to confirm, have you already solved 2 out of the 3 problems in the contest https://omegaup.com/arena/gsoc2026/? 

If so, could you please share your omegaUp username? That will help us speed up the review of your PRs, as we are prioritizing contributions from candidates who have successfully completed the first phase of GSoC.

frontend/www/js/omegaup/grader/util.ts
```diff
@@ -12,7 +13,8 @@ export function vuexGet(store: any, name: string) {
   return o;
 }
 
-export function vuexSet(store: any, name: string, value: any) {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function vuexSet(store: any, name: string, value: unknown) {
```
pabo99: To avoid adding this line, wouldn’t it be better to use the proper store typing to ensure type safety and autocompletion?
```suggestion
export function vuexGet<K extends GraderStoreKeys>(
  store: Store<GraderStore>,
  name: K
): GraderStore[K] {
```
where:
```ts
type GraderStoreKeys = keyof GraderStore;
```
and we would only need to add the necessary imports:
```ts
import { Store } from 'vuex';
import { GraderStore } from './GraderStore';
```
This way, we avoid using `any` and `unknown`, which only bypass type checking, and we also eliminate the need to disable lint rules.

The method would simply require:
```ts
return store.state[name];
```


The vuexSet method should apply the same changes:
```ts
export function vuexSet<K extends keyof GraderStore>(
  store: Store<GraderStore>,
  name: K,
  value: GraderStore[K],
) {
  store.commit(name, value);
}
```

However, after reviewing the codebase, it appears that these two functions are not used anywhere. Could you please confirm this? If that’s the case, they can be safely removed.

frontend/www/js/omegaup/grader/util.ts
```diff
@@ -88,15 +85,32 @@ export function asyncError(err: Error) {
...
+
+  document.body.appendChild(announcement);
+  setTimeout(() => {
+    try {
+      document.body.removeChild(announcement);
+    } catch {
+      /* Ignored */
+    }
+  }, 1000);
+}
```
pabo99: Just to confirm: Is this change necessary for the purpose of the PR?

If not, the remaining changes seem to be only style adjustments. Could everything be reverted entirely? That would help us move forward with the reviews more quickly.

yarn.lock
```diff
@@ -15639,4 +15639,4 @@ yocto-queue@^1.0.0:
 zwitch@^1.0.0:
   version "1.0.5"
   resolved "https://registry.npmjs.org/zwitch/-/zwitch-1.0.5.tgz"
-  integrity sha512-V50KMwwzqJV0NpZIZFwfOD5/lyny3WlSzRiXgA0G7VUnRlqttta1L6UQIHzd6EuBY/cHGfwTIck7w1yH6Q5zUw==
+  integrity sha512-V50KMwwzqJV0NpZIZFwfOD5/lyny3WlSzRiXgA0G7VUnRlqttta1L6UQIHzd6EuBY/cHGfwTIck7w1yH6Q5zUw==
```
pabo99: Could you revert this minor change as well?

ParvTyagi:
@pabo99 
Username: parvmit.blr (Parv)

PR #9236, opened 2026-02-20
Make assign_badges resilient to per-badge failures

### LINKED ISSUE #9177: [BUG] assign_badges cron aborts on single badge failure
## Improve assign_badges cron reliability by isolating failures per badge

Currently, `assign_badges.py` processes all badges in one run, but if any single badge fails (for example due to an invalid `query.sql`), the entire cron aborts and no badges are committed — even those that were processed successfully.

This makes the job fragile, since one broken badge can block all badge assignments.

The idea is to handle each badge independently:
- Commit after each successful badge.
- If a badge fails, log the error, roll back that badge’s changes, and continue with the remaining badges.

This would prevent one bad badge from stopping all others, while keeping existing logic unchanged.

Prince-Prajapati07:
This PR makes assign_badges more robust by isolating failures per badge.

- Commits after each successful badge.
- Rolls back and logs on per-badge failure, then continues processing remaining badges.
- Returns failure status and exits non-zero if any badge fails, while keeping successful badges committed.

No SQL or badge logic changes, only control flow in assign_badges.py.

Fixes #9177


stuff/cron/assign_badges.py
pabo99: I'm noticing this implementation uses `f-string` interpolation directly in the SQL query, which could lead to SQL injection.

While `badge` comes from filesystem directory names, it's still a security risk. Could you use parametrized queries instead:
```suggestion
cur_readonly.execute('''
       SELECT
           ub.user_id
       FROM
           Users_Badges ub
       WHERE
           ub.badge_alias = %s;''', (badge,))
```

stuff/cron/assign_badges.py
```diff
@@ -79,23 +79,32 @@ def save_new_owners(badge: str, users: Set[int],
...
-            all_owners = get_all_owners(badge, current_timestamp, cur_readonly)
+            all_owners = get_all_owners(badge, current_timestamp,
+                                        cur_readonly)
             current_owners = get_current_owners(badge, cur_readonly)
             new_owners = all_owners - current_owners
             logging.info('New owners: %s', new_owners)
             if new_owners:
                 save_new_owners(badge, new_owners, cur)
+            dbconn.conn.commit()
         except:  # noqa: bare-except
```
pabo99: This except statement is catching all exceptions, including KeyboardInterrupt and SystemExit. This can make the cron job hard to stop and masks critical errors, could you change to use:

```suggestion
        except Exception:
```
?

stuff/cron/assign_badges.py
```diff
@@ -79,23 +79,32 @@ def save_new_owners(badge: str, users: Set[int],
...
+            dbconn.conn.rollback()
+            has_failures = True
```
pabo99, line 108: Additionally, it would be helpful to track processing statistics here. I mean, add a counter for successful badges, and a list of failed badges. This way, you can log outside the `for` statement the number of badges processed successfully and show the list of errors if any occur.

Prince-Prajapati07:
Thanks for the feedback @pabo99 sir!
Updated to use parameterized queries, replaced bare `except` with `except Exception`, and added processing statistics for successful and failed badges.
Please let me know if anything else should be improved.

stuff/cron/assign_badges.py
```diff
@@ -43,7 +43,8 @@ def get_current_owners(
     cur_readonly: mysql.connector.cursor.MySQLCursorDict,
 ) -> Set[int]:
     '''Returns a set of ids of current badge owners'''
-    cur_readonly.execute('''
+    cur_readonly.execute(
+        '''
```
pabo99: nit:
```suggestion
    cur_readonly.execute('''
```

Prince-Prajapati07: Thanks for the fix and the review!

pabo99 (approved):

PR #9243, opened 2026-02-21
fix(Username): Scope css to fix breadcrumbs regression

### LINKED ISSUE #9174: [Bug] Fix breadcrumbs styling regression in course view
## Description

The breadcrumbs in the course view currently look broken. 

Image

## Context

This regression was introduced with PR #9020

## Posible solution

The minimal fix would be to revert the styles added in that PR. However, instead of a full revert, a better approach would be to use more specific selectors, since the current ones are leaking across components and causing conflicts.
This change should restore the intended appearance of breadcrumbs in the course view while preventing style overrides in unrelated components:

Image

SH20RAJ:
Fixes #9174

This PR resolves the breadcrumbs styling regression introduced in #9020 by applying scoped CSS correctly. Instead of leaking generic `span` styles, the component now uses the `.username` class selector to constraint its layout rules exclusively within `Username.vue`.

### DIFF
--- a/frontend/www/js/omegaup/components/user/Username.vue
+++ b/frontend/www/js/omegaup/components/user/Username.vue
@@ -1,5 +1,5 @@
 <template>
-  <span :class="classname" :title="username">
+  <span :class="['username', classname]" :title="username">
     <omegaup-countryflag
       v-if="country != null"
       :country="country"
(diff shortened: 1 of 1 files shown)


pabo99:
Closed in favor of PR #9546

Issue #9193, opened 2026-02-20
[BUG] Grader container fails to initialize ephemeral run manager: permission denied on /var/lib/omegaup/ephemeral


PrasannaMishra001:
## Current Behavior

On every fresh `docker compose up`, the grader container logs the following error on startup:

```
t=2026-02-20T12:11:24+0000 lvl=eror msg="Failed to fully initalize the ephemeral run manager" err="mkdir /var/lib/omegaup/ephemeral: permission denied"
stack=[main.main.func5(omegaup-grader/main.go:274)]
```

The grader continues running but the ephemeral run manager is silently broken. This happens on every container start because the `ephemeral` subdirectory does not exist inside the `omegaupdata` volume and the container user does not have permission to create it.

Verified with:
```bash
docker logs omegaup-grader-1 2>&1 | grep -i 'ephemeral\|permission'
```

## Expected Behavior

The grader should start up with the ephemeral run manager fully initialized, with no permission errors. The directory `/var/lib/omegaup/ephemeral` should be created and owned by the correct user before the grader starts.

## Steps to Reproduce

1. Clone the repo and run `docker compose up -d`
2. Run `docker logs omegaup-grader-1 2>&1 | grep ephemeral`
3. Observe the permission denied error on every startup

## Root Cause

In `docker-compose.yml`, the `omegaupdata` volume is mounted at `/var/lib/omegaup` for the grader container with `user: "${UID_GID}"`. The `ephemeral` subdirectory is never pre-created with the correct ownership, so the grader binary cannot `mkdir` it at runtime.

## Suggested Fix

**File to change:** `docker-compose.yml` or the grader entrypoint script

Option 1: Add an entrypoint wrapper that pre-creates the directory:
```yaml
entrypoint: ["sh", "-c", "mkdir -p /var/lib/omegaup/ephemeral && wait-for-it mysql:13306 -- /usr/bin/omegaup-grader"]
```

Option 2: Add a volume init container or an init step in the Dockerfile that creates and permissions the subdirectory.

*Relevant environment:*
- Docker Compose: all versions
- OS: Linux (WSL2 and native)
- Affects all developers running the local dev environment

PR #9248, opened 2026-02-21
fix: pre-create ephemeral dir in grader entrypoint

PrasannaMishra001:
# description

on every fresh `docker compose up`, the grader container logs:

```
msg="Failed to fully initalize the ephemeral run manager" err="mkdir /var/lib/omegaup/ephemeral: permission denied"
```

the `omegaupdata` volume is mounted at `/var/lib/omegaup` with `user: "${UID_GID}"`, but the `ephemeral` subdirectory is never pre-created with the correct ownership, so the grader binary cannot create it at runtime. verified with:

```bash
docker logs omegaup-grader-1 2>&1 | grep ephemeral
```

changed the grader entrypoint to run `mkdir -p /var/lib/omegaup/ephemeral` before starting, so the directory always exists with the correct user ownership when the grader initializes its ephemeral run manager.

Fixes: #9193

# checklist:

- [x] the code follows the [coding guidelines](https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/Coding-guidelines.md) of omegaUp.
- [x] the tests were executed and all of them passed.
- [ ] if you are creating a feature, the new tests were added.
- [x] if the change is large (> 200 lines), this pr was split into various pull requests.


pabo99 (approved):
Great!

Issue #8272, opened 2025-05-16
[FEATURE] [SQL Queries Optimization] Crear índices para las tablas `Submissions` y `Assignments`


pabo99:
Agregar los siguientes indices:

- `identity_id`, `problemset_id`, `problem_id`para la tabla de `Submissions`
- `problemset_id`, `assignment_id`, `course_id` para la tabla de `Àssignments`

para optimizar la consulta:

```sql 
SELECT
                    c.alias,
                    c.name,
                    s.name as school_name,
                    IFNULL(pr.progress, ?.?) AS progress
                FROM
                    Courses c
                LEFT JOIN
                    Schools s ON c.school_id = s.school_id
                LEFT JOIN (
                    -- we want a score even if there are no submissions yet
                    -- and that score should not be greater than ?%
                    SELECT
                        cbpr.course_id,
                        LEAST(?, ROUND(SUM(cbpr.total_assignment_score) / SUM(cbpr.max_points) * ?, ?)) AS progress
                    FROM (
                        -- Aggregate all problem scores of an assignment.
                        SELECT
                            a.alias,
                            a.course_id,
                            a.assignment_id,
                            IFNULL(
                                -- Get the best score of each problem.
                                (
                                    SELECT
                                        SUM(ps.contest_score)
                                    FROM (
                                        SELECT
                                            psp.problem_id,
                                            MAX(r.contest_score) AS contest_score
                                        FROM
                                            Problemset_Problems psp
                                        INNER JOIN
                                            Submissions s ON s.problem_id = psp.problem_id AND s.problemset_id = psp.problemset_id
                                        INNER JOIN
                                            Runs r ON r.run_id = s.current_run_id
                                        WHERE
                                            psp.problemset_id = a.problemset_id AND
                                            s.identity_id = ?
                                        GROUP BY
                                            psp.problem_id
                                    ) AS ps
                                ),
                                ?.?
                            ) AS total_assignment_score,
                            a.max_points
                        FROM
                            Assignments a
                        WHERE
                            EXISTS (
                                SELECT
                                    *
                                FROM
                                    Problemset_Problems psp
                                INNER JOIN
                                    Submissions s ON s.problemset_id = psp.problemset_id AND s.problem_id = psp.problem_id
                                WHERE
                                    psp.problemset_id = a.problemset_id AND
                                    s.identity_id = ?
                            )
                        GROUP BY
                            a.assignment_id
                    ) cbpr
                    GROUP BY cbpr.course_id
[...]

PR #9251, opened 2026-02-21
Add composite indexes for course progress query optimization

amanpandey2587:
# Description

I have added a migration to create 2 composite indexes in the 00263_add_course_progress_indexes.sql file . These indexes are Submissions(identity_id,problemset_id,problem_id) and in Assignments(problemset_id , assignment_id, course_id) in ordered to speed up the course progress query path , where joins or filters use these columns together. Since these are composite indexes , the indexes would be created for identity_id ,identity_id+problemset_id and identity_id+problemset_id+problem_id and same for Assignments table as well. 
These indexes are currently added in schema.sql part as well.

Fixes: #8272 

# Checklist:


pabo99 (commented):
In recent analyses we’ve noticed that using `EXISTS` in subqueries is not the most optimal approach. Could you check if this:

```sql
                            EXISTS (
                                SELECT
                                    *
                                FROM
                                    Problemset_Problems psp
                                INNER JOIN
                                    Submissions s ON s.problemset_id = psp.problemset_id AND s.problem_id = psp.problem_id
                                WHERE
                                    psp.problemset_id = a.problemset_id AND
                                    s.identity_id = ?
``` 
Can be replaced with a normal `INNER JOIN`?

With that change and the creation of the indexes, this query should improve its performance

amanpandey2587:
Hi @pabo99 , I have replaced the correlated filter inside getEnrolledAndFinishedCoursesForTabs():WHERE EXISTS ( Problemset_Problems + Submissions ... s.identity_id = ? ) With a join to a deduplicated derived table INNER JOIN (SELECT DISTINCT s.problemset_id ... WHERE s.identity_id = ?) attempted ON attempted.problemset_id = a.problemset_id.
The main thoughtprocess was to avoid repeated per-assignment "EXISTS" checks ,precompute attempted problemset_id's once for the identity, then join.
Please review the changes and suggest any further modifications if required.
Thanks

yarn.lock
```diff
@@ -935,14 +935,6 @@
     "@babel/helper-create-regexp-features-plugin" "^7.22.5"
     "@babel/helper-plugin-utils" "^7.22.5"
 
-"@babel/polyfill@^7.2.5":
```
pabo99: Could you update your branch? It seems these changes were already approved in another PR, and besides, they shouldn't be included here

frontend/database/00264_add_course_progress_indexes.sql
```diff
@@ -0,0 +1,6 @@
+-- Add indexes to optimize course progress queries in Courses::getEnrolledAndFinishedCoursesForTabs()
```
pabo99, line 1: Revision 263 already exists in the repo. 

After updating your branch with the latest changes, you'll see the current consecutive number. Then you can update the file name accordingly

frontend/server/src/DAO/Courses.php
```diff
@@ -379,21 +379,22 @@ public static function getEnrolledAndFinishedCoursesForTabs(
...
+                    INNER JOIN
+                        Problemset_Problems psp
+                        ON psp.problemset_id = s.problemset_id
+                    AND psp.problem_id = s.problem_id
+                    WHERE
+                        s.identity_id = ?
+                ) attempted
+                    ON attempted.problemset_id = a.problemset_id
+                GROUP BY
+                    a.assignment_id
```
pabo99: Interesting, after performing analysis, it turns out EXISTS is faster than INNER JOIN. The query optimizer handles semi-joins more efficiently than materializing and deduplicating, and the short‑circuit evaluation avoids scanning all records, which further improves performance

Could you revert the change in this query?

amanpandey2587:
Hi @pabo99 , I have reverted the change to use Exist along with changing the file name as well ,
Please check the code and suggest changes if any, also the commits in the pagedown were due to other commits that I pulled during rebasing with main.
Thanks

frontend/server/src/DAO/Courses.php
```diff
@@ -379,8 +379,7 @@ public static function getEnrolledAndFinishedCoursesForTabs(
                             ) AS total_assignment_score,
                             a.max_points
                         FROM
-                            Assignments a
-                        WHERE
+                    Assignments a
```
pabo99: nit:
```suggestion
                            Assignments a
                            WHERE
```

frontend/server/src/DAO/Courses.php
```diff
@@ -391,9 +390,10 @@ public static function getEnrolledAndFinishedCoursesForTabs(
                                 WHERE
                                     psp.problemset_id = a.problemset_id AND
                                     s.identity_id = ?
-                            )
-                        GROUP BY
-                            a.assignment_id
+                ) attempted
+                    ON attempted.problemset_id = a.problemset_id
+                GROUP BY
+                    a.assignment_id
```
pabo99: nit:
```suggestion
                            )
                        GROUP BY
                            a.assignment_id
```

amanpandey2587:
Hi @pabo99 , you have approved the changes but didn't merged it.
Is there any problem from my part ?
Thanks

pabo99:
Merged!

Issue #8653, opened 2025-12-18
Create API endpoints for User Profile READMEs


mohi-devhub:
### Context
With the DAO layer prepared, we need to extend the `\OmegaUp\Controllers\User` controller to manage and expose Profile README data. The new API methods must follow the conventions found in `APITool.php` and existing controllers.

### Scope
- **Extend Profile Payload:**
    - Modify the `@psalm-type UserProfileInfo` or `@psalm-type ExtraProfileDetails` in `frontend/server/src/Controllers/User.php`.
    - Add a new field (e.g., `readme`) that will hold the README content.
    - Update the `apiProfile` implementation to conditionally fetch and include the README data from the new DAO, respecting the `is_visible` and `is_disabled` flags.
- **Create Update Endpoint:**
    - Implement `public static function apiUpdateReadme(\OmegaUp\Request $r)`.
    - This method will handle PUT requests for the authenticated user to save their README content. It must perform input validation (e.g., size limits) and use the `UserReadmes` DAO.
- **Create Report Endpoint:**
    - Implement `public static function apiReportReadme(\OmegaUp\Request $r)`.
    - This method will handle POST requests where an authenticated user can report another user's README.
    - It will use the DAO to increment the `report_count` for the target user's README.
    - It must include logic to check if `report_count` exceeds a defined threshold. If it does, it will use the DAO to set `is_disabled` to true, effectively and automatically hiding the content pending review.

### Out of Scope
- UI implementation.
- Direct DAO or database schema changes.
- Server-side Markdown rendering. The API will handle raw Markdown and trust the client to perform sanitization.

### Next Steps
- Implement the Vue.js components to consume these new API capabilities.

arghabanerjee54:
Thanks for the detailed context. I went through Controllers/User.php and also checked the existing API patterns in APITool.php to understand how this should be implemented.

If this issue is still unassigned, I’d like to work on it. I’m planning to:

Extend the profile payload types to include a README field

Update apiProfile to include README data conditionally, respecting the visibility flags

Implement apiUpdateReadme and apiReportReadme following the existing controller conventions
I’ll keep the changes limited to the controller/API layer as mentioned, and open a draft PR once I have an initial implementation ready.

mohi-devhub:
This issue is actually one part of the original, larger issue that I posted, and I was planning to work on the overall feature myself.

PR #9279, opened 2026-02-22
Add user profile README API endpoints and tests

mohi-devhub:
Summary

  This PR implements the server-side controller and test layer for the User Profile README feature, building on the already-merged
  database schema and DAO sub-issues.

  Files changed:
  - frontend/server/src/Controllers/User.php — core API logic
  - frontend/server/src/DAO/UserReadmeReportLog.php — hand-written DAO (new)
  - frontend/tests/controllers/UserReadmeTest.php — PHPUnit tests (new)

  ---
  What was added

  apiUpdateReadme
  - Requires authentication (ensureMainUserIdentity)
  - Validates README content: non-empty, max 10,000 characters
  - Creates a new User_Readmes record if none exists, otherwise updates the existing one
  - Editing a disabled README resets is_disabled = false (restores it)

  apiReportReadme
  - Requires authentication
  - Resolves the target user by username
  - Throws NotFoundException if no README exists or it is already disabled
  - Prevents duplicate reports via UserReadmeReportLog::hasAlreadyReported()
  - Logs each report in User_Readme_Report_Log
  - Auto-disables the README once report_count reaches README_REPORT_THRESHOLD (10)

  Profile integration (apiProfile)
  - UserProfileInfo psalm type extended with readme: null|string
  - getUserProfile() fetches the README after the cached identity profile and injects it into the response — only populated when
  is_visible = true AND is_disabled = false, otherwise null
  - Private profiles always return readme: null

  UserReadmeReportLog DAO
  - Extends auto-generated Base\UserReadmeReportLog
  - Adds hasAlreadyReported(int $readmeId, int $reporterUserId): bool using the base class existsByPK

  ---
  Tests (UserReadmeTest.php — 11 cases)

  | Test | Covers |
  |------|--------|
  | testCreateReadme | Create path — new record in DB |
  | testUpdateExistingReadme | Update path — content replaced |
  | testProfileIncludesReadme | Profile returns README when visible & enabled |
  | testProfileReadmeNullWhenDisabled | Profile returns null when is_disabled = true |
  | testProfileReadmeNullWhenNotExists | Profile returns null when no README record |
  | testUpdateReadmeTooLong | >10,000 chars throws InvalidParameterException |
  | testReportReadme | Report increments report_count |
  | testAutoDisableAtReportThreshold | 10 reports triggers auto-disable |
  | testDuplicateReportPrevented | Second report from same user throws DuplicatedEntryInDatabaseException |
  | testUpdateReadmeRestoresAfterDisable | Edit after disable resets is_disabled = false |
  | testReportNonExistentReadme | Reporting with no README throws NotFoundException |

  ---
  Out of scope

  - Frontend/Vue rendering (separate sub-issue)
  - Server-side Markdown rendering or sanitization
  - DAO or schema changes (already merged)
  
  ---
  
  Fixes : #8653

### DIFF
--- a/frontend/www/docs/Development-Environment-Setup-Process.md
+++ b/frontend/www/docs/Development-Environment-Setup-Process.md
@@ -94,7 +94,7 @@ If you want to run the JavaScript/TypeScript tests locally (outside of Docker), 
 
 ### Prerequisites
 
-1. **Node.js**: Install Node.js (version 16 or higher recommended)
+1. **Node.js**: Install Node.js (version 20 or higher recommended; CI and Cypress tooling expect Node 20+)
 2. **Yarn**: Install Yarn package manager
 
 ### Setup Steps
--- a/frontend/www/js/omegaup/components/contest/Form.vue
+++ b/frontend/www/js/omegaup/components/contest/Form.vue
@@ -884,7 +884,7 @@ import { library } from '@fortawesome/fontawesome-svg-core';
 import { isValidAlias } from '../../validators';
 library.add(fas);
 
-Vue.use(VueCookies, { expire: -1 });
+Vue.use(VueCookies, { expires: -1 });
 
 export enum ScoreMode {
   AllOrNothing = 'all_or_nothing',
--- a/frontend/www/js/omegaup/components/course/Form.vue
+++ b/frontend/www/js/omegaup/components/course/Form.vue
@@ -326,7 +326,7 @@ import 'vue-multiselect/dist/vue-multiselect.min.css';
 import 'intro.js/introjs.css';
 import introJs from 'intro.js';
 import VueCookies from 'vue-cookies';
-Vue.use(VueCookies, { expire: -1 });
+Vue.use(VueCookies, { expires: -1 });
 
 import {
   FontAwesomeIcon,
--- a/frontend/www/js/omegaup/components/course/Tabs.vue
+++ b/frontend/www/js/omegaup/components/course/Tabs.vue
@@ -24,7 +24,7 @@
       </li>
     </ul>
     <div class="tab-content">
-      <div class="row m-0 mt-4">
+      <div class="row m-0 mt-4 course-filters">
         <div class="col-md-4 col-lg-3 p-0 ml-4">
           <input
             v-model="searchText"
@@ -137,7 +137,7 @@ import latinize from 'latinize';
 import 'intro.js/introjs.css';
 import introJs from 'intro.js';
 import VueCookies from 'vue-cookies';
-Vue.use(VueCookies, { expire: -1 });
+Vue.use(VueCookies, { expires: -1 });
 
 import omegaup_Markdown from '../Markdown.vue';
 import course_CardPublic from './CardPublic.vue';
@@ -356,6 +356,18 @@ export default class CourseTabs extends Vue {
     width: calc(50% - 25px);
     flex: 0 0 calc(50% - 25px);
     max-width: calc(50% - 25px);
+  }
+
+  .row.m-0.mt-4.course-filters {
+    display: flex;
+    flex-wrap: wrap;
+    row-gap: 0.75rem;
+  }
+
+  .row.m-0.mt-4.course-filters > div {
+    max-width: 100%;
+    flex: 0 0 100%;
+    margin-left: 0 !important;
   }
 }
 
--- a/frontend/www/js/omegaup/components/group/Identities.vue
+++ b/frontend/www/js/omegaup/components/group/Identities.vue
@@ -105,7 +105,7 @@ import omegaup_Markdown from '../Markdown.vue';
 import 'intro.js/introjs.css';
 import introJs from 'intro.js';
 import VueCookies from 'vue-cookies';
-Vue.use(VueCookies, { expire: -1 });
+Vue.use(VueCookies, { expires: -1 });
 
 import { library } from '@fortawesome/fontawesome-svg-core';
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome';
--- a/frontend/www/js/omegaup/components/homepage/Homepage.vue
+++ b/frontend/www/js/omegaup/components/homepage/Homepage.vue
@@ -159,7 +159,7 @@ import user_Rank from '../user/Rank.vue';
 import homepage_Sponsors from './Sponsors.vue';
 import homepage_Cookie from './CookieConsent.vue';
 import VueCookies from 'vue-cookies';
-Vue.use(VueCookies, { expire: -1 });
+Vue.use(VueCookies, { expires: -1 });
 
 @Component({
   components: {
--- a/frontend/www/js/omegaup/components/login/Login.vue
+++ b/frontend/www/js/omegaup/components/login/Login.vue
@@ -105,7 +105,7 @@ import 'intro.js/introjs.css';
 import introJs from 'intro.js';
 import VueCookies from 'vue-cookies';
 
-Vue.use(VueCookies, { expire: -1 });
+Vue.use(VueCookies, { expires: -1 });
 
 @Component({
   components: {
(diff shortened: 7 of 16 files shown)


mohi-devhub:
@pabloaguilar Can you review this please..

pabo99:
Hello @mohi-devhub!


Are you interested in participating in GSoC? If so, could you please share your omegaUp username? This will help us verify whether you’ve already completed Phase 1 of the selection process and also allows us to focus on reviewing PRs of candidates who have solved the test problems.

mohi-devhub:
@pabo99 My omegaUp username is mohi. I have solved 2/3 test problems

frontend/tests/controllers/UserReadmeTest.php
```diff
@@ -0,0 +1,332 @@
+<?php
+/**
+ * Tests para el README del perfil de usuario.
```
pabo99: Can you add this and all the comments in English?

frontend/server/src/Controllers/User.php
```diff
@@ -5316,6 +5335,115 @@ public static function getCompareDetailsForTypeScript(
...
+    public static function apiUpdateReadme(\OmegaUp\Request $r): array {
+        $r->ensureMainUserIdentity();
+
+        \OmegaUp\Validators::validateStringOfLengthInRange(
+            $r['readme'],
+            'readme',
+            1,
+            10000
+        );
+        $content = strval($r['readme']);
```
pabo99: This way, you can directly assign the value of the `readme` parameter to `$content`:
```suggestion
        $content = $r->ensureOptionalString(
            'readme',
            required: false,
            validator: fn (string $content) => \OmegaUp\Validators::stringOfLengthInRange(
                $content,
                1,
                10000
            )
        );
```

frontend/server/src/Controllers/User.php
```diff
@@ -5316,6 +5335,115 @@ public static function getCompareDetailsForTypeScript(
...
+     * @throws \OmegaUp\Exceptions\DuplicatedEntryInDatabaseException si el usuario ya reportó este README
+     *
+     * @return array{status: string}
+     *
+     * @omegaup-request-param string $username
+     */
+    public static function apiReportReadme(\OmegaUp\Request $r): array {
+        $r->ensureMainUserIdentity();
+
+        \OmegaUp\Validators::validateStringNonEmpty($r['username'], 'username');
```
pabo99: Here you can also directly assign the value of `$username`:
```suggestion
        $username = $r->ensureString(
            'username',
            fn (string $username) => \OmegaUp\Validators::usernameOrEmail(
                $username
            )
        );
```

frontend/server/src/Controllers/User.php
```diff
@@ -5316,6 +5335,115 @@ public static function getCompareDetailsForTypeScript(
...
+        $updatedReadme = \OmegaUp\DAO\UserReadmes::getByUserId(
+            $targetUser->user_id
+        );
+        if (
+            !is_null($updatedReadme) &&
+            $updatedReadme->report_count >= self::README_REPORT_THRESHOLD
+        ) {
+            \OmegaUp\DAO\UserReadmes::setDisabled(
+                intval($updatedReadme->readme_id),
+                true
```
pabo99: To make it clearer:
```suggestion
                isDisabled: true
```

frontend/server/src/Controllers/User.php
```diff
@@ -5316,6 +5335,115 @@ public static function getCompareDetailsForTypeScript(
...
+
+        $readme = \OmegaUp\DAO\UserReadmes::getByUserId($targetUser->user_id);
+        if (is_null($readme) || $readme->is_disabled) {
+            throw new \OmegaUp\Exceptions\NotFoundException(
+                'resourceNotFound'
+            );
+        }
+
+        if (
+            \OmegaUp\DAO\UserReadmeReportLog::hasAlreadyReported(
```
pabo99: I'm noticing that this function only calls `existsByPK`. How about we just call it directly?
```suggestion
            \OmegaUp\DAO\UserReadmeReportLog::existsByPK(
```
This way you avoid creating `UserReadmeReportLog` class. You might just need a comment to indicate that this checks whether the readme has already been reported by the logged-in user.

frontend/server/src/Controllers/User.php
```diff
@@ -5316,6 +5335,115 @@ public static function getCompareDetailsForTypeScript(
...
+        );
+        if (
+            !is_null($updatedReadme) &&
+            $updatedReadme->report_count >= self::README_REPORT_THRESHOLD
+        ) {
+            \OmegaUp\DAO\UserReadmes::setDisabled(
+                intval($updatedReadme->readme_id),
+                true
+            );
+        }
```
pabo99: Can you wrap all of this in a transaction? That way, if anything fails during the process, we can roll it back.

frontend/server/src/Controllers/User.php
```diff
@@ -5316,6 +5335,115 @@ public static function getCompareDetailsForTypeScript(
...
+     * Actualiza el README del perfil del usuario autenticado.
+     *
+     * @throws \OmegaUp\Exceptions\UnauthorizedException si el usuario no está autenticado
+     * @throws \OmegaUp\Exceptions\InvalidParameterException si el contenido es vacío o excede el límite
+     *
+     * @return array{status: string}
+     *
+     * @omegaup-request-param null|string $readme
+     */
+    public static function apiUpdateReadme(\OmegaUp\Request $r): array {
```
pabo99: This API is used both to update and to create the `README`. How about we name it something like `apiSaveReadme`?

frontend/tests/controllers/UserReadmeTest.php
```diff
@@ -0,0 +1,332 @@
...
+                ])
+            );
+            $this->fail('Debió lanzar DuplicatedEntryInDatabaseException');
+        } catch (\OmegaUp\Exceptions\DuplicatedEntryInDatabaseException $e) {
+            $this->assertSame('readmeAlreadyReported', $e->getMessage());
+        }
+    }
+
+    /**
+     * Prueba que editar un README deshabilitado lo vuelve a habilitar.
```
pabo99: I’m not fully convinced by this rule, because a user who violates it through misuse could re-enable their `README` just by editing it. I think a `README` that has been disabled should have some consequence, but I can’t think of one at the moment. It could be, for example, a certain waiting period before it can be re-enabled. What do you think?

mohi-devhub: That's a fair point. A few possible approaches we could consider:                  
                                                                                                                                         
  1. Cooldown period: Don't allow the user to re-enable their README for a fixed duration (e.g., 24–48 hours) after it's been auto-disabled.                                                                                                                         
  2. Reset report count only on re-enable, not on edit: Keep the README disabled until an explicit moderator action re-enables it, and only allow the user to pre-edit the content. The new content goes live only once a moderator approves.

mohi-devhub: For this PR, I've kept the current auto-re-enable behavior since implementing a cooldown or moderation queue would require additional  schema changes and moderator-facing endpoints. I'm happy to open a follow-up issue to track whichever approach you prefer, so it doesn't get lost

.gitignore
```diff
@@ -130,5 +130,8 @@ cypress/downloads/*
 stuff/inefficient_queries.csv
 
 
-# ignore files que log extension in root 
+# ignore files que log extension in root
```
pabo99: nit:
```suggestion
Ignore files with the `.log` extension in the root directory
```

frontend/server/src/DAO/UserReadmeReportLog.php
```diff
@@ -5,24 +5,10 @@
 /**
  * UserReadmeReportLog Data Access Object (DAO).
  *
- * Esta clase contiene toda la manipulacion de bases de datos que se necesita
- * para almacenar de forma permanente y recuperar instancias de objetos
- * {@link \OmegaUp\DAO\VO\UserReadmeReportLog}.
+ * This class contains all the database manipulation needed to permanently
```
pabo99: This class is not needed at the moment; is it possible to remove it? If it’s required in the future, it can be created again

frontend/tests/controllers/UserReadmeTest.php
pabo99: Let’s stick with the first approach, since it involves less human effort. Although I think it could be enhanced if we store timestamps in the logs, for now it’s fine to create a new issue specifically for this case, ideally without impacting the current schema

mohi-devhub:
@pabo99 I have made the required changes

Issue #9282, opened 2026-02-22
Add explicit return types to Vuex getters in  GraderStore.ts


ShreshthaPratapSingh:
The GraderStore.ts file located in frontend/www/js/omegaup/grader/ contains a TODO comment indicating that explicit return types should be added to each of its Vuex getters (// TODO: add return types to each of the getters).

Currently, many of these getters rely on implicit return type inference. Explicitly defining the return types will improve:

1. Type Safety: Ensure that getters always return the expected shape, catching potential bugs early.

2. Code Readability & Maintainability: Make it immediately clear to developers what data structure a getter provides without needing to trace the implementation logic.

3. IDE Support: Enhance auto-completion and type checking in components consuming this store.

Proposed Solution:

1. Add explicit TypeScript return types to all functions within the getters object of the storeOptions in frontend/www/js/omegaup/grader/GraderStore.ts.
2. Remove the corresponding TODO comment at the top of the file since it has been fulfilled.

Examples of changes:

getters: {
  alias(state: GraderStore): string {
    return state.alias;
  },
  showSubmitButton(state: GraderStore): boolean {
    return state.showSubmitButton;
  },
  // ... other getters with their respective types (e.g., GraderSessionStorageSources | null, CaseSelectorGroup[], etc.)
}

ShreshthaPratapSingh:
Already opened, it's awaiting review

pabo99:
Hello @ShreshthaPratapSingh!

If you're interested in participating in GSoC, could you please share your omegaUp username? This will help us verify that you've already completed the phase 1 and will allow us to speed up the review of your PRs.

ShreshthaPratapSingh:
shreshthap25 @pabo99

pabo99:
Great! Once you complete 2 out of the 3 problems in the test, I'll start reviewing your PR.

The reason for this is that we're currently prioritizing PR reviews from candidates who have already completed Phase 1 of the application process (the test mentioned earlier).

Once your progress is reflected there, we'll be able to move forward with the review of this PR and the others you work on.

ShreshthaPratapSingh:
Sir can you please explain about which test are you talking about? @pabo99

pabo99:
Hello, @ShreshthaPratapSingh!

It looks like you might not be aware of the Phase 1 test, which is part of the application process for GSoC with omegaUp. Please take a moment to review the process described in this document:

[...]

PR #9283, opened 2026-02-22
fix(frontend): add explicit return types to GraderStore getters

ShreshthaPratapSingh:
# Description

Add explicit return types to all Vuex getters inside GraderStore.ts to improve type safety, code maintainability, and IDE support. This resolves a known TODO in this file (// TODO: add return types to each of the getters).

Fixes: #9282

# Comments

1. This is purely a TypeScript typing improvement; no runtime behavior or Logic was modified.
2. Explicit return types like CaseSelectorGroup[] and types.ProblemSettingsDistrib inner interfaces (CustomValidatorDistrib and CaseDistrib) have been mapped correctly properties.

# Checklist:


ShreshthaPratapSingh:
@heduenas Sir can you please review this PR ?

ShreshthaPratapSingh:
@pabo99 Sir can you please approve the workflow?

frontend/www/js/omegaup/grader/GraderStore.ts
```diff
@@ -187,24 +186,24 @@ const storeOptions: StoreOptions<GraderStore> = {
...
       const filename: CaseKey = `${state.currentCase}.out`;
       return state.outputs[filename] || '';
     },
-    outputStderr(state: GraderStore) {
+    outputStderr(state: GraderStore): string {
       const filename: CaseKey = `${state.currentCase}.err`;
       return state.outputs[filename] || '';
     },
-    settingsCases(state: GraderStore) {
+    settingsCases(state: GraderStore): SettingsCasesGroup[] {
```
pabo99: Ok, let’s start with this change.

In lines 63–72 there are a couple of interfaces with a comment indicating that those types do not exist in `api_types`. It looks like they might not have been created because there is a slight inconsistency with what the backend returns. In the backend, the `Weight` element does not exist at the group level:
```php
{
Cases: list<array{
  Cases: list<array{Name: string, Weight: float}>,  // ← Weight is HERE (single case)
  Name: string                                      
                                                    // ← NO Weight present here (group)
}
```
I ran all the PHP tests and this is the only structure that appears:
```php
                (
                    [Cases] => Array
                        (
                            [0] => Array
                                (
                                    [Name] => medium.00
                                    [Weight] => 27
                                )

                            [1] => Array
                                (
                                    [Name] => medium.01
                                    [Weight] => 28
                                )

                        )

                    [Name] => medium
                )
```
So the type we create from PHP will remain as it is, since that is the source of truth.

What we need to do is the following:

- Find the declaration of the `ProblemSettings` type in the **_Problem.php_** controller.
- Extract `array{Name: string, Weight: float}` from `ProblemSettings` and save it as a new type called `SettingsCase`:
```php
 * @psalm-type SettingsCase=array{Name: string, Weight: float}
```
Then replace it inside ProblemSettings so it looks something like this:
```php
ProblemSettings=array{Cases: list<array{Cases: list<SettingsCase>, ...
```
- Extract `array{Cases: list<SettingsCase>, Name: string}` from `ProblemSettings` and save it as a new type called `SettingsCaseGroup`:
```php
 * @psalm-type SettingsCasesGroup=array{Cases: list<SettingsCase>, Name: string}
```
Then replace it inside `ProblemSettings` so it looks something like this:
```php
ProblemSettings=array{Cases: list<SettingsCasesGroup>,  ...
```
Once this change is applied, run `./stuff/lint.sh` so the new types are propagated.

Then check in `api_types.ts` that they were added correctly and replace them with the interfaces created in this file. In fact, the two interfaces currently in this file can be removed.

frontend/www/js/omegaup/grader/GraderStore.ts
pabo99: This line (224) needs to be removed

pabo99: This would also be removed

frontend/www/js/omegaup/grader/GraderStore.ts
ShreshthaPratapSingh: Done! Extracted SettingsCase and SettingsCaseGroupas named psalm-types in Problem.php, added the corresponding interfaces to api_types.ts, and replaced the manual interfaces in GraderStore.ts. Also removed Weight from the group level since the backend doesn't include it there.

frontend/www/js/omegaup/grader/GraderStore.ts
ShreshthaPratapSingh: Removed — the Weight: 0 initialization and the Weight += accumulation are both gone from the group level. @pabo99

ShreshthaPratapSingh: Removed along with the group-level Weight property.

pabo99 (approved):
Great!

pabo99:
Oops, the linter failed!

ShreshthaPratapSingh:
@pabo99 Sir I hope it should pass now, can you please check?

pabo99:
Please run:

```bash:
./stuff/lint.sh fix -- frontend/www/js/omegaup/grader/GraderStore.ts
```

ShreshthaPratapSingh:
I ran it before but I suppose it was failing silently, this time it made some changes and I suppose it might work perfectly now, can you check it once again please? @pabo99

ShreshthaPratapSingh:
Sir can you merge this PR please? All the checks have passed, @pabo99

PR #9301, opened 2026-02-23
fix: show more contests appends without resetting list

### LINKED ISSUE #9260: [BUG] "show more contests" button causes full contest list to visually reset/reload instead of appending
## expected behavior

clicking "show more" should:
1. keep all currently visible contests on screen
2. show a loading indicator only at the bottom (below existing contests)
3. append the newly fetched contests below without any scroll jump or full-list refresh

## steps to reproduce

1. go to `/arena/?tab_name=past`
2. scroll down - click "show more contests" 2-3 times until there are 30+ contests visible
3. click "show more" again
4. observe: the list flashes / reloads from scratch instead of appending new items at the bottom

*relevant file:* `frontend/www/js/omegaup/components/arena/ContestListv2.vue` (`loadMoreContests`, `fetchPage`, template skeleton condition)

PrasannaMishra001:
# Description

When clicking "Show More Contests", the existing contest cards were replaced by skeleton loaders instead of staying visible while new contests loaded below. This caused a full-list refresh on every load-more click.

**Root causes fixed:**

1. **Template condition**: All three tab templates used `v-if="loading || refreshing"` for skeleton rendering, which replaced the entire list whenever `loading=true` - including during load-more. Changed to `v-if="(loading || refreshing) && contestListEmpty"` so skeletons only appear on initial load (when the list is empty).

2. **Fire-and-forget timing**: `loadMoreContests()` was `async` and used `await this.fetchPage(...)`, but `fetchPage()` returns `void` (it only emits an event), so `await` resolved immediately. The `finally` block reset `isScrollLoading = false` before the API responded. Removed `async/try/catch/finally`, switched to fire-and-forget `fetchPage()`, and added a `@Watch('loading')` handler to reset `isScrollLoading` and recalculate `hasMore` once the API call completes.

3. **Stale hasMore calculation**: `hasMore` was computed from `contestList.length` before the API responded, so it always saw the pre-load count. Now `scrollLoadPrevCount` snapshots the list size before the request, and `hasMore` is computed from the delta (`addedCount`) in the loading watcher.

Fixes: #9260

https://github.com/user-attachments/assets/55acdb51-5df4-47a7-8f4b-e06835ae2c46

# Checklist:

### DIFF
--- a/frontend/www/js/omegaup/components/arena/ContestListv2.vue
+++ b/frontend/www/js/omegaup/components/arena/ContestListv2.vue
@@ -177,7 +177,7 @@
           :active="currentTab === ContestTab.Current"
           @click="currentTab = ContestTab.Current"
         >
-          <template v-if="loading || refreshing">
+          <template v-if="(loading || refreshing) && contestListEmpty">
             <div
               v-for="index in 3"
               :key="`current-${index}`"
@@ -251,7 +251,7 @@
           :active="currentTab === ContestTab.Future"
           @click="currentTab = ContestTab.Future"
         >
-          <template v-if="loading || refreshing">
+          <template v-if="(loading || refreshing) && contestListEmpty">
             <div
               v-for="index in 3"
               :key="`future-${index}`"
@@ -328,7 +328,7 @@
           :active="currentTab === ContestTab.Past"
           @click="currentTab = ContestTab.Past"
         >
-          <template v-if="loading || refreshing">
+          <template v-if="(loading || refreshing) && contestListEmpty">
             <div
               v-for="index in 3"
               :key="`past-${index}`"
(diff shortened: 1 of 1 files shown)


carlosabcs:
Hi @PrasannaMishra001 the video attached can not be reproduced. Can you please fix that so we can properly check what you are solving through these changes? CC @pabo99

PrasannaMishra001:
https://github.com/user-attachments/assets/56a60997-0927-4bf3-9422-8b1de172f8d0

hi @carlosabcs , thanks for checking the pr.

in this updated video you can see it more clearly, earlier the entire list was resetting and showing skeletons again on every “show more” click. with this fix, existing contest cards remain visible and only the new ones are appended below.

the change mainly ensures skeletons render only when the list is empty and that scroll loading state is cleared after the api response via the @Watch('loading') handler. let me know if you’d like any additional clarification.

PR #9308, opened 2026-02-24
Fix scroll-to-top behavior when selecting filter/sort options in the contest list.

### LINKED ISSUE #9257: [BUG] filter/sort dropdown items use href="#" causing page scroll-to-top on click in contest list
## current behavior

in `ContestListv2.vue`, every `<b-dropdown-item>` in the **order by** and **filter by** dropdowns has `href="#"`:

```html
<b-dropdown-item href="#" @click="orderByEnds">...</b-dropdown-item>
<b-dropdown-item href="#" @click="filterBySignedUp">...</b-dropdown-item>
```

there are 9 such items total (6 sort options + 3 filter options). clicking any of them causes the browser to follow the `href="#"` anchor, which scrolls the page back to the top — even though the `@click` handler runs correctly. this is especially disruptive when a user has scrolled down to "show more" contests, then tries to change the sort/filter order.

## expected behavior

clicking a filter or sort option should apply the selection with no page scroll side effect. the page position should remain unchanged.

## steps to reproduce

1. go to `/arena/` → past contests tab
2. scroll down and click "show more" several times
3. while scrolled far down, open the **order by** or **filter by** dropdown and select any option
4. observe: page jumps back to the top

## suggested fix

replace `href="#"` on all `b-dropdown-item` elements with the `button` variant or prevent the default anchor behavior:

```html
<b-dropdown-item button @click="orderByEnds">...</b-dropdown-item>
```

*relevant file:* `frontend/www/js/omegaup/components/arena/ContestListv2.vue` (lines 57-166)

pratham15541:
# Description

Fix scroll-to-top behavior when selecting filter/sort options in the contest list.

## Problem
In `ContestListv2.vue`, all `<b-dropdown-item>` elements in the "Order by" and "Filter by" dropdowns were using `href="#"`.

Clicking any option triggered the anchor navigation, causing the page to scroll back to the top — even though the Vue `@click` handler executed correctly. This disrupted the user experience, especially after using "Show more" to scroll further down the list.

## Solution
Replaced `href="#"` with the `button` variant in all affected `<b-dropdown-item>` elements to prevent anchor navigation.

Before:
```vue
<b-dropdown-item href="#" @click="orderByEnds">...</b-dropdown-item>
````

After:

```vue
<b-dropdown-item button @click="orderByEnds">...</b-dropdown-item>
```

This ensures the dropdown items behave as buttons and no longer cause unwanted scrolling.

## How was this tested?

1. Navigated to `/arena/` → Past Contests tab.
2. Clicked "Show more" multiple times to scroll down.
3. Selected different sort and filter options.
4. Verified that:

   * Sorting/filtering works correctly.
   * Page position remains unchanged.
   * No console errors are produced.

Fixes: #9257

# Comments

Small frontend-only change in a single Vue file. No backend or test updates required.

# Checklist:

* [x] The code follows the coding guidelines of omegaUp.
* [x] This is a small change (< 200 lines).


pabo99 (commented):
It seems that the scroll to top behavior is not prevented with this change. Still, it does remove unnecessary anchors.

@PrasannaMishra001, are you okay with this change? If so, we can go ahead and approve it.

PrasannaMishra001:
@pabo99 yes, the unnecessary anchors are removed correctly. however, i had raised the issue mainly to address the scroll-to-top behavior due to these dropdown buttons.

similar to what i worked on in #9301, but that case was different since the list wasn’t being cleared there. in this case (filter/sort), the scroll jump happens because 'fetchInitialContests()' clears the list before the API response arrives.

@pratham15541 would you be willing to work on this further? we may need to avoid clearing the list immediately or introduce a separate flag to preserve existing items until the new data arrives. let me know if you’re interested in taking this up.

pratham15541:
@PrasannaMishra001 yes I will do

pratham15541:
did changes @PrasannaMishra001 (review needed) is it appropriate as per requirement?
[recording.webm](https://github.com/user-attachments/assets/5cb0a0a1-7dcc-4950-a5a1-e57d0b67770b)

PrasannaMishra001:
@pratham15541 this approach looks much better now. I just tried to test these changes locally. the overlay and refreshing are a lot improved but found this one issue on `Show More` though. In loadMoreContests(), fetchPage() added just triggers that event and returns void, so `await ` resolves before api responding.

https://github.com/user-attachments/assets/b16e6c20-a8d2-43d1-a5da-b62971b46a60

@pabo99 could you also please take a look and confirm if this approach is acceptable ? I want to make sure we keep the pattern consistent across #9301 and #9308. Then @pratham15541 can simply rebase as he would also have to do same changes in `onLoadingChanged` to fix this `Show More button` and do give suggestions as you would like to proceed.

PR #9317, opened 2026-02-25
aggregate_feedback cron aborts on single problem failure

### LINKED ISSUE #9307: [BUG] aggregate_feedback cron aborts on single problem failure
Improve `aggregate_feedback` cron reliability by isolating failures per problem

Currently, aggregate_feedback.py processes all problem feedback updates in a single run under one large transaction. If updating any single problem fails (for example due to malformed data or an unexpected DB error), the entire cron aborts and no aggregated feedback is committed, even for problems that were processed successfully.

This makes the job fragile, since one broken problem can block feedback updates for all problems.

The idea is to handle each problem independently:

- Commit after each successful problem update (or use savepoints).
- If a problem update fails, log the error with problem_id, roll back only that problem’s changes, and continue with the remaining problems.
- Add clearer logging to identify which problem caused the failure.

This would prevent one bad problem from stopping the entire aggregation run, while keeping the existing aggregation logic unchanged.

Prince-Prajapati07:
This PR makes the `aggregate_feedback` cron resilient to per-problem failures.

### What changed

- Wrap each problem aggregation in its own try/except so a single failure doesn’t abort the entire cron.
- Log the failing `problem_id` and continue processing remaining problems.
- Add summary logging (processed / failed counts).
- Add a focused unit test to verify that one failing problem does not stop others from being aggregated.

### Why

Previously, one bad problem could prevent feedback aggregation for all problems in that run.  
This change improves cron robustness and debuggability while keeping existing aggregation logic unchanged.

Fixes #9307


Prince-Prajapati07:
@pabo99 sir! It looks like CI failed during yarn install with npm 403 (before any tests ran). 
This seems like a transient registry issue. Could you please rerun the failed jobs when convenient? Thanks!

stuff/cron/aggregate_feedback_test.py
```diff
@@ -0,0 +1,106 @@
...
+            aggregate_feedback.aggregate_problem_feedback = (
+                original_aggregate_problem_feedback)
+
+        self.assertEqual(called_ids, problem_ids)
+        self.assertEqual(dbconn.rollback_calls, 1)
+
+
+if __name__ == '__main__':
+    unittest.main()
+
```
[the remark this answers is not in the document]
pabo99: Could you apply this suggestion?

stuff/cron/aggregate_feedback.py
```diff
@@ -381,16 +381,37 @@ def aggregate_feedback(dbconn: lib.db.Connection) -> None:
      global_difficulty_average) = get_global_quality_and_difficulty_average(
          dbconn, rank_cutoffs)
 
+    processed_problems = 0
```
pabo99: The counter naming could be misleading. Currently `processed_problems` is incremented before attempting aggregation, but the name suggests problems that were successfully processed.

Could you use more precise counter names to improve clarity? 
```suggestion
    attempted_problems = 0
    successful_problems = 0
```

where:
- `attempted_problems` will be the total problems we tried to process
- `successful_problems`  will be the problems that were completed without errors (increment after try block)

stuff/cron/aggregate_feedback.py
```diff
@@ -381,16 +381,37 @@ def aggregate_feedback(dbconn: lib.db.Connection) -> None:
...
                        FROM `QualityNominations` as qn
                        WHERE qn.`nomination` = 'suggestion'
                          AND qn.`qualitynomination_id` > %s;""",
                     (QUALITYNOMINATION_QUESTION_CHANGE_ID,))
-        for row in cur.fetchall():
-            aggregate_problem_feedback(dbconn, row[0], rank_cutoffs,
-                                       global_quality_average,
-                                       global_difficulty_average)
+        for (problem_id,) in cur.fetchall():
+            processed_problems += 1
```
pabo99:
```suggestion
            attempted_problems += 1
```

stuff/cron/aggregate_feedback.py
```diff
@@ -381,16 +381,37 @@ def aggregate_feedback(dbconn: lib.db.Connection) -> None:
...
+            try:
+                aggregate_problem_feedback(dbconn, problem_id, rank_cutoffs,
+                                           global_quality_average,
+                                           global_difficulty_average)
```
pabo99, line 399: And after that:
```suggestion
                aggregate_problem_feedback(dbconn, problem_id, rank_cutoffs,
                                           global_quality_average,
                                           global_difficulty_average)
                successful_problems += 1
```

stuff/cron/aggregate_feedback.py
```diff
@@ -381,16 +381,37 @@ def aggregate_feedback(dbconn: lib.db.Connection) -> None:
...
+            except Exception:  # pylint: disable=broad-except
+                failed_problems += 1
+                logging.exception(
+                    'Failed to aggregate feedback for problem %d',
```
pabo99: The exception logging is good, but it could be slightly more explicit about the cron's behavior. Can you consider adding that we're continuing with remaining problems?
```suggestion
                    'Failed to aggregate feedback for problem %d. '
                    'Continuing with remaining problems.',
```

stuff/cron/aggregate_feedback.py
```diff
@@ -381,16 +381,37 @@ def aggregate_feedback(dbconn: lib.db.Connection) -> None:
...
+                    dbconn.conn.rollback()
+                except Exception:  # pylint: disable=broad-except
+                    logging.exception(
+                        'Failed to rollback transaction for problem %d',
+                        problem_id)
+
+    logging.info(
+        'Finished aggregating feedback for problems. processed=%d failed=%d',
+        processed_problems,
+        failed_problems)
```
pabo99, line 419: Here you can log problems that were attempted, solved successfully, or failed
```suggestion
    logging.info(
        'Finished aggregating feedback: '
        'attempted=%d successful=%d failed=%d',
        attempted_problems,
        successful_problems,
        failed_problems)
```

stuff/cron/aggregate_feedback_test.py
```diff
@@ -0,0 +1,114 @@
...
+feedback for a single problem does not prevent other problems from
+being processed.
+'''
+
+import unittest
+from typing import Any, List, Tuple, cast
+
+import aggregate_feedback
+
+# pylint: disable=missing-function-docstring
```
pabo99: Could you add the docstring instead of disabling the rule?

Prince-Prajapati07: > Could you add the docstring instead of disabling the rule?

Added proper docstrings and removed the pylint disable. Thanks for pointing that out!

PR #9319, opened 2026-02-25
exempted members and dependabot from the issue linking workflow

### LINKED ISSUE #9318: [FEATURE] Remove OWNER, MEMBER, COLLABORATOR roles and dependabot from the PR linked issue rule
Same as title

antedotee:
This PR exempts people with member tag and dependabot to not necessarily being assigned to the issue. They can raise PR and the check-pr-issue.yml will pass.

The people with member tag I think is much better than specifically picking CODEOWNERS, OWNERS and COLLABORATORS.

Fixes #9318


antedotee:
CC @pabo99

PR #9322, opened 2026-02-25
Refactor/controller layer

xx-Tasneem-Ahmed-xx:
# Description
### 1. Divided `apiReview()` into focused private methods

`apiReview()`:
**before:** ~60 lines of deeply nested validation and business logic.  
**after:** 10 lines reading as a clean and linear sequence of named steps.

| New method                          | Responsibility                                      |
| ----------------------------------- | --------------------------------------------------- |
| `validateReviewAction()`            | Validates `action` is `approve` or `reject`         |
| `getJobForReview()`                 | Fetches and validates the job exists                |
| `getProblemForJob()`                | Fetches and validates the associated problem        |
| `validateProblemAdminPermissions()` | Checks the caller has problem admin rights          |
| `validateJobStatus()`               | Ensures the job is in `completed` state             |
| `approveEditorial()`                | Orchestrates the full approve flow                  |
| `getSolutionMarkdownByLanguage()`   | Extracts the correct language markdown from the job |
| `isTestEnvironment()`               | Isolates the PHPUnit class-existence check          |
| `rejectEditorial()`                 | Handles the reject flow                             |

**benefits:**
- Each helper has a single responsibility, typed parameters and documented `@throws` annotations
- Individual steps are independently testable in isolation
- High cognitive complexity slows down changes and increases the cost of maintenance.

### 2.Replaced generic `\Exception` with specific OmegaUp exceptions

Two bare `\Exception` throws were replaced with \OmegaUp\Exceptions\InternalServerErrorException`, while keeping the original error messages

**benefit:** when working with a generic type of exception, the only way to distinguish between multiple exceptions is to check their message, which is error-prone and difficult to maintain.

# Comments
No functional or behavioral changes were introduced. All existing logic and outputs remain the same.

Part of: #8919 

# Checklist:


mnayaab141-arch:
I’d like to work on this refactor. The proposed breakdown of apiReview() into focused private methods makes the flow much clearer and significantly reduces cognitive complexity.

I’m comfortable handling this kind of structural refactoring, keeping behavior unchanged while improving readability and testability. I’ll follow the outlined responsibilities for each helper method and stay within the scope described for this issue.

Please let me know if I can be assigned to this.

xx-Tasneem-Ahmed-xx:
Hello @mnayaab141-arch, Thanks a lot for your interest, I really appreciate the thoughtful breakdown you described.

Just a quick update: this refactor has already been implemented in the current PR, following the same idea of splitting apiReview() into smaller focused private methods while keeping behavior unchanged.

Please feel free to take a look at the latest commits, and let me know if you think anything could be further improved. More than happy to collaborate!

frontend/server/src/Controllers/AiEditorial.php
```diff
@@ -151,14 +151,16 @@ private static function queueJobToRedis(
...
-                throw new \Exception(
-                    "Failed to connect to Redis at {$redisHost}:{$redisPort}"
+                throw new \OmegaUp\Exceptions\InternalServerErrorException(
+                    'generalError'
                 );
             }
 
             /** @psalm-suppress RedundantCondition REDIS_PASS is really a variable */
             if (REDIS_PASS !== '' && !$redis->auth(REDIS_PASS)) {
-                throw new \Exception('Redis authentication failed');
```
pabo99, line 161: Could you add this string to the localization files and use it in the appropriate place?

frontend/server/src/Controllers/AiEditorial.php
```diff
@@ -151,14 +151,16 @@ private static function queueJobToRedis(
             $connected = $redis->connect($redisHost, $redisPort, $timeout);
 
             if (!$connected) {
-                throw new \Exception(
-                    "Failed to connect to Redis at {$redisHost}:{$redisPort}"
```
pabo99, line 155: Could you also add this string to the localization files?

When passing one or more parameters, you might need to update the `InternalServerErrorException` class so that it can accept those extra parameters. You can use the `InvalidParameterException` class as a reference, since it handles something similar.

pabo99 (commented):
Just a couple of suggestions to make sure we don't lose the error messages that were being thrown before.

Issue #9254, opened 2026-02-21
[FEATURE]


vijay-x-dev:
I wanted to suggest a small usability feature,  showing a problem status indicator (e.g., Solved / Attempted) beside each problem title.
I noticed platforms like LeetCode display this, and it helps users quickly recognize which problems they have already worked on. I’ve attached a reference screenshot for context

<img width="1079" height="434" alt="Image" src="https://github.com/user-attachments/assets/113af09c-fa99-4650-b1f7-4b8af029e50c" />

vijay-x-dev:
Hi @pabo99 , I had a thought regarding the Solved / Attempted indicator. In the problem list view, most of the space is already occupied (ID, Title, Level, quality, difficulty, ratio, my score, etc.), so adding another column might clutter the layout. I was thinking it might be better to place these indicators inside the individual problem view instead. What do you think about this approach?

pabo99:
It sounds great! Feel free to work on this feature. 

Regarding the limited space we have, I'm thinking if this view can be displayed in full-screen mode.

vijay-x-dev:
Hi @pabo99, just to clarify , by full-screen mode, do you mean expanding/enlarging the existing problem list view layout so we have room to add another field, or something like a separate expanded view?

pabo99:
Exactly, expanding the layout to take better advantage of the available space so we can add an additional visual element.

By the way, have you had a chance to review the Phase 1 GSoC contest? It's necessary to complete it so we can speed up the reviews of your PRs.

PR #9328, opened 2026-02-26
Show solved/attempted/unattempted status in problem list

vijay-x-dev:
# Description

This change adds a solved/attempted/unattempted status indicator in the problem list for each problem row.
The problem table layout is widened to use the full available horizontal space so the status field fits comfortably without feeling cramped. This makes it easier for users to quickly see which problems they’ve already solved or at least attempted

Fixes: #9254

# Screenshot
image

# Checklist:


frontend/templates/en.lang
```diff
@@ -1,3 +1,6 @@
+ProblemStatusAttempted = "Attempted"
```
vijay-x-dev: I added the new Translations (ProblemStatusAttempted, ProblemStatusSolved, ProblemStatusUnattempted) in lexicographical order, but the linter keeps moving them to the top of the block. Could you help me fix the correct placement for these?

frontend/templates/en.lang
```diff
@@ -1,3 +1,6 @@
...
+ProblemStatusSolved = "Solved"
+ProblemStatusUnattempted = "Unattempted"
```
pabo99: nit:
```suggestion
problemStatusAttempted = "Attempted"
problemStatusSolved = "Solved"
problemStatusUnattempted = "Unattempted"
```

frontend/www/js/omegaup/problem/list.ts
```diff
@@ -68,7 +68,26 @@ OmegaUp.on('ready', () => {
...
+            this.solvedProblemAliases = (solvedRes.problems || []).map(
+              (p: { alias: string }) => p.alias,
+            );
+            this.unsolvedProblemAliases = (unsolvedRes.problems || []).map(
+              (p: { alias: string }) => p.alias,
+            );
+          })
+          .catch(ui.apiError);
+      }
+    },
```
pabo99: Instead of making this pair of API requests, is there a way you could add the list of solved and unsolved problem aliases to the payload?

frontend/www/js/omegaup/components/problem/List.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="container-lg p-3">
+  <div class="container-fluid p-3">
```
pabo99: Could you show a screenshot of how it looks in full‑width mode and another of how it looks with the current width?

This will help us decide, because I’m not fully convinced by how it looks this way.

vijay-x-dev: **container-lg**
image

**container-fluid** 
image


I've attached both screenshots: one with  (container-lg) and one with the (container-fluid).

frontend/www/js/omegaup/problem/list.ts
vijay-x-dev: Sure. The backend now includes solvedProblemAliases and unsolvedProblemAliases in the problem list payload `(via User::apiProblemsSolved and User::apiListUnsolvedProblems)`, and the frontend uses those instead of making the two API requests

frontend/server/src/Controllers/Problem.php
```diff
@@ -5224,6 +5225,23 @@ public static function getProblemListForTypeScript(
             authors: $authors
         );
 
+        $solvedProblemAliases = [];
+        $unsolvedProblemAliases = [];
+        if (!is_null($r->identity)) {
+            $solvedRes = \OmegaUp\Controllers\User::apiProblemsSolved($r);
```
pabo99: Normally, we don't make calls to the endpoints from PHP, because it's more complicated to trace the inputs, since the request object is always sent. Instead, we try to create an intermediate function that retrieves the same information, avoiding the duplication of validations that each function might bring.

In this case, there are the private functions `getSolvedProblems` and `getUnsolvedProblems`, which you can make public and use here in the following way:

```suggestion
            $solvedRes = \OmegaUp\Controllers\User::getSolvedProblems(
                intval($r->identity->identity_id)
            );
```

vijay-x-dev: Thanks for the suggestion! I’ve updated to use User::getSolvedProblems() and User::getUnsolvedProblems()  instead of calling the API endpoints from PHP

frontend/server/src/Controllers/Problem.php
```diff
@@ -5224,6 +5224,25 @@ public static function getProblemListForTypeScript(
...
         );
 
+        $solvedProblemAliases = [];
+        $unsolvedProblemAliases = [];
+        if (!is_null($r->identity) && !is_null($r->identity->identity_id)) {
+            $solvedProblems = \OmegaUp\Controllers\User::getSolvedProblems(
+                intval($r->identity->identity_id)
+            );
+            $solvedProblemAliases = array_map(
+                fn($p) => $p['alias'],
```
pabo99: We try to avoid abbreviations in variable names for better readability, could you use instead:

```suggestion
                fn($problem) => $problem['alias'],
```
?
The same applies to unsolved problems

frontend/server/src/Controllers/Problem.php
```diff
@@ -5224,6 +5224,25 @@ public static function getProblemListForTypeScript(
             authors: $authors
         );
 
+        $solvedProblemAliases = [];
+        $unsolvedProblemAliases = [];
+        if (!is_null($r->identity) && !is_null($r->identity->identity_id)) {
+            $solvedProblems = \OmegaUp\Controllers\User::getSolvedProblems(
+                intval($r->identity->identity_id)
```
pabo99: It seems that the condition on line 5229 rules out null in `identity_id`, so parsing is no longer necessary:


```suggestion
                $r->identity->identity_id
```
The same applies to unsolved problems

frontend/server/src/Controllers/Problem.php
```diff
@@ -5240,6 +5259,8 @@ public static function getProblemListForTypeScript(
                     'columns' => $result['columns'],
                     'tags' => $result['tags'],
                     'tagData' => $result['tagData'],
+                    'solvedProblemAliases' => $solvedProblemAliases,
+                    'unsolvedProblemAliases' => $unsolvedProblemAliases,
```
pabo99: At first I was a bit confused by this term, now I understand and I see that in the UI these problems are shown as attempted. Can we change the name to `attemptedProblemAliases` for consistency?

frontend/www/js/omegaup/components/problem/List.vue
pabo99: Can we revert this to how it was originally? It looks better that way. While reviewing the UI I noticed something:
The **_My score_** column already shows this information, so instead of the status column you added, we could place a visual indicator directly in the **_My score_** column — one for solved problems and another for attempted ones. The score already shows which are solved, but the visual indicator will make them easier to spot.
Image
 This way you can remove the status column. 

In the future we could add a filter by status, but for this PR this change is fine

frontend/server/src/Controllers/Problem.php
vijay-x-dev: sure

frontend/www/js/omegaup/components/problem/List.vue
vijay-x-dev: I agree with your suggestion about removing the extra Status column and using a visual indicator inside the “My score” column instead, and I’ve updated the implementation accordingly.

 I also had an idea to add the status in the problem view (next to the bookmark), even though it can already be inferred from the submissions . I think a small, consistent indicator there makes the status more appealing. What are your thoughts on keeping that status indicator in the problem view as well?

image

pabo99: Do you want to create a new issue for this proposal? Or do you prefer to apply it in this PR?

frontend/www/js/omegaup/components/problem/BaseList.vue
```diff
@@ -210,7 +210,27 @@
...
+              >
+                {{ problem.score.toFixed(2) }}
+              </span>
+              <span
+                v-else
+                :title="T.problemStatusUnattempted"
+                class="badge badge-secondary"
+              >
+                {{ problem.score.toFixed(2) }}
+              </span>
```
pabo99, line 218: What do you think about, instead of adding all these conditional `span` elements, creating a method that retrieves the status title for the problem and another one that retrieves the status class? This way the template remains more readable, delegating the logic to the code section.

frontend/www/js/omegaup/components/problem/List.vue
vijay-x-dev: To keep it simple, I’ll handle it in a separate PR.

frontend/www/js/omegaup/components/problem/BaseList.vue
vijay-x-dev: Agreed

PR #9342, opened 2026-02-28
added tabSize 4 for python

### LINKED ISSUE #9186: [BUG] Inconsistent indentation while writing code in Omegaup's editor
I was recently submitting the solution in python and faced problem of inconsistent indentation. 
Ideally the indentation should have been proper to make sure I don't face any issues regarding the indentation.

I think the tabSize is 2 for all languages in the editor. But for indentation language like python, I believe it should be 4 to really match other conventional editors for python.

antedotee:
### Changes made
Added support for indentation 4 in python, while keeping others unchanged.

Fixes: #9186

### DIFF
--- a/frontend/data/languages.json
+++ b/frontend/data/languages.json
@@ -75,13 +75,15 @@
     "extension": "py",
     "name": "Python (2.7)",
     "modelMapping": "python",
-    "language": "py2"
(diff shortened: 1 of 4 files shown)


antedotee:
PTAL @pabo99.

Issue #9345, opened 2026-02-28
[BUG]Enforce strict date format in backend validator


akhil06232:
## Current Behavior

*Date validation accepts ambiguous natural-language inputs (for example, “next Thursday”) as valid dates because parsing is too permissive. This can cause inconsistent results and unintended values in profile/account-related flows.*

## Expected Behavior

*The platform should only accept strict, explicit date formats for validated date fields (for example, YYYY-MM-DD, and numeric timestamps only where explicitly supported). Ambiguous human-language strings should be rejected with a clear validation error.*

akhil06232:
@Yashagarwal9798  could you check is it good issue to work on

Yashagarwal9798:
Hey @akhil06232! Yes, this is a fantastic backend issue. Enforcing strict date formats like YYYY-MM-DD prevents a lot of weird edge cases. I see you have  already assigned it to yourself awesome!
Just a heads-up on a couple of things you might want to investigate while working on this:
You might want to check if some endpoints actually require full time components (like YYYY-MM-DD HH:MM:SS) instead of just the date. For example, contest start times usually need the exact hour and minute. If they do, just make sure your strict validation accounts for that so it doesn't accidentally block valid timestamps!
Double-check the PHPUnit tests. Some older tests might currently be passing loose strings (like 'tomorrow') to the API, so you might need to update a few test files if the new strict validation causes them to fail.
Good luck, and feel free to open a PR whenever you are ready!

PR #9346, opened 2026-02-28
Enforce strict date format in backend validator

akhil06232:
## Description
*This PR fixes date validation to reject ambiguous natural-language inputs (for example, next Thursday) and only accept strict calendar dates in YYYY-MM-DD format.*

## Changes made:
Updated backend date validation logic to enforce strict YYYY-MM-DD format.
Added robust parsing checks to reject invalid/ambiguous dates.
Added regression tests in user profile update flows to ensure:
birth_date rejects natural-language values.
graduation_date rejects natural-language values.
Preserved existing behavior for valid explicit date strings.

Fixes: #9345 

## Comments
This is a backend validation hardening change; no frontend/UI behavior was modified directly.
Please pay special attention to date-format expectations in any endpoint using validateDate() / validateOptionalDate().
If desired, I can split follow-up work to add dedicated validator-level unit tests in a separate PR.

### DIFF
--- a/frontend/server/src/Validators.php
+++ b/frontend/server/src/Validators.php
@@ -474,7 +474,8 @@ class Validators {
         if (!self::isPresent($parameter, $parameterName, required: true)) {
             return;
         }
-
+ // Validate that we are working with a date
+    // @TODO This strtotime() allows nice strings like "next Thursday".
         if (
             !is_string($parameter) ||
             !preg_match('/^\d{4}-\d{2}-\d{2}$/', $parameter)


akhil06232:
@pabo99 could you please review these pr

frontend/server/src/Validators.php
```diff
@@ -475,9 +475,30 @@ public static function validateDate(
             return;
         }
 
-        // Validate that we are working with a date
-        // @TODO This strtotime() allows nice strings like "next Thursday".
-        if (!is_string($parameter) || strtotime($parameter) === false) {
+        if (
```
pabo99, line 479: Excellent! Would it be possible to add comments at each step to make the changes easier to follow?

pabo99 (commented):
Same as I've mentioned in other PRs, ideally changes shouldn’t be mixed in the same PR. Here I can see three different ones. Let’s try to keep this PR focused on “Enforce strict date format in backend validator.”

That said, the change looks good and the tests help validate that it works correctly.

While reviewing, I’d like to know if we can make some extra improvements that would help simplify actions in the code. 

Let me explain a bit of the context:
- The functions with the `validate...` prefix were the first ones we introduced in omegaUp, and they helped validate most of the data types we use in the platform.
- Later, we started requiring these types to be declared in `Psalm`, and at the same time, return the validated value so it could be used as a variable instead of relying on the request parameter.
- So, intermediate functions were created in` frontend/server/src/Request.php` with the `ensure...` suffix. These already cover what we needed.

Therefore, would it be possible to use these functions instead of `validateDate` in the `User` controller:
https://github.com/omegaup/omegaup/blob/7d9db977ceff9087d517705bf0e2ae2cd9795ebc/frontend/server/src/Controllers/User.php#L2623-L2638
so that you use something like this:
```php
$graduationDateTimestamp = $r->ensureOptionalTimestamp('graduation_date');
if (!is_null($graduationDateTimestamp)) {
    $newGraduationDate = $graduationDateTimestamp;
}
```
and let the `ensureTimestamp` function handle identifying the format being received and passing it to the validate function.

If you have any doubts, I’d be happy to help you review them

frontend/server/src/Validators.php
akhil06232: Okay

frontend/server/src/Validators.php
```diff
@@ -475,9 +475,30 @@ public static function validateDate(
             return;
         }
 
-        // Validate that we are working with a date
-        // @TODO This strtotime() allows nice strings like "next Thursday".
```
akhil06232: Comment on line R479

PR #9348, opened 2026-02-28
Add a composite index on Groups_Identities(group_id, identity_id)

### LINKED ISSUE #9347: [FEATURE] Optimize course assignment problem statistics query
The current statistics query expands `Groups_Identities × problems` and aggregates per identity, which is expensive and leads to join buffer usage.

antedotee:
- Rewrote the stats query to avoid the full `identity × problem` expansion and compute variance/percentages using total identity counts.
- Added `Groups_Identities(group_id, identity_id)` index to speed joins.

I ran the boostrap environment script with synthetic data and env as development to really see whether my changes work ot not. I ran the 10 queries on main and got 13.63 ms avg time and on this branch, I got 9.16 ms average time. It means the changes work.

Fixes #9347


antedotee:
PTAL @pabo99.

Issue #8270, opened 2025-05-16
[FEATURE] [SQL Queries Optimization] Crear índice para el campo `name` en la tabla de `Courses`


pabo99:
Agregar el índice para optimizar la consulta:

```sql
                SELECT DISTINCT
                `c`.`course_id`, `c`.`name`, `c`.`description`, `c`.`objective`, `c`.`alias`, `c`.`group_id`, `c`.`acl_id`, `c`.`level`, `c`.`start_time`, `c`.`finish_time`, `c`.`admission_mode`, `c`.`school_id`, `c`.`needs_basic_information`, `c`.`requests_user_information`, `c`.`show_scoreboard`, `c`.`languages`, `c`.`archived`, `c`.`minimum_progress_for_certificate`, `c`.`certificates_status`, `c`.`recommended`
                FROM Courses c
                WHERE c.name
                LIKE CONCAT(?, ?, ?) LIMIT ?
```

Además se necesita quitar el patrón `LIKE CONCAT('%', ? '%')` en favor de usar `FULLTEXT` index en `name` para búsquedas más eficientes con `MATCH ... AGAINST`

Ref:
https://github.com/omegaup/omegaup/blob/b59b3717cee2a550be32994122c7e4828c719f5b/frontend/server/src/DAO/Courses.php#L34

antedotee:
If the issue is still unresolved, I would like to take this up.

pabo99:
> If the issue is still unresolved, I would like to take this up.

Hello @antedotee. 

We've had some async discussions regarding this issue. What remains to be defined is which search approach is more efficient: using `LIKE CONCAT(...)` or `MATCH ... AGAINST`.

If you have a proposal, feel free to open the discussion here so we can decide on the optimal solution together

akhil06232:
If the issue is still unresolved, I would like to take this up.

akhil06232:
@pabo99 I already raise pr for these:https://github.com/omegaup/omegaup/pull/9407
@pabo99 could you assign to me these issue.

PR #9340, opened 2026-02-27
Add FULLTEXT index on Courses.name (Fixes #8270)

akhil06232:
## Change 1: Add FULLTEXT index on Courses.name (Fixes #8270)

Add a FULLTEXT index on `Courses.name` and switch `Courses::findByName()` 
to use `MATCH … AGAINST` in boolean mode with a trailing `*` for prefix search.  
Wildcard `LIKE` queries were preventing the database from using any index and 
scaled poorly on large course tables.  The new implementation is much faster; 
note that only word prefixes are matched (substrings inside a word no longer 
return results).

A migration file (00033_add_fulltext_index_courses_name.sql) creates the index, 
and `schema.sql` is updated accordingly for new installs.

Tests have been adjusted to exercise the new behaviour and assert that the 
index exists.

## Change 2: Fix CSP violation for golden-layout source maps (Fixes #9335)

Fixes the CSP violation that was being logged in the console when the frontend 
tried to load `https://golden-layout.com/assets/css/goldenlayout-base.css.map`.
The `connect-src` directive now allows that origin (or the source‑map reference 
is removed altogether when the file is not needed in production).

# Comments

**Change 1:** Search-by-name queries are now indexed; reviewers should pay 
attention to the boolean‑mode query and the changed matching semantics.  No UI 
changes are visible.

**Change 2:** This is purely a configuration tweak; no UI changes are visible 
to the user. The update was applied at the point where the policy is 
constructed so that other CSP headers remain unaffected. Please pay attention 
to the `csp` helper in `src/server/...` where the extra entry was added.

# Checklist:

- [x] The code follows the [coding guidelines](https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/Coding-guidelines.md) of omegaUp.
- [x] The tests were executed and all of them passed.
- [x] New tests were added where appropriate.
- [x] PR changes are appropriately scoped (database/DAO + configuration fix).

PR #9351, opened 2026-03-01
Optimize course name search with FULLTEXT index and MATCH query

akhil06232:
## Description
Optimize course name search in the Courses DAO by replacing the LIKE CONCAT('%', ?, '%') filter with a MATCH ... AGAINST (... IN BOOLEAN MODE) query, and add a FULLTEXT index on Courses.name to support that search efficiently.

## Database changes included:

New migration to add FULLTEXT index on Courses.name.
Schema updated to include the new FULLTEXT key for consistency.
Fixes: #8270

## Comments
This PR focuses only on SQL query optimization for course name lookup and the required database index.
No UI changes were made.

## Checklist:
 The code follows the [coding guidelines](vscode-file://vscode-app/c:/Users/akhil/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) of omegaUp.
 The tests were executed and all of them passed.
 If you are creating a feature, the new tests were added.
 If the change is large (> 200 lines), this PR was split into various Pull Requests. It's preferred to create one PR for changes in controllers + unit tests in PHPUnit, and then another Pull Request for UI + tests in Jest, Cypress or both.


akhil06232:
@pabo99 could you please review these pr

akhil06232:
Closes #9352
Issue: [https://github.com/omegaup/omegaup/issues/9352](vscode-file://vscode-app/c:/Users/akhil/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

## What this PR changes
->Replaces the hand-maintained transliteration map in [alias.generate.js](vscode-file://vscode-app/c:/Users/akhil/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) with the maintained [latinize](vscode-file://vscode-app/c:/Users/akhil/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) library.
->Keeps the alias-generation pipeline unchanged:
 -transliterate accents/special letters
 -replace whitespace with -
 -remove invalid alias chars
 -truncate to alias max length
->Adds stronger regression coverage in [Form.test.ts](vscode-file://vscode-app/c:/Users/akhil/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) with accented/multilingual/ligature samples.
## Why the large deletion is required
->The deleted block is a custom map (~800+ lines) that duplicates functionality provided by a maintained dependency.
->Keeping that map would continue technical debt and increase maintenance cost/risk of drift.
->This issue’s goal is specifically to remove that custom complexity and rely on maintained transliteration logic.
->Behavior compatibility was validated against the legacy map: mapped characters matched (diff count: 0 in comparison check), so deletion is structural cleanup, not feature loss.
## Why additions are required
->The test additions document expected alias behavior for multilingual/accented inputs and protect against regressions.
->They provide confidence that replacing the implementation does not change expected output for covered cases.
## Validation done
->Focused unit tests passed for [Form.test.ts](vscode-file://vscode-app/c:/Users/akhil/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html), including new alias-generation cases.
## Impact
Reduces file complexity substantially (large map removed).
Keeps existing alias behavior for tested inputs.
No intended frontend UX/API change; this is a refactor/maintenance improvement.

akhil06232:
@pabo99  could you please review these pr and provide me guidance for improvement.

PrasannaMishra001:
@akhil06232 i can see you dropped `*` for prefix matching that u added in older #9340 which was better and the changes in `alias.generate.js` & `Form.test.ts` are unrelated so you could have raised another pr if you want .other changes and split look good to me.
@pabo99 who is the maintainer and already reviewing so he will have final say on this approach.

frontend/server/src/DAO/Courses.php
```diff
@@ -32,8 +32,8 @@ public static function findByName(string $name): array {
         $sql = "SELECT DISTINCT
                 {$fields}
                 FROM Courses c
-                WHERE c.name
-                LIKE CONCAT('%', ?, '%') LIMIT 10";
+            WHERE MATCH(c.name) AGAINST (? IN BOOLEAN MODE)
```
pabo99, line 35: Here you can use the wildcard you had added in PR #9340, right?

Still, I was checking that `FULLTEXT` indexes in `BOOLEAN MODE` don't tokenize random strings (like the ones used in the tests), which causes failures.

So, it would be worth making a fallback so that, in case no records are found, it uses the method with `LIKE CONCAT(...)`.

This way, no test would break.

frontend/www/js/alias.generate.js
```diff
@@ -1,835 +1,4 @@
...
-  ᵤ: 'u',
-  ᵥ: 'v',
-  ₓ: 'x',
-};
-function latinize(str) {
-  return str.replace(/[^A-Za-z0-9\[\] ]/g, function (a) {
-    return latin_map[a] || a;
-  });
-}
+import latinize from 'latinize';
```
pabo99, line 1: It seems these changes are not related to the query optimization described in the PR. Could you move them to a separate change? That would help speed up reviews.

Also, I’ve checked that this file is already deprecated and not used anywhere in the code. So before applying the change, could you verify if it's actually being used? If it isn’t, it could be removed in favor of creating a common `generateAlias` function, which is used in `frontend/www/js/omegaup/components/group/Form.vue` and `frontend/www/js/omegaup/components/teamsgroup/FormCreate.vue`

The test could be included as part of that change.

pabo99 (commented):
Wait a minute. I’m noticing that the query being optimized in this change is only relevant in the tests (in fact, it's only used in the tests).

So, it doesn't make sense to add a `FULLTEXT` index that will just consume memory without actually being used in production.

Therefore, to keep this PR, what I would recommend is creating a normal index for `Courses.name` and having the query perform a simple search (`WHERE c.name = ?`). This way, MySQL would no longer throw false positives since it would be using a reference-based search.

Could you apply those changes?

PR #9369, opened 2026-03-02
Arreglando un par de redirects temporalmente

pabo99:
# Description

Deshabilitando la redirección de `redaccion.php`.

También se actualiza la url del concurso que está teniendo porblemas con las redireciones y usando la ruta correcta.

Estos cambios se pueden revertir una vez que los redirects funcionen correctamente


xx-Tasneem-Ahmed-xx:
I take it that the redirection issue wasn't resolved in the sandbox environment?

pabo99:
> I take it that the redirection issue wasn't resolved in the sandbox environment?

That's right. The redirection issue still isn't fixed, not in the sandbox and not in production. We'll need to dig in further to figure out what's causing it and get it properly sorted out

Issue #9372, opened 2026-03-02
[BUG]405 ‘GET not allowed’ popup when running/submitting code 


shiavm006:
## Summary

When running or submitting code in the arena, omegaUp shows a popup error:
 “This API endpoint does not allow GET requests. Use POST, PUT, or PATCH.”

even though the run/submission itself succeeds. This started after PR #9005, which added a check to reject GET requests for mutating API endpoints.

## Expected Behavior

- The run/submit should work (as it already does).
- **No error popup** should appear, as the frontend should be using POST (or the endpoint should be allowed to be called with GET if it’s read-only).

PR #9373, opened 2026-03-02
Fix false 405 errors for execute endpoints

shiavm006:
## Summary

Fixes a regression introduced by PR #9005 where users see:

 “This API endpoint does not allow GET requests. Use POST, PUT, or PATCH.”

while running/submitting code, even though the run itself succeeds.

Fixes: #9372 

image

i have tested it locally now not coming anything like it


shiavm006:
cc @pabo99

Issue #9395, opened 2026-03-02
[BUG] grader container crashes on fresh docker compose up (Permission denied)


Yashagarwal9798:
## Description

On a fresh `docker compose up`, the grader container enters an infinite restart loop with:

    mkdir: cannot create directory '/var/lib/omegaup/ephemeral': Permission denied

## Steps to reproduce

1. Clone the repo
2. Set `UID_GID=1000:1000` in `.env`
3. Run `docker compose pull && docker compose up`
4. Observe grader-1 endlessly restarting

### Root Cause Analysis: Grader Container Infinite Restart Loop

**The Problem**
On a fresh `docker compose up`, the `grader` container fails to start, immediately crashing with a `Permission denied` error and entering an infinite restart loop (0% CPU, 0B Memory). This completely blocks new developers from running the local environment.

**The Root Cause**

1. **Volume Ownership:** By default, Docker initializes the `omegaupdata` volume (mounted at `/var/lib/omegaup`) with `root:root` ownership and `755` permissions. This means non-root users cannot write to the root of this directory.
2. **Entrypoint User:** The `docker-compose.yml` explicitly set the grader container to run as `user: "${UID_GID}"` (typically `1000:1000`).
3. **The Fatal Chain (`&&`):**  added `mkdir -p /var/lib/omegaup/ephemeral && ...` to the entrypoint. Because the container is running as user `1000`, the `mkdir` command hits a `Permission denied` error when trying to write to the `root`-owned volume. Because the command is chained with `&&`, the `mkdir` failure causes the entire entrypoint to exit with code 1. 
4. **The Loop:** Docker sees the container crash and, following the `restart: always` policy, immediately attempts to restart it, creating the infinite loop.

**The Solution (Privilege Drop Pattern)**
To fix this without requiring developers to run manual `chown` commands on their host machines, we must implement a **Privilege Drop**:

* **Start as Root:** Remove the `user: "${UID_GID}"` directive so the entrypoint initially runs as `root`.
* **Fix Permissions:** Allow the `root` user to successfully run the `mkdir` command and then immediately `chown` the `/var/lib/omegaup` directory to the correct `${UID_GID}`.
* **Drop Privileges:** Use `exec setpriv --reuid=... --regid=... --clear-groups` to securely drop the container's privileges back down to the non-root user (`1000:1000`) right before executing the actual `omegaup-grader` binary.

## Expected behavior

The grader container should start successfully on a fresh Docker setup
[...]

PR #9397, opened 2026-03-02
fix: fix grader container crash loop on fresh Docker setup due to volume permissions

Yashagarwal9798:
# Description

On a fresh `docker compose up`, the grader container enters an infinite restart loop with:

```
mkdir: cannot create directory '/var/lib/omegaup/ephemeral': Permission denied
```

### Root Cause

 added `mkdir -p /var/lib/omegaup/ephemeral` to the grader entrypoint, but the command runs as user `${UID_GID}` (1000:1000). The `omegaupdata` Docker volume is initialized by Docker with `root:root` ownership (`drwxr-xr-x`), so the `mkdir` fails. the `&&` in the entrypoint makes the failure fatal, crashing the container on every startup.

### Fix

- Start the grader container as `root` (via `user: root` in compose)
- Create `/var/lib/omegaup/ephemeral` and `chown` the volume to `${UID_GID}`
- Drop privileges to `${UID_GID}` via `setpriv` before executing the grader binary

The grader process (PID 1) still runs as the non-root user (`ubuntu`, uid 1000).

### Verified

- Fresh `docker compose up` grader starts without errors
- `Ephemeral run manager ready` in grader logs
- Grader process runs as `ubuntu` (uid 1000), not root
- `/var/lib/omegaup/ephemeral` correctly owned by `ubuntu:ubuntu`
- All 8 containers running
- No impact on other compose files ([local-backend](cci:7://file:///c:/Users/BIT/Desktop/omegaup/stuff/docker/Dockerfile.local-backend:0:0-0:0), `k8s`) or CI

Fixes: #9395

# Checklist:

- [ ] the tests were executed and all of them passed.
- [ ] if you are creating a feature, the new tests were added.
- [ ] if the change is large (> 200 lines), this PR was split into various Pull Requests.


Yashagarwal9798:
@pabo99 Could you review this? It fixes a critical bug from previous pr the grader is completely broken on any fresh `docker compose up` (infinite restart loop due to volume permissions). 2-line fix, tested and verified. Thanks!

antedotee:
It would be great, I think if you can upload before vs after comparison to make sure the changes work as they are intended to be, I think.

Yashagarwal9798:
@antedotee! , here is the before and after comparison testing on a completely fresh docker volume.

###  BEFORE (Current `main` branch)
The container gets stuck in an infinite restart loop because it cannot create the ephemeral directory on the root-owned volume. 

**Container State (Restarting Loop):**

Screenshot 2026-03-04 021156


**Container Logs (Permission Denied):**

Screenshot 2026-03-04 022005



---

### AFTER (With this PR)
The entrypoint successfully uses the privilege drop pattern to fix the volume permissions, drops down to the normal `ubuntu` user, and the grader starts perfectly.

**Container State (Running & Healthy):**

Screenshot 2026-03-04 020059


**Container Logs (Grader Ready):**

Screenshot 2026-03-04 020126

Let me know if you need any other tests run!

Issue #9384, opened 2026-03-02
[BUG] Authorization bypass in `/api/run/download` using `show_diff` allows hidden test output exfiltration


pratham15541:
# Critical: Authorization bypass in `/api/run/download` using `show_diff` allows hidden test output exfiltration

## Current Behavior

A non-admin contestant can directly call the run download endpoint with `show_diff=true` and successfully download `files.zip` for their own run, even when they are not a problem admin.

This bypasses intended authorization checks and can expose outputs generated from hidden/private test cases. In practice, a contestant can submit code that echoes input and then recover hidden-case content from downloaded `.out` files.

## Expected Behavior

The backend should enforce authorization server-side and reject non-admin access to run artifact downloads, regardless of user-controlled request parameters.

Setting `show_diff=true` in the request must not disable authorization checks. Non-admin users should only be able to access data explicitly allowed by policy (e.g., sample/example visibility), never protected run artifacts.

## If necessary, add the following information (otherwise delete this section):

_Provide a link to an example, or a detailed list of steps to reproduce this bug._

    1. Create or use a problem with hidden/private test cases and set feedback mode so contestants are not expected to access all diffs.
    2. Log in as a normal contestant (not problem admin) and submit a solution that copies stdin to stdout (PoC below).
    3. Obtain the run alias (guid) from the submissions page/network requests.
    4. While authenticated as the same contestant, request:
       `/api/run/download/run_alias/[guid]/show_diff/true/`
  **guid** is found from /status/ response
    5. Observe that the server returns a zip file (`files.zip`) instead of a forbidden response.
    6. Extract `.out` files and observe hidden-case input can be accessed.

_Relevant details about the environment in which you encountered this bug:_

    * Browser name (e.g., Chrome 39, node.js 5.4): Chrome 133
    * Operating System and version (desktop or mobile): Windows 11 (desktop)
    * Link to the site where this issue occurs: https://omegaup.com/
    * Screenshot: N/A

## PoC submission code (C++):

```cpp
#include <iostream>
#include <string>

int main() {
    std::ios_base::sync_with_stdio(false);
    std::cin.tie(nullptr);

    std::string line;
    while (std::getline(std::cin, line)) {
        std::cout << line << '\n';
    }
    return 0;
}
```

<img width="1919" height="507" alt="Image" src="https://github.com/user-attachments/assets/151419e2-332a-40b7-8bf2-679b634766f3" />

[...]

PR #9400, opened 2026-03-03
Added restriction during contest to hide input test cases

pratham15541:
# Description

The backend should enforce authorization server-side and reject non-admin access to run artifact downloads, regardless of user-controlled request parameters.

Fixes: #9384 

# Checklist:

## I created contest and tested

Local development zip file => [95169cb7285434180408ad199eae35c1.zip](https://github.com/user-attachments/files/25706518/95169cb7285434180408ad199eae35c1.zip)

Production environment zip file:

[ce1f2c52163ccaffa6d8defe2523b560.zip](https://github.com/user-attachments/files/25706541/ce1f2c52163ccaffa6d8defe2523b560.zip)

### DIFF
--- a/frontend/www/js/omegaup/components/arena/RunDetailsPopup.vue
+++ b/frontend/www/js/omegaup/components/arena/RunDetailsPopup.vue
@@ -148,22 +148,22 @@
                 >{{ T.wordsDownloadCode }}</a
               >
             </li>
-            <li>
-              <a
-                v-if="data.admin"
-                class="output"
-                :href="`/api/run/download/run_alias/${data.guid}/`"
-                >{{ T.wordsDownloadOutput }}</a
-              >
-            </li>
-            <li>
-              <a
-                v-if="data.admin"
-                class="details"
-                :href="`/api/run/download/run_alias/${data.guid}/complete/true/`"
-                >{{ T.wordsDownloadDetails }}</a
-              >
-            </li>
+            <template v-if="data.admin">
+              <li>
+                <a
+                  class="output"
+                  :href="`/api/run/download/run_alias/${data.guid}/`"
+                  >{{ T.wordsDownloadOutput }}</a
+                >
+              </li>
+              <li>
+                <a
+                  class="details"
+                  :href="`/api/run/download/run_alias/${data.guid}/complete/true/`"
+                  >{{ T.wordsDownloadDetails }}</a
+                >
+              </li>
+            </template>
           </ul>
         </div>
         <div v-if="data.judged_by" class="judged_by">


frontend/tests/controllers/RunDetailsTest.php
```diff
@@ -198,6 +206,29 @@ public function testDownload() {
...
+            \OmegaUp\Controllers\Run::apiDownload(new \OmegaUp\Request([
+                'run_alias' => $runData['response']['guid'],
+                'auth_token' => $login->auth_token,
+                'show_diff' => true,
+            ]));
+            $this->fail('Should have thrown ForbiddenAccessException');
+        } catch (\OmegaUp\Exceptions\ForbiddenAccessException $e) {
+            $this->assertSame('userNotAllowed', $e->getMessage());
+        }
+    }
```
pabo99, line 246: Could you add another test with the following characteristics?

The test needs to verify that downloading submissions with `show_diff=true` is blocked during an active contest, even when the problem has `show_diff` enabled. The test should:

- Create an active contest
- Create a problem with `show_diff='all'`
- Add the problem to the active contest
- Create a contestant and add them to the contest
- Contestant opens the contest and submits a run that gets graded (AC)
- Attempt to download with `show_diff=true` during the active contest - This should throw `ForbiddenAccessException` with message `userNotAllowed`
- Fast-forward time to after the contest ends
- Attempt download again - This should succeed 

It seems that we may need to add extra validations in `apiDownload`, because in the test I ran it did allow it.

pabo99 (commented):
Thanks for finding this bug and fixing it. I left you a comment to add an extra test and apply any validations that may be needed.

frontend/tests/controllers/RunDetailsTest.php
pratham15541: @pabo99 Done. Tell me if any changes needed.

frontend/tests/controllers/RunDetailsTest.php
```diff
@@ -229,6 +229,59 @@ public function testDownloadBlockedWhenShowDiffNone() {
...
+            $problemDataWithDiff,
+            $this->contestData,
+            $this->identity
+        );
+        \OmegaUp\Test\Factories\Run::gradeRun($runData, 1, 'AC', 60);
+
+        $login = self::login($this->identity);
+
+    // During the active contest, download should be blocked
+        try {
```
pabo99, line 267:
```suggestion
        $login = self::login($this->identity);

        // During the active contest, download should be blocked
        try {
```

frontend/tests/controllers/RunDetailsTest.php
```diff
@@ -198,6 +206,82 @@ public function testDownload() {
...
+
+    public function testDownloadBlockedDuringActiveContest() {
+        $adminLogin = self::login($this->admin);
+
+        $problemDataWithDiff = \OmegaUp\Test\Factories\Problem::createProblem(
+            new \OmegaUp\Test\Factories\ProblemParams([
+            'show_diff' => 'all',
+            ]),
+            $adminLogin,
+        );
```
pabo99: I see that in more than one test a problem is created with `show_diff='all'`. It would be better to add it in setUp and then just use it in the test functions.

frontend/tests/controllers/RunDetailsTest.php
```diff
@@ -198,6 +206,82 @@ public function testDownload() {
...
+
+        \OmegaUp\Test\Factories\Contest::addProblemToContest(
+            $problemDataWithDiff,
+            $this->contestData
+        );
```
pabo99, line 255: It would also be necessary to add the problem to the contest in the `setUp`. Additionally, in the `setUp` you could perform the complete user process as well:

- Add the contestant to the contest
- Contestant opens the contest
- And maybe contestant creates a run

frontend/tests/controllers/RunDetailsTest.php
```diff
@@ -198,6 +206,82 @@ public function testDownload() {
...
+        try {
+            \OmegaUp\Controllers\Run::apiDownload(new \OmegaUp\Request([
+            'run_alias' => $runData['response']['guid'],
+            'auth_token' => $login->auth_token,
+            'show_diff' => true,
+            ]));
+        } catch (\OmegaUp\Exceptions\ExitException $e) {
+            // Expected — apiDownload always exits this way on success.
+        }
+        ob_end_clean();
```
pabo99: Could you get the contents and then add an assertion to ensure they are not empty?

pabo99 (commented):
Just a few adjustments in the tests. If the tests still pass with these changes, we can go ahead and approve this change

pratham15541:
@pabo99  Any other changes required?

Issue #9239, opened 2026-02-21
Improve visual alignment and spacing in Coders Ranking table


AMRITESH-BAKSHI:
**Description**

While reviewing the Coders with the most points ranking page, I noticed a minor visual alignment inconsistency between the # (rank) column and the Coder column.

The rank numbers appear visually detached from the coder names due to narrow column width and spacing/padding differences. This makes the table feel slightly misaligned and less polished, especially when scanning vertically.

Although functionality is unaffected, improving the alignment would enhance readability and visual consistency.

**Current Behavior**

- The # column is very narrow.
- Rank numbers appear visually separated from the coder names.
- Spacing between the # and Coder columns feels inconsistent.
- Header alignment and row alignment could be visually tighter.

**Expected Behavior**

- The rank column should have consistent spacing and alignment.
- Proper horizontal padding should visually group the rank with the coder.
- Header and row alignment should feel consistent and balanced.
- The table should look visually clean and structured when scanning top-to-bottom.

**Proposed Improvement**

- Adjust column width for the # column.
- Ensure consistent padding between columns.
- Review text alignment (e.g., right-align rank numbers if appropriate).
- Improve vertical alignment consistency between header and rows.
- These changes would be limited to UI/CSS adjustments and would not affect any ranking logic.

**Environment**

- Browser: Chrome
- OS: Windows (desktop)
- Page: Ranking → Coders with the most points

**Screenshot of the problem**

<img width="1849" height="920" alt="Image" src="https://github.com/user-attachments/assets/21a83374-4182-47d3-80eb-35c11202fd9e" />

AMRITESH-BAKSHI:
Hi 
I’ve assigned this issue to myself and would like to confirm the approach before starting implementation.
**My plan is to:**

- Adjust the width of the # (rank) column
- Improve horizontal padding between the rank and coder columns
- Ensure consistent alignment between <th> and <td>
- Limit changes strictly to CSS/UI (no structural or logic changes)

Please let me know if this approach aligns with your expectations, or if you would prefer a different solution

Yashagarwal9798:
Hi @AMRITESH-BAKSHI! Welcome, and thanks for proposing this it looks like a really great issue. Since this involves UI changes, could you please provide a quick mockup or sketch of what the final result would look like? Having a visual makes it much easier for the maintainers to review the design early on, which saves everyone time before you start writing the code. Let us know if you have any questions!

AMRITESH-BAKSHI:
[...]

PR #9419, opened 2026-03-05
Fix alignment of username and real name in ranking table

AMRITESH-BAKSHI:
# Description

This PR fixes the alignment issue in the user ranking table where the username and real name appeared on separate lines.

Previously, the real name was displayed using a `<br>` element, which caused the following elements to appear misaligned:

* Country flag
* Username
* Real name

# Solution

This change removes the `<br>` element and introduces a flex container to properly align these elements on the same line.

The following CSS classes were added to ensure correct layout and spacing:

* `coder-cell`
* `coder-column`
* `coder-name`

The flex layout ensures that the country flag, username, and real name appear horizontally aligned and vertically centered.

# Before
image

Username and real name appeared on separate lines which caused visual misalignment.

# After
final for improve ranking page

Country flag, username, and real name appear on the same line with proper spacing and vertical alignment.

# Files Modified

```
frontend/www/js/omegaup/components/user/Rank.vue
```

# Testing

* Verified locally using the omegaUp development environment.
* Confirmed that the ranking table now renders the username and real name on the same line.
* Checked that spacing and alignment remain consistent.

Fixes: #9239 

# Checklist

* [x] The code follows the coding guidelines of omegaUp.
* [x] If the change is small (< 200 lines).

### DIFF
--- a/.gha-reversemap.yml
+++ b/.gha-reversemap.yml
@@ -31,11 +31,6 @@ actions/github-script:
   sha-url: https://github.com/actions/github-script/commit/f28e40c7f34bde8b3046d885e986cb6290c5673b
   tag: v7
   tag-url: https://github.com/actions/github-script/tree/v7
-takanome-dev/assign-issue-action:
-  sha: d9515870cda43f62505e7c3617f16965f1d40df0
-  sha-url: https://github.com/takanome-dev/assign-issue-action/commit/d9515870cda43f62505e7c3617f16965f1d40df0
-  tag: beta
-  tag-url: https://github.com/takanome-dev/assign-issue-action/tree/beta
 actions/labeler:
   sha: 634933edcd8ababfe52f92936142cc22ac488b1b
   sha-url: https://github.com/actions/labeler/commit/634933edcd8ababfe52f92936142cc22ac488b1b
--- a/.github/workflows/assign-issues.yml
+++ b/.github/workflows/assign-issues.yml
@@ -1,12 +1,12 @@
-name: "Assign Issues"
+name: "Auto Assign & Unassign"
 
 on:
     issue_comment:
         types: [created]
-    schedule:
-        # Run daily to handle reminders and automatic unassignments
-        - cron: "0 3 * * *"
-    workflow_dispatch:
+
+env:
+    # Maximum number of open issues a user can have assigned in THIS repository.
+    MAX_OPEN_ASSIGNMENTS: 2
 
 # Prevent race condition when multiple users comment /assign simultaneously
 concurrency:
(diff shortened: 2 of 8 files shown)


AMRITESH-BAKSHI:
It looks like the CI failure is caused by a dependency installation error
from the npm registry (500 Internal Server Error during yarn install).

I will re-run the workflow.

frontend/www/js/omegaup/components/user/Rank.vue
```diff
@@ -163,20 +163,23 @@
...
+      :classname="user.classname"
+      :linkify="true"
+      :username="user.username"
+    ></omegaup-user-username>
+    
+    <span v-if="user.name && length !== 5" class="coder-name">
+  
+      {{ user.name }}
+    </span>
+  </div>
```
pabo99: This block looks misaligned. You can fix it by running `./stuff/lint.sh` outside the container, but inside the WSL instance if you're using Windows.

frontend/www/js/omegaup/components/user/Rank.vue
```diff
@@ -234,7 +237,7 @@ interface Rank {
   username: string;
   name?: string;
   score: number;
-  problemsSolvedUser: number;
+  problems_solved: number;
```
pabo99, line 242: Thanks for fixing this. In the future we should remove this interface and use the correct one that is autogenerated in the types. But for now, this is fine.

frontend/www/js/omegaup/components/user/Rank.vue
```diff
@@ -386,4 +389,28 @@ export default class UserRank extends Vue {
   width: 40px;
   font-size: 0.75rem;
 }
+
+.coder-column{
+  vertical-align: middle;
+}
+.coder-name {
+  margin-left: 6px;
```
pabo99: In the `coder-cell` class you are already adding a gap between each cell. Is it necessary to add even more margin to this element? 

If so, could you add it using the Bootstrap class `ml-1`? It’s the closest value to these 6 pixels.

frontend/www/js/omegaup/components/user/Rank.vue
```diff
@@ -386,4 +389,28 @@ export default class UserRank extends Vue {
...
+
+}
+.coder-cell {
+  display: inline-flex;
+  align-items: center;
+  gap: 6px;
+
```
pabo99: Here we could use relative sizes. It could be something similar to:
```suggestion
  gap: 0.25rem;
```

AMRITESH-BAKSHI:
Hi @pabo99 
Thanks for the suggestions! I’ve updated the implementation to address the feedback.

• Added a dedicated "coder-flag" wrapper with a fixed width so that the country flag space is reserved even when a user has not configured a flag, keeping usernames aligned.
• Updated the "coder-cell" layout to use flex for consistent alignment between the flag, username, and name.
• Limited the username width and added "overflow: hidden", "text-overflow: ellipsis", and "white-space: nowrap" so long usernames do not break the layout while preserving alignment.
• Increased the width of the rank column to prevent the rank values from wrapping into multiple lines.
• Replaced the fixed pixel gap with a relative value ("0.25rem") for better consistency with the existing styles.


image

Please let me know if any further adjustments are needed.

PR #9420, opened 2026-03-05
Optimize update_problem_accepted_stats to eliminate per-problem UPDATE queries

### LINKED ISSUE #9343: [FEATURE] Optimize update_problem_accepted_stats to reduce per-problem UPDATE queries
### Improve performance of `update_problem_accepted_stats` in update_ranks.py

Hi,

While going through `stuff/cron/update_ranks.py`, I noticed something in `update_problem_accepted_stats()` that might be worth revisiting from a performance perspective.

Right now, the function:

- Runs a correlated subquery to compute the number of accepted users per problem.
- Then loops over all problems in Python.
- Executes one `UPDATE` per problem.
- Commits at the end.

So effectively, if there are N problems, we end up doing N updates from Python.

This works correctly, but as the number of problems and submissions grows, this could become expensive, especially since the `Submissions` table can get quite large. It also creates many round-trips from Python to MySQL.

I was wondering whether this could be rewritten as a single set-based query that:

- Computes `COUNT(DISTINCT identity_id)` grouped by `problem_id`
- Applies the same filters (AC verdict, excluding forfeited users and owners)
- Updates `Problems.accepted` in one statement using a `LEFT JOIN`
- Sets `accepted = 0` for problems with no AC submissions

That would reduce the number of UPDATE statements from O(N) to 1, and let MySQL handle the aggregation more efficiently.

Before attempting a PR, I wanted to check:

- Is the current structure intentional (for example, to distribute load between replica and primary)?
- Are there concerns around locking or replication if we switch to a single large UPDATE?
- Would this kind of change align with the ongoing Cronjob Optimization effort?

Happy to prototype and benchmark an alternative version if this sounds reasonable.

Prince-Prajapati07:
## Summary

This PR optimizes `update_problem_accepted_stats()` in `stuff/cron/update_ranks.py` by replacing the per-problem update loop with a single set-based SQL query.

Previously, the cronjob computed accepted users per problem using a correlated subquery and then issued **one `UPDATE` per problem from Python**, resulting in **O(N) UPDATE queries** and multiple database round-trips.

This change replaces that pattern with a **single `UPDATE ... LEFT JOIN` aggregation query**, allowing MySQL to compute accepted counts in one pass and update all problems at once.

## Impact

- Eliminates the **N+1 query pattern**
- Reduces database round-trips
- Lets MySQL handle aggregation efficiently
- Keeps the same filtering logic and behavior

## Benchmark (local)

Using `EXPLAIN ANALYZE`:

Old plan: ~2.32 ms  
New plan: ~1.65 ms  

While the local dataset is empty, the query plan shows the structural improvement from **O(N) updates to a single set-based operation**.

Closes #9343


Prince-Prajapati07:
Hi @pabo99 sir! following up here. This PR implements the set-based approach we discussed in [#9343](https://github.com/omegaup/omegaup/issues/9343) (replacing the per-problem UPDATE loop). I also added a small local benchmark using EXPLAIN ANALYZE to compare the query plans.
Would be great to get your thoughts on whether this looks good to move forward, or if you'd like any changes in the query/approach.

stuff/cron/update_ranks.py
```diff
@@ -69,54 +62,51 @@ def update_problem_accepted_stats(
...
-        cur.execute(
-            '''
-                UPDATE
-                    `Problems` AS `p`
-                SET
-                    `p`.`accepted` = %s
-                WHERE
-                    `p`.`problem_id` = %s;
-            ''', (row['accepted'], row['problem_id']))
+    _ = cur_readonly
```
pabo99: If `cur_readonly` is no longer used, consider removing it

pabo99 (commented):
Excellent optimization! Verified with production data and the improvement is significant.

Production benchmarks (EXPLAIN ANALYZE):
- Old: ~30-40s (correlated subquery with 23,324 loops + N UPDATEs)
- New: ~18.5s (single materialization + one UPDATE)
- ~40-50% performance improvement

Just a minor suggestion and that's fine

Prince-Prajapati07:
Hi @pabo99 sir,
I have addressed the feedback by removing the unused `cur_readonly` and keeping the change minimal.
Could you please take a look when you get a chance?
Thanks!

pabo99 (approved):
Great!

pabo99:
Can you take a look at the conflict? Once it's resolved, we can merge this PR

PR #9431, opened 2026-03-06
[FEATURE] Improve readability of problem statements when using classroom projectors

### LINKED ISSUE #8907: [FEATURE] Improve readability of problem statements when using classroom projectors
When projecting problem statements in class, it is necessary to zoom in so students can read the text clearly. However, zooming in causes several UI elements to become obtrusive:
- The navbar menu grows disproportionately and takes up too much space.
- The “scroll to top” button overlaps the content and distracts from reading.
- Hiding the button after inactivity is not effective, since scrolling happens constantly while reading.
- The constant appearance/disappearance of elements is distracting during lectures.

Image

## Possible solution

Explore solutions that improve readability without requiring zoom (e.g., adjustable font size or presentation mode)

## Context

Reported by @darkmx, who regularly projects omegaUp in class. Current UI elements interfere with readability, making the experience less effective for teaching

Prashant-thakur77:
# Description

Add Presentation Mode for problem statements to improve readability during classroom projections (Fixes #8907).

Changes made:

frontend/www/js/omegaup/components/problem/Print.vue
1) Added a presentation controls toolbar with:
2)Font size decrease (-) and increase (+) buttons (default 18px, range 12–36px)
3) A toggle button to show/hide the problem limits table

frontend/www/js/omegaup/components/problem/SettingsSummary.vue
1) Added a "Classroom View" projector icon link  next to the edit icon in the problem title header, which links to the printable/presentation version of the problem. Controlled by a new showClassroomViewLink prop (default true; set to false inside 

problemClassroomView
problemPresentationModeFontDec
problemPresentationModeFontInc
problemPresentationModeToggleLimits
Fixes: #8907 
image
Screenshot From 2026-03-06 11-04-23

# Comments

# Checklist:


Prashant-thakur77:
hlo @pabo99 I have created thr pr  and i would also love to participate in gsoc 2026,i have joined the discord and currenlty doing the gsoc 2026 test .do tell if there are any other requirements i need to follow.Would love to be part of this org ;)

pabo99:
Hello @Prashant-thakur77 ! Glad to have you back!

The only thing we’re missing is for you to share your omegaUp username. That way, once you complete 2 out of the 3 problems, we can start reviewing this PR and any others you create.

Prashant-thakur77:
Prashant_thakur77         this is my username in omegaup.thanks for the update.glad to be a part ;)

Prashant-thakur77:
Hlo @pabo99 i have solved 2 problems in the contest;) thanks for your info regarding this.

frontend/www/js/omegaup/components/problem/Print.vue
```diff
@@ -1,12 +1,49 @@
 <template>
-  <div class="mt-4">
-    <a href="" @click="() => $emit('print-page')">
-      <font-awesome-icon
-        :title="T.contestAndProblemPrintButtonDesc"
-        :icon="['fas', 'print']"
-    /></a>
+  <div class="mt-4 presentation-wrapper" :style="{ fontSize: fontSize + 'px' }">
```
pabo99: Okay, the Classroom mode view should have its own component so it doesn’t affect the print view.

What you could do is check whether both will share many things in common. If so, we could create a parent component with slots to include the parts that change between them.

Due to time constraints, you can focus on moving this to a new component for now, and later we can look into reusing code

frontend/www/js/omegaup/components/problem/Print.vue
```diff
@@ -1,12 +1,49 @@
...
+          :title="T.contestAndProblemPrintButtonDesc"
+          :icon="['fas', 'print']"
+      /></a>
+
+      <div class="d-flex align-items-center mr-3">
+        <span class="mr-2 text-muted small">{{ T.fontSize }}:</span>
+        <button
+          class="btn btn-sm btn-outline-secondary mr-1"
+          :title="T.problemPresentationModeFontDec"
+          :disabled="fontSize <= 12"
```
pabo99: All these arbitrary values should be added as constants, and preferably these operations should be obtained through computed properties.

This way it becomes easier to modify them if necessary, instead of having to search through the templates for all the places where they are used.

frontend/www/js/omegaup/components/problem/Print.vue
```diff
@@ -1,12 +1,49 @@
...
+          @click="fontSize = Math.max(12, fontSize - 2)"
+        >
+          <font-awesome-icon :icon="['fas', 'search-minus']" />
+        </button>
+        <span class="px-2">{{ fontSize }}px</span>
+        <button
+          class="btn btn-sm btn-outline-secondary ml-1"
+          :title="T.problemPresentationModeFontInc"
+          :disabled="fontSize >= 36"
```
pabo99: From what I recall, when using a projector the professor needs to zoom up to 300%, so a 36px font size is still too small. Could we use a range with higher values? For example, from 48px to 76px, increasing in 6px steps

frontend/www/js/omegaup/components/problem/Print.vue
```diff
@@ -1,12 +1,49 @@
...
-    /></a>
+  <div class="mt-4 presentation-wrapper" :style="{ fontSize: fontSize + 'px' }">
+    <div
+      class="presentation-controls d-flex align-items-center flex-wrap mb-3 p-2 bg-light border rounded"
+    >
+      <a href="" class="mr-3" @click.prevent="$emit('print-page')">
+        <font-awesome-icon
+          :title="T.contestAndProblemPrintButtonDesc"
+          :icon="['fas', 'print']"
+      /></a>
```
pabo99: This view shouldn’t include anything related to printing; Could you remove this and the associated action?

frontend/www/js/omegaup/components/problem/Print.vue
pabo99: Is there a way to avoid using inline styles? I’m thinking of using classes for the available font sizes or something more dynamic, but without using the `style` attribute.

frontend/www/js/omegaup/components/problem/Print.vue
```diff
@@ -1,12 +1,49 @@
...
+        <span class="mr-2 text-muted small">{{ T.fontSize }}:</span>
+        <button
+          class="btn btn-sm btn-outline-secondary mr-1"
+          :title="T.problemPresentationModeFontDec"
+          :disabled="fontSize <= 12"
+          @click="fontSize = Math.max(12, fontSize - 2)"
+        >
+          <font-awesome-icon :icon="['fas', 'search-minus']" />
+        </button>
+        <span class="px-2">{{ fontSize }}px</span>
```
pabo99: We should find a way to exclude this element from the class that changes the font size, because it also starts growing and shifts the element.

In fact, this could be placed outside the presentation-wrapper so that it stays isolated in a spot where it’s less visible.

frontend/www/js/omegaup/components/problem/Print.vue
```diff
@@ -1,12 +1,49 @@
...
+        :title="T.problemPresentationModeToggleLimits"
+        @click="showLimits = !showLimits"
+      >
+        <font-awesome-icon :icon="['fas', 'table']" />
+        {{ T.problemPresentationModeToggleLimits }}
+      </button>
+    </div>
+
     <omegaup-problem-settings-summary
+      v-show="showLimits"
```
pabo99: I would like the limits table to have a fixed size and only change with the problem statement; we could try 36px. Additionally, we could try adding different styles for this view in particular, since here we use the full available screen space

frontend/www/js/omegaup/components/problem/Print.vue
```diff
@@ -40,6 +82,8 @@ export default class ProblemPrint extends Vue {
   @Prop() problem!: types.ProblemInfo;
 
   T = T;
+  fontSize = 18;
+  showLimits = true;
```
pabo99: By default, could the limits table be hidden?

In fact, we could just show a text indicating "Limits Table" that expands when clicked, giving us more space on the screen.

pabo99 (commented):
We should evaluate whether, for this particular view, we use a different container in the main tag, since it has a maximum width and here it would be worth using the full available space.

Image

We should also consider hiding the "Scroll to top" element, as requested, because it is still appearing:

Image

frontend/www/js/omegaup/components/problem/Print.vue
Prashant-thakur77: done

Prashant-thakur77:
@pabo99 I have a doubt regarding the scroll-to-top button. In 

template.tpl, the 
>{% jsInclude 'common_scroll_to_top' %} 

and its div id="scroll-to-top" are rendered unconditionally, even when hideFooterAndHeader is true. This means the scroll-to-top button was appearing on the presentation page (and the existing print page too).

To fix it, I wrapped it in a {% if not hideFooterAndHeader %} check, same as the navbar and footer:

>{% if not hideFooterAndHeader %}
  {% jsInclude 'common_scroll_to_top' omitRuntime %}
 > div id="scroll-to-top"></div
>{% endif %}

This affects all pages that set hideFooterAndHeader: true (currently: problem print, contest print, grader, and the new presentation page). Is this the right approach, or would you prefer a different solution — such as hiding it only for the presentation page specifically?

frontend/www/js/omegaup/components/problem/Print.vue
Prashant-thakur77: done

frontend/www/js/omegaup/components/problem/Print.vue
Prashant-thakur77: range kept form 48px to 78 px ,a 6px increment on each.

frontend/www/js/omegaup/components/problem/Print.vue
Prashant-thakur77: done

frontend/www/js/omegaup/components/problem/Print.vue
Prashant-thakur77: CSS classes .presentation-font-48 through .presentation-font-78 used instead of :style , @pabo99 is this change fine?

frontend/www/js/omegaup/components/problem/Print.vue
Prashant-thakur77: done

frontend/www/js/omegaup/components/problem/Print.vue
Prashant-thakur77: done kept at 36px

frontend/www/js/omegaup/components/problem/Print.vue
Prashant-thakur77: done

Prashant-thakur77:
> We should evaluate whether, for this particular view, we use a different container in the main tag, since it has a maximum width and here it would be worth using the full available space.

For this i have added full-width=true to controller for the presentation mode.

Prashant-thakur77:
after updates:
[Screencast From 2026-03-07 10-22-52.webm](https://github.com/user-attachments/assets/630abd67-0d57-445d-95c3-cd880aaff814)

Prashant-thakur77:
@pabo99 I have made the asked changes do tell if any other changes are required(specially regarding the scroll to top button) as we can also use mounted and destroyed lifecycle hooks to hide the scroll to top specifically in presentation mode

Prashant-thakur77:
Hello @pabo99 ,I am very interested on working on the task for vue2 to vue 3 migration.

Would it be okay if I opened a tracking issue to outline a phased migration strategy using `@vue/compat`? I'd also love to pick up a small starter task (tooling update or an isolated component) to start contributing toward this goal.
Happy to discuss the approach before opening anything !

frontend/server/src/Controllers/Problem.php
```diff
@@ -6878,6 +6878,76 @@ public static function getProblemPrintDetailsForTypeScript(\OmegaUp\Request $r)
...
+
+        return [
+            'templateProperties' => [
+                'payload' => [
+                    'details' => $details,
+                ],
+                'hideFooterAndHeader' => true,
+                'fullWidth' => true,
+                'title' => new \OmegaUp\TranslationString(
+                    'omegaupTitleProblemPrint',
```
pabo99, line 6941: It will be necessary to add an entry in the localization files for `omegaupTitleProblemPresentationMode` (or similar) in order to use it here

frontend/templates/en.lang
```diff
@@ -1187,6 +1187,7 @@ problemBookmarkAdd = "Add bookmark"
 problemBookmarkAdded = "Bookmark added successfully"
 problemBookmarkRemove = "Remove bookmark"
 problemBookmarkRemoved = "Bookmark removed successfully"
+problemClassroomView = "Classroom View"
```
pabo99, line 1190: We’ll need to decide which term fits better:

_**Presentation mode**_ or **_Classroom view_**

frontend/templates/es.lang
```diff
@@ -1421,6 +1422,9 @@ problemNotFound = "Problema no encontrado"
 problemNotFoundInContest = "Problema no encontrado en concurso"
 problemNotFoundInProblemset = "Problema no encontrado en conjunto de problemas"
 problemNotPartOfAssignment = "El problema no es parte del contenido."
+problemPresentationModeFontDec = "Disminuir tamaño de fuente"
+problemPresentationModeFontInc = "Aumentar tamaño de fuente"
+problemPresentationModeToggleLimits = "Alternar tabla de límites"
```
pabo99, line 1427: In Spanish, this term doesn’t tell us much. You can use the following instead:
```suggestion
problemPresentationModeToggleLimits = "Mostrar/Ocultar tabla de límites"
```

frontend/templates/pt.lang
```diff
@@ -1421,6 +1422,9 @@ problemNotFound = "Problema não encontrado"
 problemNotFoundInContest = "Problema não encontrado neste concurso"
 problemNotFoundInProblemset = "Problema não encontrado neste conjunto do problemas"
 problemNotPartOfAssignment = "O problema não faz parte do conteúdo."
+problemPresentationModeFontDec = "Diminuir tamanho da fonte"
+problemPresentationModeFontInc = "Aumentar tamanho da fonte"
+problemPresentationModeToggleLimits = "Alternar tabela de limites"
```
pabo99, line 1427: Same as the Spanish translation
```suggestion
problemPresentationModeToggleLimits = "Mostrar/Ocultar tabela de limites"
```

frontend/templates/template.tpl
```diff
@@ -62,8 +62,10 @@
       <script async src="https://www.googletagmanager.com/gtag/js?id=G-PBDCQK1GEQ"></script>
       <script type="text/javascript" src="{% versionHash '/js/analytics.js' %}"></script>
     {% endif %}
+    {% if not hideFooterAndHeader %}
```
pabo99, line 65: Indeed, this will hide the element in all views that implement `hideFooterAndHeader`.

If you prefer, to avoid conflicts, you can revert the changes in this file, since there is already a proposal to remove the `scrollToTop` button completely

frontend/www/js/omegaup/components/problem/PresentationMode.vue
```diff
@@ -0,0 +1,163 @@
...
+}
+.presentation-font-66 {
+  font-size: 66px;
+}
+.presentation-font-72 {
+  font-size: 72px;
+}
+.presentation-font-78 {
+  font-size: 78px;
+}
```
pabo99, line 162: I recently found a way to make this more fancy using `SCSS`. Do you want to try adding this in a new file that can be imported from `main.scss`?
```suggestion
$font-start: 48;
$font-end: 78;
$font-step: 6;

@for $i from $font-start through $font-end {
  @if ($i - $font-start) % $font-step == 0 {
    .presentation-font-#{$i} {
      font-size: #{$i}px;
    }
  }
}
```

Or, if we prefer to use constant values, it could look like this:

```suggestion
$font-sizes: 48, 54, 60, 66, 72, 78;

@each $size in $font-sizes {
  .presentation-font-#{$size} {
    font-size: #{$size}px;
  }
}
```

This way we keep compatibility with the constants in the script.

frontend/www/js/omegaup/components/problem/SettingsSummary.vue
```diff
@@ -34,6 +34,15 @@
       <a v-if="showEditLink" :href="`/problem/${problem.alias}/edit/`">
         <font-awesome-icon :icon="['fas', 'edit']" />
       </a>
+      <a
```
pabo99, line 37: Oh, I see now why there are changes in this file! Strictly speaking, this component shouldn’t be moved for this change. But it turns out that the problem title is displayed inside it.

We should look for a way (in a future change) to move it out of here and place it in the `ProblemDetails` component so that all the actions we’re implementing can be added there

frontend/www/js/omegaup/components/problem/SettingsSummary.vue
```diff
@@ -139,6 +150,7 @@ export default class ProblemSettingsSummary extends Vue {
   @Prop({ default: false }) userLoggedIn!: boolean;
   @Prop({ default: false }) isBookmarked!: boolean;
   @Prop({ default: false }) inContestOrCourse!: boolean;
+  @Prop({ default: true }) showClassroomViewLink!: boolean;
```
pabo99, line 153: This will cause the presentation mode icon to appear in all components that use it, even in the print view.

We could keep it hidden by default and, for now, only enable it in `Details.vue` (and possibly in `Detailsv2.vue`). Either way, that covers the Standalone Problem, Courses, and Contests views

frontend/server/src/Controllers/Problem.php
```diff
@@ -6878,6 +6878,76 @@ public static function getProblemPrintDetailsForTypeScript(\OmegaUp\Request $r)
         ];
     }
 
+    /**
+     * @return array{entrypoint: string, templateProperties: array{payload: ProblemPrintDetailsPayload, title: \OmegaUp\TranslationString}}
```
pabo99, line 6882: Even though they seem similar, could you add a new type for the payload that will be obtained here?

Perhaps in the future each one will have different properties

pabo99 (commented):
Excellent, it looks much better this way!

I left you a few suggestions to improve it a little more

pabo99:
> Hello @pabo99 ,I am very interested on working on the task for vue2 to vue 3 migration.
> 
> Would it be okay if I opened a tracking issue to outline a phased migration strategy using `@vue/compat`? I'd also love to pick up a small starter task (tooling update or an isolated component) to start contributing toward this goal. Happy to discuss the approach before opening anything !

That sounds great! Working on an isolated component is a good way to start contributing and get familiar with `@vue/compat`.

Feel free to open a tracking issue with your migration strategy, and we can review it together later

Prashant-thakur77:
Sure @pabo99

frontend/templates/en.lang
Prashant-thakur77: I suppose we should go with presentation mode?what are your thoughts about it?

frontend/templates/template.tpl
Prashant-thakur77: I will remove this from here then..

frontend/www/js/omegaup/components/problem/PresentationMode.vue
Prashant-thakur77: Sure will work on it .

frontend/www/js/omegaup/components/problem/SettingsSummary.vue
Prashant-thakur77: Yeah , would love to do that (will raise it in coming issues as you proposed it as a future change)? What are your thoughts about it.

pabo99: Excellent idea!

PR #9432, opened 2026-03-06
fix #9386: return contest_alias in clarification create response

### LINKED ISSUE #9386: [BUG] Clarification::apiCreate always returns null for contest_alias
## Current Behavior

POST /api/clarification/create/ always returns contest_alias: null in the response, even when the clarification was created for a specific contest. The variable $contestAlias is set correctly inside the function but is never written into the return array (frontend/server/src/Controllers/Clarification.php line 211 hardcodes null).

## Expected Behavior

The response should include the actual contest alias when the clarification belongs to a contest, so callers can identify which contest the clarification was created for without needing a separate fetch.

## If necessary, add the following information

Steps to reproduce:

1. Log in as any user
2. POST /api/clarification/create/ with a valid contest_alias, problem_alias, and message
3. Inspect the response - contest_alias field is always null

The frontend currently works around this by ignoring the create response and refreshing the full list, so there is no visible breakage now.

PrasannaMishra001:
# Description

`POST /api/clarification/create/` was always returning `contest_alias: null` in its response because the return statement had `'contest_alias' => null` hardcoded instead of using the `$contestAlias` variable that was already being populated from the request.

https://github.com/user-attachments/assets/a101b206-de8f-4a46-b244-f0106892df09

Fixed by returning `$contestAlias` (which is `null` for course clarifications and the actual alias string for contest clarifications).

Added a regression test `testCreateClarificationReturnsContestAlias` to verify the contest alias is returned correctly.

Fixes: #9386

# Checklist:


PrasannaMishra001:
@pabo99, I have also attached the video in this PR for before and after changes and verification. Please review these changes whenever you have time.

frontend/server/src/Controllers/Clarification.php
```diff
@@ -208,7 +208,7 @@ public static function apiCreate(\OmegaUp\Request $r): array {
             'answer' => $clarification->answer,
             'author' => $r->identity->username,
             'clarification_id' => intval($clarification->clarification_id),
-            'contest_alias' => null,
+            'contest_alias' => $contestAlias,
```
pabo99: Looks like this change was added in PR #5212 when the course clarifications endpoint was introduced. It makes sense it was added as `null`, but after ~4 years without needing to update it, I think the best approach is to remove this field from the endpoint signature. Clarifications can only be created from the same problemset the user is in, so returning it doesn’t add value.

Another option would be to include the `assignment_alias` or the `problemset` if created in a course, but I’d prefer to remove unused fields.

If you remove it and verify nothing breaks, we can merge faster.

pabo99 (commented):
One minor suggestion!

frontend/server/src/Controllers/Clarification.php
PrasannaMishra001: ok, removed `contest_alias` and also changed psalm inconsistency.

PrasannaMishra001:
Rebased onto latest `main` to resolve stale branch. @pabo99, when you have time, could you please take another look? Thanks!

pabo99 (approved):

Issue #9433, opened 2026-03-06
[BUG] Private clarification receiver can’t open details (receiver_id not checked in canViewClarification)


shiavm006:
Receivers of private clarifications cannot view clarification details, even when they are explicitly invited via receiver_id.

Looks like an auth inconsistency between listing and details endpoint.

## Current Behavior

<img width="896" height="204" alt="Image" src="https://github.com/user-attachments/assets/49b317fd-b55f-402f-96e2-f47784f16d3a" />

- Clarification::apiDetails() checks private access using Authorization::canViewClarification().
- canViewClarification() currently allows:
author
problemset admin
course TA
- It does not allow receiver_id.
## Expected Behavior

If user is receiver_id of a private clarification, user should be allowed to view that clarification’s details.

shiavm006:
cc @pabo99 just check it out once and pls assign it

pabo99:
Done!

PR #9457, opened 2026-03-07
Fix clarification view permissions for invited receivers

shiavm006:
# Description
Private clarifications can currently be viewed by the author and by problemset/course admins, but not by the **invited receiver**, even though the DAO queries and comments indicate receivers should have access.

Fixes: #9433 

Tested and its working as expected
Screenshot 2026-03-07 at 4 35 31 PM


shiavm006:
cc @pabo99  i think the solution is correct have tested it locally else would love to know ur thought on it

pabo99 (approved):
It looks like at some point we stopped using  `Clarification::apiDetails` endpoint in the production environment, and now it’s only used in tests. But the change is still good in case we decide to use it again at some point.

Thanks for fixing it.

Issue #9439, opened 2026-03-06
[BUG] Fix login redirect after logging out


pabo99:
## Current Behavior
When a user logs out, they are redirected to the homepage. However, the URL might still contains the `logout` path when the platform responds slow.

If the user immediately clicks **Log in**, the current URL is passed as the redirect parameter.

After logging in successfully, the system redirects the user to the `logout` route again, effectively logging them out and returning them to the homepage.

## Expected Behavior
After logging out, the login flow should not redirect back to the `logout` route.

The user should be able to log in normally and remain authenticated.

## Steps to Reproduce
This issue is easier to reproduce when the page loads slowly.

1. Log out of your account.
2. Immediately click **Log in** from the homepage.
3. After logging in, notice that you are redirected to the `logout` route and returned to the homepage without being logged in.

If it is difficult to reproduce due to normal page speed, you can simulate the behavior by manually adding `logout` as the redirect parameter in the URL.

For example:

https://github.com/user-attachments/assets/1d41a4d3-c0ed-4d25-be1f-96a33e1c3580

Rudrika-Singhal:
Hi! 
I'd love to work on this issue.
Could you please assign it to me if it's still available? Thank you!

PR #9459, opened 2026-03-07
Fix login redirect after logging out

Prashant-thakur77:
# Description

After a user logs out and immediately clicks "Log In" (before the page fully redirects), the current URL still contains /logout/. This path gets passed as the redirect parameter to the login form. After logging in successfully, the system would redirect the user back to /logout/, immediately logging them out again and returning them to the homepage without being authenticated.

Fix: In signin.ts, the redirect()  function now checks if the redirect target is /logout or /logout/ and falls back to / (homepage) in that case , the same way it already handles invalid or relative paths.
[Screencast From 2026-03-07 16-39-55.webm](https://github.com/user-attachments/assets/99829613-a389-4cd3-b496-19862e2d2b27)

Fixes: #9439 

# Checklist:


pabo99 (approved):
Excellent! Thank you, @Prashant-thakur77!

PR #9471, opened 2026-03-07
Fix: Browser Back/Forward buttons do not update active tab on Edit Course page

### LINKED ISSUE #9445: [Bug] Browser back button does not update displayed tab on edit course page
## Current Behavior
When navigating between tabs in the **Edit course** page, using the browser's **Back button** does not update the visible tab correctly.

## Expected Behavior
Using the browser's Back button should correctly update the displayed tab.

## Additional Context
A similar issue was previously fixed in another part of the platform.

Reference issue:
#9392

Reference PR:
#9396

The solution implemented there can be used as a reference.

abhinavsingh-hub:
## Description

Fixes an issue where using the browser Back/Forward buttons on the **Edit Course** page updates the URL hash but does not update the visible tab.

The component initializes the tab state from `initialTab`, and tab changes triggered via click handlers correctly update `showTab`. However, when navigating using the browser history (Back/Forward), only the URL hash changes while the component state remains unchanged, causing the UI to stay on the last viewed tab.

## Changes
- Added a `hashchange` event listener to synchronize `showTab` with the URL hash when the user navigates using browser Back/Forward buttons.
- Initialized the active tab from the URL hash when the component is mounted.
- Removed the event listener in `beforeDestroy` to prevent potential memory leaks.

## Additional Fix
While implementing the fix, it was noticed that the `archive` tab exists in the template but was missing from the `availableTabs` list. This prevented proper validation when syncing the tab state with the URL hash. The `archive` tab has been added to `availableTabs` to ensure consistent behavior.

## Reference
This follows the same pattern implemented in #9396 for handling hash-based tab navigation.

Fixes #9445

## Screen Recording of Fix

https://github.com/user-attachments/assets/d1e803c1-6125-479e-8794-cb42197ded2d

## Checklist

- [x] The code follows the existing code style of the project
- [x] The changes were tested locally
- [x] `npm run lint` was executed and passes
- [x] The change is limited to the scope of the issue
- [x] The PR description clearly explains the change


abhinavsingh-hub:
@pabo99 I have implemented the solution and also checked with ./stuff/lint.sh , please go through it once and let me know if anything is ought to be altered
Regards

abhinavsingh-hub:
Thanks for the suggestion! I moved the hashchange handling to edit.ts following the same pattern used in profile.ts. I also removed the listener from the Vue component and verified the behavior locally. Please let me know if any further adjustments are needed. Apologies for the delay.
Regards

abhinavsingh-hub:
Hey @pabo99!

In my last push, CI was failing in the IDE Cypress test because the test attempted to upload the downloaded ZIP before it was fully written to `cypress/downloads/` in the CI environment.

Added a `cy.readFile(filePath, { timeout: 15000 })` before `selectFile` so Cypress waits for the download to complete before uploading it. This preserves the original test flow (download → upload → verify ZIP contents) while making it stable in CI.

PR #9474, opened 2026-03-07
Optimize scoreboard performance using SQL Window Functions and simpli…

### LINKED ISSUE #9472: Performance: Enhance Scoreboard Performance using SQL Window Functions
## Description:

Currently, the scoreboard generation in Scoreboard.php and Runs.php fetches all submissions for a problemset and processes them within PHP to find the best run per user. This leads to high memory consumption and slow response times in large contests with thousands of submissions.

## Proposed Solution:
    1. SQL Optimization: Refactored Runs::getProblemsetRuns to use MySQL 8.0 Window Functions (ROW_NUMBER()). This allows the database to return only the best run per identity per problem, significantly reducing the data transferred to PHP.
    2. Logic Simplification: Simplified the loops in Scoreboard::getScoreboardFromRuns and Scoreboard::calculateEvents to reflect that incoming data is already pre-filtered and ranked by the database.

## Database Indexing:
To maximize the performance of the PARTITION BY and ORDER BY operations within the Window Functions, I recommend adding a composite index on the Submissions table. This will allow the database engine to perform an index scan instead of a full table scan:

SQL
ALTER TABLE Submissions 
ADD INDEX idx_problemset_ranking (problemset_id, status, type, identity_id, problem_id, submission_id);

## Expected Impact:
    1. Time Complexity: Reduced from $O(N)$ processing in PHP (where $N$ is total runs) to $O(M)$ (where $M$ is total unique student-problem pairs).
    2. Memory Footprint: Significant reduction in RAM usage for the web server during scoreboard updates.

## Update:
I have refactored the SQL queries in getProblemsetRuns to ensure full compatibility with MySQL's only_full_group_by mode, which was causing the CI failures.
    - The optimized path now uses a subquery to rank submissions before selecting specific columns.
    - The standard path has also been updated to avoid incorrect grouping.
    - All scoreboard-related calls now correctly utilize the onlyBest parameter for maximum performance.

Wafaamohamed2:
## Description

This PR optimizes the scoreboard generation process by offloading data ranking and filtering to the database layer. By utilizing MySQL 8.0 Window Functions (ROW_NUMBER()), we now fetch only the best submission per user per problem, significantly reducing memory consumption and improving response times for large contests.

## Changes:
    - Modified Runs::getProblemsetRuns to use a CTE and ROW_NUMBER() for efficient data retrieval.
    - Refactored Scoreboard::getScoreboardFromRuns and Scoreboard::calculateEvents to simplify processing logic.
    - Removed redundant $O(N)$ comparisons in PHP that are now handled by SQL.

Fixes: #9472

## Comments:
     - Reviewer Note: Please pay attention to the new SQL query in Runs::getProblemsetRuns. It assumes MySQL 8.0 support (already used in other DAO files like UserRank.php). Also, I recommend adding a composite index on the Submissions table to further optimize this query's performance.

## Checklist

[x] The code follows the coding guidelines of omegaUp.

[ ] The tests were executed and all of them passed. 

[ ] If you are creating a feature, the new tests were added.

[ ] If the change is large (> 200 lines), this PR was split.

### DIFF
--- a/frontend/server/src/DAO/Runs.php
+++ b/frontend/server/src/DAO/Runs.php
@@ -177,7 +177,7 @@ class Runs extends \OmegaUp\DAO\Base\Runs
                 "EXECUTION_FINISHED"
             )
         LIMIT 1
-            ), "EXECUTION_COMPILATION_ERROR")AS execution,
+            ), "EXECUTION_COMPILATION_ERROR") AS execution,
             COALESCE(
         ( SELECT
             IF(
@@ -193,7 +193,7 @@ class Runs extends \OmegaUp\DAO\Base\Runs
         ORDER BY
             field(status_runtime, "RUNTIME_NOT_AVAILABLE", "RUNTIME_EXCEEDED", "RUNTIME_AVAILABLE")
                 LIMIT 1
-            ), "RUNTIME_NOT_AVAILABLE" )AS status_runtime,
+            ), "RUNTIME_NOT_AVAILABLE") AS status_runtime,
             COALESCE(
         ( SELECT
             IF(
@@ -209,7 +209,7 @@ class Runs extends \OmegaUp\DAO\Base\Runs
         ORDER BY
             field(status_memory, "MEMORY_NOT_AVAILABLE", "MEMORY_EXCEEDED", "MEMORY_AVAILABLE")
         LIMIT 1
-            ), "MEMORY_NOT_AVAILABLE" )AS status_memory';
+            ), "MEMORY_NOT_AVAILABLE") AS status_memory';
     }
 
     /**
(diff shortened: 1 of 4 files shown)


pabo99 (commented):
Hmm, the errors seem to be legit. So to focus on the relevant changes, could you run `./stuff/lint.sh` outside the container but inside the WSL instance?

That will fix the linter errors and reduce the number of changes made in the PR. This way, we can concentrate on the changes to the queries.

pabo99 (commented):
This PR might not be the best fit for GSoC given its complexity and how disruptive it could be. The review process may take a significant amount of time, which is currently limited.

I would recommend focusing on issues that are easier to reproduce and resolve, so they can better contribute to your progress

Wafaamohamed2:
> This PR might not be the best fit for GSoC given its complexity and how disruptive it could be. The review process may take a significant amount of time, which is currently limited.
> 
> I would recommend focusing on issues that are easier to reproduce and resolve, so they can better contribute to your progress

Thank you for the feedback. I understand this PR might be too complex for the current timeline. I'll focus on simpler and easier to reproduce issues for now, and come back to this when there's more time for review.

PR #9475, opened 2026-03-08
added color scheme for the verdicts

### LINKED ISSUE #9426: [BUG] Incorrect color scheme used in verdicts section of the profile
## Current Behavior
In competitive programming, Green is used to signify AC and any other error we can use red color
But on my profile WAs are shown in green color. Currently we are using Blue for ACs and Green for WAs.
I understand this is not technically a bug, but it does feel a bit out of picture.
Is there some specific color scheme we are using strictly?

Image

## Expected Behavior

I feel AC should be green, WA be red, other errors can get yellow/black shade. 

## Steps to reproduce this bug
	1. Just navigate to the my profile section on omegaup after logging in and scroll down to the bar graph.
	
Relevant details about the environment in which I encountered this bug:
	* Browser name: Chrome
	* Operating System: Windows 11
	* Link to the [site](https://omegaup.com/profile/#charts) where this issue occurs.

pranav-kukreja:
# Description
Added color scheme for the 5 verdict categories AC, RTE, WA, PA, MLE

Fixes: #9426 

# Comments
I have chosen colors according to my understanding, let me know if something needs to be changed. 
I have not been able to test this feature on the local because the local doesn't reflect data on the graph as this is a dummy user with no real problem solving stats, I am not aware of how developers test the features on omegaup. If you can help me with some docs/resources to follow, I will be happy to test this as well.

### DIFF
--- a/frontend/www/js/omegaup/components/user/Chartsv2.vue
+++ b/frontend/www/js/omegaup/components/user/Chartsv2.vue
@@ -1,4 +1,4 @@
-<template>
+﻿<template>
   <div class="card-body text-center">
     <label class="pr-4"
       ><input v-model="type" type="radio" value="delta" />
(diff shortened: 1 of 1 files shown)


pabo99 (commented):
Could you share a screenshot so we can see the change you are proposing? Usually, when UI changes are introduced, it is important that the PR description includes either a screenshot or a screencast, depending on the change being made. Since all submissions in the local environment will result in AC, you can update some runs with different verdicts in the database directly for this case.

Also, you can run `./stuff/lint.sh` from outside the container, but inside the WSL instance, to fix the errors that appear in the workflows.

pranav-kukreja:
yes I understand
I definitely need to test this atleast locally
I will figure this out and try to come up with changes by tomorrow.

PR #9479, opened 2026-03-08
Use FULLTEXT MATCH/AGAINST for problem title/alias search

### LINKED ISSUE #8273: [FEATURE] [SQL Queries Optimization] Usar FULLTEXT index para búsquedas en `title` y `alias` de la tabla `Problems`
Sacar provecho al índice `FULLTEXT` haciendo `WHERE MATCH... AGAINST` en lugar de `LIKE CONCAT(...)` en la consulta:

```sql
SELECT COUNT(*) 
                FROM
                    Problems p
                LEFT JOIN (
                    SELECT
                        Submissions.problem_id,
                        MAX(Runs.score) AS score
                    FROM
                        Submissions
                    INNER JOIN
                        Runs ON Runs.run_id = Submissions.current_run_id
                    WHERE
                        Submissions.identity_id = ?
                    GROUP BY
                        Submissions.problem_id
                    ) ps ON ps.problem_id = p.problem_id 
WHERE
(
                    p.title LIKE CONCAT(?, ?, ?) OR
                    p.alias LIKE CONCAT(?, ?, ?) OR
                    p.problem_id = ?
                    ) AND p.visibility > -?
```

Y de ser posible, separar casos con UNION para evitar filtros con indices multiples.

Ref:
https://github.com/omegaup/omegaup/blob/b59b3717cee2a550be32994122c7e4828c719f5b/frontend/server/src/DAO/Problems.php#L262

amanpandey2587:
# Description
I have replaced LIKE CONCAT('%', ?, '%') search on title/alias with MATCH(p.title, p.alias) AGAINST (? IN BOOLEAN MODE) in byIdentityType(), along with preservingnumeric query behavior by keeping p.problem_id = ? as an exact-match fallback for numeric input and added query curation (preg_replace('/\W+/', ' ', $query)) and a numeric-only fallback when curated text is empty.

Fixes: #8273

# Checklist:


amanpandey2587:
Hi @pabo99 , please review the changes and if there any modifications to do , I would be happy to do that.
I also want to ask you that in the issue on a single query was asked to be use FULLTEXT , what about the 3 other methods getAllWithCount,getAllProblemsAdminedByIdentity and getAllProblemsOwnedByUser. Should i work on them as well ?? 
Thanks

amanpandey2587:
Hi @pabo99 , I wanted to mention that this PR is part of my GSoC application requirement for OmegaUp. The deadline is approaching and this would be my 3rd contribution. I would really appreciate a review when you get a chance.

pabo99 (commented):
Thanks for the time you’ve spent looking into this issue.

After reviewing it more carefully, we realized that replacing `LIKE CONCAT` with `MATCH ... AGAINST` involves several aspects that require deeper analysis (sanitizing search strings, ranking results, evaluating how relevance scoring should work, etc.). Because of that, it may take us some time to fully design the final solution.

However, one possible way to move forward for now would be to introduce `MATCH ... AGAINST` with a temporary fallback. For example, if the new query returns no results, we could fall back to the previous LIKE-based search. This would help ensure that users are not affected while we continue refining the search behavior.

With this approach, the change would be safer and tests should not break, which could help us move forward more quickly. Let me know what you think.

amanpandey2587:
Hi @pabo99 , first of all thanks for the detailed feedback , I have implemented the fallback approach but only for the non numeric queries since for numeric ones the fulltext union uses the exact problem_id ,so if the problem_id exists it guarantees a search result.
I ran all the tests and all of them were passing as well.
Please review this and if any changes tell me I would be happy to do it.
Thanks

pabo99 (approved):
LGTM! The fallback approach looks solid and all tests pass.

Optional follow-ups for future issues:

- Use `$curatedQuery` in the `LIKE` fallback for consistency
- Add metrics to track fallback usage frequency
- Consider pre-determining FULLTEXT vs LIKE based on query characteristics

None of these should block merging

PR #9480, opened 2026-03-08
Add empty state when course has no enrolled students

### LINKED ISSUE #9447: [FEATURE] Improve empty state for student list in courses
## Description
As a course administrator, I want a clearer interface when a course has no students yet so that I understand how to add them.

## Context
When there are no students enrolled in a course, the interface could display an **empty state** that includes:

- An icon
- A title
- A short explanation
- A centered button to add students

This pattern would match the improved empty states proposed in other parts of the course interface similar Github does in some sections
Image

xviraj248275-beep:
Description

Fixes #9447

This PR introduces an empty state for the course students progress page when a course has no enrolled students.

Previously, the page displayed an empty table, which could be confusing for course administrators. With this change, the interface now clearly shows a message indicating that no students are enrolled yet and suggests adding students to start tracking their progress.

This improves usability and provides clearer feedback in cases where the course has not yet been populated with students.

Comments

The change only affects the UI and adds a conditional check to display an empty state message when sortedStudents.length === 0.


frontend/www/js/omegaup/components/course/ViewProgress.vue
```diff
@@ -16,7 +16,30 @@
...
+          <div v-if="sortedStudents.length === 0" class="text-center p-5">
+
+              <img
+               src="/media/empty-state.svg"
+               alt="No students"
+               width="80"
+               class="mb-3"
+            />
+
+            <h3>No students enrolled yet</h3>
```
pabo99: All text strings displayed in the UI should go through the localization files so they appear in the logged-in user’s preferred language. You can add the translations in the languages we have (en.lang, es.lang, pt.lang).

frontend/www/js/omegaup/components/course/ViewProgress.vue
```diff
@@ -16,7 +16,30 @@
...
           </div>
-          <div class="table-responsive">
+          <div v-if="sortedStudents.length === 0" class="text-center p-5">
+
+              <img
+               src="/media/empty-state.svg"
+               alt="No students"
+               width="80"
+               class="mb-3"
+            />
```
pabo99: Would it be possible to add a Font Awesome icon? There should be plenty of examples in the code

pabo99 (commented):
I left you a couple of suggestions. Other points to clarify:

- Whenever you make a change in the UI, we encourage adding a screenshot or screencast showing the changes, or a before-and-after if something is being fixed. This helps us understand what we’ll be reviewing.
- Before pushing your changes, run `./stuff/lint.sh` from outside the container, but within the `WSL` instance if you’re using Windows. In theory, this command runs in the `pre-push` hooks via Husky, but there’s usually a reason why many contributors skip it (in my case, running it from a Mac takes too long).

frontend/www/js/omegaup/components/course/ViewProgress.vue
xviraj248275-beep: Thanks for the review! I will update the UI text to use the localization files and replace the image with a Font Awesome icon as suggested.

frontend/www/js/omegaup/components/course/ViewProgress.vue
xviraj248275-beep: Thanks for the suggestion! I've updated the PR to replace the SVG with a Font Awesome icon.

frontend/templates/en.lang
```diff
@@ -2309,3 +2309,6 @@ wordsWallTimeInSeconds = "Wall-time (Sec)"
 wordsWarningProblem = "Keep with a warning"
 wordsYes = "Yes"
 zipPrepare = "Prepare zip file"
+noStudentsEnrolled=No students enrolled yet
+addStudentsToTrackProgress=Add students to this course to start tracking their progress.
+addStudents=Add students
```
pabo99: Could the new added strings have the same suffix? So that when the linter runs, these translations stay together. For example, I’m thinking of `studentProgress…`.

Also, try to use the same style as the rest for consistency and to avoid potential linter errors, e.g.:
```bash
studentProgressAddStudents = "Add students"

File: frontend/www/js/omegaup/components/course/ViewProgress.vue [hunk 4]
```diff
@@ -18,24 +18,18 @@
           </div>
           <div v-if="sortedStudents.length === 0" class="text-center p-5">
 
-              <img
-               src="/media/empty-state.svg"
-               alt="No students"
-               width="80"
-               class="mb-3"
-            />
+              <i class="fa fa-users fa-3x mb-3 text-muted"></i>
```
pabo99: Could you use the `FontAwesomeIcon` component? Here’s an example of how it’s used with the same icon:
https://github.com/omegaup/omegaup/blob/3c9b3cf6cc2a9b3cdeb35e5ccdc914fe3b0c944d/frontend/www/js/omegaup/components/arena/ContestCard.vue#L75

https://github.com/omegaup/omegaup/blob/3c9b3cf6cc2a9b3cdeb35e5ccdc914fe3b0c944d/frontend/www/js/omegaup/components/arena/ContestCard.vue#L140-L149

pabo99 (commented):
Can you add a screenshot of how this change will look?

The linter also needs to be run:
```bash
./stuff/linter.sh
``` 
If you have trouble running it, let me know so we can understand what’s going on with this and many other PRs that are skipping these checks.

frontend/www/js/omegaup/components/course/ViewProgress.vue
xviraj248275-beep: Thanks for the suggestion! I updated the translation keys to follow the `studentProgress` prefix and replaced the `<i>` icon with the `FontAwesomeIcon` component as recommended.

pabo99 (commented):
I just wanted to point out two important things:

Please make sure to run `./stuff/lint.sh`, as it validates that all language files contain the same elements. Without this, the changes won’t propagate correctly.

The location where you applied the corrections doesn’t match the issue description. That’s why the screenshots are important—to ensure the changes are applied in the right place.

If you want, I can help quickly review where the changes should go so everything aligns with the issue description

pabo99 (commented):
Ok, you still need to run `./stuff/lint.sh` so the translation strings propagate through TypeScript and become visible in the UI.

Also, you’re currently applying the change in the _**Course Student Progress**_ view. So you could update the PR description, or even better, also apply the change in the `Add students to the course` view. That’s where the button you included here would actually make sense.

Issue #9453, opened 2026-03-06
[BUG] Double back-button required to exit Coder of the Month page


Yashagarwal9798:
## Current Behavior

When navigating to the Coder of the Month page from the OmegaUp main page, the user must click the browser's Back button **twice** to return to the main page. This affects **both** pages:
- **Overall Coder of the Month** → `/coderofthemonth/`
- **Coder of the Month (Her)** → `/coderofthemonth/?category=female`

Both pages share the same entry point: `frontend/www/js/omegaup/coderofthemonth/index.ts` (lines 13-15), which automatically sets a default hash on the URL when the page loads without one:
```typescript
  const locationHash = window.location.hash.substring(1).split('/')[0];
  const selectedTab = getSelectedValidTab(locationHash);
  if (selectedTab !== locationHash) {
    window.location.hash = selectedTab; // ← Pushes a new history entry
  }
```
Setting `window.location.hash` directly forces the browser to push a new entry to the session history. This means immediately upon loading either page, the history becomes:
1. `omegaup.com`
2. `omegaup.com/coderofthemonth/`
3. `omegaup.com/coderofthemonth/#codersOfTheMonth`

Clicking Back once only resolves the hash (returning to step 2). A second click is needed to return to the main page.

## Expected Behavior

Navigating to either Coder of the Month page should only create a single history entry. Clicking the Back button once should return the user to the previous page.

## If necessary, add the following information (otherwise delete this section):

**Reproduce on Overall Coder of the Month:**
1. Go to `omegaup.com`
2. Navigate to the Overall Coder of the Month page 
3. Notice the URL instantly changes to append `#codersOfTheMonth`
4. Click the browser's Back button once — notice you are still on the Coder of the Month page
5. Click Back a second time to actually return to the previous page

**Reproduce on Coder of the Month (Her):**
1. Go to `omegaup.com`
2. Navigate to the Coder of the Month (Her) page 
3. Same behavior — the URL gets `#codersOfTheMonth` appended, and Back must be clicked twice

https://github.com/user-attachments/assets/bc7201c9-5735-44f1-a84f-abcf25f3af72

Yashagarwal9798:
If you would like to work on this issue, please state your proposed approach or solution in the comments below. To give everyone a fair chance, we will review the proposals, and the maintainers will assign the issue to the person with the best approach!

*Note: This is intended to be a "good first issue" for newcomers. Please only request to work on this if you currently have fewer than 2 active issues.*

[...]

PR #9481, opened 2026-03-08
Bug: Fix back button navigation in Coders of the Month tabs

Tharunkunamalla:
This PR fixes the browser history behavior in the Coders of the Month page.

Previously, when switching between tabs, the application used history.replaceState, which does not create a new entry in the browser history. As a result, users had to press the browser Back button twice to return to the previous page.

This change updates the tab selection logic to use window.location.hash instead. This ensures that each tab switch creates a proper history entry, allowing the browser back button to work correctly.

Changes made:

Updated tab selection logic in frontend/www/js/omegaup/components/coderofthemonth/List.vue

Replaced history.replaceState with window.location.hash

Ensured tab navigation updates browser history correctly

Fixes: #9453
This change only affects the client-side navigation logic and does not modify any backend behavior.

The fix was tested locally by:
1 .Opening /coderofthemonth
2. Switching between tabs
3. Using the browser back button
4. Confirming that navigation works with a single click
# Checklist:

https://github.com/user-attachments/assets/e1da4c82-631a-4f6f-b30f-41ded7e83382

~Thank you


Tharunkunamalla:
@pabo99 could you please check this PR, im ready to develop more if any issues

frontend/www/js/omegaup/components/coderofthemonth/List.vue
```diff
@@ -104,60 +106,83 @@ export default class CoderOfTheMonthList extends Vue {
...
+  if (hash) {
+    this.currentSelectedTab = hash;
+  }
+
+  window.addEventListener('hashchange', () => {
+    const newHash = window.location.hash.replace('#', '');
+    if (newHash) {
+      this.currentSelectedTab = newHash;
+    }
+  });
```
pabo99: Could we move this logic to the component’s `.ts` file and wrap the `hashchange` listener in a method? 

Also, it would be good to remove the event listener in `beforeDestroy` to avoid leaving dangling listeners.

You can check PR #8776 for reference.

Tharunkunamalla: thanks for the reply @pabo99,  i will make the changes

Tharunkunamalla:
@pabo99 I’ve made the requested changes. I moved the hash handling logic into a new helper file `hashHandler.ts` inside the `components/coderofthemonth` folder.

This file now contains the logic for reading the current hash and registering the `hashchange` listener. The `List.vue` component now imports these helper functions instead of directly using `window.addEventListener`.

https://github.com/user-attachments/assets/cfdd12d0-53e7-43ff-9cf6-fb259ab946e7

I also wrapped the listener logic and added cleanup in `beforeDestroy` to ensure the event listener is removed properly.
~thank you

frontend/www/js/omegaup/components/coderofthemonth/List.vue
```diff
@@ -104,37 +109,53 @@ export default class CoderOfTheMonthList extends Vue {
...
+    this.removeHashListener = addHashChangeListener((newHash: string) => {
+      this.currentSelectedTab = newHash;
+    });
+  }
+
+  beforeDestroy(): void {
+    if (this.removeHashListener) {
+      this.removeHashListener();
+    }
+  }
```
pabo99: The idea is to move this logic to the `frontend/www/js/omegaup/coderofthemonth/index.ts` file. Is it possible to make this change?

frontend/www/js/omegaup/components/coderofthemonth/List.vue
```diff
@@ -144,20 +165,28 @@ export default class CoderOfTheMonthList extends Vue {
   border: none;
   border-left: 0.0625rem solid #dee2e6;
   border-right: 0.0625rem solid #dee2e6;
-  border-top-left-radius: 0rem;
-  border-top-right-radius: 0rem;
 }
+
```
pabo99: It seems that this and the other style changes belong in a different PR. Could you revert them?

pabo99 (commented):
I left you a couple of suggestions, and with those the change looks perfect. 

One more thing: don’t forget to run `./stuff/lin.sh` before pushing your changes. If you run into any issues executing it, please let me know

frontend/www/js/omegaup/components/coderofthemonth/List.vue
Tharunkunamalla: yes, I will I'm moving that logic from new hashHandler.js to index.js

frontend/www/js/omegaup/components/coderofthemonth/List.vue
Tharunkunamalla: Thanks for the feedback! I have reverted the unrelated style changes and kept only the hash handling changes in this PR.

Tharunkunamalla:
> I left you a couple of suggestions, and with those the change looks perfect.
> 
> One more thing: don’t forget to run `./stuff/lin.sh` before pushing your changes. If you run into any issues executing it, please let me know

Hi @pabo99, I couldn't find `lin.sh`, but I did find `lint.sh` and ran it. It modified a large number of files due to Windows line-ending fixes. Since those changes are unrelated to my PR, I did not include them. Could you confirm if I should only commit the changes relevant to my modification

pabo99:
> > I left you a couple of suggestions, and with those the change looks perfect.
> > One more thing: don’t forget to run `./stuff/lin.sh` before pushing your changes. If you run into any issues executing it, please let me know
> 
> Hi @pabo99, I couldn't find `lin.sh`, but I did find `lint.sh` and ran it. It modified a large number of files due to Windows line-ending fixes. Since those changes are unrelated to my PR, I did not include them. Could you confirm if I should only commit the changes relevant to my modification

My bad, apologies. That’s exactly the command you shared, thank you.

The line-ending issue is interesting. Are you using any external code validator? I don’t recall this happening to us recently. Ideally, you should commit all the changes that appeared—perhaps you updated some libraries, and it would be important to know if any of them are changing the linter’s behavior.

Also, could you please resolve the conflicts before pushing?

Tharunkunamalla:
> > > I left you a couple of suggestions, and with those the change looks perfect.
> > > One more thing: don’t forget to run `./stuff/lin.sh` before pushing your changes. If you run into any issues executing it, please let me know
> > 
> > 
> > Hi @pabo99, I couldn't find `lin.sh`, but I did find `lint.sh` and ran it. It modified a large number of files due to Windows line-ending fixes. Since those changes are unrelated to my PR, I did not include them. Could you confirm if I should only commit the changes relevant to my modification
> 
> My bad, apologies. That’s exactly the command you shared, thank you.
> 
> The line-ending issue is interesting. Are you using any external code validator? I don’t recall this happening to us recently. Ideally, you should commit all the changes that appeared—perhaps you updated some libraries, and it would be important to know if any of them are changing the linter’s behavior.
> 
> Also, could you please resolve the conflicts before pushing?

Thanks for the clarification! I ran `lint.sh`, which modified several files due to line-ending changes. I’ve committed all the changes generated by the linter and resolved the conflicts before pushing.

frontend/www/js/omegaup/components/coderofthemonth/List.vue
```diff
@@ -1,11 +1,11 @@
 <template>
-  <div class="card w-100">
-    <ul class="nav nav-tabs justify-content-arround" role="tablist">
+  <div class="card ranking-width">
```
pabo99: There are changes that are overwriting recently approved/merged PRs.

Could you check if it’s possible to revert those changes? Mainly the changes in PR #9508

frontend/www/js/omegaup/coderofthemonth/index.ts
```diff
@@ -29,40 +33,29 @@ OmegaUp.on('ready', () => {
...
-          },
-        },
       });
     },
   });
 
+  window.addEventListener('hashchange', () => {
+    const newHash = window.location.hash.substring(1).split('/')[0];
+    const validTab = getSelectedValidTab(newHash);
+    coderOfTheMonthList.selectedTab = validTab;
```
pabo99: Could you move this logic into a function or constant? Right now the event listener is using an anonymous function, which makes it difficult to remove later.

If this component is destroyed and mounted again, it could end up registering multiple `hashchange` listeners. Using a named function would allow you to properly remove it in `beforeDestroy` with `removeEventListener`.

Tharunkunamalla:
> It seems like it’s not working correctly.
> 
>  Grabacion.de.pantalla.2026-03-16.a.la.s.12.06.21.p.m.mov 
> Have you tried checking the navigation behavior in your local environment?

Hi @pabo99, thanks for the video!

I noticed something while navigating through **Coder of the Month**. When we first go to the parent route (`coderofmonth/#codeofthemonth`), it loads correctly. However, when we move to the other sections (the child routes) and then return back, it again shows the parent (`coderofmonth/#codeofthemonth`).

After that, if we try to navigate again, it redirects to the **home page**, as if there is no parent route for `coderofmonth`.

I’ve also attached a video for reference earlier while submitting the **PR**. And also in **issue #9481**, it was mentioned that when we click once, the `#coderofthemonth` hash disappears, and when we double-click it redirects to the home page. I made a fix addressing that behavior.

Could you please confirm whether my fix is correct, or if I might be misunderstanding the expected behavior?
Thanks!

Tharunkunamalla:
@pabo99, I forgot to include the video in my latest commit, sorry about that,  I'll add it right away 


https://github.com/user-attachments/assets/ae3d635a-0108-4b5a-8d74-3d19c0d2a096

pabo99:
I’m a bit confused about this PR. What’s reported in issue #9453 is not reproducible in my local environment, and the tab navigation behavior hasn’t been fixed in this one.

@Yashagarwal9798, could you help review this PR to determine whether it should be approved or requires changes?

Yashagarwal9798:
> I’m a bit confused about this PR. What’s reported in issue #9453 is not reproducible in my local environment, and the tab navigation behavior hasn’t been fixed in this one.
> 
> @Yashagarwal9798, could you help review this PR to determine whether it should be approved or requires changes?

Ok @pabo99 sir.

frontend/www/js/omegaup/coderofthemonth/index.ts
```diff
@@ -1,18 +1,18 @@
...
 
 OmegaUp.on('ready', () => {
   const payload = types.payloadParsers.CoderOfTheMonthPayload();
+
   const locationHash = window.location.hash.substring(1).split('/')[0];
   const selectedTab = getSelectedValidTab(locationHash);
+
   if (selectedTab !== locationHash) {
-    window.location.hash = selectedTab;
+    history.replaceState(null, '', `#${selectedTab}`);
```
Yashagarwal9798, line 14: Hey, thanks for working on this! The core fix is spot on  using `history.replaceState` instead of `window.location.hash` is exactly the right approach.

However, I think this PR would be better if we keep only that one-line change and revert the other modifications. Here's why:

1. **The `select-coder` event handler was removed** this means mentors won't be able to select a Coder of the Month anymore. The button in the template still emits the event, but nothing handles it now, so clicking it would do nothing.

2. **The `hashchange` listener doesn't work as expected**  it updates the `selectedTab` prop on the root Vue instance, but the component's `currentSelectedTab` is only set once during initialization and doesn't watch for prop changes. So browser back/forward between tabs won't actually update the view.

3. **Removing imports (`api`, `ui`, `T`) and `data-toggle="tab"`** are side effects of the above changes and would also need to be reverted.

4. **The cosmetic changes** (reformatting switch cases, simplifying variables, CSS tweaks) are unrelated to the bug and add unnecessary noise to the diff.

Tharunkunamalla:
> The main fix using history.replaceState is perfect! However, it looks like a few of the other modifications accidentally broke the mentor selection feature and added some out-of-scope formatting changes.
> 
> If we can just revert those extra bits and keep your core one line fix, I think this PR will be good to go.

i have made changes only in the 2 files regarding the hash handler so could you please say me wht modifications can i revert

Yashagarwal9798:
> > The main fix using history.replaceState is perfect! However, it looks like a few of the other modifications accidentally broke the mentor selection feature and added some out-of-scope formatting changes.
> > If we can just revert those extra bits and keep your core one line fix, I think this PR will be good to go.
> 
> i have made changes only in the 2 files regarding the hash handler so could you please say me wht modifications can i revert

Hey @Tharunkunamalla, just revert both files back to how they are in the main branch, and re-apply only your single history.replaceState line. That one line is all we need to fix the bug!

Tharunkunamalla:
> > > The main fix using history.replaceState is perfect! However, it looks like a few of the other modifications accidentally broke the mentor selection feature and added some out-of-scope formatting changes.
> > > If we can just revert those extra bits and keep your core one line fix, I think this PR will be good to go.
> > 
> > 
> > i have made changes only in the 2 files regarding the hash handler so could you please say me wht modifications can i revert
> 
> Hey @Tharunkunamalla, just revert both files back to how they are in the main branch, and re-apply only your single history.replaceState line. That one line is all we need to fix the bug!

ok bro... So should i revert total things, like pabo sir also mentioned, i will revert all by tomorrow..👍

Tharunkunamalla:
@Yashagarwal9798 I have reverted all the changes back to normal and done only the necessary changes in the file, I hope this solves the issue 

https://github.com/user-attachments/assets/765ca210-bdcc-4d81-a7f3-69ada81205e5

~Thank You

Yashagarwal9798:
Hi @pabo99, the back button fix seems to be working perfectly now. Let me know if it looks good to you!


https://github.com/user-attachments/assets/691aa360-7f29-4d05-a426-3a790610b7b3

Additionally, I noticed a bug similar to this https://github.com/omegaup/omegaup/pull/9396 bug on the "Overall Coder of the Month" and "Overall Female Coder of the Month" pages. If you switch between the tabs (e.g., from Coder of the Month -> Top 100... -> List of candidates...) and then click the browser's back button, only the URL hash changes. The actual UI remains stuck on the last selected tab.

Should I go ahead and open a new issue to track this, or would you prefer it be addressed in this PR?

PR #9484, opened 2026-03-09
Replace ended contest banner with global notification

### LINKED ISSUE #9370: Global Notifications: Add optional closable property to toast component
Description
Currently, the global notification toast component always displays a close (dismiss) icon, allowing users to manually dismiss notifications.

However, in some scenarios, we need notifications to remain visible without providing a dismiss option.

Expected Behavior
The toast component should support an optional property (e.g., `closable` or `dismissible`) that determines whether the notification can be closed.

- By default, the toast should remain closable (to preserve existing behavior).
- When `closable = false`, the close icon should be hidden and the user should not be able to dismiss the notification manually.

Additional Notes
This change will allow better control over notification visibility and will be useful for system-critical messages.

Image

vidhannnn:
## Description
'Fixes #9370 
This change replaces the yellow banner shown when a contest ends with a global notification.

When the contest is finished (isContestFinished), a notification appears informing users that the contest has ended and that they can continue submitting solutions in Practice Mode.

## Changes
- Removed the banner displayed after contest ends.
- Added a global notification using ui.notifications.

## Testing
Open a finished contest and verify that the notification appears instead of the banner.

### DIFF
--- a/frontend/www/js/omegaup/components/arena/Contest.vue
+++ b/frontend/www/js/omegaup/components/arena/Contest.vue
@@ -33,9 +33,11 @@
         </a>
       </template>
       <template #clock>
+      <!--
         <div v-if="isContestFinished" class="alert alert-warning" role="alert">
           <a :href="urlPractice">{{ T.arenaContestEndedUsePractice }}</a>
         </div>
+        -->
         <omegaup-countdown
           v-show="!isContestStarted"
           :countdown-format="omegaup.CountdownFormat.ContestHasNotStarted"
@@ -309,8 +311,19 @@ export default class ArenaContest extends Vue {
   @Prop({ default: null }) blockedMessage!: string | null;
 
   T = T;
-  ui = ui;
-  omegaup = omegaup;
+ui = ui;
+omegaup = omegaup;
+
+mounted(): void {
+  if (this.isContestFinished) {
+    ui.notifications.add({
+      message: this.T.arenaContestEndedUsePractice,
+      type: 'info',
+    });
+  }
+}
+    }
+  
   AdmissionMode = AdmissionMode;
   PopupDisplayed = PopupDisplayed;
   ContestClarificationType = ContestClarificationType;


vidhannnn:
Hi! This PR implements the change described in #9370.
All checks are passing. Please review when possible. Thanks!

vidhannnn:
Hi @pabo99 

This PR implements the change described in #9370 by replacing the ended contest banner with a global notification.

All checks are passing on my side. Could you please review it when you have a moment?

Thank you!

frontend/www/js/omegaup/components/arena/Contest.vue
```diff
@@ -33,9 +33,11 @@
         </a>
       </template>
       <template #clock>
+      <!--
```
pabo99, line 36: If this code is no longer going to be used, could you remove it?

frontend/www/js/omegaup/components/arena/Contest.vue
```diff
@@ -309,8 +311,19 @@ export default class ArenaContest extends Vue {
   @Prop({ default: null }) blockedMessage!: string | null;
 
   T = T;
-  ui = ui;
-  omegaup = omegaup;
+ui = ui;
+omegaup = omegaup;
+
+mounted(): void {
```
pabo99, line 317: At omegaUp, we prefer to avoid using `mounted()` unless it’s necessary to interact directly with the DOM or initialize a third-party library. To display a notification when the contest ends, it’s better to use a watcher on the reactive property that indicates the contest state (`isContestFinished`). This ensures the notification appears both when the component loads and if the state changes dynamically, keeping the code cleaner and more reactive.

pabo99 (commented):
I left a couple of suggestions so we can approve this PR. Also, you’ll need to run `./stuff/lint.sh` outside the container but inside the WSL instance to fix the errors shown by GitHub Actions.

Additionally, if you can fix the image, because the one you uploaded to the PR description isn’t visible.

pabo99:
I just noticed you pushed your changes directly to the main branch. If possible, please create a new PR from a branch other than main. For reference, you can follow the steps on how to create a PR in the following link:

[frontend/www/docs/How-to-Make-a-Pull-Request.md](https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/How-to-Make-a-Pull-Request.md)

Issue #9374, opened 2026-03-02
[BUG] Problem Upload Returns 400 Bad Request + Editor/Statement Not Loading Correctly


pratham15541:
<html>
<body>
<h2>Description</h2><p>When creating a problem using the ZIP generated via <strong>“Create ZIP file”</strong>, the problem is successfully created and appears in the problem list (<code inline="">/browse/</code>), but uploading the ZIP redirects to <strong>400 (Bad Request)</strong>.</p><p>Additionally, the problem page behavior differs from the reference test ZIP located in <code inline="">frontend/tests/resources</code>.</p><hr><h2>Current Behavior</h2><h3>1️⃣ Problem Upload Redirects to 400</h3><ul><li><p>After creating a problem and uploading my generated ZIP (<code inline="">demo.zip</code>), the system redirects to:</p><pre><code>400 (Bad Request)
</code></pre></li><li><p>However:</p><ul><li><p>The problem <strong>is actually created</strong></p></li><li><p>It appears in <code inline="">/browse/</code></p></li><li><p>It is accessible manually</p></li></ul></li></ul><p>This suggests the upload may be partially succeeding but failing during validation or post-processing.</p><hr><h3>2️⃣ Statement &amp; Code Editor Not Loading on First Load</h3><p>On first load of the problem:</p><ul><li><p>The <strong>problem statement does not render immediately</strong></p></li><li><p>The <strong>code editor does not load automatically</strong></p></li><li><p>I must manually click on the "Code" tab to initialize the editor</p></li></ul><p>Expected behavior:</p><ul><li><p>Statement should render immediately</p></li><li><p>Code editor should initialize on load (default tab)</p></li></ul><hr><h3>3️⃣ Structural Differences Between ZIP Files</h3><p>I compared:</p><ul><li><p>✅ <code inline="">testproblem.zip</code> (from <code inline="">frontend/tests/resources</code>) → Works correctly</p></li><li><p>❌ <code inline="">demo.zip</code> (generated using “Create ZIP file”) → Causes 400 redirect</p></li></ul><p>Observed differences:</p>
Item | demo.zip | testproblem.zip
-- | -- | --
cdp.data | ✅ Present | ❌ Not present
unknownfile | ❌ Not present | ✅ Present
Multiple test cases | ❌ Single case | ✅ Multiple
examples/ folder | ❌ No | ✅ Yes
tests/ config | ❌ No | ✅ Yes

[...]

PR #9333, opened 2026-02-26
fix: local gitserver issue where directory not created on first run

pratham15541:
# Description

Fix local development failure when creating a new problem from `/problem/new/`.

On a fresh Docker setup following the official Development Environment Setup guide, submitting the create-problem form fails with:

```
Failed to create temporary directory
stat /var/lib/omegaup/problems.git: no such file or directory
```

## Root Cause

The shared Docker volume `omegaupdata` does not initialize the required base directory `/var/lib/omegaup/problems.git`.
As a result, `git-upload-zip` fails during problem deployment and the frontend shows a generic internal deploy error.

## Changes Made

* Added a dedicated `init-omegaupdata` service in `docker-compose.yml`.
* The init service:

  * Creates `/var/lib/omegaup/problems.git` if missing.
  * Sets correct ownership for `/var/lib/omegaup`.
* Configured `gitserver` to depend on the successful completion of the init service.
* Preserved the non-root execution model for `gitserver` and other services.

This ensures correct volume initialization while maintaining stricter container security.

Fixes: #9330

---

# Comments

This issue occurs on a fresh clone using Docker with an empty `omegaupdata` volume.

This PR only affects local development setup.
No production behavior is modified.

Please let me know if the initialization should instead be handled inside the gitserver image itself.

---

# Checklist:

* [x] The code follows the coding guidelines of omegaUp.
* [x] The tests were executed and all of them passed.
* [ ] If you are creating a feature, the new tests were added. (Not applicable – development environment fix)
* [x] If the change is large (> 200 lines), this PR was split appropriately. (Change is localized to docker-compose)

PR #9487, opened 2026-03-10
fix: 400 create problem issue

pratham15541:
# Description

Fix the issue where uploading a problem ZIP generated using **"Create ZIP file"** resulted in a **400 (Bad Request) redirect**, even though the problem was successfully created in the system.

Previously, when users uploaded the generated ZIP file:

* The backend processed the upload and created the problem successfully.
* However, the request incorrectly redirected to a **400 error page**, creating the impression that the upload failed.

This PR fixes the upload flow so that:

* The ZIP upload completes successfully without triggering a 400 redirect.
* The user is redirected correctly after the problem is created.

This ensures that ZIP files generated through the **Create ZIP file** feature work correctly during problem upload.

Fixes: #9374

And On first run this is also issue which is cover in pr: https://github.com/omegaup/omegaup/pull/9333
# Comments

This PR focuses specifically on resolving the **incorrect 400 redirect after a successful ZIP upload**.

The issue occurred even though the problem was correctly created and visible in `/problem/new`.
The fix ensures that the upload flow returns the correct response and prevents the erroneous 400 redirect.

Other observations mentioned in the issue (such as differences in ZIP structure, editor initialization behavior, and collection page refresh behavior) are outside the scope of this PR and can be addressed separately if needed.

# Checklist:

* [x] The code follows the [[coding guidelines](https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/Coding-guidelines.md)](https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/Coding-guidelines.md) of omegaUp.
* [x] If the change is large (> 200 lines), this PR was split into various Pull Requests. It's preferred to create one PR for changes in controllers + unit tests in PHPUnit, and then another Pull Request for UI + tests in Jest, Cypress or both.


pratham15541:
> Could you revert all the changes that are not directly related to the problem described in the PR?

@pabo99  Done.

pabo99 (approved):
Thanks. We still need to address the issue of making the errors in the .zip file more discoverable. That way we’ll know which data is actually required and which is not.

But for now, this is excellent.

PR #9488, opened 2026-03-10
docs: add requirements.txt for cron tools

Jami436:
Description
docs: add requirements.txt for cron tools

While setting up a development environment on Windows, I encountered a ModuleNotFoundError for python-json-logger when attempting to run the backend scripts in stuff/cron/. Currently, these dependencies are not documented within the stuff/cron directory.

This PR adds a requirements.txt file containing python-json-logger and mysql-connector-python. This ensures that new contributors can easily set up their environment and that the cron tools are ready for cross-platform execution.

Fixes: #9488

Comments
I discovered this while verifying the update_ranks.py script. By adding this file, we improve the onboarding experience for developers not using the full Docker environment or those working on specific Python backend tasks.

Checklist:
[x] The code follows the [coding guidelines](https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/Coding-guidelines.md) of omegaUp.

[x] The tests were executed and all of them passed. (Since you ran --help and verified the module loads, this is true for a documentation/requirements PR).

[ ] If you are creating a feature, the new tests were added.

[ ] If the change is large (> 200 lines), this PR was split into various Pull Requests.


pabo99 (commented):
Thanks for the detailed explanation! I understand the use case for developers not using the full Docker environment. However, I have some concerns about this approach:

**Our recommended development workflow**
Even for Python-specific tasks, we strongly recommend using the Docker environment because it's the only way we guarantee all dependencies are properly satisfied. This includes:

- Correct Python version
- System-level dependencies (MySQL client libraries, etc.)
- Environment variables and configuration
- Database schema and test data
- Proper isolation from host system quirks

The Docker setup is designed to handle these complexities automatically, which is especially important for cross-platform development (Windows/Mac/Linux).

**The specific issue with this PR**
While I appreciate the attempt to make setup easier, this creates several problems:

Incomplete dependencies: The proposed file only lists 2 packages, but the cron scripts actually depend on the entire [requirements.txt](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) (20+ packages) through imports from [lib](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html).

Maintenance burden: We'd need to keep two files in sync. When security updates or version bumps happen, both files must be updated.

False sense of standalone-ness: This suggests [cron](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) can work independently, but it actually requires the entire [stuff](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) module structure.

**Alternative solutions**
If Docker isn't an option for your specific workflow, I'd suggest:

Option 1 - Use the parent requirements:
```bash
cd stuff
pip install -r requirements.txt
python cron/update_ranks.py --help
```
Option 2 - If we want to improve discoverability, let's add a stuff/cron/README.md:
```md
## Setup

These scripts require dependencies from the parent directory:


pip install -r ../requirements.txt


**Note**: For full development environment setup, we recommend using Docker 
as documented in the main CONTRIBUTING.md. This ensures all system dependencies
and configurations are properly set up.
```

This approach:

- Documents the setup clearly
- Points developers to the correct requirements file
- Recommends Docker for full setup
- Avoids duplication and maintenance burden

Would you be interested in submitting a documentation PR instead? That would solve your onboarding issue without creating technical debt.

Jami436:
> Thanks for the detailed explanation! I understand the use case for developers not using the full Docker environment. However, I have some concerns about this approach:
> 
> **Our recommended development workflow** Even for Python-specific tasks, we strongly recommend using the Docker environment because it's the only way we guarantee all dependencies are properly satisfied. This includes:
> 
> * Correct Python version
> * System-level dependencies (MySQL client libraries, etc.)
> * Environment variables and configuration
> * Database schema and test data
> * Proper isolation from host system quirks
> 
> The Docker setup is designed to handle these complexities automatically, which is especially important for cross-platform development (Windows/Mac/Linux).
> 
> **The specific issue with this PR** While I appreciate the attempt to make setup easier, this creates several problems:
> 
> Incomplete dependencies: The proposed file only lists 2 packages, but the cron scripts actually depend on the entire requirements.txt (20+ packages) through imports from lib.
> 
> Maintenance burden: We'd need to keep two files in sync. When security updates or version bumps happen, both files must be updated.
> 
> False sense of standalone-ness: This suggests cron can work independently, but it actually requires the entire stuff module structure.
> 
> **Alternative solutions** If Docker isn't an option for your specific workflow, I'd suggest:
> 
> Option 1 - Use the parent requirements:
> 
> ```shell
> cd stuff
> pip install -r requirements.txt
> python cron/update_ranks.py --help
> ```
> 
> Option 2 - If we want to improve discoverability, let's add a stuff/cron/README.md:
> 
> ```
> ## Setup
> 
> These scripts require dependencies from the parent directory:
> 
> 
> pip install -r ../requirements.txt
> 
> 
> **Note**: For full development environment setup, we recommend using Docker 
> as documented in the main CONTRIBUTING.md. This ensures all system dependencies
> and configurations are properly set up.
> ```
> 
> This approach:
> 
> * Documents the setup clearly
> * Points developers to the correct requirements file
> * Recommends Docker for full setup
> * Avoids duplication and maintenance burden
> 
> Would you be interested in submitting a documentation PR instead? That would solve your onboarding issue without creating technical debt.

Hi Juanpa! Thank you for the detailed feedback. I completely see your point about the maintenance burden of keeping two dependency files in sync and the potential 'false sense of standalone-ness' it creates.

I'd be very happy to pivot this to a documentation-focused PR. I will remove the requirements.txt and instead add a README.md in stuff/cron/ following your suggestion. This will improve discoverability for new contributors while maintaining the single source of truth for dependencies. I'll push those changes shortly!

Jami436:
@omegaup-bot assign me

Issue #9451, opened 2026-03-06
[BUG] `preview` prop in ProblemMarkdown.vue is a no-op (hardcoded to `false`)


Yashagarwal9798:
## Current Behavior

The `ProblemMarkdown.vue` component accepts a `preview` prop on **line 44**, but the markdown converter on **line 49** ignores it and hardcodes `preview: false`:

- Line 44 declares the prop:
  ```typescript
  @Prop({ default: false }) preview!: boolean;
  ```
- Line 49 hardcodes it:
  ```typescript
  markdownConverter = new markdown.Converter({ preview: false });
  ```

When a parent passes `preview=true`, the converter still renders full download panels (`panel panel-default` divs) instead of the simpler icon-based preview format defined in [markdown.ts](https://github.com/omegaup/omegaup/blob/main/frontend/www/js/omegaup/markdown.ts) (lines 66–72). The prop is silently ignored.

## Expected Behavior

When `preview=true` is passed, the converter should use the simplified icon-based templates (e.g., a `<code class="libinteractive-download">` element with a download glyph icon) instead of the full panel HTML.

*Relevant details about the environment in which you encountered this bug:*

* Not environment-specific — this is a code-level bug
* File: `frontend/www/js/omegaup/components/problem/ProblemMarkdown.vue` (lines 44, 49)

Yashagarwal9798:
If you would like to work on this issue, please state your proposed approach or solution in the comments below. To give everyone a fair chance, we will review the proposals, and the maintainers will assign the issue to the person with the best approach!

*Note: This is intended to be a "good first issue" for newcomers. Please only request to work on this if you currently have fewer than 2 active issues.*

UraModi:
@Yashagarwal9798 Hi! I investigated this issue and the problem occurs because the `preview` prop declared in `ProblemMarkdown.vue` is not passed to the markdown converter.

Currently the converter is initialized with:

markdownConverter = new markdown.Converter({ preview: false });
This hardcodes the preview mode to `false`, so even when a parent component passes `preview=true`, the converter still renders the full download panel layout.

**Proposed fix:**

Pass the prop value to the converter instead:

markdownConverter = new markdown.Converter({
  preview: this.preview,
});

This allows the component to correctly render the simplified preview templates defined in `markdown.ts` when `preview=true`.

@pabo99 If this approach looks good, I'd be happy to submit a PR implementing the fix.

Rudrika-Singhal:
[...]

PR #9498, opened 2026-03-11
Pass preview prop to markdown converter in ProblemMarkdown

UraModi:
Fixes an issue where the preview prop in ProblemMarkdown.vue was ignored.
The markdown converter was initialized with preview: false, so even when a parent component passed preview=true, the simplified preview templates were never used.
This change passes the preview prop to the converter.

Changes
1 file changed
1 line modified

Fixes: #9451 
@pabo99 Please check when convenient.


UraModi:
@pabo99 Noted, will make the above change , crosscheck and update soon.

pabo99 (approved):
Thanks for fixing this:

Image

PR #9499, opened 2026-03-11
Fix responsive table layout on school profile page

### LINKED ISSUE #9466: [FEATURE] Table layout breaks on mobile in school profile page
## Description
On the school profile page (`/schools/profile/{id}`), the users table does not render properly on small screens. When viewed on mobile devices, the table layout collapses and the **"Place" column header appears vertically stacked**, making the UI difficult to read.

## Steps to Reproduce
1. Open https://omegaup.com/schools/profile/{id}/
2. Enable mobile view in the browser (e.g., iPhone SE – 375px width using Chrome DevTools).
3. Scroll to the **"Users who belong to this school"** section.

## Current Behavior
- The table becomes poorly formatted on small screens.
- The **"Place" column header breaks into multiple lines**, showing each letter vertically.
- Overall table readability is reduced and the layout appears broken.

## Expected Behavior
The table should remain readable on small screens. Column headers should not collapse vertically and the layout should adjust responsively to fit mobile devices.

## Possible Cause
This may be due to insufficient responsive styles for the table layout or column width constraints on small screen sizes.

## Screenshot

Image

AMRITESH-BAKSHI:
# Description

This PR fixes the responsive table layout issue on the school profile page where the "Place" column header was stacking vertically on mobile devices.

## Changes made:
- Added `table-responsive` wrapper around the table to enable horizontal scrolling on mobile devices
- Removed `d-table col-12` classes from `thead` and `tbody` that were breaking the table layout
- Removed problematic `@media (max-width: 550px)` query that was forcing `display: block` on table headers, causing vertical stacking
- Added `:key` attributes to `th` elements for better Vue rendering optimization
- Desktop layout remains completely unchanged

## Before:

before

## After:
final

Fixes: #9466 

# Comments

The issue was caused by two main problems:
1. A media query that set `display: block` on `th` and `td` elements on screens below 550px
2. Incorrect Bootstrap classes (`d-table col-12`) that interfered with responsive behavior

The solution uses Bootstrap's built-in `table-responsive` class which is the standard approach for making tables scrollable on mobile devices while preserving the desktop layout. The changes are minimal and have been visually tested on iPhone SE viewport (375×667).

# Checklist:


AMRITESH-BAKSHI:
Hi @pabo99 i hope you are doing good can you review the pr for this issue and also suggest me if some changes are required . Thank you

frontend/www/js/omegaup/components/common/TablePaginator.vue
```diff
@@ -157,15 +162,6 @@ export default class TablePaginator extends Vue {
 </script>
 
 <style>
-@media (max-width: 550px) {
```
pabo99, line 160: Ok, I’m trying to understand why we had this here, and I realized that this component was originally created from the **_GridPaginator_** component to split them, since in some places it made sense to use tables and in others to use grids.

So the styles you are removing remained without being properly validated, and they were causing the issue that you fixed.

By removing them in this PR, it seems like the problem was fixed, but not completely, because it turns out that it is still inheriting those styles from the _**GridPaginator**_ component. There are a couple of pending things to finally fix it:

- Add `scoped` to the `style` tag so the styles are not shared between components.
- Remove the `table-typo` class from this component, since it is no longer needed after removing those styles.

After these changes, the table headers should now display correctly.

pabo99 (commented):
If you can add `scoped` to the `style` tag in the **_GridPaginator_** component, that would be great

AMRITESH-BAKSHI:
@pabo99  Thank you so much for the detailed feedback and explanation! 

I've made the changes you suggested:

- Added scoped to the <style> tag to prevent styles from leaking into other components (like GridPaginator)
- Removed the table-typo class from the table since it's no longer needed after removing the media query

I understand now why those styles were causing issues - the component inheritance from GridPaginator explains the problem perfectly. The changes have been tested locally and the table headers are displaying correctly on mobile view.

Please let me know if there's anything else that needs adjustment!

AMRITESH-BAKSHI:
hi @pabo99 please can you review this pr. Please let me know if some changes are required. thank you

AMRITESH-BAKSHI:
hi @pabo99 please can you review this pr. Please let me know if some changes are required. thank you

pabo99 (approved):
Excellent! Thank you, @AMRITESH-BAKSHI!

Issue #9470, opened 2026-03-07
[FEATURE] Improve performance of update_user_rank in update_ranks.py


Prince-Prajapati07:
# Improve performance of `update_user_rank` in `update_ranks.py`

While going through `stuff/cron/update_ranks.py`, I noticed something in `update_user_rank()` (and similarly `update_author_rank()`) that might be worth revisiting from a performance perspective.

Right now, the function:

- Computes ranking information in Python.
- Iterates over all ranked users.
- Executes one `INSERT ... ON DUPLICATE KEY UPDATE` per user.
- Commits at the end.

So effectively, if there are **N ranked users**, we end up doing **N individual database writes from Python**.

This works correctly, but as the number of users grows, this could become expensive. For large installations with tens of thousands of users, this results in a large number of round-trips between Python and MySQL during a single cron execution.

## Possible Improvement

Since the ranking logic is already computed in Python, the database writes could potentially be **batched** instead of executing one query per user.

For example, instead of executing one query per row from Python, the writes could potentially be batched using `executemany()` while keeping the SQL and ranking logic unchanged.

This could reduce the number of database round-trips and improve cron execution time for installations with many users.

Prince-Prajapati07:
Hi @pabo99 sir! I wanted to get your thoughts on this before starting any implementation.
Would batching the writes (for example using `executemany()` or chunked batches) be an acceptable approach here while keeping the ranking logic unchanged?
If this sounds reasonable, I'd be happy to work on a PR for it.

Aditya060806:
Hi @Prince-Prajapati07 pabo99 sir!

While reading through stuff/cron/update_ranks.py, I noticed something in update_user_rank() (and also in update_author_rank()) that made me curious from a performance point of view.

Right now the ranking is calculated in Python and then an INSERT ... ON DUPLICATE KEY UPDATE query is executed for each user individually. It works perfectly, but I was wondering if this might become a bit heavy for larger installations where there could be thousands of ranked users, since that would mean many database round-trips during a single cron run.

I was thinking that since the ranking values are already computed in Python, maybe the database writes could be batched (for example using executemany() or chunked batches) while keeping the existing SQL logic the same. That might help reduce the number of DB calls and slightly improve the cron execution time.

Before trying anything, I just wanted to ask for your thoughts on it. Do you think batching the writes would be an acceptable approach here?

If it sounds reasonable, I’d be really happy to try working on a small PR for it. I’m eager to contribute and learn more about the codebase.

[...]

PR #9502, opened 2026-03-11
Batch inserts in update_user_rank and update_author_rank to reduce DB round-trips

Prince-Prajapati07:
# Description

This PR improves the performance of the ranking cron job by batching database writes in `update_user_rank` and `update_author_rank` in `stuff/cron/update_ranks.py`.

Previously, both functions computed the rankings in Python and executed one `INSERT` query per user:

```python
for row in rows:
    cur.execute(...)
```

This resulted in **one database round-trip per ranked user**. For installations with many users, this could generate tens of thousands of queries during a single cron execution.

This change keeps the existing ranking logic exactly the same but batches the database writes using `executemany()`.

Rows are first collected in memory and then inserted in chunks:

```python
batch = []
...
cur.executemany(sql, batch_chunk)
```

To avoid very large statements, inserts are executed in **chunks of 1000 rows**.

As a result, the number of database round-trips is reduced from:

```
N queries → ceil(N / 1000) queries
```

while preserving:

- ranking semantics
- SQL statements
- transaction behavior

### Performance validation

A small local experiment was performed to measure the effect of batching.  
Using a temporary table with the same schema as `User_Rank`, 10,000 rows were inserted using both strategies while MySQL `general_log` was enabled to count executed queries.

| Method | Rows Inserted | INSERT Queries |
|------|------|------|
| Per-row inserts (previous implementation) | 10,000 | 10,000 |
| Batched inserts (`executemany`, chunk size 1000) | 10,000 | 10 |

**Result:** ~1000× fewer database round-trips for the insert workload.

This demonstrates the expected improvement for large installations where tens of thousands of users may be ranked during cron execution.

Fixes: #9470 

---

# Comments

The ranking computation and SQL statements were intentionally left unchanged to keep the behavior identical.  
The change only modifies how the inserts are executed (batched instead of per-row).

---

# Checklist:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -36,13 +36,33 @@ services:
       REDIS_HOST: "redis"
       REDIS_PORT: "6379"
 
+  init-omegaupdata:
+    image: alpine:3.20
+    user: "0:0"
+    restart: "no"
+    command:
+      [
+        "/bin/sh",
+        "-lc",
+        "mkdir -p /var/lib/omegaup/problems.git && chown -R ${UID_GID:-1000:1000} /var/lib/omegaup",
+      ]
+    volumes:
+      - type: volume
+        source: omegaupdata
+        target: /var/lib/omegaup
+        volume:
+          nocopy: true
+
   gitserver:
     image: omegaup/gitserver:v1.9.13
     entrypoint: ["wait-for-it", "mysql:13306", "--", "/usr/bin/omegaup-gitserver"]
     user: "${UID_GID}"
     restart: always
     depends_on:
-      - mysql
+      mysql:
+        condition: service_started
+      init-omegaupdata:
+        condition: service_completed_successfully
     volumes:
       - type: bind
         source: ./stuff/docker/etc/omegaup/gitserver
--- a/frontend/server/src/Controllers/Problem.php
+++ b/frontend/server/src/Controllers/Problem.php
@@ -5444,10 +5444,19 @@ class Problem extends \OmegaUp\Controllers\Controller {
             $problem->commit,
             $lang
         );
-        $cdp = \OmegaUp\Controllers\Problem::getProblemCDP(
-            $problem,
-            strval($problem->commit)
-        );
+        $cdp = null;
+        try {
+            $cdp = \OmegaUp\Controllers\Problem::getProblemCDP(
+                $problem,
+                strval($problem->commit)
+            );
+        } catch (\Exception $e) {
+            self::$log->error(
+                "Failed to load CDP data for {$problem->alias}",
+                ['exception' => $e],
+            );
+            // CDP data is optional for the edit page to function
+        }
         $publishedRevision = null;
         foreach ($versions['log'] as $revision) {
             if ($versions['published'] === $revision['commit']) {
--- a/frontend/www/js/omegaup/components/coderofthemonth/List.vue
+++ b/frontend/www/js/omegaup/components/coderofthemonth/List.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="card ranking-width">
+  <div class="card w-100">
     <ul class="nav nav-tabs justify-content-arround" role="tablist">
       <li v-for="tab in availableTabs" :key="tab.id" class="nav-item">
         <a
@@ -147,17 +147,18 @@ export default class CoderOfTheMonthList extends Vue {
   border-top-left-radius: 0rem;
   border-top-right-radius: 0rem;
 }
-.nav .nav-tabs {
+
+.nav-tabs {
   border-bottom: 0rem;
+  flex-wrap: nowrap !important;
+  overflow-x: auto;
+  overflow-y: hidden;
 }
 
 .nav-link {
   font-weight: medium;
   letter-spacing: 0.022rem;
   padding: 0.65rem 1rem;
-}
-.ranking-width {
-  max-width: 55rem;
-  margin: 0 auto;
+  white-space: nowrap;
 }
 </style>
--- a/frontend/www/js/omegaup/components/course/Tabs.vue
+++ b/frontend/www/js/omegaup/components/course/Tabs.vue
@@ -24,7 +24,7 @@
       </li>
     </ul>
     <div class="tab-content">
-      <div class="row m-0 mt-4">
+      <div class="row m-0 mt-4 course-filters">
         <div class="col-md-4 col-lg-3 p-0 ml-4">
           <input
             v-model="searchText"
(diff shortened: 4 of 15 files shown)


stuff/cron/update_ranks.py
```diff
@@ -1,13 +1,37 @@
...
+AuthorRankRow = Tuple[
+    int,
+    str,
+    float,
+    int,
+    str,
+    Optional[str],
+    Optional[str],
+    Optional[int],
+]
```
pabo99: Oh, it’s hard to identify which data is received in each position. How about using `NamedTuple`s instead? This way it’s easier to identify each of the fields.

```suggestion

class UserRankRow(NamedTuple):
    '''Row data for User_Rank table insertion.'''
    user_id: int
    ranking: int
    problems_solved_count: int
    score: float
    username: str
    name: str
    country_id: Optional[str]
    state_id: Optional[str]
    school_id: Optional[int]


class AuthorRankRow(NamedTuple):
    '''Row data for User_Rank author fields insertion.'''
    user_id: int
    username: str
    author_score: float
    author_ranking: int
    name: str
    country_id: Optional[str]
    state_id: Optional[str]
    school_id: Optional[int]

```

stuff/cron/update_ranks.py
```diff
@@ -205,29 +229,44 @@ def update_user_rank(
...
                         `User_Rank` (`user_id`, `ranking`,
                                      `problems_solved_count`, `score`,
                                      `username`, `name`, `country_id`,
                                      `state_id`, `school_id`)
-                    VALUES(%s, %s, %s, %s, %s, %s, %s, %s, %s);''',
-            (row['user_id'], rank, row['problems_solved_count'], score,
-             row['username'], row['name'], row['country_id'], row['state_id'],
-             row['school_id']))
+                    VALUES(%s, %s, %s, %s, %s, %s, %s, %s, %s);'''
+    batch: List[UserRankRow] = []
```
pabo99: Could you use a more descriptive name here? something like:
```suggestion
    user_rank_rows: List[UserRankRow] = []
```

stuff/cron/update_ranks.py
```diff
@@ -205,29 +229,44 @@ def update_user_rank(
...
+            row['user_id'],
+            rank,
+            row['problems_solved_count'],
+            score,
+            row['username'],
+            row['name'],
+            row['country_id'],
+            row['state_id'],
+            row['school_id'],
+        ))
```
pabo99: Now it will be easier to identify each piece of data that will be stored here:
```suggestion
        user_rank_rows.append(UserRankRow(
            user_id=row['user_id'],
            ranking=rank,
            problems_solved_count=row['problems_solved_count'],
            score=score,
            username=row['username'],
            name=row['name'],
            country_id=row['country_id'],
            state_id=row['state_id'],
            school_id=row['school_id'],
        ))
```

stuff/cron/update_ranks.py
```diff
@@ -205,29 +229,44 @@ def update_user_rank(
...
+    chunk_size = 1000
```
pabo99: If this indicates the batch size, icould it be named:

```suggestion
    batch_size = 1000
```
?

stuff/cron/update_ranks.py
```diff
@@ -205,29 +229,44 @@ def update_user_rank(
...
+    for index in range(0, len(batch), chunk_size):
+        cur.executemany(
+            insert_user_rank_sql,
+            batch[index:index + chunk_size],
+        )
```
pabo99: As copilot explained, this implementation collects all rows into `batch` before inserting. Since `fetchall()` already loads everything into memory, you're temporarily duplicating the data (O(2N) memory usage).

So the suggestion is flush batches as you iterate to keep memory bounded at O(batch_size):
```suggestion
    batch = []
    for row in rows:
        batch.append(row)
        # Insert when batch reaches batch_size to keep memory bounded
        if len(batch) >= batch_size:
            cur.executemany(insert_user_rank_sql, batch)
            # Using the method `clear()` to keep the reference
            batch.clear()

    # Insert remaining rows if any
    if batch:
        cur.executemany(insert_sql, batch)
```

stuff/cron/update_ranks.py
```diff
@@ -278,26 +317,40 @@ def update_author_rank(
             `author_score` DESC
     ''')
 
+    rows = cur_readonly.fetchall()
```
pabo99: The same needs to be done for the authors’ rank.

stuff/cron/update_ranks.py
```diff
@@ -1,13 +1,37 @@
 #!/usr/bin/env python3
 '''Updates the user ranking.'''
 
+# pylint: disable=too-many-lines
```
pabo99, line 4: Ok, this file is growing too much. But we can fix it.

A bit of context: in this cronjob, we’ve been trying to separate the business logic from the data access layer. As was done previously, all the queries related to `coder_of_the_month` and `school_of_the_month` were moved to a separate file inside the `database` directory.

So, for this change we can try something similar. Would it be possible for you to move all the queries in the modified functions to a new file called `user_author_rank.py`, for example?

This way you would only need to import all the functions you move, just like it’s done with `school_of_the_month`, and you could also remove this line that disables a pylint rule.
```suggestion
```

pabo99 (commented):
Great optimization! 

The batching approach significantly reduces database round-trips from N queries to ceil(N/1000). I've tested this locally and have some suggestions to make it even better:

stuff/cron/update_ranks.py
Prince-Prajapati07: Thanks for the suggestion @pabo99 sir!

I tried removing the pylint pragma, but pylint fails because the module currently exceeds the 1000-line limit. Since this PR focuses only on the batching improvements and readability changes, I kept the pragma to avoid expanding the scope of this PR.

If you'd prefer, I can split the module (moving the queries into a separate file like `user_author_rank.py`) in a follow-up PR.

stuff/cron/update_ranks.py
pabo99: This is the key update in the review. Can you implement it, or do you have any questions? We can clarify them

stuff/cron/update_ranks.py
pabo99: Thanks for the clarification. Keeping the pragma for now is fine to keep this PR focused. We can split the module in a follow-up PR.

stuff/cron/update_ranks.py
Prince-Prajapati07: Thanks for the suggestion! I’ve updated both `update_user_rank` and `update_author_rank` to stream rows directly from the cursor and flush inserts in batches while iterating, removing the `fetchall()` usage.

This should keep memory bounded to `O(batch_size)`. CI is running now.

stuff/cron/update_ranks.py
Prince-Prajapati07: Implemented the change sir!

pabo99 (approved):
Thank you!

Issue #9503, opened 2026-03-11
[BUG] Avoid N+1 queries in get_users_ids() in contest.py


Prince-Prajapati07:
## Checklist
- [x] I have searched existing issues and confirmed this is not a duplicate

## Current Behavior

The function `get_users_ids()` in `contest.py` executes one database query per username:

```python
for username in usernames:
    cur.execute(
        '''
        SELECT user_id
        FROM Identities
        WHERE username = %s
        ''',
        (username,),
    )
```
This results in N database queries for N usernames, creating an *N+1* query pattern.

## Expected Behavior

All user IDs should be fetched using a single query, for example:

```
SELECT username, user_id
FROM Identities
WHERE username IN (...)
```
This would reduce the number of queries from N → 1, improving performance when processing contests with many participants.

chiraglaheru:
Hi, I'd like to work on this issue. 
My proposed fix is to replace the loop with a single batch query using 
WHERE username IN (...) to fetch all user IDs in one database call, 
reducing N queries to 1.

Could a maintainer please assign this to me?

Prince-Prajapati07:
Thanks for showing your interest in this @chiraglaheru ! I’m planning to work on this optimization and submit a PR shortly.

chiraglaheru:
Hi @Prince-Prajapati07, I have already submitted PR #9506 
with the fix. Feel free to review it!

Prince-Prajapati07:
Thanks for the PR @chiraglaheru !

One thing to note is that mysql doesn't expand `IN %s` automatically.
Usually the placeholders need to be constructed like:

WHERE username IN (%s, %s, %s)

Otherwise the query may fail depending on the driver.

Also, the current implementation silently skips usernames that are not
found in the result set. The previous implementation would raise an
error in that case, so it might be worth preserving the original behavior.

Let me know what you think.

chiraglaheru:
Hi @Prince-Prajapati07, thanks for the feedback!

I've updated the PR to:
1. Use proper placeholders `(%s, %s, %s)` for the MySQL IN clause
2. Removed the silent skip so it raises KeyError for missing usernames, 
   preserving the original behavior

chiraglaheru:
Hi, could a maintainer please assign this issue to me? 
I have submitted PR #9506 which has been reviewed 
and approved by @Prince-Prajapati07. Thank you!

pabo99:
Hi @chiraglaheru! The issue has been assigned to you. 

[...]

PR #9506, opened 2026-03-11
Fix: replace N+1 queries with single batch query in get_users_ids()

chiraglaheru:
# Description

Replaced the N+1 query loop in `get_users_ids()` with a single batch 
query using `WHERE username IN %s`, fetching all user IDs in one 
database call. Also added an empty list guard for edge cases.

Fixes: #9503

# Comments

Reduces database queries from N to 1 when processing contests 
with many participants, improving performance significantly.

# Checklist:


Prince-Prajapati07:
Thanks for the update @chiraglaheru! The change to use proper placeholders for the IN clause looks correct, and the behavior now matches the original function while avoiding the **N+1** queries.

This resolves the issue from my side. Thanks for working on it!

pabo99 (approved):
Excellent! Thank you, @chiraglaheru!

Issue #9505, opened 2026-03-11
[BUG]:events_socket: onclose() fails to clear clarificationInterval and rankingInterval, causing duplicate polls


AAdIprog:
## Checklist

- [x] I have searched existing issues and confirmed this is not a duplicate (Required)

## Current Behavior

In `frontend/www/js/omegaup/arena/events_socket.ts`, when a WebSocket connection closes and reconnects, the `onclose()` method only clears the `socketKeepalive` interval but does **not** clear `clarificationInterval` or `rankingInterval`. When the socket reconnects successfully, `setupPolls()` is called again, creating **new** intervals without clearing the old ones. This causes duplicate intervals to accumulate, firing redundant API requests (`Problemset.scoreboard`, `Problemset.scoreboardEvents`, `refreshContestClarifications`) indefinitely — one extra set per reconnection cycle.

## Expected Behavior

When the WebSocket connection closes, **all** polling intervals (`socketKeepalive`, `clarificationInterval`, and `rankingInterval`) should be cleared in `onclose()`. On reconnect, `setupPolls()` should start fresh without duplicate intervals running in the background. Only one set of polling intervals should ever be active at a time.

**Steps to reproduce:**

1. Open a contest or problemset page that uses the events socket (e.g., any active contest scoreboard).
2. Trigger a WebSocket disconnection (e.g., network interruption, or wait for the server to close the connection).
3. Allow the socket to automatically reconnect (it retries with exponential backoff).
4. Observe in the browser Network tab that API calls to `Problemset.scoreboard`, `Problemset.scoreboardEvents`, and clarification refresh endpoints now fire at **double** the expected rate. Each subsequent reconnect adds another set.

**Relevant code:**

- `onclose()` at line 243 only clears `socketKeepalive`: https://github.com/omegaup/omegaup/blob/main/frontend/www/js/omegaup/arena/events_socket.ts#L243
- `clarificationInterval` set at line 354 — never cleared on close
- `rankingInterval` set at line 366 — never cleared on close

**Environment:**

- Browser name: All browsers (logic bug, not browser-specific)
- Operating System and version: All platforms
- Link to the site where this issue occurs: Any active contest/problemset page on omegaUp that uses WebSocket events

ratatall:
Hi! I'd like to work on this issue. @pabo99 

[...]

PR #9507, opened 2026-03-11
Fix duplicate polling intervals in Arena

AAdIprog:
# Description

Fixed a bug where duplicate polling intervals for rankings and clarifications accumulated every time the WebSocket reconnected. The fix explicitly clears `clarificationInterval` and `rankingInterval` inside the `EventsSocket.onclose()` method to prevent leaks. 

Additionally, a comprehensive unit test (`should not leak polling intervals during reconnect cycles with setupPolls`) was added to simulate the connection fallback/reconnect loop and verify that all intervals are correctly disposed of across multiple cycles.

Fixes: #9505 

# Checklist:

### DIFF
--- a/.github/workflows/discord-issue-notification.yml
+++ b/.github/workflows/discord-issue-notification.yml
@@ -1,0 +1,277 @@
+name: Discord Issue Thread Notification
+
+on:
+  issues:
+    types: [labeled, opened]
+
+jobs:
+  notify-discord:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Check Label Conditions
+        id: check-label
+        run: |
+          ALLOWED_LABELS=("GSoC" "good first issue")
+          CURRENT_LABEL="${{ github.event.label.name }}"
+
+          # For 'opened' events, check all labels
+          if [ "${{ github.event.action }}" == "opened" ]; then
+            ISSUE_LABELS='${{ toJSON(github.event.issue.labels.*.name) }}'
+            for label in "${ALLOWED_LABELS[@]}"; do
+              if echo "$ISSUE_LABELS" | grep -q "\"$label\""; then
+                echo "should_notify=true" >> $GITHUB_OUTPUT
+                echo "matched_label=$label" >> $GITHUB_OUTPUT
+                exit 0
+              fi
+            done
+          fi
+
+          # For 'labeled' events, check the specific label
+          if [ "${{ github.event.action }}" == "labeled" ]; then
+            for label in "${ALLOWED_LABELS[@]}"; do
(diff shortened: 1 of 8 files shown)


AAdIprog:
@heduenas Please take a look at this pr

frontend/www/js/omegaup/arena/events_socket.test.ts
```diff
@@ -154,6 +154,69 @@ describe('EventsSocket', () => {
     expect(client.socketStatus).toEqual(SocketStatus.Failed);
   });
 
+  it('should clear intervals when the connection is closed', async () => {
```
pabo99, line 157: For this test, the real scenario should be the following:

- The socket fails to connect
- The fallback creates the polling
- The clarificationInterval and rankingInterval intervals are activated if the socket is not connected
- The socket connects
- The polling should be cleared

Could you adjust it so that it reflects the real scenario?

frontend/www/js/omegaup/arena/events_socket.test.ts
```diff
@@ -154,6 +154,69 @@ describe('EventsSocket', () => {
...
+    expect((client as any).rankingInterval).not.toBeNull();
+
+    client.shouldRetry = false;
+
+    (client as any).onclose();
+    expect((client as any).clarificationInterval).toBeNull();
+    expect((client as any).rankingInterval).toBeNull();
+  });
+
+  it('should not leak polling intervals during reconnect cycles with setupPolls', async () => {
```
pabo99, line 186: This test can be adjusted using the same cycle as the previous test, but completing it with the other two cycles. In this one, you should verify that the newly created intervals are different (so there is no leak) and that the entire reconnection flow is validated.

frontend/www/js/omegaup/arena/events_socket.ts
```diff
@@ -246,6 +246,14 @@ export class EventsSocket {
...
       this.socketKeepalive = null;
     }
+    if (this.clarificationInterval) {
+      clearInterval(this.clarificationInterval);
+      this.clarificationInterval = null;
+    }
+    if (this.rankingInterval) {
+      clearInterval(this.rankingInterval);
+      this.rankingInterval = null;
+    }
```
pabo99: According to the flow described in the tests, this is not the place where the intervals should be cleared, since `clarificationInterval` and `rankingInterval` are created when the connection fails. Therefore, the correct place to clear them should be when the `socket` connects. So they can be moved to just before the `socketKeepalive` interval is created.

AAdIprog:
@pabo99 Please take a look i have made the changes that were suggested in the review

AAdIprog:
@pabo99 Please take a look at this

Issue #9518, opened 2026-03-12
Narrow problems-solved/scoreboard cache invalidation to AC grader callback only


shiavm006:
deleteProblemsSolvedRankCacheList() currently invalidates several global caches in more situations than intended, causing unnecessary cache churn and load.

The function’s own TODO says it should only be called from the grader → frontend callback and only when the verdict is AC and the run is not a test run, but it is also called directly from the run submission API and other code paths.

shiavm006:
cc @pabo99  could u see this issue would like to know ur thoughts on this and work on it

pabo99:
It sounds interesting to solve it, go ahead!

PR #9535, opened 2026-03-13
Narrow rank/scoreboard cache invalidation to AC events

shiavm006:
# Description
deleteProblemsSolvedRankCacheList() was being called in more situations than necessary, causing global rank/scoreboard cache invalidation even when standings could not have changed.

This PR narrows cache invalidation in apiRejudge(), apiDisqualify(), and apiRequalify() to only trigger when the previous verdict was AC (the only case where rankings can change).

Per reviewer feedback, apiCreate() keeps calling deleteProblemsSolvedRankCacheList() for now, until we add a backend callback to invalidate caches only when a run transitions to AC.

Partially addresses: #9518


shiavm006:
cc @pabo99 
currently facing issue when running frontend/tests/controllers/RunCreateTest.php inside the frontend container, all 29 tests fail 

When running:

  cd /opt/omegaup
  vendor/bin/phpunit --configuration frontend/tests/phpunit.xml frontend/tests/controllers/RunCreateTest.php

I get:
image

shiavm006:
cc @pabo99 could u review it

frontend/server/src/Controllers/Run.php
```diff
@@ -643,8 +643,12 @@ public static function apiCreate(\OmegaUp\Request $r): array {
...
         }
 
-        // Expire rank cache
-        \OmegaUp\Controllers\User::deleteProblemsSolvedRankCacheList();
+        \OmegaUp\Cache::invalidateAllKeys(
+            \OmegaUp\Cache::CONTESTANT_SCOREBOARD_PREFIX
+        );
+        \OmegaUp\Cache::invalidateAllKeys(
+            \OmegaUp\Cache::ADMIN_SCOREBOARD_PREFIX
+        );
```
pabo99: Currently, what is happening here is that every time there is a submission, all existing cache into the indexes are invalidated until we have a better strategy to invalidate them only when the verdict is AC (which needs to be solved from the backend).

So for now, we need to leave this as it was to avoid affecting rankings where relevant data might remain cached.

The rest of the changes can stay because they slightly mitigate the cache cleanup load, but ultimately we need to find the callback invoked by the backend to fix this at the root.

shiavm006: reverted apiCreate() back to calling deleteProblemsSolvedRankCacheList() to keep the current safe behavior until we have an AC-only invalidation triggered from the backend grader callback.

shiavm006:
CC @pabo99 updated the PR description to Partially addresses #9518 (instead of Fixes) since the root cause needs the backend callback path.

frontend/server/src/Controllers/User.php
```diff
@@ -3018,10 +3018,15 @@ public static function getAuthorRankForTypeScript(\OmegaUp\Request $r) {
...
-     * verdict = AC (and not test run)
+     * Expires rank caches and scoreboard caches.
+     *
+     * Rank caches (PROBLEMS_SOLVED_RANK, USER_COMPARE_DATA) only
+     * change when a submission receives an AC verdict, so callers
+     * must guard invocations accordingly.  Scoreboard caches are
+     * invalidated separately in Run::apiCreate() on every new
+     * submission (verdict is unknown at that point).  This function
+     * is now called conditionally in apiRejudge, apiDisqualify and
+     * apiRequalify (AC verdict + normal submission type).
```
pabo99, line 3032: Could the TODO remain? This will help us keep track of this pending task.

shiavm006: added the todo

frontend/server/src/Controllers/User.php
```diff
@@ -3018,10 +3018,16 @@ public static function getAuthorRankForTypeScript(\OmegaUp\Request $r) {
...
-     * verdict = AC (and not test run)
+     * Expires rank caches and scoreboard caches.
+     * 
+     * Rank caches (PROBLEMS_SOLVED_RANK, USER_COMPARE_DATA) only
+     * change when a submission receives an AC verdict, so callers
+     * must guard invocations accordingly.  Scoreboard caches are
+     * invalidated separately in Run::apiCreate() on every new
+     * submission (verdict is unknown at that point).  This function
+     * is now called conditionally in apiRejudge, apiDisqualify and
+     * apiRequalify (AC verdict + normal submission type).
```
pabo99, line 3032: nit:
```suggestion
     * verdict = AC (and not test run)     
     * 
     * Expires rank caches and scoreboard caches.
     * 
     * Rank caches (PROBLEMS_SOLVED_RANK, USER_COMPARE_DATA) only
     * change when a submission receives an AC verdict, so callers
     * must guard invocations accordingly.  Scoreboard caches are
     * invalidated separately in Run::apiCreate() on every new
     * submission (verdict is unknown at that point).  This function
     * is now called conditionally in apiRejudge, apiDisqualify and
     * apiRequalify (AC verdict + normal submission type).
```

shiavm006:
thanks @pabo99

PR #9539, opened 2026-03-13
Backend Controller Changes 

thunderbird7413:
### Title

Expose admin course/contest targets in problem details and problem list payloads

### Summary

This PR adds backend support for the **quick-add-problem workflow** by including eligible **admin course and contest targets** in the page payloads consumed by the frontend problem screens.

### Why

The frontend quick-add UI requires a **preloaded list of targets** where the current identity can add a problem.

These targets include:

* Courses where the user is an **admin or TA** and the assignment is still open.
* **Ongoing contests** where the user is an administrator.

Providing these targets in the payload removes the need for additional requests from the frontend.

### 1. Added shared Psalm types for quick-add targets

**File:** `Problem.php`

* Introduced new types:

  * `AdminCourseForProblem`
  * `AdminContestForProblem`
* These types represent courses and contests that are valid targets for adding a problem.

### 2. Extended payload contracts

**File:** `Problem.php`

* `ProblemDetailsPayload`

  * Added optional fields:

    * `adminCourses`
    * `adminContests`

* `ProblemListPayload`

  * Added fields:

    * `adminCourses`
    * `adminContests`

These fields allow the frontend to access eligible targets directly from the payload.

### 3. Added shared helper to fetch eligible targets

**File:** `Problem.php`

Added a helper function:

```
getAdminCoursesAndContestsForProblem(identity)
```

This helper:

* Fetches courses where the identity is **admin or TA**
* Filters assignments to include only those that are **pending or still open**
* Fetches contests administered by the identity
* Filters contests to include only **ongoing contests**

### 4. Wired helper into payload builders

**File:** `Problem.php`

Integrated the helper into:

* **Problem details payload builder**
* **Public problem list payload builder**

Both payloads now include:

* `adminCourses`
* `adminContests`

### 5. Updated generated frontend payload interfaces

**File:** `api_types.ts`

* Added or updated **TypeScript interfaces** corresponding to the new payload fields.
* Ensures that frontend types remain **consistent with backend Psalm definitions**.

### `Problem.php`

* Source of truth for:

  * payload structure
  * server-side data retrieval
  * helper logic for retrieving admin targets.

### `api_types.ts`

* Updated **TypeScript payload contracts** to match backend changes.

### Validation

* Lint checks pass locally.
* Type checks pass locally.
* Payload types validated with updated interfaces.

## **Related Issue**

Fixes **#1318**


pabo99:
Could you resolve the conflicts? Preferably revert the changes in the Cypress tests if what you're trying to do is fix a flaky test. It’s better to report that in a separate issue and resolve it with a dedicated PR

thunderbird7413:
> Could you resolve the conflicts? Preferably revert the changes in the Cypress tests if what you're trying to do is fix a flaky test. It’s better to report that in a separate issue and resolve it with a dedicated PR

Okay @pabo99, I have reverted the file.

thunderbird7413:
@pabo99, You can review the changes. I have done the changes as suggested by you.

frontend/server/src/Controllers/Problem.php
```diff
@@ -4772,9 +4774,91 @@ public static function getProblemDetailsForTypeScript(
...
         return $response;
     }
 
+    /**
+     * Returns the list of courses (with pending assignments) and contests
+     * (still ongoing) that the current user administers, for the quick-add feature.
+     *
+     * @return array{adminCourses: list<AdminCourseForProblem>, adminContests: list<AdminContestForProblem>}
+     */
+    private static function getAdminCoursesAndContestsForProblem(
```
pabo99, line 4820: This implementation works but is doing a lot of roundtrips to DB: one per course to fetch assignments plus separate queries for courses and contests.

Since all assignments courses and contests are problemsets, you could fetch everything in a single query filtering by type (assignment vs contest) and finish_time.

Also, this data doesn’t change on every request, so it would be worth caching it per user and invalidating only when a course/contest is created, deleted, or admin rights change. This would make the method much more efficient and scalable.

pabo99 (commented):
In addition to the change I suggested, could you add a simple test where a user has courses and contests, but some are expired, and the test only fetches the active content?

frontend/server/src/Controllers/Problem.php
thunderbird7413: Got it, I have done these changes as the current logic is doing a lot of roundtrips to DB: one per course to fetch assignments plus separate queries for courses and contests.

### Changes done are below with reasons->

* **Problem.php**
  Replaced old multi-query logic with a **DAO-based consolidated query** to remove the **N+1 pattern**. Added `ADMIN_CONTAINERS_CACHE_TTL`, a **TTL helper based on nearest `finish_time`**, and `invalidateAdminCoursesAndContestsForProblemCache()` to ensure timely cache expiration and centralized cache invalidation.

* **Problemsets.php**
  Added `getAdminContainersForProblemQuickAdd(int $identityId, int $currentTimestamp)` which returns unified results for **manageable, non-archived courses with active assignments** and **ongoing contests**. This moves **filtering and access checks to SQL** and fetches all quick-add data in a **single query**.

* **Cache.php**
  Introduced cache key constant `PROBLEM_ADMIN_CONTAINERS` to create a **dedicated cache namespace** for quick-add admin container data.

* **ACLs.php**
  Added `getTypeByAclId(int $aclId): ?string` to determine whether **ACL role changes affect course or contest quick-add visibility**.

* **ACL.php**
  Added `invalidateAdminProblemsetCacheIfNeeded(int $aclId)` and triggered it in **addUser, removeUser, addGroup, and removeGroup** since **ACL changes affect admin/TA visibility in quick-add**.

* **Course.php**
  Added quick-add cache invalidation after **course creation, assignment creation/update, course updates, and archive/unarchive**, as these affect visible **courses and active assignments**.

* **Contest.php**
  Added quick-add cache invalidation after **contest creation, updates, and archive/unarchive** since these operations change **ongoing contests available in quick-add**.

* **ProblemDetailsTest.php**
  Added `testQuickAddPayloadInvalidatesAfterManagedContainersChange` to verify that **creating assignments or contests updates the cached quick-add payload**, preventing regression.

### Result

* **Reduced DB round trips** from `1 + N + 1` to **a single main query**.
* Applied **SQL-side filtering** for active entities (`finish_time`, `archived`).
* Added **per-user caching with proper invalidation** on both **data changes and ACL permission updates**.

pabo99: This change hasn’t been pushed yet, right?

thunderbird7413: Yes, I am doing it right now

thunderbird7413: I have pushed the code, you can review them.  Added the test and changed the logic to remove the round trip to DB, Also implemented cache.

thunderbird7413:
@pabo99, Can you plss help me solve the lint errors.

frontend/server/src/Controllers/ACL.php
```diff
@@ -3,6 +3,16 @@
...
+    private static function invalidateAdminProblemsetCacheIfNeeded(
+        int $aclId
+    ): void {
+        $aclType = \OmegaUp\DAO\ACLs::getTypeByAclId($aclId);
+        if ($aclType !== 'contest' && $aclType !== 'course') {
+            return;
+        }
+        \OmegaUp\Controllers\Problem::invalidateAdminCoursesAndContestsForProblemCache();
+    }
+
```
pabo99, line 15: This and everything related to invalidating all the keys will be reverted. We need to find a way to invalidate the cache by identity, since this will cause the data to be fetched again from the DB.

For now, we’ll need to revert the changes in this file

frontend/server/src/Controllers/Problem.php
```diff
@@ -46,10 +46,12 @@
...
  * @psalm-type ProblemsMineInfoPayload=array{isSysadmin: bool, privateProblemsAlert: bool, visibilityStatuses: array<string, int>, query: string|null}
- * @psalm-type ProblemListPayload=array{selectedTags: list<string>, loggedIn: bool, pagerItems: list<PageItem>, problems: list<ProblemListItem>, keyword: string, language: string, mode: string, column: string, languages: list<string>, columns: list<string>, modes: list<string>, tagData: list<array{name: null|string}>, tags: list<string>, solvedProblemAliases: list<string>, attemptedProblemAliases: list<string>}
+ * @psalm-type ProblemListPayload=array{selectedTags: list<string>, loggedIn: bool, pagerItems: list<PageItem>, problems: list<ProblemListItem>, keyword: string, language: string, mode: string, column: string, languages: list<string>, columns: list<string>, modes: list<string>, tagData: list<array{name: null|string}>, tags: list<string>, solvedProblemAliases: list<string>, attemptedProblemAliases: list<string>, adminCourses: list<AdminCourseForProblem>, adminContests: list<AdminContestForProblem>}
```
pabo99, line 54: Now that I think about it, this particular change is not directly related to problems. The part about assigning problems to courses and contests will be another change. 

So this sounds more like general user information. Therefore, the place to obtain it could be in the User controller. Even better, we could include it in the `commonPayload` that is loaded at startup—I believe that is located in the `UITools.php` file. 

Let me know if you have any questions

frontend/server/src/Controllers/Contest.php
```diff
@@ -2643,6 +2643,7 @@ private static function createContest(
         \OmegaUp\Cache::invalidateAllKeys(
             \OmegaUp\Cache::CONTESTS_LIST_SYSTEM_ADMIN
         );
+        \OmegaUp\Controllers\Problem::invalidateAdminCoursesAndContestsForProblemCache();
```
pabo99, line 2646: Here, the cache should be invalidated only for the contest admin:
```suggestion
        \OmegaUp\Cache::deleteFromCache(
            \OmegaUp\Cache::PROBLEM_ADMIN_CONTAINERS,
            strval($currentUserId)
        );
```

Also, since I previously suggested reverting all the changes in the ACL controller, you should look for the `apiAddAdmin` and `apiRemoveAdmin` endpoints so you can also invalidate the cache for the user being added/removed as an admin.

Regarding the `apiAddGroupAdmin` and `apiRemoveGroupAdmin` endpoints, you can create a function to invalidate the cache for a group of members simply iterating over each of them using the same function described above. That function can also be reused in Courses.

frontend/server/src/Controllers/Contest.php
```diff
@@ -5137,6 +5138,7 @@ function ($value): ?string {
...
```
pabo99, line 5141: For the `apiUpdate` and `apiArchive` endpoints, it will be necessary to add a function that retrieves the `user_id` of all contest admins. In this case, I would prefer to keep this function separate from the one used in `Courses`, since they have different structures. Trying to merge them could introduce inefficient queries that may negatively impact performance.

I created the following query:
```sql

SELECT acl.owner_id
FROM Contests c
INNER JOIN ACLs acl ON acl.acl_id = c.acl_id
WHERE c.contest_id = ?

UNION

SELECT ur.user_id
FROM Contests c
INNER JOIN ACLs acl ON acl.acl_id = c.acl_id
INNER JOIN User_Roles ur ON ur.acl_id = c.acl_id AND ur.role_id = 1
WHERE c.contest_id = ?

UNION

SELECT i.user_id
FROM Contests c
INNER JOIN ACLs acl ON acl.acl_id = c.acl_id
INNER JOIN Group_Roles gr ON gr.acl_id = c.acl_id AND gr.role_id = 1
INNER JOIN Groups_Identities gi ON gi.group_id = gr.group_id
INNER JOIN Identities i ON gi.identity_id = i.identity_id
WHERE c.contest_id = ?;
```

This query appears to make good use of indexes and primary keys in most cases. During the analysis, I identified a couple of indexes that could be added in a separate change:

```sql
-- For User_Roles
CREATE INDEX idx_ur_acl_role ON User_Roles (acl_id, role_id);

-- For Group_Roles
CREATE INDEX idx_gr_acl_role ON Group_Roles (acl_id, role_id);
```

frontend/server/src/Controllers/Course.php
```diff
@@ -971,6 +971,7 @@ private static function createCourseAndGroup(
             }
             throw $e;
         }
+        \OmegaUp\Controllers\Problem::invalidateAdminCoursesAndContestsForProblemCache();
```
pabo99, line 974: Same as in `Contest`, only the course owner’s cache is invalidated

frontend/server/src/Controllers/Course.php
```diff
@@ -1249,6 +1251,7 @@ public static function apiCreateAssignment(\OmegaUp\Request $r): array {
         \OmegaUp\Cache::invalidateAllKeys(
             \OmegaUp\Cache::SCHOOL_STUDENTS_PROGRESS
         );
+        \OmegaUp\Controllers\Problem::invalidateAdminCoursesAndContestsForProblemCache();
```
pabo99, line 1254: An assignment can be created after additional admins have already been added to the course, so in this case we should call a function that retrieves all course admins. The query can be very similar to the one used for contests:
```sql

SELECT acl.owner_id
FROM Courses c
INNER JOIN ACLs acl ON acl.acl_id = c.acl_id
WHERE c.course_id = ?

UNION

SELECT ur.user_id
FROM Courses c
INNER JOIN ACLs acl ON acl.acl_id = c.acl_id
INNER JOIN User_Roles ur ON ur.acl_id = c.acl_id AND ur.role_id = 1
WHERE c.course_id = ?

UNION
         
SELECT i.user_id
FROM Courses c
INNER JOIN ACLs acl ON acl.acl_id = c.acl_id
INNER JOIN Group_Roles gr ON gr.acl_id = c.acl_id AND gr.role_id = 1
INNER JOIN Groups_Identities gi ON gi.group_id = gr.group_id
INNER JOIN Identities i ON gi.identity_id = i.identity_id
WHERE c.course_id = ?;
```

The same composite indexes suggested for contests would also work for this query.

frontend/server/src/Controllers/Course.php
```diff
@@ -1053,6 +1054,7 @@ private static function createAssignment(
             }
             throw $e;
         }
+        \OmegaUp\Controllers\Problem::invalidateAdminCoursesAndContestsForProblemCache();
```
pabo99, line 1057: It seems that nothing needs to be invalidated here because it’s already handled in `apiCreateAssignment`. Unless you want to handle it here instead and avoid doing it in both that endpoint and `apiClone`

frontend/server/src/Controllers/Course.php
```diff
@@ -1410,6 +1413,7 @@ public static function apiUpdateAssignment(\OmegaUp\Request $r): array {
         \OmegaUp\Cache::invalidateAllKeys(
             \OmegaUp\Cache::SCHOOL_STUDENTS_PROGRESS
         );
+        \OmegaUp\Controllers\Problem::invalidateAdminCoursesAndContestsForProblemCache();
```
pabo99, line 1416: Here you reuse the same function used to create an assignment

frontend/server/src/Controllers/Course.php
```diff
@@ -5578,6 +5582,7 @@ public static function apiArchive(\OmegaUp\Request $r): array {
         $archive = $r->ensureOptionalBool('archive') ?? true;
         $course->archived = $archive;
         \OmegaUp\DAO\Courses::update($course);
+        \OmegaUp\Controllers\Problem::invalidateAdminCoursesAndContestsForProblemCache();
```
pabo99, line 5585: Same as in contests

frontend/server/src/Controllers/Course.php
```diff
@@ -6189,6 +6194,7 @@ function ($identity): int {
         \OmegaUp\Cache::invalidateAllKeys(
             \OmegaUp\Cache::SCHOOL_STUDENTS_PROGRESS
         );
+        \OmegaUp\Controllers\Problem::invalidateAdminCoursesAndContestsForProblemCache();
```
pabo99, line 6197: Same as in contests

frontend/server/src/Controllers/Problem.php
```diff
@@ -4772,9 +4776,127 @@ public static function getProblemDetailsForTypeScript(
...
+        }
+
+        return max(
+            1,
+            min(
+                self::ADMIN_CONTAINERS_CACHE_TTL,
+                $nextChangeAt - $now + 1
+            )
+        );
+    }
```
pabo99, line 4812: All this caching logic can be removed in favor of using the methods provided by the `\OmegaUp\Cache` class

frontend/server/src/Controllers/Problem.php
```diff
@@ -4772,9 +4776,127 @@ public static function getProblemDetailsForTypeScript(
...
+        $cache = new \OmegaUp\Cache(
+            \OmegaUp\Cache::PROBLEM_ADMIN_CONTAINERS,
+            strval($identity->identity_id)
+        );
+
+        /** @var array{adminCourses: list<AdminCourseForProblem>, adminContests: list<AdminContestForProblem>}|null */
+        $cachedPayload = $cache->get();
+        if (is_array($cachedPayload)) {
+            return $cachedPayload;
+        }
```
pabo99, line 4837: All of this can be removed as well

frontend/server/src/Controllers/Problem.php
```diff
@@ -4772,9 +4776,127 @@ public static function getProblemDetailsForTypeScript(
...
+            'adminContests' => $adminContests,
+        ];
+
+        $cache->set(
+            $result,
+            self::getAdminCoursesAndContestsForProblemCacheTimeout(
+                $nextChangeAt,
+                $now
+            )
+        );
```
pabo99, line 4895: This and everything that was removed related to caching can be replaced with:

```suggestion\OmegaUp\Cache::getFromCacheOrSet(
            \OmegaUp\Cache::PROBLEM_ADMIN_CONTAINERS,
            strval($identity->identity_id),
            $callback,
            self::ADMIN_CONTAINERS_CACHE_TTL
        )
```
Where `$callback` is the call to the `getAdminContainersForProblemQuickAdd` function and all the data processing it involves

frontend/server/src/DAO/ACLs.php
```diff
@@ -50,4 +50,35 @@ public static function getUserOwnedAclTypesWithAliases(int $userId): array {
...
+                1;';
+
+        /** @var array{type: string}|null */
+        $row = \OmegaUp\MySQLConnection::getInstance()->GetRow(
+            $sql,
+            [$aclId]
+        );
+
+        return $row['type'] ?? null;
+    }
```
pabo99, line 83: With the suggestion to revert all changes related to ACLs, this function should also be reverted

frontend/server/src/DAO/Problemsets.php
```diff
@@ -12,6 +12,141 @@
...
 class Problemsets extends \OmegaUp\DAO\Base\Problemsets {
+    /**
+     * @return list<array{assignment_type: null|string, container_alias: string, container_id: int, container_name: string, entity_type: 'assignment'|'contest', finish_time: \OmegaUp\Timestamp|null, item_alias: string, item_name: string, item_order: int|null}>
+     */
+    public static function getAdminContainersForProblemQuickAdd(
+        int $identityId,
+        int $currentTimestamp
+    ): array {
+        $adminRole = \OmegaUp\Authorization::ADMIN_ROLE;
+        $taRole = \OmegaUp\Authorization::TEACHING_ASSISTANT_ROLE;
```
pabo99, line 23: I was reviewing this, and users with the `TA` role do not have permission to modify course content, so it is not necessary to include them in this function.

frontend/server/src/DAO/Problemsets.php
```diff
@@ -12,6 +12,141 @@
...
+
+        $sql = '
+            SELECT
```
pabo99, line 26: I was reviewing this query and noticed that the current implementation relies on multiple JOINs (`User_Roles`, `Group_Roles`, etc.) combined with `DISTINCT` to remove duplicates. This can lead to incorrect results and unnecessary overhead, since the joins may generate duplicate rows depending on how many roles or groups are associated with the same `ACL`.

I refactored the query to use `EXISTS` instead. This approach avoids generating intermediate duplicates and expresses the intent more clearly: we only need to check whether the user has access through any of the possible paths (owner, user role, or group role).

Additionally, this removes the need for `DISTINCT` and avoids temporary tables, which improves performance. I tested the change and it reduced execution time while also producing the expected results.

Let me know what you think

frontend/server/src/Cache.php
```diff
@@ -582,6 +582,7 @@ class Cache {
     const DATA_CASES_FILES = 'data-cases-files-';
     const PROBLEM_CASES_METADATA = 'problem-cases-metadata-';
     const PROBLEM_IDENTITY_TYPE = 'problems-identity-type-';
+    const PROBLEM_ADMIN_CONTAINERS = 'problem-admin-containers-';
```
pabo99, line 585: Now that we better understand the real goal of this change, this could be renamed:
```suggestion
    const PROBLEMSET_ADMINS = 'problemset-admins-';
```
since it is not directly related to problems, and a container of problems is commonly referred to as a problemset

pabo99 (commented):
I’ve added some comments. Feel free to go through them when you have time, and let me know if you have any questions

Issue #9516, opened 2026-03-12
[BUG] Cookie consent banner does not notify the server and `cookieClickedPostpone` does nothing


ratatall:
## Summary
The cookie consent banner on the homepage has two functional bugs: the server is never informed of the user's accept/decline decision, and the postpone handler is an empty function. Together these make the consent banner non-compliant with privacy regulations such as GDPR.

## Affected Files
- Homepage.vue
- CookieConsent.vue

### Bug A - Server Never Notified of Consent Choice

```ts
// Homepage.vue
cookieClickedAccept() {
  // TODO: make an API to send the response to the server  ← never implemented
  this.$cookies.set('accept-cookies', true, -1);
}

cookieClickedDecline() {
  // TODO: make an API to send the response to the server  ← never implemented
  this.$cookies.set('accept-cookies', false, -1);
}
```

The user's decision is written only to a **client-side browser cookie**. The server has no knowledge of whether the user accepted or declined. If any server-side processing, analytics, or tracking is gated on this consent, it cannot respect the user's choice.

### Bug B - `cookieClickedPostpone` is not yet implemented

```ts
cookieClickedPostpone() {
  // TODO: make an API to send the response to the server
  // (nothing else — function body is empty)
}
```

If the postpone button were ever enabled (currently hidden via `:show-postpone-button="false"` in CookieConsent.vue), clicking it would silently do nothing. The banner would disappear without recording any decision, and would not re-appear on the next visit.

### Expected Behavior
- Clicking Accept or Decline triggers an API call to persist the user's choice server-side.
- Clicking Postpone stores a short-lived preference so the banner re-appears on the next visit.

### Actual Behavior
- No API call is made; only a client-side cookie is written.
- Postpone does nothing at all.

### Proposed Fix
1. Implement the missing API endpoint and call it from both `cookieClickedAccept` and `cookieClickedDecline`.
2. Implement `cookieClickedPostpone` to set a short-lived session cookie (e.g. `'1d'`) so the banner reappears next visit, without recording a permanent decision.

AAdIprog:
@pabo99 Can i work on this issue

ratatall:
@pabo99 Sorry for forgetting to ping you, can I work on this issue myself?

pabo99:
> [@pabo99](https://github.com/pabo99) Sorry for forgetting to ping you, can I work on this issue myself?

Sure!

PR #9543, opened 2026-03-14
Issue/cookie consent banner fix

ratatall:
# Description

Updated the cookie consent banner. 

Fixes: #9516 
# Comments

Cookie decision has been recorded server-side, but not yet in the database since there is no db for that yet (look at comments at User.php).

# Checklist:


frontend/server/src/Controllers/User.php
```diff
@@ -5316,6 +5316,32 @@ public static function getCompareDetailsForTypeScript(
...
+    public static function apiRecordCookieConsent(\OmegaUp\Request $r): array {
+        // Validate required parameter
+        if (!isset($r['accepted'])) {
+            throw new \OmegaUp\Exceptions\InvalidParameterException(
+                'parameterNotFound',
+                'accepted'
+            );
+        }
+
+        $accepted = $r->ensureOptionalBool('accepted', required: true);
```
pabo99: These two instructions are duplicated. While in L5329 you indicate that `accepted` must exist, in L5336 you validate again that it is required.

In this case, the correct approach is to use:
```suggestion
        $accepted = $r->ensureBool('accepted');
```
This internally takes care of validating whether the parameter exists and is of the correct type, and also returns the value so it can be stored in the variable you specify.

frontend/server/src/Controllers/User.php
```diff
@@ -5316,6 +5316,32 @@ public static function getCompareDetailsForTypeScript(
...
+
+        // Cookie consent has been recorded server-side
+        // However, it is not yet recorded to the database
```
pabo99: If this endpoint doesn’t do anything st this point, could it at least send a message to the log?

frontend/www/js/omegaup/components/contest/Form.vue
```diff
@@ -884,7 +884,7 @@ import { library } from '@fortawesome/fontawesome-svg-core';
 import { isValidAlias } from '../../validators';
 library.add(fas);
 
-Vue.use(VueCookies, { expire: -1 });
+Vue.use(VueCookies, { expires: -1 });
```
pabo99, line 905: Hmmm, this sounds like it was resolved in another PR, which has actually already been merged. Could you update your branch so these changes disappear from here?

frontend/www/js/omegaup/components/homepage/Homepage.vue
```diff
@@ -184,18 +184,37 @@ export default class Homepage extends Vue {
...
+    try {
+      // Send user's consent acceptance to the server
+      await omegaup.API.call('user/recordCookieConsent', {
+        accepted: true,
+      });
+    } catch (error) {
+      console.error('Failed to record cookie consent acceptance:', error);
+    }
+    // Set client-side cookie with Secure flag and year-9999 expiry
+    this.$cookies.set('accept-cookies', true, -1, '/', undefined, true);
```
pabo99: Is there a way for you to emit an action so that this can be executed from the `homepage.ts` file, which is the one that controls this component? In general, we try to keep components as isolated as possible from the data layer.

Once you're in the `homepage.ts` file, the correct way to call the endpoint is:
```ts

            api.User.recordCookieConsent({ accpeted: true })
              .then(() => {
                // You can add a message to the user indicating that their response was stored
              })
              .catch((error) => {
                ui.apiError(error);
              });
```

You might see this call fail at the moment. That’s because you haven’t properly propagated the controller changes so that `TypeScript` knows there is a new endpoint. 

You can do that by running: `./stuff/lint.sh`

ratatall:
I decided to leave the cookie decision on a client's side for now, as creating the db better be done in another PR.

pabo99:
Can you review the PR? It looks like there were some conflicts that need to be resolved

ratatall:
@pabo99 I resolved a conflict, please take a look

Issue #9528, opened 2026-03-13
[ENHANCEMENT] Use bootstrap <b-modal> component for confirmation instead of native browser dialog


mazen-fahim:
## Checklist

- [X] I have searched existing issues and confirmed this is not a duplicate (Required)

## Current Behavior

When creating a new contest, if the preset is changed, it shows a native browser confirmation dialog: 
- <img width="450" height="176" alt="Image" src="https://github.com/user-attachments/assets/02b0414c-f963-4353-9731-cbce61b55191" />

instead of using the bootstrap `<b-modal>` modal component used across the codebase (e.g. when creating a new problem):
- <img width="450" alt="Image" src="https://github.com/user-attachments/assets/c7ba4cd6-92bf-4048-bdb2-9cf35a3b8e9e" />

## Expected Behavior

When changing preset it should show the  bootstrap `<b-modal>` component modal for confirmation instead of native browser dialog

mazarona:
should I go on with this? and if so could you please assign it to me @pabo99 🌷

pabo99:
Assigned to you, @mazen-fahim. Feel free to open a PR whenever you have time

PR #9547, opened 2026-03-14
Fix/contest preset modal

mazen-fahim:
# Description

- Preserved behavior for initial selection to directly apply the preset without prompting the user for confirmation.
- Add `<b-modal>` bootstrap modal component as confirmation dialog for changing preset when creating a new contest.
- Add all necessary translations for confirmation modal title.

https://github.com/user-attachments/assets/d121ee73-e717-44ea-96e0-35d52fd79448

Fixes: #9528 

# Comments
I have also finished my first phase for GSoC, my omegaup username is: `mazarona` and would like to contribute more to this codebase 🌷 

# Checklist:

### DIFF
--- a/frontend/server/src/Controllers/Contest.php
+++ b/frontend/server/src/Controllers/Contest.php
@@ -55,7 +55,7 @@ namespace OmegaUp\Controllers;
  * @psalm-type Contestant=array{name: null|string, username: string, email: null|string, gender: null|string, state: null|string, country: null|string, school: null|string}
  * @psalm-type ListItem=array{key: string, value: string}
  * @psalm-type ScoreboardMergePayload=array{contests: list<ContestListItem>}
- * @psalm-type MergedScoreboardEntry=array{name: null|string, username: string, contests: array<string, array{points: float, penalty: float}>, total: array{points: float, penalty: float}, place?: int}
+ * @psalm-type MergedScoreboardEntry=array{name: null|string, username: string, classname: string, contests: array<string, array{points: float, penalty: float}>, total: array{points: float, penalty: float}, place?: int}
  * @psalm-type ContestReport=array{country: null|string, is_invited: bool, name: null|string, place?: int, problems: list<ScoreboardRankingProblem>, total: array{penalty: float, points: float}, username: string}
  * @psalm-type ContestReportDetailsPayload=array{contestAlias: string, contestReport: list<ContestReport>}
  * @psalm-type SettingLimits=array{input_limit: string, memory_limit: string, overall_wall_time_limit: string, time_limit: string}
@@ -4500,6 +4500,7 @@ class Contest extends \OmegaUp\Controllers\Controller {
                     $mergedScoreboard[$username] = [
                         'name' => $userResults['name'],
                         'username' => $username,
+                        'classname' => $userResults['classname'],
                         'contests' => [],
                         'total' => [
                             'points' => 0.0,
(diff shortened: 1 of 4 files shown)


frontend/www/js/omegaup/components/contest/Form.vue
```diff
@@ -1307,13 +1323,12 @@ export default class Form extends Vue {
 
   confirmPresetChange(presetType: PresetType): void {
     if (this.hasFormChanged && !this.update) {
-      if (!confirm(T.contestNewFormPresetOverwriteWarning)) {
-        return;
-      }
+      this.changePresetTo = presetType;
+      this.showModal = true;
```
pabo99, line 1332: The early return could have been kept to avoid the `else` statement.

mazarona:
> Now that you are adding this modal, it would be a good moment to add an extra validation:
> 
> When the selected preset is pressed, it shouldn’t do anything.
> 
> This could be solved in both ways—choose the one you find most convenient:
> 
>     * Disable the button for the selected preset.
> 
>     * Validate that it’s the current preset and prevent the modal from opening.
> 
> 
> There is also another detail: the color of the buttons is confusing. Usually green is used to accept a change, but in this case it is used to cancel the action. Let me take a look at how that could be improved.
> 
> I’m noticing that the modal itself already has a button to close it, so adding the confirm button should be enough (we still need to decide on the color).

Ah okay, 

- for the validation, we can go with disabling choosing the same preset, but I think the user may want to reapply the same preset if he messes up setting the contest and wants to start a clean slate. So choosing the same preset again should be possible as a reset mechanism. Or we may disable it and implement a separate reset button to be more explicit. Let me know what you think 🤔 

- For the colors, Yea it's confusing and it makes sense for the confirm button to be green. I found that there are two potential colors I can use for the `confirm` button.
    - The green color `$omegaup-green: #35b835;` defined in the `frontend/www/sass/base/_colors.scss` file.
    - Ex: image

    - The color defined for `ok` buttons `--btn-ok-background-color: #007bff;` defined in `frontend/www/sass/base/_variables.scss`  file (but it's a bluish color tho).
    -  Ex: image


- And finally, yea removing the cancel button makes sense, it feels redundant.

Let me know if I should go on with these changes 🌷 @pabo99

pabo99:
- Let’s keep the button enabled so the template can be reset
- Let’s use the `--btn-ok-background-color` variable
- Let’s remove the cancel button in favor of the modal icon

If that sounds good to you, you can start applying the changes.

mazarona:
The problem was with bootstrap btn and btn-secondary default :focus & :hover states styling. I had to overwrite it to defaults and introduce a new state of the currently selected preset and conditionally apply my own custom styling.
Here is the current behavior:

demo

frontend/www/js/omegaup/components/contest/Form.vue
```diff
@@ -1526,4 +1547,31 @@ export default class Form extends Vue {
...
+  border-color: var(--btn-ok-background-color);
+  color: var(--btn-ok-font-color);
+
+  &:hover {
+    color: var(--btn-ok-font-color--hover);
+  }
+}
+
+.btn-group .btn.selected {
+  outline: 3px solid #c0c4c7 !important;
```
pabo99: Could you add this color to `_variables.scss` and use it like the others?

mazarona:
Done. Thanks for all the feedback 🌷

pabo99 (approved):
Thanks!

PR #9009, opened 2026-02-07
[FEATURE]: Add DAO support for managing GSoC submission records

isabelccc:
# Description

Feature : Add DAO support for managing GSoC submission records
Fixes: #9008 

# Comments
need to approve the first PR #9007 to proceed this one 

update dao_schema.sql to align with schema.sql and generate VO/ base files

# Checklist:

- [x] The code follows the [coding guidelines](https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/Coding-guidelines.md) of omegaUp.
- [ ] The tests were executed and all of them passed.
- [x] If you are creating a feature, the new tests were added.
- [ ] If the change is large (> 200 lines), this PR was split into various Pull Requests. It's preferred to create one PR for changes in controllers + unit tests in PHPUnit,  and then another Pull Request for UI + tests in Jest, Cypress or both.

Issue #8871, opened 2026-01-24
[FEATURE]Update UI to display and edit social media fields on user profile


Prashant-thakur77:
## Parent Issue
Part of #8692 - Add Social Media Links to User Profile

## Description
Update the user interface to allow users to add, edit, delete, and display social media links on their profiles.

## Technical Details
- Location: `frontend/www/profile/` and related Vue components
- Follow existing UI patterns in the codebase
- Use existing icon library or add new icons as needed
- Maintain consistent styling with current profile page

## Related
- This is step 4 of 4 for implementing #8692

abhinavsingh-hub:
Hey @pabo99 ! 

I would love to work on this issue and complete all the remaining steps

Regards

pabo99:
Assigned  to you, @abhinavsingh-hub

Issue #8867, opened 2026-01-24
[FEATURE][Sub-issue]Update the controllers to store/edit/delete the social media fields


Prashant-thakur77:
## Parent Issue
Part of #8692 - Add Social Media Links to User Profile

## Description
Update application controllers to handle storing, editing, and deleting social media fields on user profiles.

## Technical Details
- Location: `frontend/server/src/Controllers/User.php`
- Follow existing controller patterns in the codebase
- Use existing validation utilities where applicable
- Maintain backward compatibility

## Dependencies
- Depends on: #8860 (DAO updates must be completed first)

## Acceptance Criteria
- [ ] Users can successfully add social media links when creating/editing profiles
- [ ] Social media fields can be updated independently
- [ ] Proper validation prevents invalid data from being saved
- [ ] All existing tests pass
- [ ] New tests cover all CRUD operations for social media fields
- [ ] API responses include social media fields appropriately
- [ ] Code follows omegaUp coding standards

## Related
- This is step 3 of 4 for implementing #8692

waygeance:
Hi @pabo99 this issue looks stale to me, can you assign it to me as no progress is made from significant time ☺️

Prashant-thakur77:
> Hi @pabo99 this issue looks stale to me, can you assign it to me as no progress is made from significant time ☺️ 

Hlo @waygeance i am currently working on it.as its sub issue pr is currently being reviewed which is necessary for this issue.

abhinavsingh-hub:
Hey @pabo99 !

I feel this is a very good feature to be added and would to complete the remaining steps to finalise this feature

Regards

pabo99:
Assigned!

Issue #8692, opened 2025-12-22
[Feature] Add Social Media Links to User Profile


Akarshkushwaha:
Context

Currently, OmegaUp user profiles do not provide an option to add external social media or developer profile links. On platforms like Leetcode, users can link their GitHub or LinkedIn profiles, which helps showcase their work and professional presence.

Feature Request
Add support for social media links in the user profile, such as:
GitHub
LinkedIn
Personal website (optional)
These links can be added from profile settings and displayed on the public profile page.

Why This Is Useful
Improves user profile personalization
Helps users showcase their developer presence
Makes OmegaUp profiles more comparable with other competitive programming platform

Reference
Similar functionality exists on Leetcode user profiles.

Akarshkushwaha:
@pabloaguilar  I want to work on these issue

Piyushrathoree:
@pabloaguilar could you please assign ?

pabo99:
Assigned to @Akarshkushwaha

Tanvir-7793:
I want to work on these issue

PrinceLogic:
I want to work on this feature implementation. If you are not working on these implementation then could you assign to me @pabo99 @Akarshkushwaha

anuragdubey137:
Hello @pabo99, I’m interested in working on this issue and would like to contribute. Could you kindly assign it to me? I’d really appreciate it.

pabo99:
Hello @Akarshkushwaha, are you still interested in working on this issue? If not, please let us know so we can assign it to someone else.

PrinceLogic:
If no one is working on this issue, could you assign to me. # @pabo99 @Akarshkushwaha

pabo99:
This change can be split into the following sub-issues:

- Add the social media fields to the `Users` table
- Update the `DAOs` with the social media fields
- Update the controllers to `store`/`edit`/`delete` the social media fields
- Update the UI to `store`/`edit`/`delete`/`show` them on the user profile

For those interested in contributing to this issue, you can create the sub-issues, self-assign them, and link them to this master-issue to get started.

If you have any questions, we’re happy to help

PR #9554, opened 2026-03-15
feature: Split GSOC custom DAO methods into an isolated change 

### LINKED ISSUE #9008: [FEATURE] GSOC Feature - 2 Creation of DAO
_GSOC Feature(#8207) sub-task - 2 Creation of DAO:_

## Context

DAO (GSoC.php): Database access layer with:
Proper SQL queries with joins and filtering
Parameterized queries for security
Ordered results (newest first)

isabelccc:
feature: Split GSOC custom DAO methods into an isolated change 

# Description

Split custom DAO methods into an isolated change 
frontend/server/src/DAO/GSoCIdeaEdition.php
 frontend/server/src/DAO/GSoCIdea.php
frontend/server/src/DAO/GSoCEdition.php

Fixes: #9008

# Comments

# Checklist:

### DIFF
--- a/frontend/server/src/DAO/GSoCEdition.php
+++ b/frontend/server/src/DAO/GSoCEdition.php
@@ -1,0 +1,72 @@
+<?php
+
+namespace OmegaUp\DAO;
+
+/**
+ * GSoCEdition Data Access Object (DAO).
+ */
+class GSoCEdition extends \OmegaUp\DAO\Base\GSoCEdition {
+    /**
+     * @return array{edition_id: int, year: int, is_active: bool, application_deadline: string|null, created_at: string, updated_at: string}
+     */
+    private static function toPublicArray(
+        \OmegaUp\DAO\VO\GSoCEdition $edition
+    ): array {
+        return [
+            'edition_id' => intval($edition->edition_id),
+            'year' => intval($edition->year),
+            'is_active' => boolval($edition->is_active),
+            'application_deadline' => is_null($edition->application_deadline)
+                ? null
+                : \OmegaUp\DAO\DAO::toMySQLTimestamp($edition->application_deadline),
+            'created_at' => \OmegaUp\DAO\DAO::toMySQLTimestamp($edition->created_at),
+            'updated_at' => \OmegaUp\DAO\DAO::toMySQLTimestamp($edition->updated_at),
+        ];
+    }
+
+    /**
+     * @return list<array{edition_id: int, year: int, is_active: bool, application_deadline: string|null, created_at: string, updated_at: string}>
+     */
+    public static function getEditions(): array {
+        $result = [];
+        foreach (self::getAll() as $edition) {
+            $result[] = self::toPublicArray($edition);
+        }
+        usort(
+            $result,
+            /**
+             * @param array{year: int} $a
+             * @param array{year: int} $b
+             */
+            function (array $a, array $b): int {
+                return $b['year'] <=> $a['year'];
+            }
+        );
+        return $result;
+    }
+
+    /**
+     * @return array{edition_id: int, year: int, is_active: bool, application_deadline: string|null, created_at: string, updated_at: string}|null
+     */
+    public static function getEditionByYear(int $year): ?array {
+        $sql = '
+            SELECT
+                edition_id
+            FROM
+                GSoC_Edition
+            WHERE
+                year = ?
+            LIMIT 1;
+        ';
+        /** @var array{edition_id: int|string}|null */
+        $row = \OmegaUp\MySQLConnection::getInstance()->GetRow($sql, [$year]);
+        if (empty($row)) {
+            return null;
+        }
+        $edition = self::getByPK(intval($row['edition_id']));
+        if (is_null($edition)) {
+            return null;
+        }
+        return self::toPublicArray($edition);
+    }
+}
--- a/frontend/server/src/DAO/GSoCEdition.php
+++ b/frontend/server/src/DAO/GSoCEdition.php
@@ -18,9 +18,15 @@ class GSoCEdition extends \OmegaUp\DAO\Base\GSoCEdition {
             'is_active' => boolval($edition->is_active),
             'application_deadline' => is_null($edition->application_deadline)
                 ? null
-                : \OmegaUp\DAO\DAO::toMySQLTimestamp($edition->application_deadline),
-            'created_at' => \OmegaUp\DAO\DAO::toMySQLTimestamp($edition->created_at),
-            'updated_at' => \OmegaUp\DAO\DAO::toMySQLTimestamp($edition->updated_at),
+                : \OmegaUp\DAO\DAO::toMySQLTimestamp(
+                    $edition->application_deadline
+                ),
+            'created_at' => \OmegaUp\DAO\DAO::toMySQLTimestamp(
+                $edition->created_at
+            ),
+            'updated_at' => \OmegaUp\DAO\DAO::toMySQLTimestamp(
+                $edition->updated_at
+            ),
         ];
     }
 
--- a/frontend/server/src/DAO/GSoCEdition.php
+++ b/frontend/server/src/DAO/GSoCEdition.php
@@ -7,72 +7,42 @@ namespace OmegaUp\DAO;
  */
 class GSoCEdition extends \OmegaUp\DAO\Base\GSoCEdition {
     /**
-     * @return array{edition_id: int, year: int, is_active: bool, application_deadline: string|null, created_at: string, updated_at: string}
-     */
-    private static function toPublicArray(
-        \OmegaUp\DAO\VO\GSoCEdition $edition
-    ): array {
-        return [
-            'edition_id' => intval($edition->edition_id),
-            'year' => intval($edition->year),
-            'is_active' => boolval($edition->is_active),
-            'application_deadline' => is_null($edition->application_deadline)
-                ? null
-                : \OmegaUp\DAO\DAO::toMySQLTimestamp(
-                    $edition->application_deadline
-                ),
-            'created_at' => \OmegaUp\DAO\DAO::toMySQLTimestamp(
-                $edition->created_at
-            ),
-            'updated_at' => \OmegaUp\DAO\DAO::toMySQLTimestamp(
-                $edition->updated_at
-            ),
-        ];
-    }
-
-    /**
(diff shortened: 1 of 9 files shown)


frontend/server/src/DAO/GSoCIdea.php
```diff
@@ -0,0 +1,263 @@
...
+            'contributor_username' => $row['contributor_username'],
+            'created_at' => $row['created_at'],
+            'updated_at' => $row['updated_at'],
+        ];
+    }
+
+    /**
+     * @return list<array{idea_id: int, edition_id: int, title: string, brief_description: string|null, expected_results: string|null, preferred_skills: string|null, possible_mentors: string|null, estimated_hours: int|null, skill_level: string|null, status: string, blog_link: string|null, contributor_username: string|null, created_at: string, updated_at: string}>
+     */
+    public static function getIdeas(
```
pabo99, line 12: I don’t see a practical case where we’d need to fetch ideas from all editions, but we can keep it if you have a justification. What does make sense is fetching all ideas from a specific edition. That function could be called `getIdeasByEdition`, could be added to _**GSoCIdeaEdition.php**_, and could reuse this same query, only making the `$editionId` parameter required.

frontend/server/src/DAO/GSoCIdea.php
```diff
@@ -0,0 +1,263 @@
...
+    ): array {
+        $sql = '
+            SELECT
+                i.idea_id,
+                ie.edition_id,
+                i.title,
+                i.brief_description,
+                i.expected_results,
+                i.preferred_skills,
+                i.possible_mentors,
```
pabo99, line 24: Now that I look at it more closely, this field shouldn’t be in this table. Even if we have the same idea across different editions, the potential mentors could change in each one, so ideally this should belong in the `GSoCIdeaEdition` table. Even better, we could create a separate relational table like `GSoCIdeaEditionMentor` to dynamically assign mentors per edition, instead of using a static text field.

What’s currently shown in the document with the **_Ideas List_** is the mentors’ _**GitHub**_ usernames, so it would be a good idea to revisit issue #8692, which introduces social media links to the user profile, including **_GitHub_**. In fact, the `DAOs` and `VOs` are already created, we would just need to implement the endpoints (#8867) and the UI changes (#8871).


This is a good opportunity to apply this change (i.e., remove this field), since the DAOs and VOs for the GSoC tables haven’t been generated yet.

frontend/server/src/DAO/GSoCIdea.php
```diff
@@ -0,0 +1,263 @@
...
+                i.estimated_hours,
+                i.skill_level,
+                ie.status,
+                i.blog_link,
+                i.contributor_username,
```
pabo99, line 29: Since this is an ideas catalog, it shouldn’t be linked to a contributor. If the intention is to store the username of the contributor who will work on the project, it could instead be added to the `GSoCIdeaEdition` table and removed from here.

frontend/server/src/DAO/GSoCIdea.php
```diff
@@ -0,0 +1,263 @@
...
+        $sql .= ' ORDER BY e.year DESC, i.created_at DESC;';
+
+        /** @var list<array{idea_id: int|string, edition_id: int|string|null, title: string, brief_description: string|null, expected_results: string|null, preferred_skills: string|null, possible_mentors: string|null, estimated_hours: int|string|null, skill_level: string|null, status: string|null, blog_link: string|null, contributor_username: string|null, created_at: string, updated_at: string}> */
+        $rows = \OmegaUp\MySQLConnection::getInstance()->GetAll($sql, $params);
+
+        $result = [];
+        foreach ($rows as $row) {
+            $result[] = self::toPublicArray($row);
+        }
+        return $result;
```
pabo99: The `GetAll` function actually returns a list of arrays, so this should be enough:
```suggestion
        return \OmegaUp\MySQLConnection::getInstance()->GetAll($sql, $params);
```

frontend/server/src/DAO/GSoCIdea.php
```diff
@@ -0,0 +1,263 @@
...
+                e.year DESC
+            LIMIT 1;
+        ';
+
+        /** @var array{idea_id: int|string, edition_id: int|string|null, title: string, brief_description: string|null, expected_results: string|null, preferred_skills: string|null, possible_mentors: string|null, estimated_hours: int|string|null, skill_level: string|null, status: string|null, blog_link: string|null, contributor_username: string|null, created_at: string, updated_at: string}|null */
+        $row = \OmegaUp\MySQLConnection::getInstance()->GetRow($sql, [$ideaId]);
+        if (empty($row)) {
+            return null;
+        }
+        return self::toPublicArray($row);
```
pabo99: Same case here: `GetRow` already returns an array or null, so this is enough:
```suggestion
        return \OmegaUp\MySQLConnection::getInstance()->GetRow($sql, [$ideaId]);
```

frontend/server/src/DAO/GSoCIdea.php
```diff
@@ -0,0 +1,263 @@
...
+                } elseif (!is_null($status)) {
+                    $ideaEdition->status = $status;
+                    $affectedRows += \OmegaUp\DAO\GSoCIdeaEdition::update(
+                        $ideaEdition
+                    );
+                }
+            }
+        }
+        return $affectedRows;
+    }
```
pabo99, line 208: Since this doesn’t involve any custom queries and only contains the logic to create/update ideas, it could  be added to the controllers in a later change.

frontend/server/src/DAO/GSoCIdea.php
```diff
@@ -0,0 +1,263 @@
...
+                $row['estimated_hours']
+            ),
+            'skill_level' => $row['skill_level'],
+            'status' => is_null($row['status']) ? 'Proposed' : $row['status'],
+            'blog_link' => $row['blog_link'],
+            'contributor_username' => $row['contributor_username'],
+            'created_at' => $row['created_at'],
+            'updated_at' => $row['updated_at'],
+        ];
+    }
```
pabo99: This function would no longer be necessary and could be removed:
```suggestion
```

frontend/server/src/DAO/GSoCIdeaEdition.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace OmegaUp\DAO;
+
+/**
+ * GSoCIdeaEdition Data Access Object (DAO).
+ */
+class GSoCIdeaEdition extends \OmegaUp\DAO\Base\GSoCIdeaEdition {
+    public static function getByIdeaAndEdition(
```
pabo99, line 9: This will only fetch IDs, status, and decision notes. The queries you added in `GSoCIdea` would make more sense here

frontend/server/src/DAO/GSoCEdition.php
```diff
@@ -0,0 +1,78 @@
...
+            $result,
+            /**
+             * @param array{year: int} $a
+             * @param array{year: int} $b
+             */
+            function (array $a, array $b): int {
+                return $b['year'] <=> $a['year'];
+            }
+        );
+        return $result;
```
pabo99: Instead of fetching a list of `Edition` objects and then convert to array, it would be better to get the data you want to return in this function as an array, and perform the sorting directly in the query:
```suggestion
        $fields = \OmegaUp\DAO\DAO::getFields(
            \OmegaUp\DAO\VO\GSoCEdition::FIELD_NAMES,
            'ge'
        );
        $sql = "SELECT {$fields} FROM `GSoC_Edition` `ge` ORDER BY `ge`.`year` DESC LIMIT 50;";
        /** @var list<array{application_deadline: null|string, created_at: null|string, edition_id: int, is_active: bool, updated_at: null|string, year: int}> */
        return \OmegaUp\MySQLConnection::getInstance()->GetAll($sql);
```

frontend/server/src/DAO/GSoCEdition.php
```diff
@@ -0,0 +1,78 @@
...
+        /** @var array{edition_id: int|string}|null */
+        $row = \OmegaUp\MySQLConnection::getInstance()->GetRow($sql, [$year]);
+        if (empty($row)) {
+            return null;
+        }
+        $edition = self::getByPK(intval($row['edition_id']));
+        if (is_null($edition)) {
+            return null;
+        }
+        return self::toPublicArray($edition);
```
pabo99: Same case here:
```suggestion
        $fields = \OmegaUp\DAO\DAO::getFields(
            \OmegaUp\DAO\VO\GSoCEdition::FIELD_NAMES,
            'ge'
        );
        $sql = "SELECT {$fields} FROM `GSoC_Edition` `ge` WHERE `ge`.`year` = ? LIMIT 1;";
        /** @var array{application_deadline: null|string, created_at: null|string, edition_id: int, is_active: bool, updated_at: null|string, year: int}|null */
        return \OmegaUp\MySQLConnection::getInstance()->GetRow($sql, [$year]);
```
This also helps to make a single round trip to the DB instead of two.

frontend/server/src/DAO/GSoCEdition.php
```diff
@@ -0,0 +1,78 @@
...
+                    $edition->application_deadline
+                ),
+            'created_at' => \OmegaUp\DAO\DAO::toMySQLTimestamp(
+                $edition->created_at
+            ),
+            'updated_at' => \OmegaUp\DAO\DAO::toMySQLTimestamp(
+                $edition->updated_at
+            ),
+        ];
+    }
```
pabo99: With the suggested changes, this function can also be removed:
```suggestion
```

pabo99 (commented):
Can you take a look at https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/Useful-Commands-for-Development.md#regenerate-dao-files-from-schema

Once it’s clear, could you run the command described there?

This should make some of the errors you encountered in the workflows disappear.

But wait— with the suggested changes, it might be worth putting this PR on standby until the recommendations to remove some fields and create others are applied.

frontend/server/src/DAO/GSoCIdea.php
isabelccc: Great point, thanks! I agree possible_mentors should be edition-scoped, not idea-scoped. I’ll update the schema proposal to remove it from GSoC_Idea and move toward an edition-level model (preferably a relational mentor table per idea-edition). I’ll keep this PR on standby and regenerate DAOs/VOs only after we align on this design.see #9009 for my new schema design proposal

Issue #8654, opened 2025-12-18
Implement Vue.js components for User Profile READMEs


mohi-devhub:
### Context
The backend API is now ready to serve, store, and manage Profile READMEs. The frontend needs the necessary Vue.js components to display, edit, and report this information, integrating smoothly into the existing user profile page.

### Scope
- **Display Component:**
    - Modify `frontend/www/js/omegaup/components/user/ViewProfile.vue` to handle the new `readme` field within its profile or data prop.
    - When content is present, render it through `marked` and sanitize it with `dompurify` before display.
    - Add a "Report" button that calls the new `omegaup.User.reportReadme` API method. Provide user feedback on success.
    - If the README is disabled, display an appropriate message instead of the content.
- **Editing Interface:**
    - On a user's own profile, provide an "Edit README" button.
    - This button should toggle an editing view (e.g., a modal or an inline textarea component).
    - A "Save" button will call the `omegaup.User.updateReadme` API method with the new content.
- **API Integration:** Ensure all new UI elements correctly call the auto-generated API client methods (e.g., `omegaup.User.updateReadme`) and handle their promises for feedback.

### Out of Scope
- Any backend (PHP) or database changes.
- Major refactoring of the profile page layout; new elements should fit the existing design.

### Next Steps
- End-to-end testing of the complete feature, from editing and saving to viewing and reporting.

arghabanerjee54:
Hi @mohi-devhub,

I’ve reviewed `ViewProfile.vue` and how profile sections are currently structured (tabs + conditional rendering based on privacy).

My plan is to integrate the README as a new tab (e.g. `ViewProfileTabs.Readme`) to avoid impacting existing sections and keep changes localized. The tab would:
- Render the README content using `marked` + `DOMPurify`
- Handle the disabled state with a message instead of content
- Expose a “Report” action wired to `omegaup.User.reportReadme`
- Show an “Edit README” action only on `profile.is_own_profile`

Before I start coding, I just want to confirm:
1. Is adding a dedicated README tab the preferred UX, rather than embedding it inline?
2. For reporting, is a simple success/error toast sufficient, consistent with other profile actions?

Once confirmed, I’ll open a PR scoped only to the display logic first.

mohi-devhub:
This issue is actually one part of the original, larger issue that I posted, and I was planning to work on the overall feature myself.

PR #9560, opened 2026-03-16
Add profile README display, edit, and report UI 

mohi-devhub:
## Summary

Implements the frontend for the **User Profile README** feature.  
The backend API (`saveReadme`, `reportReadme`, and `readme` field in `apiProfile`) was already merged via **#9279**.

- **Display:** README is rendered as Markdown above the profile tabs using the existing `omegaup-markdown` component  
- **Own profile:** **Edit** button toggles an inline textarea editor with **Save / Cancel**  
- **Visitor view:** **Report README** button submits a report and is replaced with a success message  
- **Empty state:** Own profile with no README sees a prompt to add one  
- **i18n:** 5 new strings added for **EN, ES, PT, and pseudo** locales  

---
---

## Screenshots

- Edit mode  

Screenshot 2026-03-16 at 2 07 02 PM

---

- Own profile (rendered)

Screenshot 2026-03-16 at 2 07 10 PM

---

- Visitor view

Screenshot 2026-03-16 at 2 09 57 PM

---

- After reporting

Screenshot 2026-03-16 at 2 10 06 PM

---
---

## Files changed

- `components/user/ViewProfile.vue` — README section + edit/report logic  
- `frontend/templates/{en,es,pt,pseudo}.lang` — i18n template files  
- `lang.{en,es,pt,pseudo}.ts / .json` — auto-generated i18n files  

---

Fixes : #8654


frontend/www/js/omegaup/components/user/ViewProfile.vue
```diff
@@ -401,6 +479,33 @@ export default class ViewProfile extends Vue {
...
+  }
+
+  saveReadme(): void {
+    api.User.saveReadme({ readme: this.readmeEditContent })
+      .then(() => {
+        this.currentReadme = this.readmeEditContent;
+        this.isEditingReadme = false;
+      })
+      .catch(ui.apiError);
+  }
```
pabo99: Could you move this and all the calls to the entrypoints to the `profile.ts` file and emit the actions from there? That way we can keep the components separated from the data layer.

mohi-devhub:
@pabo99 Have made the necessary changes

frontend/www/js/omegaup/components/user/ViewProfile.vue
```diff
@@ -1,5 +1,73 @@
...
+              ></omegaup-markdown>
+              <p v-else-if="profile.is_own_profile" class="text-muted mb-0">
+                {{ T.profileReadmeAddPrompt }}
+              </p>
+              <div
+                v-if="readmeReportSubmitted"
+                class="alert alert-success mt-2 mb-0"
+              >
+                {{ T.profileReadmeReportSuccess }}
+              </div>
```
pabo99: Could you use the global notifications instead of adding a div that does the same thing?

frontend/www/js/omegaup/components/user/ViewProfile.vue
```diff
@@ -339,6 +409,10 @@ export default class ViewProfile extends Vue {
   columns = 3;
   currentSelectedTab = getInitialSelectedTab(this.profile, this.selectedTab);
   normalizedRunCounts: Highcharts.PointOptionsObject[] = [];
+  currentReadme: string | null = this.profile.readme ?? null;
+  isEditingReadme = false;
+  readmeEditContent = '';
```
pabo99: Could this variable accept null? That way there’s no need to assign it magic values:
```suggestion
  readmeEditContent: string | null = null;
```

frontend/www/js/omegaup/components/user/ViewProfile.vue
```diff
@@ -401,6 +475,35 @@ export default class ViewProfile extends Vue {
     }
   }
 
+  startEditReadme(): void {
+    this.readmeEditContent = this.currentReadme ?? '';
```
pabo99: Same here:
```suggestion
    this.readmeEditContent = this.currentReadme;
```

frontend/www/js/omegaup/components/user/ViewProfile.vue
```diff
@@ -401,6 +475,35 @@ export default class ViewProfile extends Vue {
...
+    this.isEditingReadme = true;
+  }
+
+  cancelEditReadme(): void {
+    this.isEditingReadme = false;
+    this.readmeEditContent = '';
```
pabo99: It’s better to assign null here:
```suggestion
    this.readmeEditContent = null;
```

frontend/www/js/omegaup/components/user/ViewProfile.vue
```diff
@@ -401,6 +475,35 @@ export default class ViewProfile extends Vue {
...
+        this.currentReadme = this.readmeEditContent;
+        this.isEditingReadme = false;
+      },
+    });
+  }
+
+  reportReadme(): void {
+    this.$emit('report-readme', {
+      username: this.profile.username ?? '',
+      onSuccess: () => {
```
pabo99, line 500: Oh, this is the first time we pass a callback as a parameter. It keeps the component’s data better encapsulated. Thanks for introducing it, we’ll definitely keep using this approach

mohi-devhub:
@pabo99 I hope everything looks good now !!

pabo99 (commented):
The PR looks good, but I’m not completely convinced it’s ready to be merged yet, since there are many improvements I’d like to implement. What we can do is find a way to temporarily hide it, and regarding the current changes, I can give you several suggestions that can be turned into smaller, separate issues so they can be reviewed and approved quickly.

What do you think? Maybe we can add a feature flag so it’s easy to enable the feature once the new improvements are ready. Can you think of a way to hide it? Once you have an idea, go ahead and push it and we’ll merge it.

mohi-devhub:
@pabo99 
Great idea! I've added a feature flag using the existing Experiments system. The README feature is now gated behind the user_readme experiment:

- Backend: The profile API only returns readme data when the experiment is enabled, and the saveReadme/reportReadme endpoints return 404 if it's not enabled.
- Frontend: The entire README section in ViewProfile.vue is hidden when the experiment is off. 
- To enable it, you can either define EXPERIMENT_USER_README in the config, or add it per-user via the admin panel /Users_Experiments table. Once the improvements are ready, we can enable it globally and then remove the flag.
- I'm also happy to turn your improvement suggestions into smaller, separate issues , please share them and I'll get started!

Issue #9455, opened 2026-03-06
[BUG] "Or upload the file" label and file input break onto separate lines on mobile


Yashagarwal9798:
## Current Behavior

On the Problem Creator page (`/problem/creator/`), when viewing the Code tab on mobile (e.g., iPhone 12 Pro, 390px width), the **"Or upload the file"** label and the **"Choose File"** button are on **separate lines** instead of sitting next to each other on the same line.

This happens because the label uses `col-12` for small screens, which forces it to take the full width and pushes the file input below it.

In `CodeTab.vue` (lines 42–57):
```html
<div class="form-group row mt-3 align-items-center">
  <label class="col-12 col-sm-auto ...">    <!-- col-12 = full width on mobile -->
    Or upload the file
  </label>
  <div class="col-12 col-sm-auto ...">       <!-- also col-12 = drops to next line -->
    <input type="file" ... />
  </div>
</div>
```

On desktop (`col-sm-auto`), both sit side by side. On mobile (`col-12`), each takes the full row width, so they stack vertically.

## Expected Behavior

The label ("Or upload the file") and the file input ("Choose File") should appear on the **same line** on all screen sizes, just like the "Language" label and its dropdown above it.

## If necessary, add the following information (otherwise delete this section):

1. Go to `omegaup.com/problem/creator/`
2. Click the "Code" tab
3. Open browser DevTools and switch to a mobile view (e.g., iPhone 12 Pro, 390px)
4. Scroll down to the "Or upload the file" section
5. Notice that the label and file input are stacked on two lines instead of one

*Relevant details about the environment in which you encountered this bug:*

* Happens on mobile screen widths (below 576px where `col-sm` breakpoint kicks in)
* File: `frontend/www/js/omegaup/components/problem/creator/code/CodeTab.vue` (lines 42–57)

<img width="1069" height="1021" alt="Image" src="https://github.com/user-attachments/assets/d9531880-bba1-4964-92f4-cd54ed7db692" />

Yashagarwal9798:
If you would like to work on this issue, please state your proposed approach or solution in the comments below. To give everyone a fair chance, we will review the proposals, and the maintainers will assign the issue to the person with the best approach!

*Note: This is intended to be a "good first issue" for newcomers. Please only request to work on this if you currently have fewer than 2 active issues.*

UraModi:
[...]

PR #9569, opened 2026-03-18
fix: keep upload label and input on same line on mobile

mxnueh:
# Description
On mobile screens (below 576px), the "Or upload the file" label and the "Choose File" input were stacking on separate lines due to Bootstrap's `col-12` class forcing full width on small screens.

Replaced the Bootstrap grid classes with `d-flex flex-nowrap` on the container so the label and file input always stay on the same line regardless of screen size.

Fixes: #9455

# Comments
The change is limited to `CodeTab.vue` (3 lines modified). Below are screenshots showing the before and after on mobile (iPhone 12 Pro, 390px):

**Before:** label and input stacked on separate lines

image

**After:** label and input on the same line

image

# Checklist:


pabo99 (commented):
Puedes actualizar tu rama? parece que hay cambios de otros PRs

Yashagarwal9798:
hello @mxnueh , The label and input staying on the same line is an improvement, but there's a UX issue with the current approach:

On narrow screens, the filename gets truncated (only the first letter or "N" from "No file chosen" is visible). This means the user can't verify which file they selected.

I think you should try this:

When a file is selected, to show the full filename we can add a small bar that the user can slide/scroll sideways to read the complete name. This way nothing gets cut off, even on small phones.

also one more thing that i noticed:
we can Add a small delete/trash icon next to the filename that only shows up after a file is picked. Tapping it clears the selection and goes back to showing "No file chosen." Right now there's no easy way to undo a file selection.

The main thing is: the user needs to be able to see the full filename of what they selected. A scrollable bar solves that nicely on mobile.

@pabo99 what do you think about this approch?

mxnueh:
Thanks for the great suggestions! You're absolutely right — showing the full filename is important for the user experience. I'll go ahead and implement the scrollable bar for the filename so it doesn't get cut off on narrow screens.

Regarding the delete/trash icon, I think that's also a nice touch. I'll add it as well if @pabo99 agrees with that approach. 

I'll update the PR shortly!

frontend/www/js/omegaup/components/problem/creator/code/CodeTab.vue
```diff
@@ -39,20 +39,37 @@
...
             @change="handleInputFile"
           />
+          <button
+            class="btn btn-secondary btn-sm text-nowrap"
+            type="button"
+            @click="$refs.fileInput.click()"
+          >
+            Choose File
+          </button>
+          <div class="mx-2 text-nowrap" style="overflow-x: auto; max-width: 100px;">
```
pabo99: It seems the idea is for the entire row content to have horizontal scrolling, so these styles wouldn’t belong here. Also, we usually avoid inline styles due to potential XSS or CSS injection risks. Could you remove them from here?
```suggestion
          <div class="mx-2 text-nowrap">
```

frontend/www/js/omegaup/components/problem/creator/code/CodeTab.vue
```diff
@@ -39,20 +39,37 @@
           </div>
         </div>
       </div>
-      <div class="form-group row mt-3 align-items-center">
-        <label class="col-12 col-sm-auto col-form-label mb-2 mb-sm-0 pr-sm-2">
+      <div class="form-group d-flex flex-nowrap mt-3 align-items-center">
```
pabo99: Instead, can you add the `overflow-auto` class here?
```suggestion
      <div class="form-group d-flex flex-nowrap overflow-auto mt-3 align-items-center">
```

frontend/www/js/omegaup/components/problem/creator/code/CodeTab.vue
```diff
@@ -39,20 +39,37 @@
...
+        <label class="col-form-label mb-0 mr-2 text-nowrap">
           {{ T.problemCreatorCodeUpload }}
         </label>
-        <div
-          class="col-12 col-sm-auto pl-sm-0 d-flex align-items-center overflow-hidden"
-        >
+        <div class="d-flex align-items-center flex-grow-1 overflow-hidden">
```
pabo99: This `overflow-hidden` would cause the horizontal scroll to disappear on small devices. Could you remove it?
```suggestion
        <div class="d-flex align-items-center flex-grow-1">
```

frontend/www/js/omegaup/components/problem/creator/code/CodeTab.vue
```diff
@@ -210,6 +229,13 @@ export default class CodeTab extends Vue {
     this.selectedLanguage = language;
   }
 
+  clearFile(): void {
+    this.selectedFileName = '';
+    (this.$refs.fileInput as HTMLInputElement).value = '';
```
pabo99: There’s a more Vue-friendly way to use refs. `vue-property-decorator` provides the `@Ref` decorator, which you can add near the props like this:

```ts
  @Ref('fileInput') fileInput!: HTMLInputElement;
```

This way, you won’t need to cast it here anymore since it’s already an `HTMLInputElement`:

```suggestion
    this.fileInput.value = '';
```
You should also update the other places where it’s used.

And don’t forget to import it from `vue-property-decorator`, like the rest of the decorators.

frontend/www/js/omegaup/components/problem/creator/code/CodeTab.vue
pabo99: Here you could also simulate a click on the `fileInput` to make it behave as closely as possible to the native file upload element

frontend/www/js/omegaup/components/problem/creator/code/CodeTab.vue
```diff
@@ -39,20 +39,37 @@
...
             type="file"
             name="file"
             @change="handleInputFile"
           />
+          <button
+            class="btn btn-secondary btn-sm text-nowrap"
+            type="button"
+            @click="$refs.fileInput.click()"
+          >
+            Choose File
```
pabo99: Could you add this and all the other text strings to the localization files?

frontend/www/js/omegaup/components/problem/creator/code/CodeTab.vue
```diff
@@ -39,20 +39,37 @@
...
+          <div class="mx-2 text-nowrap" style="overflow-x: auto; max-width: 100px;">
+            {{ selectedFileName || 'No file chosen' }}
+          </div>
+          <button
+            v-if="selectedFileName"
+            class="btn btn-sm text-danger p-0 flex-shrink-0"
+            type="button"
+            @click="clearFile"
+          >
+            🗑️
```
pabo99: Could you use a FontAwesome icon? There should be plenty of examples in the code of how to include a delete button.

mxnueh:
Done! @pabo99 Thank you for the feedback. I Applied all the suggested changes:
- Added overflow-auto to the main container
- Removed overflow-hidden from the inner div
- Removed inline styles from the filename div
- Used @Ref decorator instead of $refs cast
- Button now simulates click on fileInput via @Ref
- Replaced emoji 🗑️ with FontAwesome trash icon
- Added "Choose File" and "No file chosen" strings to all localization files (en, es, pt, pseudo)

## How it looks Now

image

image

frontend/www/js/omegaup/components/problem/creator/code/CodeTab.vue
```diff
@@ -55,20 +55,20 @@
           <button
             class="btn btn-secondary btn-sm text-nowrap"
             type="button"
-            @click="$refs.fileInput.click()"
+            @click="fileInput.click()"
           >
-            Choose File
+            {{ T.problemCreatorCodeChooseFile }}
```
pabo99, line 62: The right way to generate the new translation strings is by adding them to the files:

- _**frontend/templates/en.lang**_
- _**frontend/templates/es.lang**_
- **_frontend/templates/pt.lang_**

Once that’s done, you need to run _**./stuff/lint.sh**_ to propagate the changes to the frontend files.

pabo99:
The linter is still failing. Could you run:
```bash
  ./stuff/lint.sh fix -- frontend/templates/en.lang frontend/templates/es.lang frontend/templates/pseudo.lang frontend/templates/pt.lang frontend/www/js/omegaup/components/problem/creator/code/CodeTab.vue frontend/www/js/omegaup/lang.en.json frontend/www/js/omegaup/lang.en.ts frontend/www/js/omegaup/lang.es.json frontend/www/js/omegaup/lang.es.ts frontend/www/js/omegaup/lang.pseudo.json frontend/www/js/omegaup/lang.pseudo.ts frontend/www/js/omegaup/lang.pt.json frontend/www/js/omegaup/lang.pt.ts
```
one more time? If you’re having trouble running this command, let me know and I can help you find a final solution.

mxnueh:
Done! @pabo99 I was having trouble running ./stuff/lint.sh on Windows due to a Docker path issue, but I managed to run it through WSL and all the lang files have been synced. The Jest test selector has also been updated.

pabo99:
Could you push your latest changes? This way we can check if everything is okay

Issue #9570, opened 2026-03-18
[BUG] Scoreboard merge should display usernames with ranking-based colors


pabo99:
## Checklist

- [x] I have searched existing issues and confirmed this is not a duplicate (Required)

## Current Behavior

In the Scoreboard merge view, all usernames are displayed using a single color, regardless of their ranking.

## Expected Behavior

Usernames should be displayed using the same color scheme based on ranking as in the global rankings view (e.g., top users in red, then blue, green, etc.), to maintain consistency across the platform.

## Steps to Reproduce

- Go to the **_Scoreboard merge_** in the **_Contests_** menu.
- Choose two contests at least to merge their scoreboards (in local environment you can use **_pasado_** and **_prueba_** contests)
- Observe the list of usernames in the table.
- Compare it with the Coders with the most points in **_Ranking > User ranking_** view.

## Context

In other parts of the platform (such as the global rankings), usernames are color-coded according to their rank. This visual distinction is missing in the Scoreboard merge view, which makes it inconsistent and less intuitive.

## Screenshots

<img width="1080" height="741" alt="Image" src="https://github.com/user-attachments/assets/77e27994-2186-4f9d-8345-cfb1c561d2ae" />

<img width="1080" height="741" alt="Image" src="https://github.com/user-attachments/assets/49691942-783b-4612-8837-e720e249d883" />

## Possible Solution

Reuse the same logic or helper used to assign ranking-based colors in the global rankings and apply it to the usernames in the _**Scoreboard merge**_ view.

johanaguilar12:
Hi @pabo99, I'd like to work on this issue.

UraModi:
@pabo99 Hi! I was able to reproduce this issue locally and looked into how username coloring is handled across the platform. It seems that the Scoreboard merge view is not reusing the existing rank-based color logic that is already implemented in the global rankings.

A clean approach would be to reuse the same helper/utility (used in the rankings view) that assigns colors based on user rank, and apply it to the usernames in the Scoreboard merge table. This would ensure consistency in UX and avoid duplicating styling logic.

I can take care of identifying the shared logic and refactoring the Scoreboard merge component to use it in a minimal and maintainable way.

If this sounds good, I’d be happy to work on a PR for this.

PR #9573, opened 2026-03-18
Fix scoreboard merge usernames not showing ranking-based colors

johanaguilar12:
# Description

Fixes an issue where usernames in the scoreboard merge view were not
displayed with ranking-based colors. This change propagates the existing `classname` field from the
scoreboard API to the merged scoreboard and applies it in the frontend.

Usernames are now rendered using an anchor element with the appropriate CSS class, ensuring consistency with other views such as the global ranking.

image

Fixes: #9570

# Comments

Additionally, the username element was updated to use an <a> tag to match existing CSS selectors used in other views.

# Checklist:

- [ ✅] The code follows the [coding guidelines](https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/Coding-guidelines.md) of omegaUp.
- [ ✅] The tests were executed and all of them passed.
- [  ] If you are creating a feature, the new tests were added.
- [ ✅] If the change is large (> 200 lines), this PR was split into various Pull Requests. It's preferred to create one PR for changes in controllers + unit tests in PHPUnit,  and then another Pull Request for UI + tests in Jest, Cypress or both.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -7616,9 +7616,9 @@ flat-cache@^3.0.4:
     rimraf "^3.0.2"
 
 flatted@^3.1.0:
-  version "3.1.1"
-  resolved "https://registry.npmjs.org/flatted/-/flatted-3.1.1.tgz"
-  integrity sha512-zAoAQiudy+r5SvnSw3KJy5os/oRJYHzrzja/tBDqrZtNhUw8bt6y8OBzMWcjWr+8liV8Eb6yOhw8WZ7VFZ5ZzA==
+  version "3.4.2"
+  resolved "https://registry.yarnpkg.com/flatted/-/flatted-3.4.2.tgz#f5c23c107f0f37de8dbdf24f13722b3b98d52726"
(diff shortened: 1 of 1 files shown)


frontend/www/js/omegaup/components/contest/ScoreboardMerge.vue
```diff
@@ -50,7 +50,12 @@
               <th>{{ rank.place }}</th>
               <th>
                 <div class="username" data-test-merged-username>
-                  {{ rank.username }}
+                  <a
+                    :href="`/profile/${rank.username}/`"
+                    :class="rank.classname"
+                  >
+                    {{ rank.username }}
+                  </a>
```
pabo99: Great, you’re already getting the classname for each user in the ranking. Now there’s just one detail left:

The color classes are encapsulated in the `Username.vue` component, which is responsible for formatting the usernames displayed on the platform. So, would it be possible to use that component the same way it’s used in the ranking view?

It would be something like:
```suggestion
                <omegaup-user-username
                  :classname="rank.classname"
                  :linkify="true"
                  :username="rank.username"
                ></omegaup-user-username>
```
And then you would just need to import and register the component

johanaguilar12: I've updated the code to use the omegaup-user-username component as suggested!

PR #9586, opened 2026-03-19
Fix course link overflow

### LINKED ISSUE #9541: [BUG] Course repository link overflows outside the card on mobile view
## Description
When viewing the course page in a mobile viewport, the repository link inside the course description overflows outside the card container. This causes the layout to break and the link extends beyond the card boundaries.

This appears to happen because the URL is long and does not wrap properly within the card element.

### Steps to Reproduce
1. Open the course page:  
   https://omegaup.com/course/Curso-OMI/
2. Open browser DevTools.
3. Switch to a mobile viewport (e.g., iPhone SE – 375px width).
4. Scroll to the course description card.
5. Observe the GitHub repository link.

### Actual Behavior
The repository link overflows outside the card instead of wrapping within the card container.

### Expected Behavior
The link should wrap inside the card and remain within the container boundaries so the layout stays consistent on mobile devices.

### Screenshot

Image

UraModi:
Fixes overflow issue where long URLs in course description break layout on mobile view.

Adds CSS rules to allow long links to wrap within the container.

1 file changed
2 lines added

Fixes #9541


UraModi:
Before:
before
After:
after
@pabo99 I have added the screenshots for review.
Thanks🌻

PR #9592, opened 2026-03-19
Implement infinite scroll across global submissions view

### LINKED ISSUE #8026: [FEATURE] Implement infinite scrolling in the Global Submissions view
As a sysadmin, I want to see a higher number of submissions for quick search in:

http://localhost:8001/arena/admin/

## Context
It seems like the pagination in this view is not working properly, and it's not worth maintaining it. It would be better to migrate to infinite scrolling.

## Example
Here is a [PR](https://github.com/omegaup/omegaup/pull/7553) with an example of how to migrate to infinite scrolling.

thunderbird7413:
## Problem

The Global Submissions admin page (`/arena/admin/`) used page-based navigation (prev/next buttons) to browse through runs. This required manual clicking to load each page of 100 results — a poor UX for admins reviewing large volumes of submissions.

## Approach

Replace page-based pagination with **infinite scroll**: the page loads an initial batch of rows, and automatically fetches more when the user scrolls to the bottom. During loading, **3 animated skeleton rows** appear at the bottom of the table as a loading indicator.

I used the existing `vue-infinite-scroll` library (already a project dependency, used in `submissions/List.vue`) to detect when the user reaches the bottom of the page. The library's directive fires a `fetch-more-data` event, which triggers an API call for the next page of results and appends them to the Vuex store.

### 1. `runsStore.ts` — New `appendRuns` Vuex Mutation

The existing flow to load runs was `clear() → addRun() × N`, which **wipes the entire store** each time. Infinite scroll needs to **append** without clearing.

- Added `appendRuns(state, runs)` mutation that iterates over incoming runs
- If a run already exists (matched by `guid` via the index), it updates in-place
- If new, it appends to the array and registers in the index
- Uses `Vue.set()` to ensure reactivity

### 2. `submissions.ts` — New `onAppendRuns()` Helper

Added a counterpart to the existing `onRefreshRuns()`:

- Commits `setTotalRuns` to update the total count
- Commits `appendRuns` to add new runs without clearing existing ones

### 3. `Runs.vue` — Scroll Directive, Filters Separation, Skeleton Rows

**Infinite scroll directive:**
- Added `v-infinite-scroll` on the root `<div>` that emits `fetch-more-data`
- `infinite-scroll-disabled="isScrollDisabled"` prevents duplicate fetches
- `infinite-scroll-distance="10"` triggers 10px before reaching the bottom

**New props:**
- `loading` (boolean) — whether a fetch is in progress
- `endOfResults` (boolean) — whether all data has been loaded
- `showFilters` (boolean | null) — independently controls filter visibility

**New computed properties:**
- `isScrollDisabled` — returns `true` when `loading || endOfResults`
- `shouldShowFilters` — defaults to `showPager` when `showFilters` is null (backward compatible)

**Separated pager from filters:** Previously, filters were nested inside `v-if="showPager"`. Now the pager buttons use `v-if="showPager"` and filters use `v-if="shouldShowFilters"` independently, so we can hide the pager but keep filters visible.

**Skeleton loading:** 3 animated `<tr>` elements with the existing `loading-line` CSS class appear at the bottom of the table when `loading` is `true`.

### 4. `global_runs.ts` — Event Handler and State Management

**New reactive state:**
- `loading: false` — passed as prop, controls skeleton display and scroll blocking
- `endOfResults: false` — passed as prop, permanently disables scroll when all data loaded
- `currentOffset: 0` — tracks the current page number (0-indexed)

**Component configuration:**
- `showPager: false` — hides prev/next buttons
- `showFilters: true` — keeps filter dropdowns visible
- `rowCount = 20` — rows per page

**`fetch-more-data` event handler:**
1. Guards against duplicate calls (`loading || endOfResults`)
2. Sets `loading = true`
3. Fetches next page: `api.Run.list({ offset: currentOffset + 1, rowcount: 20 })`
4. On success: appends via `onAppendRuns()`, increments `currentOffset`
5. On empty response: sets `endOfResults = true`
6. On error: sets `endOfResults = true`, shows error
7. Finally: sets `loading = false`

**`filter-changed` handler:** Resets `currentOffset = 0` and `endOfResults = false`, then calls `refreshRuns()`.

**Auto-refresh (5-min interval):** Resets `currentOffset` and `endOfResults` before refreshing to ensure infinite scroll works again after data reload.

### 5. `Runs.test.ts` — Test Mock

Added `jest.mock('vue-infinite-scroll', () => ({}))` to mock the directive in tests. All 18 existing tests pass unchanged.

Fixes #8026


frontend/www/js/omegaup/arena/global_runs.ts
```diff
@@ -9,13 +9,16 @@ import * as time from '../time';
...
   onRefreshRuns,
+  onAppendRuns,
   showSubmission,
   SubmissionRequest,
   updateRunFallback,
 } from './submissions';
 import { types } from '../api_types';
 import { getOptionsFromLocation, getProblemAndRunDetails } from './location';
 
+const rowCount = 20;
```
pabo99: Can this be adjusted to 100 submissions?

thunderbird7413: yes, It can be. I have changed it to 100.

thunderbird7413:
> Can you check why the endpoint is being called twice when the component loads? Ideally, it should only fetch the first batch of runs:
> Screencast.From.2026-03-19.18-15-15.webm

I have fixed the double API call. 
The root cause was that `vue-infinite-scroll` fires immediately on mount when the content doesn't fill the viewport (empty table). This was happening because initially loading was set as false, which fires the `vue-infinite-scroll` which calls the api endpoint and other happens because of the `refreshRuns()` which also calls the endpoint again. Due to which the endpoint getting called twice. 

How I solved it->
1. `loading: true` (was false) — `isScrollDisabled` now blocks the scroll directive on mount
2. `finally(() => { vueInstance.$data.loading = false; })` on `refreshRuns()` — unlocks scroll after the initial data loads

Now the flow is: mount → `loading=true` blocks scroll → `refreshRuns()` fetches page 0 →` .finally` sets `loading=false` → scroll is now enabled → user scrolls → page 1 loads. Only one API call on load.

thunderbird7413:
I have done both the changes and pushed the code. You can review the changes and suggest me for any further improvements

frontend/www/js/omegaup/arena/global_runs.ts
pabo99: It’s still initialized to 20.

thunderbird7413:
Ohh my bad, I forgot to push that. 
Just did it and pushed.

thunderbird7413:
Thanks @pabo99 ❤️

Issue #9571, opened 2026-03-18
[BUG] Course assignment clarifications is not paginated


mazen-fahim:
## Checklist

- [X] I have searched existing issues and confirmed this is not a duplicate (Required)

## Current Behavior

The current implementation of a course clarification page doesn't have pagination.

https://github.com/user-attachments/assets/c9474ac6-10a3-485f-870a-49bbc28b0272

## Expected Behavior

When I visit the clarification page of a particular course assignment, I get the entire table.
There should be pagination, as it's best practice to never hit the database with no `LIMIT` clause to avoid performance issues on large tables.

Route: `/course/{course-alias}/assignment/{assignment-alias}#clarifications`

mazarona:
could you assign this to me if you don't mind 🌷 @pabo99

pabo99:
Assigned!

PR #9595, opened 2026-03-20
Fix: Course assignment clarifications page pagination

mazen-fahim:
# Description
- Modify DAO `getProblemsetClarifications` function to always apply pagination, similar to other DAO functions found in the codebase.
- Modify parameter name from `$offset` to `$page` for clarity and consistency
- Modify parameter name from `$rowcount` to `$pageSize` for clarity and consistency
- Add optional parameters `assignmentAlias` & `clarificationsPage` & `clarificationsPageSize` to helper methods used by the controller method `getCourseDetailsForTypeScript` which is the corresponding controller method for the problematic not-paginated route.
- The reason these newly added parameters are optional so that to not break other methods using these helper functions already. (I have noticed the pagination might be broken in multiple places, I would open separate issues/PRs for each one)
- The payload now returns a new key `clarificationsPagerItems` with all pager items used by the `omegaup-common-paginator` component for pagination of the clarifications table.
- The reason why I'm very explicit about prefixing the naming of these newly added parameters and the newly added payload key by `clarifications` and not just `page`, `pageSize`, `pagerItems` like other endpoints is that, this endpoint doesn't return a single `clarifications` array but it also return an `assignments` array. So I have to be explicit about to which array the `pagerItems` belong to.  
- Fortunately the `omegaup-arena-clarification-list` already has a `pagerItems` prop implmented, all I had to do is pass it the newly added `clarificationsPagerItems`  from the payload.
- I have also noticed that the  `omegaup-arena-clarification-list` component supports two props `page` and `page-size` to show a heading to the table.
- So I have  added two new keys to the returned payload `clarificationsPage` and `clarificationsPageSize`. I pass the value of these keys to the  `omegaup-arena-clarification-list` component to show a heading with pagination info of the table similar to this: 

Image

Fixes: #9571 

---
# Screen recording showing the fix

https://github.com/user-attachments/assets/5363e168-c85e-4794-bc76-a90d6ff49a12

---
# Checklist:

### DIFF
--- a/frontend/server/src/Controllers/README.md
+++ b/frontend/server/src/Controllers/README.md
@@ -277,7 +277,6 @@
   - [`/api/user/problemsSolved/`](#apiuserproblemssolved)
   - [`/api/user/profile/`](#apiuserprofile)
   - [`/api/user/profileStatistics/`](#apiuserprofilestatistics)
-  - [`/api/user/recordCookieConsent/`](#apiuserrecordcookieconsent)
   - [`/api/user/removeExperiment/`](#apiuserremoveexperiment)
   - [`/api/user/removeGroup/`](#apiuserremovegroup)
   - [`/api/user/removeRole/`](#apiuserremoverole)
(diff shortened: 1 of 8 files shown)


mazarona:
Also, What do you think about opening another issue where I modify the New Clarification button to be at the far right end instead of it being awkwardly on top of filters? It will be a small UI enhancement. let me know what you think. 

I have also spotted another pagination error regarding the clarifications but in a different route, I will open a separate issue for it so that I don't clutter this PR @pabo99 🌷

mazarona:
So I have traced the failing tests and fixed all type annotations. I tried the test locally and they running fine. Sorry I forgot to add them before . I have also made an additional modification which was failing creating a clarification in one of the tests. The reason why it was failing is that I have confused making the new added parameter by making them nullable instead of optional  😅 Now they are optional as I intended in the beginning to not break any existing calls to these modified methods. Thank you for you feedback and let me know if there are any additional changes you’d like me to make 🌷 @pabo99

PR #9607, opened 2026-03-21
Add incremental processing to aggregate_feedback using watermark

Prince-Prajapati07:
# Description

This PR introduces a minimal incremental processing mechanism in `aggregate_feedback` to avoid full recomputation on every run.

Currently, `aggregate_feedback` recomputes feedback for all problems with suggestions on each execution, even if no new data has been added. As the `QualityNominations` table grows, this leads to unnecessary work and increased load.

This change introduces a simple watermark based on `qualitynomination_id`:

- On the first run (or when no watermark is present), the existing full recomputation behavior is preserved.
- On subsequent runs, only problems with new `QualityNominations` (i.e., `qualitynomination_id > last_processed_id`) are selected.
- For each selected problem, the aggregation logic remains unchanged and still processes the full history of nominations.

The watermark is updated only after a successful execution to ensure correctness and idempotency.

This approach reduces unnecessary processing while keeping behavior consistent and safe.

---

# Comments

- The implementation is intentionally minimal and localized to problem selection logic.
- No changes were made to aggregation logic, ranking semantics, or SQL queries inside `aggregate_problem_feedback`.
- Full recomputation fallback is preserved to ensure safe deployment.
- Watermark updates are performed only after successful execution to avoid partial state issues.

## Testing

- Added a unit test to verify that `aggregate_feedback` executes correctly in incremental mode.
- Additionally, the behavior was validated locally by running the cron twice and confirming that only problems with new nominations are processed in the second run.

---

# Checklist:

- [x] If the change is large (> 200 lines), this PR was split appropriately.

### DIFF
--- a/stuff/cron/aggregate_feedback.py
+++ b/stuff/cron/aggregate_feedback.py
@@ -419,7 +419,7 @@ def aggregate_feedback(dbconn: lib.db.Connection) -> None:
     failed_problems = 0
 
     with dbconn.cursor() as cur:
-        if last_processed_qualitynomination_id is None:
+        if not last_processed_qualitynomination_id:
             cur.execute("""SELECT DISTINCT qn.`problem_id`
                            FROM `QualityNominations` as qn
                            WHERE qn.`nomination` = 'suggestion'
--- a/stuff/cron/aggregate_feedback.py
+++ b/stuff/cron/aggregate_feedback.py
@@ -433,9 +433,13 @@ def aggregate_feedback(dbconn: lib.db.Connection) -> None:
                            WHERE qn.`nomination` = 'suggestion'
                              AND qn.`qualitynomination_id` > %s
                              AND qn.`qualitynomination_id` <= %s;""",
-                        (max(QUALITYNOMINATION_QUESTION_CHANGE_ID,
-                             last_processed_qualitynomination_id),
-                         max_qualitynomination_id))
+                        (
+                            max(
+                                QUALITYNOMINATION_QUESTION_CHANGE_ID,
+                                last_processed_qualitynomination_id,
+                            ),
+                            max_qualitynomination_id,
+                        ))
         for (problem_id,) in cur.fetchall():
             attempted_problems += 1
             try:
--- a/stuff/cron/aggregate_feedback.py
+++ b/stuff/cron/aggregate_feedback.py
@@ -419,21 +419,18 @@ def aggregate_feedback(dbconn: lib.db.Connection) -> None:
     failed_problems = 0
 
     with dbconn.cursor() as cur:
-        if not last_processed_qualitynomination_id:
+        if (last_processed_qualitynomination_id is None or
+                last_processed_qualitynomination_id == 0):
             cur.execute("""SELECT DISTINCT qn.`problem_id`
                            FROM `QualityNominations` as qn
-                           WHERE qn.`nomination` = 'suggestion'
-                             AND qn.`qualitynomination_id` > %s;""",
-                        (QUALITYNOMINATION_QUESTION_CHANGE_ID,))
+                           WHERE qn.`nomination` = 'suggestion';""")
         else:
             cur.execute("""SELECT DISTINCT qn.`problem_id`
                            FROM `QualityNominations` as qn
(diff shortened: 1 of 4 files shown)


Prince-Prajapati07:
Hi @pabo99 sir! just wanted to follow up on this one. This change adds incremental processing to avoid full recomputation in `aggregate_feedback`. I tried to keep it minimal and not touch the existing aggregation logic.
Wanted to check if this approach (watermark-based filtering) makes sense from your side, or if you would prefer a different direction here.

stuff/cron/aggregate_feedback.py
```diff
@@ -104,6 +107,65 @@ class RankCutoff(NamedTuple):
...
+    try:
+        with dbconn.cursor() as cur:
+            cur.execute(
+                """SELECT
+                           `last_processed_qualitynomination_id`
+                       FROM
+                           `Cron_AggregateFeedback_State`
+                       WHERE
+                           `singleton_id` = 1;""")
+            rows = list(cur.fetchall())
```
pabo99: Please consider using `fetchone()` here, since the query returns only a single row

stuff/cron/aggregate_feedback.py
```diff
@@ -104,6 +107,65 @@ class RankCutoff(NamedTuple):
...
+                       COALESCE(
+                           MAX(qn.`qualitynomination_id`),
+                           %s
+                       )
+                   FROM
+                       `QualityNominations` AS qn
+                   WHERE
+                       qn.`nomination` = 'suggestion';""",
+            (QUALITYNOMINATION_QUESTION_CHANGE_ID,))
+        rows = list(cur.fetchall())
```
pabo99: Same case here

pabo99 (commented):
Great work! The incremental approach looks solid.

As Copilot suggests, before merging, we need a database migration for the `Cron_AggregateFeedback_State` table referenced in the code.

Could you add a migration file and regenerate the schema files? This could also be done in a separate PR if you prefer.

Prince-Prajapati07:
Hi @pabo99 sir!
I have been trying to fix the incremental logic here so that it matches the full recompute behavior, but I’m still running into issues with the PHP tests (counts/difficulty/tags don’t match).
Right now i am using:
- lower bound: last_processed_qualitynomination_id  
- upper bound: max_qualitynomination_id  
I also tried including QUALITYNOMINATION_QUESTION_CHANGE_ID as a lower bound, but that seemed to include extra rows and made things worse.

Before I keep iterating on this, I wanted to check a couple of things:
- Should incremental processing give exactly the same result as a full recompute for the same range?
- Is QUALITYNOMINATION_QUESTION_CHANGE_ID supposed to be part of the filtering logic here, or not?
- Am I missing any condition that affects which rows should be included?
any direction would really help.
Thanks!

pabo99:
I've been testing your PR and the incremental logic looks good. The test failures are due to the watermark persisting between test runs, causing incremental mode when tests expect full recomputation.

To answer your questions:
- **Should incremental processing give exactly the same result as a full recompute for the same range?** Yes, incremental processing should give identical results because while it only selects problems with new nominations, `aggregate_problem_feedback()` always processes the full history for each selected problem starting from `QUALITYNOMINATION_QUESTION_CHANGE_ID`.
   
- **Is `QUALITYNOMINATION_QUESTION_CHANGE_ID` supposed to be part of the filtering logic here, or not?** Yes, `QUALITYNOMINATION_QUESTION_CHANGE_ID` must always be in the filtering logic since it marks where the question format changed. Your current approach is correct.
   
-  **Am I missing any condition that affects which rows should be included?** The issue is test isolation - the `Cron_AggregateFeedback_State` table needs to be cleared before each test run to ensure full recomputation mode.

Suggested fix: In frontend/tests/Utils.php, modify `runAggregateFeedback()` to truncate the state table before calling the Python script:

```php
public static function runAggregateFeedback(): void {
    self::commit();
    
    // Clear watermark to ensure full recomputation in tests
    $conn = \OmegaUp\MySQLConnection::getInstance();
    $conn->Execute('TRUNCATE TABLE `Cron_AggregateFeedback_State`;');
    $conn->CompleteTrans();
    $conn->StartTrans();
    
    // ... rest of the method
}

Issue #9614, opened 2026-03-23
[BUG] Fix deadlock retry logic in isInsideSubmissionGap


Wafaamohamed2:
## Problem

The current implementation of `isInsideSubmissionGap` has a bug in the deadlock retry logic:

1. Non-deadlock errors are being retried unnecessarily before being thrown
2. The exponential backoff uses complex jitter calculation that can be simplified
3. `$retryCount` is incremented before checking if the error is a deadlock

## Current Behavior
```php
} catch (\mysqli_sql_exception $e) {
    $retryCount++;  // incremented even for non-deadlock errors
    $isDeadlock = in_array($errorCode, [1205, 1213]);
    if (!$isDeadlock || $retryCount >= $maxRetries) {
        throw $e;
    }
```

## Expected Behavior
```php
} catch (\mysqli_sql_exception $e) {
    if (!$isDeadlock) {
        throw $e;  // throw immediately for non-deadlock errors
    }
    $retryCount++;
    if ($retryCount >= $maxRetries) {
        throw $e;
    }
```

## Fix

- Throw immediately for non-deadlock errors
- Increment `$retryCount` only after confirming it's a deadlock
- Simplify exponential backoff by removing unnecessary jitter

PR #9615, opened 2026-03-23
Fix deadlock retry logic in isInsideSubmissionGap

Wafaamohamed2:
# Description

Fix deadlock retry logic in `isInsideSubmissionGap` .

Previously, non-deadlock errors were being retried unnecessarily before 
being thrown, and `$retryCount` was incremented even for non-deadlock errors.

Changes:
- Throw immediately if the error is not a deadlock (error codes 1205, 1213)
- Increment `$retryCount` only after confirming it's a deadlock
- Simplify exponential backoff by removing unnecessary jitter

Fixes: #9614

# Comments

Please pay attention to the retry logic change in the catch block 
of `isInsideSubmissionGap`. The main fix is ensuring non-deadlock 
errors are thrown immediately without incrementing the retry counter.

# Checklist:

✅ The code follows the coding guidelines
✅ The tests were executed and all of them passed


frontend/server/src/DAO/Submissions.php
```diff
@@ -178,25 +179,19 @@ final public static function isInsideSubmissionGap(
...
-                $retryCount++;
-
                 // Check if this is a deadlock or lock timeout error
                 $errorCode = $e->getCode();
                 $isDeadlock = in_array($errorCode, [1205, 1213]);
 
-                if (!$isDeadlock || $retryCount >= $maxRetries) {
-                    throw $e;
+                if (!$isDeadlock) {
+                    throw $e;// throw immediately for non-deadlock errors
```
pabo99: Nit:
```suggestion
                    throw $e; // throw immediately for non-deadlock errors
```

PR #9627, opened 2026-03-24
fix: ensure exactly one Discord notification per issue regardless of label count

Ankitsinghsisodya:
## Description

Follow-up to #9624. As reported by @pabo99, the workflow sent 3 messages when an issue was opened with both qualifying labels.

**Root cause:** GitHub fires both an `opened` event and a `labeled` event for each label applied at issue creation. An issue opened with 2 qualifying labels triggered 3 workflow runs.

After #9627 (removing `opened`), this was reduced to 2 messages — but still not exactly 1, since `labeled` fires once per qualifying label.

**Fix — two-trigger strategy with a 60-second age check:**

- **`opened` event:** Scans all labels on the issue and sends **exactly one** notification for the first qualifying label found. Handles issues created with multiple qualifying labels in a single shot.
- **`labeled` event:** Only fires if the issue is **older than 60 seconds**. GitHub fires `labeled` events even for labels applied at creation — the age check suppresses those (already handled by `opened`), while still notifying when a label is added to an existing issue.

## Scenarios

| Scenario | Result |
|---|---|
| Issue opened with GSoC + good first issue | `opened` sends 1, both `labeled` events skipped (< 60s) → **1 notification** |
| Issue opened with only GSoC | `opened` sends 1, `labeled` skipped (< 60s) → **1 notification** |
| Issue opened without labels, GSoC added later | `opened` skipped (no labels), `labeled` sends 1 (> 60s) → **1 notification** |
| Label added to existing issue | `labeled` sends 1 → **1 notification** |

## Test plan

- [ ] Open an issue with both `GSoC` and `good first issue` labels — should send exactly **1** Discord message
- [ ] Open an issue with only `GSoC` — should send exactly **1** message
- [ ] Add a qualifying label to an existing issue — should send exactly **1** message
- [ ] Confirm no errors in the Actions run log

Fixes duplicate notifications reported in #9624 (comment).

### DIFF
--- a/.github/workflows/discord-issue-notification.yml
+++ b/.github/workflows/discord-issue-notification.yml
@@ -2,7 +2,7 @@ name: Discord Issue Notification
 
 on:
   issues:
-    types: [labeled]
+    types: [opened, labeled]
 
 jobs:
   notify-discord:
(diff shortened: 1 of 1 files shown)


Ankitsinghsisodya:
@pabo99 I have done the fix.

pabo99 (approved):

Issue #9549, opened 2026-03-14
[BUG] Browser Back button does not update displayed tab on edit of a group via My Groups page


abhinavsingh-hub:
## Current Behavior

On the `/groups/<alias>/edit/` page, navigating between tabs updates the URL hash and displays the corresponding content correctly.

However, when pressing the browser Back button, the URL hash changes to the previous tab but the displayed tab content does not update.

## Expected Behavior

Pressing the browser Back button should update both:
- the URL hash
- the visible tab content

so that the previously viewed tab is displayed.

## Steps to Reproduce

1. Navigate to `/groups/<alias>/edit`
2. Navigate to another tab if available
3. Press the browser Back button

Result:
- URL hash changes
- tab content remains stuck on the last viewed tab

## Root Cause

The component updates the tab state through click handlers but the page entry `.ts` file does not listen for `hashchange` events. As a result, browser Back/Forward navigation updates the URL but does not update the Vue state.

A similar issue was previously fixed in:

- Issue: #9445 
- PR: #9471 

## Environment

Browser: Chrome (reproducible across browsers)  
Page: `/groups/<alias>/edit/`

## Screen Recording

https://github.com/user-attachments/assets/ea36b986-76e2-46bf-a3da-22fbb7e5c37e

abhinavsingh-hub:
Hi @pabo99!

While exploring omegaUp further, I noticed a similar issue to #9445 and #9392 on the My Groups page, where the browser Back button updates the URL hash but does not update the displayed tab.

I have opened an issue describing the bug and would be happy to work on fixing it following the same approach used in #9396.

I'd love if you’d like to assign the issue to me. Thanks!

abhinavsingh-hub:
Hi @pabo99!

While investigating this further, I noticed a similar issue on the Edit My Team Group page as well. When navigating between tabs, the URL hash updates correctly, but using the browser Back button does not update the displayed tab, leaving the UI stuck on the last viewed tab.

I’ve attached a screen recording demonstrating the behavior.

If this is something worth addressing as part of this fix, I’d be happy to include it in this PR. Otherwise, I can open a separate issue and work on it in another PR — whichever you’d prefer.

https://github.com/user-attachments/assets/2141b865-0d8a-44f4-b4f9-daacdc800294

PR #9638, opened 2026-03-25
Fix: Back button tab sync issue in group and teamsgroup edit pages

abhinavsingh-hub:
## Summary
This PR fixes an issue where the browser back/forward navigation did not correctly update the active tab in:
Fixes: https://github.com/omegaup/omegaup/pull/9549

- `group/edit`
- `teamsgroup/edit`

## Problem
When navigating between tabs (Edit / Members / Teams / Upload) and using the browser back button:
- The URL hash (`#edit`, `#teams`, etc.) changed
- But the UI tab (`selectedTab`) did not update accordingly

This caused inconsistency between the URL and the displayed content.

## Root Cause
The Vue component state (`selectedTab`) was not properly synchronized with:
- The `tab` prop from parent
- The browser `window.location.hash`

## Solution
- Synced `tab` (prop) → `selectedTab` using Vue `@Watch`
- Synced `selectedTab` → parent via `$emit('update:tab')`
- Ensured correct initialization of tab state on mount
- Ensured hash changes properly reflect in UI state

## Result
- Browser back/forward navigation works correctly
- URL hash and UI tab remain in sync
- Consistent behavior across both `group` and `teamsgroup` modules

## Consistency
This follows the same pattern used in previous fixes (e.g., #9445), and applies it consistently across similar pages.

---

## Checklist

- [x] Fix applied to `group/edit`
- [x] Fix applied to `teamsgroup/edit`
- [x] URL hash and UI tab stay in sync
- [x] Browser back/forward navigation works correctly
- [x] No regression in tab switching behavior
- [x] Code follows existing patterns in codebase
- [x] Manually tested locally

---

### Manual Testing
- Navigated between tabs (Edit / Teams / Upload)
- Used browser back and forward buttons
- Verified correct tab is shown based on URL

---

## Before / After 
Before:
- URL changed but tab UI did not update

https://github.com/user-attachments/assets/d1cc7f17-a4d3-474d-bfe3-2a752179ee9b

After:
- URL and tab UI remain fully synchronized

https://github.com/user-attachments/assets/beb90231-cfa0-42b7-8e9d-d7d5fd1c2a26

---

## 🙌 Notes
Happy to split into separate PRs if preferred, but since this is the same issue across similar modules, keeping it unified for consistency.


abhinavsingh-hub:
Hey @pabo99 @heduenas  👋

I’ve fixed the back button tab synchronization issue in both `group/edit` and `teamsgroup/edit`, following the same approach as previous fixes like #9445.

Would appreciate your review!  
Also happy to split this into two separate PRs (group and teamsgroup) if you’d prefer smaller, more focused changes.

Thanks!

frontend/www/js/omegaup/components/teamsgroup/Edit.vue
```diff
@@ -162,13 +162,23 @@ export default class TeamsGroupEdit extends Vue {
...
   }
 
+  @Watch('selectedTab')
+  onSelectedTabChanged(newValue: AvailableTabs): void {
+    this.$emit('update:tab', newValue);
+  }
+
+  mounted(): void {
+    this.onTabChanged(this.tab);
+  }
```
pabo99: This hook is redundant. It only calls `onTabChanged`, but `selectedTab` is already initialized with `this.tab`. The `@Watch('tab')` decorator only triggers on changes after mounting, not on initial value, so the mounted hook was attempting to handle initial validation.

However, the initial validation was never needed in practice. The parent always passes valid tab values.

Could you remove it?

frontend/www/js/omegaup/teamsgroup/edit.ts
```diff
@@ -385,4 +389,17 @@ OmegaUp.on('ready', () => {
...
+
+    if (!Object.values(AvailableTabs).includes(hash as AvailableTabs)) {
+      teamsGroupEdit.tab = AvailableTabs.Teams;
+      return;
+    }
+
+    teamsGroupEdit.tab = hash as AvailableTabs;
+  };
+
+  window.addEventListener('hashchange', onHashChange);
```
pabo99, line 403: The listener is created but never removes it, causing a memory leak when the vue instance is destroyed

Could you add a `beforeDestroy` hook to remove the `hashchange` listener.

frontend/www/js/omegaup/group/edit.ts
```diff
@@ -325,4 +329,17 @@ OmegaUp.on('ready', () => {
...
+
+    if (!Object.values(AvailableTabs).includes(hash as AvailableTabs)) {
+      groupEdit.tab = AvailableTabs.Members;
+      return;
+    }
+
+    groupEdit.tab = hash as AvailableTabs;
+  };
+
+  window.addEventListener('hashchange', onHashChange);
```
pabo99, line 343: Same case here, could you add the `beforeDestroy` hook

abhinavsingh-hub:
You are absolutely right @pabo99, 

I was trying to implement mounted logic because of a few failures I was facing previously, but I have removed it now, and added beforeDestroy in both the files as well

Please review and let me know if anything else ought to be changed, also let me know if you come across any other such files or some other issues, would love to contribute as much as I can!!

Regards

Issue #9639, opened 2026-03-25
[BUG] Buttons in modernized contest list view are incorrect


pabo99:
## Checklist

- [x] I have searched existing issues and confirmed this is not a duplicate (Required)

## Current Behavior

The buttons displayed in the modernized contest list view do not match the expected ones.

- _**Current contests**_: The buttons do not correspond to the intended actions.

- _**Future contests**_: The buttons are also incorrect.

Past contests appear to be displayed correctly.

## Expected Behavior

The buttons should match the expected actions for each contest type (Current, Future, Past). Users should be able to interact with contests as intended, according to their state.

## If necessary, add the following information (otherwise delete this section):

Steps to reproduce:

1. Go to `/arenav2/` on the local environment.
2. Observe the buttons under "Current" contests.
3. Scroll down to "Future" contests and check the buttons.
4. Compare with the expected behavior (refer to the current `/arena/` view ).

    * Screenshots:

<img width="1849" height="605" alt="Image" src="https://github.com/user-attachments/assets/bbb2fb15-f392-4a62-9c06-b545fd68164a" />

<img width="1849" height="669" alt="Image" src="https://github.com/user-attachments/assets/7f7be1c7-b7eb-46b0-be4a-6349dc2875cf" />

anshjalan:
Hello @pabo99 ! Can you please assign me this issue?

neerajbachani:
hey @pabo99 , I would like to work on this issue

pabo99:
Hello @neerajbachani, @anshjalan. Thank you for your interest in collaborating on this issue. I’d like to remind you that you can claim the issue directly using /assign, as long as you meet the necessary requirements (not being a first-time contributor and having fewer than two issues currently assigned).

Throughout the day, I’ll continue creating issues similar to this one, so stay tuned to the dev-training channel on Discord.

PR #9642, opened 2026-03-25
Fix arenav2 contest actions for current and future tabs

rahmanfaizur:
Hide Virtual/Practice actions for current and future contests to match `/arena/`. Added regression tests for both list variants.

Fixes: #9639

## Screenshots

**/arena/**

**/arenav2/**

# Description

- Hide Virtual/Practice for ongoing and upcoming contests  
- Match behavior with legacy `/arena/`  
- Add regression tests for both list variants


frontend/www/js/omegaup/components/arena/ContestListv2.vue
```diff
@@ -209,6 +209,9 @@
               <template #contest-dropdown>
                 <div></div>
               </template>
+              <template #contest-enroll-status>
+                <div></div>
+              </template>
```
pabo99: Are the changes in this file for any specific reason? Since this PR is focused on fixing the new view, this component shouldn’t be modified. If it’s to address some inconsistency, it would be better to handle it in a separate issue/PR

rahmanfaizur:
Thanks @pabo99 ! I’ve removed the unnecessary #contest-enroll-status overrides from ContestListv2.vue and updated the corresponding tests. This PR is now focused only on the Virtual/Practice dropdown fix.

pabo99 (approved):
Great!

PR #9672, opened 2026-03-27
added the reusable styles and modified the styles

### LINKED ISSUE #9440: [FEATURE] Create reusable modern styles for modal forms
## Description
As a developer, I want reusable modern styles for modal forms so that UI components across the platform look consistent and easier to maintain.

## Context
The modal used to **report problems** currently looks outdated and uses a very basic layout.

Image

Instead of updating only that modal, the goal is to create **general reusable styles** that can be applied to similar forms across the platform.

These styles could modernize the layout and improve the visual consistency of modal forms.

AgrapujyaLashkari:
# Description

Changed the form styles and also made reusable class for form

Fixes: #9440 
Screenshot 2026-03-27 at 5 07 51 PM

### DIFF
--- a/frontend/www/js/omegaup/components/OverlayPopup.vue
+++ b/frontend/www/js/omegaup/components/OverlayPopup.vue
@@ -22,11 +22,13 @@ div[data-overlay-popup] {
   height: fit-content;
   max-height: 90%;
   margin: auto;
-  border: 2px solid var(--overlay-popup-border-color);
-  padding: 1em;
+  border: 1px solid var(--overlay-popup-border-color);
+  padding: var(--overlay-popup-padding);
   position: absolute;
   overflow-y: auto;
   overflow-x: hidden;
+  border-radius: var(--overlay-popup-radius);
+  box-shadow: var(--overlay-popup-shadow);
   top: 0;
   bottom: 0;
   left: 0;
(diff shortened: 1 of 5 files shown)


frontend/www/js/omegaup/components/OverlayPopup.vue
```diff
@@ -22,11 +22,13 @@ div[data-overlay-popup] {
   height: fit-content;
   max-height: 90%;
   margin: auto;
-  border: 2px solid var(--overlay-popup-border-color);
-  padding: 1em;
+  border: 1px solid var(--overlay-popup-border-color);
```
pabo99, line 25: With these new styles, I can see that there’s too much space on the sides of the form. Could you adjust the width to 50% only on desktop? On devices with resolutions below the 992px breakpoint, it’s fine for it to be 80% wide

frontend/www/js/omegaup/components/OverlayPopup.vue
```diff
@@ -38,12 +40,21 @@ div[data-overlay-popup] {
     position: -webkit-sticky;
     top: 0;
     z-index: 100;
+    width: 32px;
+    height: 32px;
+    border-radius: 999px;
```
pabo99, line 45: I like that the button has a more rounded style, but it somewhat breaks the look of the other buttons in these modals. 

Could you revert this for now while we work on a comprehensive solution for all of them? That can be done in a separate issue

pabo99 (commented):
Wow! This PR brings the modals up to date in a big way! Thank you so much for the change!

As I mentioned in one of the suggestions, we need to standardize the style, positioning, and size of the buttons and input elements within the modals. This can be done in a separate issue. The modals I’ve been able to identify are the following:

- _**PromotionPopup.vue**_: Shown when you solve a problem and return to the problem view later.

Image

- **_DemotionPopup.vue_**: The one where you modified the button style; it appears in the problem details view.

Image

- **_ReviewerPopup.vue_**: Also shown in the problem details view, but only for reviewers/admins.

Image

- _**RunDetailsPopup.vue**_: Shown when making a submission and clicking the icon in the **_Actions_** column.

Image

- **_RunSubmitPopup.vue_**: Shown when pressing the _**New submission**_ button.

Image

- **_NewClarificationPopup.vue_**: Shown within courses/contests in the last tab when entering as a participant or admin and pressing _**New clarification**_ button

Image

Issue #9636, opened 2026-03-24
[BUG] Teams Group Edit: Number of contestants does not update immediately after save


xx-Tasneem-Ahmed-xx:
## Checklist

- [x] I have searched existing issues and confirmed this is not a duplicate (Required)

## Current Behavior

When creating a teams group then updating the number of contestants, the update is lost when navigating to another tab (like: teams or upload teams) and isn't reflected back unless after a refresh. This is clearly a frontend issue as the backend persistence is not a problem here.

Possible cause: a static initial payload is used and the local state isn't being updated.

## Expected Behavior

The update shouldn't be lost when navigating to another tab.

## Steps to reproduce:

    1. Navigate to https://omegaup.com/teamsgroup (ensure you are logged in)
    2. Click on `create new teams group`
    3. Fill any data inside the fields then click create `create new teams group`
    4. Click on Edit tab and change the number of contestants
    5. navigate to another tab (teams or upload teams) then back to edit tab

_Relevant details about the environment in which you encountered this bug:_

    * Browser name: chrome
    * Operating System and version (desktop or mobile): any mobile or desktop
    * Link to the site where this issue occurs: https://omegaup.com/teamsgroup
  
Video: 
https://github.com/user-attachments/assets/7922bfa9-3416-40df-9edb-b131466a1f2d

**Note**: I believe this is a `good first issue` therefore I'm intentionally leaving it for newcomers. Good luck!

iAmMazapan:
Hi, I'd like to work on this. I already have my dev environment set up and can submit a PR soon. Could you assign this to me please @pabo99 ?

SunilkumarT56:
hi , Could you please assign this to me @pabo99

PR #9675, opened 2026-03-27
Fix teams group edit contestant count staying stale across tabs after save

dhruvv16-hash:
Description

Implemented a frontend state fix for the teams group edit flow so the saved contestant limit stays in sync across the edit, teams, and upload tabs without requiring a page refresh.

Added reactive `teamGroup` page state in `frontend/www/js/omegaup/teamsgroup/edit.ts`.
Updated rendered props to use reactive state instead of the immutable initial `payload.teamGroup` object.
Updated the save success path to persist the new `name`, `description`, and `numberOfContestants` into local Vue state as soon as `api.TeamsGroup.update()` resolves.
Added a targeted regression test in `frontend/www/js/omegaup/components/teamsgroup/Upload.test.ts` to verify `read-csv` receives the updated contestant limit after the prop changes.
Added a targeted regression test in `frontend/www/js/omegaup/components/teamsgroup/Edit.test.ts` to verify the updated contestant limit persists when navigating from Edit to Upload and back.
Kept the change UI-only and limited to the existing Vue page flow.
UI change: after saving a teams group, the updated contestant limit is immediately reflected when navigating across tabs.

Fixes: #9636

Comments

Please verify that changing `Number of contestants` in the Edit tab is preserved when navigating to Teams or Upload and back.
Ensure the upload/self-generated identities flow keeps using the updated value after the save succeeds.
This is a UI-only change; backend and controller logic is untouched.

Checklist:


pabo99:
Can you create an issue reporting this behavior? That way I’ll be able to assign it to you, and the **_Check PR Linked Issue ..._** workflow won’t raise any errors

dhruvv16-hash:
Created the linked issue: #9678.

I also updated this PR to reflect the actual affected flow:

- removed the mock PNG artifacts from the branch
- updated the PR title/description to clarify that the stale value affects the upload/self-generated-identities flow, not the edit input itself
- added a targeted regression test in `Upload.test.ts` to verify `read-csv` uses the updated contestant limit after the teams group is saved

The frontend change itself is still the same: after `api.TeamsGroup.update()` resolves, the page keeps the updated teams-group state so dependent tabs stop using the initial payload.

dhruvv16-hash:
I checked the failing workflow and the blocker is the linked-issue assignment requirement from `.github/workflows/check-pr-issue.yml`.

Issue #9678 is linked correctly, but it is still unassigned. I tried assigning it to myself via the GitHub API, but the upstream repo returned `403 Must have admin rights to Repository`, so I don't have enough permission to do that directly.

@pabo99 could you please assign #9678 to me? Once that's done, the `Check PR Linked Issue & Assignee` workflow should stop failing on the next rerun/synchronize event.

dhruvv16-hash:
Attaching a screencast of the change here.

This is a local reproduction of the exact frontend flow that was fixed: after saving the teams group with a new `Number of contestants`, the upload/self-generated-identities tab immediately uses the updated limit instead of the initial payload value.

teams-group-upload-flow-screencast

Separately, I verified the failing `Check PR Linked Issue & Assignee` workflow is blocked because issue #9678 still needs to be assigned to me. I tried to assign it through the GitHub API, but the upstream repo returned `403 Must have admin rights to Repository`, so that part still needs a maintainer-side assignment.

pabo99 (commented):
Now I understand what the issue is referring to — this is why it’s important to include the description before opening the PR.

Before approving the change, could you please remove the GIF file you uploaded? It’s not a screencast of the page and doesn’t add much value. In the ideal case, we could create a Storybook for this component, but we can leave that for a future change.

dhruvv16-hash:
Removed the GIF asset from the branch and pushed the update. I will leave the Storybook idea for a separate follow-up, as suggested.

dhruvv16-hash:
Quick update: I retargeted this PR to the original issue #9636 and added an explicit regression test for the tab-navigation case described there.

The remaining workflow blocker is now issue assignment on #9636. I don''t have permission to assign myself in the upstream repo, so @pabo99 could you please assign #9636 to me when you get a chance? That should satisfy the `Check PR Linked Issue & Assignee` workflow on the next rerun/synchronize event.

Issue #9678, opened 2026-03-28
[BUG] Teams group upload uses stale contestant limit after saving edit changes


dhruvv16-hash:
Description

In the teams group edit flow, after updating `Number of contestants` and saving the group, the upload tab still uses the previous contestant limit until the page is refreshed.

This does not affect the edit input itself. The problem is that the upload flow keeps reading the initial teams-group payload, so self-generated identities continue using the old limit for that page session.

Steps to Reproduce

1. Go to `/teamsgroup/<alias>/edit/#edit`.
2. Change `Number of contestants` to a new value and save the teams group.
3. Navigate to the upload tab.
4. Enable the option that self-generates identities for missing team members.
5. Upload a CSV where at least one team has fewer members than the newly saved limit.

Expected Behavior

The upload flow should immediately use the newly saved contestant limit when generating missing identities.

Actual Behavior

The upload flow still uses the previous contestant limit until the page is refreshed.

Additional Context

This appears to be a frontend state synchronization issue in the teams-group edit page. The save request succeeds, but dependent tabs keep using the initial payload instead of the updated teams-group state.

Issue #9620, opened 2026-03-23
[FEATURE] UX UPDATES


ciara-gospel:
## Checklist

- [ ] I have searched existing issues and confirmed this is not a duplicate (Required)

 As a developer, I want update the search logic so that we have good user experience and design

## Context

Many times i have tried to search something i always had to click on the button beside for the search to start but is not supposed to be so i didn't always need to click on the button before the search should work. Even when i click enter on my machine it is supposed to work. 

Now when searching we have to use debounce so that when i enter a letter it is supposed to map and give the answer if the answer is not there i continue type the word letter by letter until i find my result and when i type it the search is filtering to find the result if there is no result it has to show not found and we can also give time for search or looping like say when i enter for example "Mov" after 2 second it's shows the result of all the words with mov

SunilkumarT56:
can you assign this to me ?

pabo99:
Hello @ciara-gospel! Are you interested in working on this issue, or would you prefer someone else to take it? 

Thanks in advance

pabo99:
Since @SunilkumarT56 was the first to request it, the issue has been assigned to them

PR #9682, opened 2026-03-28
[FEATURE] Improve problem mine search UX with debounce and Enter handling

dhruvv16-hash:
Description

Implemented live search UX improvements on `/problem/mine/` so searching no longer requires clicking the button each time.

- Replaced the link-based search action in `Mine.vue` with a form-driven search flow.
- Added debounced search emission while typing (300ms delay).
- Added immediate search on Enter key press.
- Added loading feedback while a search request is in flight.
- Added a `No results found` empty state when a query returns no problems.
- Updated `problem/mine.ts` to refresh results dynamically, reset to page 1 on a new search, and keep the `query` URL parameter in sync without a full page reload.
- Added regression tests covering debounce, Enter-triggered search, and empty search results.

Addresses: #9620
Reference issue for assignment workflow: #9678

Comments

- This is a frontend-only change; backend endpoints were reused as-is.
- Debounce is used to avoid excessive API calls while typing.
- Enter still triggers the search immediately for keyboard users.
- URL query state now stays synchronized through `history.replaceState`.

Checklist:


frontend/www/js/omegaup/components/problem/Mine.vue
```diff
@@ -18,26 +18,23 @@
     <div class="card">
       <h5 class="card-header">{{ T.myproblemsListMyProblems }}</h5>
       <div class="card-body px-2 px-sm-4">
-        <div class="row align-items-center mb-3">
+        <form class="row align-items-center mb-3" @submit.prevent="emitSearch">
```
pabo99, line 21: Since search is now triggered while typing, I think we can remove the explicit Search button. This also avoids a duplicate search when pressing Enter: the input currently has `@keyup.enter.prevent="emitSearch"` and the wrapping form also has `@submit.prevent="emitSearch"`, so Enter can trigger both paths.

Could we keep the debounced search + Enter behavior, but remove the form/button?

frontend/www/js/omegaup/components/problem/Mine.vue
```diff
@@ -345,5 +349,33 @@ export default class ProblemMine extends Vue {
...
+    }, 300);
+  }
+
+  emitSearch(): void {
+    if (this.debounceTimer) {
+      clearTimeout(this.debounceTimer);
+      this.debounceTimer = null;
+    }
+    this.$emit('search-problems', this.currentQuery.trim());
+  }
```
pabo99, line 368: Now that search no longer reloads the page, selected problems can survive across searches. A user could select some problems, type a new query that hides them, and still apply a bulk action to the old invisible selection.

Could we clear the bulk selection whenever a new search is emitted?

frontend/www/js/omegaup/problem/mine.ts
```diff
@@ -111,21 +118,36 @@ OmegaUp.on('ready', () => {
   });
 
   function showProblems(showAllProblems: boolean, pageNumber?: number): void {
+    problemsMine.isSearching = true;
```
pabo99, line 121: With live search, multiple requests can be in flight at the same time. If an older request finishes after a newer one, it can overwrite the latest results. It can also set `isSearching = false` while the latest request is still pending.

Could you guard updates with a request id so only the latest request updates the UI?

pabo99 (commented):
Found while testing: searching by problem ID doesn't return the expected results. This seems unrelated to this change, since the PR reuses the existing `Problem.myList` / `Problem.adminList` query behavior.

Could you please report it in a separate issue?

PR #9687, opened 2026-03-28
Feat/9679 arena v2 contest single request and debouncing

### LINKED ISSUE #9679: [BUG] Optimize ArenaV2 contest search: avoid multiple requests and add debounce
## Checklist

## Current Behavior

Currently, every time a search is performed in the ArenaV2 contests list, three separate requests are sent to the server: one for current, one for past, and one for future contests.

This results in unnecessary network overhead and may impact performance, especially as the number of users or requests increases.

## Expected Behavior

Ideally, only a single request should be made, and the query should handle all three periods (current, past, and future) on the backend.

Additionally, the search input does not implement any debounce mechanism, causing multiple rapid requests to be triggered as the user types. Introducing a debounce would help reduce the number of requests and improve overall performance.

https://github.com/user-attachments/assets/cf28f1b7-2687-4d59-aca8-95b7c9a8771e

rahmanfaizur:
## Description

Optimize ArenaV2 contest search by reducing multiple requests and adding debounce.

- Add `listAllTabs` API to fetch current, past, and future contests in a single request  
- Update Arena v2 to use one request instead of three on initial/search load  
- Add ~300ms debounce to search input to prevent excessive requests  

Fixes: #9679

## Notes

- Improves performance by reducing network calls  
- Video attached showing single request + debounced search behavior  

https://github.com/user-attachments/assets/1b5d3abb-7c7e-4eb3-b1f6-7360931b5138

### DIFF
--- a/frontend/server/src/Controllers/Contest.php
+++ b/frontend/server/src/Controllers/Contest.php
@@ -207,6 +207,127 @@ class Contest extends \OmegaUp\Controllers\Controller {
     }
 
     /**
+     * Returns paginated contest lists for current, past, and future in one API call.
+     *
+     * @return array{current: array{number_of_results: int, results: list<ContestListItem>}, future: array{number_of_results: int, results: list<ContestListItem>}, past: array{number_of_results: int, results: list<ContestListItem>}}
+     *
+     * @omegaup-request-param 'private'|'public'|'registration'|null $admission_mode
+     * @omegaup-request-param 'all'|'recommended'|'signedup'|null $filter
+     * @omegaup-request-param int $page
+     * @omegaup-request-param int $page_size
+     * @omegaup-request-param int|null $participating
+     * @omegaup-request-param string $query
+     * @omegaup-request-param int|null $recommended
+     * @omegaup-request-param null|string $sort_order
+     */
+    public static function apiListAllTabs(\OmegaUp\Request $r): array {
+        try {
+            $r->ensureIdentity();
+        } catch (\OmegaUp\Exceptions\UnauthorizedException $e) {
+            // Do nothing.
+            /** @var null $r->identity */
+        }
+
+        $recommended = $r->ensureOptionalInt(
+            'recommended'
+        ) ?? \OmegaUp\DAO\Enum\RecommendedStatus::ALL;
+        $participating = $r->ensureOptionalInt(
+            'participating'
+        ) ?? \OmegaUp\DAO\Enum\ParticipatingStatus::NO;
+        $filter = $r->ensureOptionalEnum(
+            'filter',
+            \OmegaUp\DAO\Enum\ContestFilterStatus::NAME_FOR_STATUS
+        );
+        $activeFilter = \OmegaUp\DAO\Enum\ContestFilterStatus::convertToInt(
+            fieldName: 'filter',
+            field: $filter,
+            defaultValue: \OmegaUp\DAO\Enum\ContestFilterStatus::ALL
+        );
+        if ($activeFilter === \OmegaUp\DAO\Enum\ContestFilterStatus::ONLY_RECOMMENDED) {
+            $recommended = \OmegaUp\DAO\Enum\RecommendedStatus::RECOMMENDED;
+        } elseif ($activeFilter === \OmegaUp\DAO\Enum\ContestFilterStatus::SIGNED_UP) {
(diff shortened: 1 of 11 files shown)


rahmanfaizur:
Hi @pabo99,
Could you please take a look at this PR when you have some time?
Thanks!

frontend/server/src/Controllers/Contest.php
```diff
@@ -206,6 +206,127 @@ public static function apiList(\OmegaUp\Request $r): array {
...
+            'current' => ['number_of_results' => 0, 'results' => []],
+            'past' => ['number_of_results' => 0, 'results' => []],
+            'future' => ['number_of_results' => 0, 'results' => []],
+        ];
+        $tabStatuses = [
+            'current' => \OmegaUp\DAO\Enum\ContestTabStatus::CURRENT,
+            'past' => \OmegaUp\DAO\Enum\ContestTabStatus::PAST,
+            'future' => \OmegaUp\DAO\Enum\ContestTabStatus::FUTURE,
+        ];
+        foreach ($tabStatuses as $tabKey => $activeContests) {
```
pabo99, line 306: I understand that the goal of the new endpoint is to reuse existing functions as much as possible, which is why you're iterating over each tab to gather the required data. However, this approach doesn’t provide much benefit, since it still results in multiple roundtrips (N+1 queries) to fetch the data.

A better approach might be to introduce a variant of `getContestList` that does not require `$activeContests`. Most DAO functions already support retrieving all contests by default (`\OmegaUp\DAO\Enum\ActiveStatus::ALL`), except for `getRecentPublicContests`, which might just be an omission. It would be good to add a test around this to confirm in which cases this function is used and whether it returns the expected results.

This new variant could then be used exclusively by `apiListAllTabs`, keeping the logic simpler and avoiding the need to iterate over each tab manually.

frontend/server/src/Controllers/Contest.php
```diff
@@ -206,6 +206,127 @@ public static function apiList(\OmegaUp\Request $r): array {
...
+        $order = $r->ensureOptionalEnum(
+            'sort_order',
+            \OmegaUp\DAO\Enum\ContestOrderStatus::NAME_FOR_STATUS,
+            required: false
+        );
+        $orderBy = \OmegaUp\DAO\Enum\ContestOrderStatus::convertToInt(
+            fieldName: 'order',
+            field: $order,
+            defaultValue: \OmegaUp\DAO\Enum\ContestOrderStatus::NONE
+        );
```
pabo99, line 293: It seems this shares a lot of similarities with what `apiList` already does. Would it be possible to extract the shared logic into a helper function and reuse it in both endpoints?

The only difference is that `apiListAllTabs` does not require `$activeContests`, but that could be handled within that endpoint to avoid fetching it unnecessarily.

frontend/www/js/omegaup/arena/contest_listv2.ts
```diff
@@ -178,10 +178,16 @@ OmegaUp.on('ready', () => {
             } else {
               window.history.pushState({}, '', urlObj);
             }
-            await contestStore.dispatch('fetchContestList', {
-              requestParams: params,
-              name: params.tab_name,
-            });
+            if (params.page === 1) {
```
pabo99, line 181: Could we leave this and all UI-related changes for a separate PR? This would help speed up the review process.

For now, I’d prefer that the changes only affect the new Contest List view. However, if we can guarantee that the data remains correct for the current UI as well, we can keep both in sync.

frontend/www/js/omegaup/components/arena/ContestListv2.vue
```diff
@@ -415,6 +416,18 @@
 </template>
 
 <script lang="ts">
+const debounce = (fn: (...args: any[]) => void, waitTime: number) => {
```
pabo99, line 419: This and everything related to the debounce logic doesn’t seem directly related to the single request change. Would it be possible to move it to a separate PR? This would also help speed up the review process. If needed, you can also create a separate issue so I can assign it to you

Just something to consider: we already have some debounce utilities in place. Could you try to align this change with the existing approach? Mainly because `any` is being used in several places here, and we generally discourage that in favor of more precise typing.

pabo99 (commented):
I left you some suggestions. Let me know if anything isn’t clear

rahmanfaizur:
Hi @pabo99, changes are split as discussed: backend #9773 and UI #9775. Please review when you can.

PR #9694, opened 2026-03-30
[docs] Update contest endpoint controller docstrings and align generated API docs

dhruvv16-hash:
Description
Updates contest endpoint controller docstrings to be the source of truth for generated API docs.
Clarifies contest timing semantics (window_length vs global start_time / finish_time).
Improves endpoint descriptions for list/details/create/update and related contest admin endpoints.

Scope
Docs-only change in frontend/server/src/Controllers/Contest.php.
No runtime logic changes.

Notes
Could not run local API doc generation in this environment because php is not installed in the runner.

### DIFF
--- a/frontend/server/src/Controllers/Contest.php
+++ b/frontend/server/src/Controllers/Contest.php
@@ -83,7 +83,15 @@ class Contest extends \OmegaUp\Controllers\Controller {
     const CONTEST_LIST_PAGE_SIZE = 10;
 
     /**
-     * Returns a list of contests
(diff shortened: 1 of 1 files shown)


PrasannaMishra001:
hi @dhruvv16-hash, the lint failure is because the API docs `README` gets auto-generated from the PHP docstrings. Since you updated the docstrings, the `README` needs to be regenerated. Run this outside the Docker container:
```
./stuff/lint.sh fix -- frontend/server/src/Controllers/README.md
```
Then commit the updated `README` alongside your changes. For the check-issue-link failure, try linking issue along with this.

Issue #9708, opened 2026-03-31
[Security] Open redirect via protocol-relative URL in login redirect parameter


neerajbachani:
## Checklist
- [X] I have searched existing issues and confirmed this is not a duplicate (Required)

## Current Behavior
The `redirect` query parameter in the login flow is validated by checking `pathname.indexOf('/') === 0`. This check accepts protocol-relative URLs such as `//evil.example.com` because that string starts with `/`.

Browsers treat `//host` as a scheme-relative URL and navigate off-site using the current protocol (https), effectively redirecting the user to an external domain after login.

Affected files:
- `frontend/www/js/omegaup/login/signin.ts`
- `frontend/www/js/omegaup/login/logout.ts`

## Expected Behavior
The redirect parameter should only accept paths that are genuinely local to omegaUp (e.g. `/contest/list/`). 
Protocol-relative URLs like `//evil.example.com` should be rejected and the user should be redirected to the 
default page instead.

## If necessary, add the following information:

1. Go to `https://omegaup.com/login/#signin?redirect=%2F%2Fevil.example.com`
2. Log in with valid credentials
3. Observe the URL bar during and after login

## Root Cause
```typescript
// signin.ts - CURRENT (potentially vulnerable)
if (pathname && pathname.indexOf('/') === 0 && !isLogoutPath) {
  window.location.href = pathname + '?fromLogin';
}
// pathname.indexOf('/') === 0 also matches '//evil.example.com'
// because that string starts with '/'
```

**Proposed fix - additionally reject protocol-relative URLs:**
```typescript
if (
  pathname &&
  pathname.startsWith('/') &&
  !pathname.startsWith('//') &&  // reject //evil.example.com
  !isLogoutPath
) {
  window.location.href = pathname + '?fromLogin';
}
```

The same fix should be applied to `logout.ts`.

## Security Impact
This class of vulnerability (CWE-601: URL Redirection to Untrusted Site) can be used for phishing attacks. An attacker crafts a login link with a malicious redirect parameter and shares it with omegaUp users. After logging in legitimately on omegaUp, the user is silently redirected to an attacker-controlled site that can mimic omegaUp to steal credentials or display malicious content.

**Relevant details:**
* Browser name: Chrome 145
* Operating System: Windows 11
* Link: `https://omegaup.com/login/#signin?redirect=%2F%2Fevil.example.com`
* Screenshot:

[...]

PR #9710, opened 2026-03-31
fix: #9708 reject protocol-relative URLs in login/logout redirect 

neerajbachani:
# Description

Reject protocol-relative URLs in login/logout redirect parameter.

Both `signin.ts` and `logout.ts` validated the `redirect` query parameter using `pathname.indexOf('/') === 0`, which incorrectly 
accepts protocol-relative URLs like `//evil.example.com`. Browsers treat `//host` as a scheme-relative URL and navigate off-site using the current protocol (https), enabling a potential open redirect attack (CWE-601) usable for phishing.

Added `!pathname.startsWith('//')` check to both files. Only genuine local paths (starting with `/` but not `//`) are now accepted as redirect targets.

**Files changed:**
- `frontend/www/js/omegaup/login/signin.ts`
- `frontend/www/js/omegaup/login/logout.ts`

**Before:**
```typescript
if (pathname && pathname.indexOf('/') === 0 && !isLogoutPath)
```

**After:**
```typescript
if (
  pathname &&
  pathname.startsWith('/') &&
  !pathname.startsWith('//') &&
  !isLogoutPath
)
```
**Screenshot after fix, perfectly redirect to home url after signin and third party domain gets completely neglected**
image

Fixes: #9708

# Comments

The vulnerability can be reproduced by visiting: `/login/#signin?redirect=%2F%2Fevil.example.com`

After logging in, the old code would follow the `//evil.example.com` redirect. The fix ensures only paths like `/contest/list/` are 
accepted. Both `signin.ts` and `logout.ts` had the same pattern and are fixed together since they share identical redirect logic.

Please pay attention to the condition inversion in `logout.ts` the logic is negated compared to `signin.ts` so the fix reads 
slightly differently but is semantically equivalent.

# Checklist:
- [X] The code follows the coding guidelines of omegaUp.
- [X] If the change is large (> 200 lines), this PR was split into various Pull Requests.


pabo99 (approved):
Great!

PR #9713, opened 2026-03-31
fix: add real-time date validation to course form to prevent end date before start date

### LINKED ISSUE #9705: [Bug] course/Form.vue allows end date before start date with no frontend validation
## Checklist

- [X] I have searched existing issues and confirmed this is not a duplicate (Required)

## Description
The course creation and edit form (`course/Form.vue`) allows submitting with an end date that is before the start date. No inline error is shown to the user — the form submits and only fails with a server-side error response, leaving the user confused with no clear feedback.

In contrast, `contest/Form.vue` already has a `validateDates()` method that shows an inline error before submission. The same validation is missing from the course form.

## Steps to Reproduce
1. Go to https://omegaup.com/course/new/ (or local /course/new/)
2. Enter any course name and required fields
3. Set Start date to today (e.g. 31-03-2026)
4. Set End date to a past date (e.g. 04-03-2026) a date BEFORE the start date
5. Click submit

## Expected Behavior
An inline error message should appear immediately, preventing submission identical to the behavior in contest/Form.vue.

## Actual Behavior
The form submits with no frontend error. The user receives only a generic server-side error response with no clear 
explanation of what went wrong.

## Screenshots

**Course form - no validation (BUG):**

Image

**Contest form - correct inline validation (REFERENCE):**

Image

## Root Cause
`contest/Form.vue` has a `validateDates()` method called inside `onSubmit()` that checks if `finishTime <= startTime` and shows an inline error. The equivalent check does not exist in `course/Form.vue`.

**contest/Form.vue (existing, working validation):**
```typescript
private validateDates(): boolean {
  if (
    this.startTime !== null &&
    this.finishTime !== null &&
    this.finishTime <= this.startTime
  ) {
    ui.error(T.contestNewEndDateMustBeAfterStartDate);
    return false;
  }
  return true;
}

// Called at top of onSubmit():
if (!this.validateDates()) return;
```

**course/Form.vue (missing — no equivalent check exists)**

## Proposed Fix
Add the same `validateDates()` check to `course/Form.vue`'s 
`onSubmit()` method:
```typescript
// Add inside onSubmit() at the top, before any API call:
if (
  this.startTime !== null &&
  this.finishTime !== null &&
  this.finishTime <= this.startTime
) {
  ui.error(T.contestNewEndDateMustBeAfterStartDate);
  return;
}
```

The existing translation string `T.contestNewEndDateMustBeAfterStartDate` can be reused since the message is identical for both forms.

## Impact
Every teacher or admin creating a course with incorrect dates gets a confusing server-side error instead of a clear, 
immediate inline message. This inconsistency between the course and contest forms creates a poor user experience.

## Related
- `contest/Form.vue` already handles this correctly and can serve as the reference implementation

neerajbachani:
# Description

The course creation and edit form (`course/Form.vue`) allowed submitting with an end date before the start date. No frontend 
validation existed the form would only fail with a server-side error, leaving users confused with no clear feedback.

This fix mirrors the validation pattern already used in 
`contest/Form.vue`:

- Added `validateDates()` method that sets an `invalidDate` flag
- Added `@input="validateDates"` on both start and end date pickers so the error appears in real time as the user changes 
  dates not just on submit
- Added inline error message directly under the end date field using Bootstrap's `invalid-feedback` pattern
- Added `onSubmit()` guard that calls `validateDates()` and returns early if dates are invalid
- Added `!this.unlimitedDuration` guard courses with no end date are completely unaffected
- Reset `invalidDate = false` in the reset method to clear stale error state

**Before:** End date error appears on submit 

**After:** Inline error "The end date is previous to the course start date" appears immediately under the end date field when 
an invalid date is selected.
image

Fixes: #9705 

# Comments

The fix uses `courseAssignmentEndDateBeforeCourseStartDate` translation key which already exists in all language files 
(en, es, pt, pseudo). No new translation strings required.

The `invalidDate` flag approach was chosen over calling `ui.error()` directly to match the Bootstrap inline validation 
pattern used throughout the codebase (consistent with how `invalidParameterName` works in the same component).

# Checklist:

- [X] If the change is large (> 200 lines), this PR was split into various Pull Requests. It's preferred to create one PR for changes in controllers + unit tests in PHPUnit,  and then another Pull Request for UI + tests in Jest, Cypress or both.

### DIFF
--- a/frontend/www/js/omegaup/components/course/Form.vue
+++ b/frontend/www/js/omegaup/components/course/Form.vue
@@ -86,7 +86,6 @@
                 name="start-date"
                 :disabled="readOnly"
                 :min="update ? null : new Date()"
-                @input="validateDates"
               ></omegaup-datepicker
             ></label>
           </div>
@@ -116,14 +115,14 @@
                 :disabled="readOnly"
                 name="end-date"
                 :enabled="!unlimitedDuration"
-                :is-invalid="invalidParameterName === 'finish_time' || invalidDate"
-                @input="validateDates"
+                :is-invalid="
+                  invalidParameterName === 'finish_time' || invalidDate
+                "
               ></omegaup-datepicker
             ></label>
             <div
               v-if="invalidDate"
-              class="invalid-feedback"
-              style="display: block"
+              class="invalid-feedback invalid-date-feedback"
             >
               {{ T.courseAssignmentEndDateBeforeCourseStartDate }}
             </div>
@@ -406,7 +405,6 @@ export default class CourseDetails extends Vue {
   selectedLanguages = this.course.languages;
   levelOptions = levelOptions;
   MAX_LENGTH = MAX_LENGTH;
-  invalidDate = false;
 
   // Computed properties to track if required fields are complete
   get isNameComplete(): boolean {
@@ -563,25 +561,17 @@ export default class CourseDetails extends Vue {
     this.needsBasicInformation = this.course.needs_basic_information;
     this.requestsUserInformation = this.course.requests_user_information;
     this.unlimitedDuration = this.course.finish_time === null;
-    this.invalidDate = false;
   }
 
-  validateDates(): boolean {
-    if (
-      !this.unlimitedDuration &&
-      this.startTime !== null &&
-      this.finishTime !== null &&
-      this.finishTime <= this.startTime
-    ) {
-      this.invalidDate = true;
-      return false;
-    }
-    this.invalidDate = false;
-    return true;
+  get invalidDate(): boolean {
+    if (this.unlimitedDuration) return false;
+    if (!this.startTime || !this.finishTime) return false;
+    if (this.finishTime <= this.startTime) return true;
+    return false;
   }
 
   onSubmit(): void {
-    if (!this.validateDates()) {
+    if (this.invalidDate) {
       return;
     }
     if (!this.selectedLanguages || this.selectedLanguages.length === 0) {
@@ -647,5 +637,9 @@ export default class CourseDetails extends Vue {
   color: var(--form-character-counter-color, #6c757d);
   font-size: 0.8rem;
   margin-top: 0.25rem;
+}
+
+.invalid-date-feedback {
+  display: block;
 }
 </style>
--- a/frontend/www/js/omegaup/components/course/Form.vue
+++ b/frontend/www/js/omegaup/components/course/Form.vue
@@ -86,6 +86,7 @@
                 name="start-date"
                 :disabled="readOnly"
                 :min="update ? null : new Date()"
+                @input="validateDates"
               ></omegaup-datepicker
             ></label>
           </div>
@@ -115,9 +116,17 @@
                 :disabled="readOnly"
                 name="end-date"
                 :enabled="!unlimitedDuration"
-                :is-invalid="invalidParameterName === 'finish_time'"
+                :is-invalid="invalidParameterName === 'finish_time' || invalidDate"
+                @input="validateDates"
               ></omegaup-datepicker
             ></label>
+            <div
+              v-if="invalidDate"
+              class="invalid-feedback"
+              style="display: block"
+            >
+              {{ T.courseAssignmentEndDateBeforeCourseStartDate }}
+            </div>
           </div>
         </div>
         <div class="row">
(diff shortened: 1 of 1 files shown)


frontend/www/js/omegaup/components/course/Form.vue
```diff
@@ -553,9 +563,27 @@ export default class CourseDetails extends Vue {
...
+      !this.unlimitedDuration &&
+      this.startTime !== null &&
+      this.finishTime !== null &&
+      this.finishTime <= this.startTime
+    ) {
+      this.invalidDate = true;
+      return false;
+    }
+    this.invalidDate = false;
+    return true;
```
pabo99: I understand that the goal is to replicate the validation from `Contest/Form`, but there’s a slightly different behavior here:

In contests, the `localErrors` object is used to mark the fields that contain errors.

Here, since it doesn’t exist, you tried to work around it with `invalidDate`, which will invariably be the opposite of what `validDates` returns. What do you think if we simplify this a bit and just create a computed property `invalidDate`?
```suggestion
  get invalidDate(): boolean {
    if (this.unlimitedDuration) return false; 
    if (!this.startTime || !this.finishTime) return true;
    if (this.finishTime <= this.startTime) return true;

    return false;
```

This way it’s clearer and we don’t mix responsibilities

frontend/www/js/omegaup/components/course/Form.vue
```diff
@@ -553,9 +563,27 @@ export default class CourseDetails extends Vue {
...
   }
 
   onSubmit(): void {
+    if (!this.validateDates()) {
```
pabo99: Here it would be:

```suggestion
    if (this.invalidDate) {
```

frontend/www/js/omegaup/components/course/Form.vue
```diff
@@ -397,6 +406,7 @@ export default class CourseDetails extends Vue {
   selectedLanguages = this.course.languages;
   levelOptions = levelOptions;
   MAX_LENGTH = MAX_LENGTH;
+  invalidDate = false;
```
pabo99: This would no longer be needed:
```suggestion
```

frontend/www/js/omegaup/components/course/Form.vue
```diff
@@ -553,9 +563,27 @@ export default class CourseDetails extends Vue {
     this.needsBasicInformation = this.course.needs_basic_information;
     this.requestsUserInformation = this.course.requests_user_information;
     this.unlimitedDuration = this.course.finish_time === null;
+    this.invalidDate = false;
```
pabo99: This wouldn’t be needed either:

```suggestion
```

frontend/www/js/omegaup/components/course/Form.vue
```diff
@@ -86,6 +86,7 @@
                 name="start-date"
                 :disabled="readOnly"
                 :min="update ? null : new Date()"
+                @input="validateDates"
```
pabo99: This isn’t needed either.
```suggestion
```

frontend/www/js/omegaup/components/course/Form.vue
```diff
@@ -115,9 +116,17 @@
                 :disabled="readOnly"
                 name="end-date"
                 :enabled="!unlimitedDuration"
-                :is-invalid="invalidParameterName === 'finish_time'"
+                :is-invalid="invalidParameterName === 'finish_time' || invalidDate"
+                @input="validateDates"
```
pabo99: And this isn’t needed either:
```suggestion
```

frontend/www/js/omegaup/components/course/Form.vue
```diff
@@ -115,9 +116,17 @@
...
               ></omegaup-datepicker
             ></label>
+            <div
+              v-if="invalidDate"
+              class="invalid-feedback"
+              style="display: block"
```
pabo99: Could you add this style to the class you created on line 125? That’s how it’s done in `Contest/Form`

PrasannaMishra001:
@neerajbachani, for the lint error you can run this outside the docker container (inside WSL if on Windows):
```
./stuff/lint.sh fix -- frontend/www/js/omegaup/components/course/Form.vue
```
But I'd suggest applying @pabo99 's review suggestions first (converting to a computed property and all) and then running lint after at the end. 
it is just a quick reminder, i hope you already know this, but I happened to notice this in many pr's so thought I’d mention it.

neerajbachani:
Hi @pabo99, I've applied all your suggestions, converted validateDates() to a computed property, removed the data property and reset calls, dropped the @input handlers, and moved the inline style to a CSS class.

One small intentional deviation: for the missing dates case I return false instead of true as suggested, so the error doesn't appear before the user has even picked an end date. Felt like better UX but happy to change it if you prefer consistency with the suggestion.

pabo99 (approved):
LGTM!

Issue #9737, opened 2026-04-05
[BUG] Emails::getByEmail() returns object instead of null when email not found


Wafaamohamed2:
## Checklist
- [x] I have searched existing issues and confirmed this is not a duplicate (Required)

## Current Behavior
`Emails::getByEmail()` always returns an `Emails` object even when no record
is found in the database. When `GetRow()` returns `null` (email not found),
the method passes `null` directly to the `Emails` constructor instead of
returning `null`:
```php
$row = \OmegaUp\MySQLConnection::getInstance()->GetRow($sql, $params);
return new \OmegaUp\DAO\VO\Emails($row); // always returns an object, never null
```

## Expected Behavior
`getByEmail()` should return `null` when no matching email exists, as declared
in its own `@return null|\OmegaUp\DAO\VO\Emails` docblock. Any caller checking
`is_null(Emails::getByEmail($email))` will incorrectly get `false` even when
the email does not exist, which can cause silent failures in registration and
login flows.

## Steps to Reproduce
1. Call `\OmegaUp\DAO\Emails::getByEmail('nonexistent@example.com')`
2. Check the return value with `is_null()`
3. Result is `false` — an `Emails` object is returned instead of `null`

## Notes
The fix is consistent with the pattern already used throughout the codebase in
`Tags::getByName()`, `TeamGroups::getByAlias()`, and `Teams::getByIdentityId()`:
```php
if (empty($row)) {
    return null;
}
return new \OmegaUp\DAO\VO\Emails($row);
```

console-log-life:
Hey, I'm interested to solve the issue so can you assign me?

PR #9740, opened 2026-04-06
Fix: Emails::getByEmail() returns null when no record found

console-log-life:
Fixed Emails::getByEmail() to return null instead of an object when no record is found in the database. Previously, the method was passing a null row directly to the Emails constructor, which always resulted in an instantiated object. This fix ensures consistency with other DAO methods like Tags::getByName() and prevents silent failures in registration and login flows.

Also updated the database collation in frontend/database/schema.sql to utf8mb4_unicode_ci for better compatibility.

Fixes: #9737

Comments
The logic was verified with a manual test script to ensure that:

It returns null when the email does not exist.

It returns a valid Emails object when the record exists.

Checklist:
[x] The code follows the [coding guidelines](https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/Coding-guidelines.md) of omegaUp.

[x] The tests were executed and all of them passed (Verified logic via manual mock testing as environment PHPUnit was unavailable).


pabo99:
Closed in favor of PR #9738

PR #9763, opened 2026-04-12
Fix starTime typo in arena Summary scoreboard cutoff

### LINKED ISSUE #9760: [BUG]  Fix typo starTime → startTime in arena Summary.vue (scoreboard cutoff row)
## Checklist

- [X] I have searched existing issues and confirmed this is not a duplicate (Required)

## Current Behavior

`Summary.vue` uses `starTime.getTime()` in the scoreboard cutoff row, but the prop is `startTime`. If that table row renders, Vue throws at runtime (`ReferenceError` / cannot read property of undefined), breaking the summary.

## Expected Behavior

The cutoff time should be computed from `startTime` (and existing `duration` / `scoreboard` logic), matching the prop defined in the script.

**File:** `frontend/www/js/omegaup/components/arena/Summary.vue` (template, scoreboard cutoff `<tr>`).

**Note:** With current parents (`Contest.vue` / `ContestPractice.vue` pass `show-ranking="false"`), this row may not appear in production; the bug is still incorrect and unsafe for any caller that enables ranking + numeric scoreboard.

**Repro (test / dev):** Mount `Summary` with `showRanking: true`, numeric `scoreboard`, finite `duration`, non-null `finishTime`.

_Relevant details about the environment in which you encountered this bug:_
- Codebase

neerajbachani:
# Description

The contest summary template referenced `starTime.getTime()` while the component prop is `startTime`. If the scoreboard cutoff row were rendered (`showRanking` true, numeric `scoreboard`, finite duration), this would throw at runtime. Replace `starTime` with `startTime` so the cutoff timestamp uses the correct value.

Fixes: #9760 

# Comments

With current parents (`show-ranking="false"` on arena summary in contest/practice), this row may not appear in production; the fix is still required for correctness and any future caller that enables the row.

# Checklist:

- [X] If the change is large (> 200 lines), this PR was split into various Pull Requests. It's preferred to create one PR for changes in controllers + unit tests in PHPUnit,  and then another Pull Request for UI + tests in Jest, Cypress or both.

### DIFF
--- a/frontend/www/js/omegaup/components/arena/Summary.vue
+++ b/frontend/www/js/omegaup/components/arena/Summary.vue
@@ -35,7 +35,7 @@
         <td>
           {{
             time.formatTimestamp(
-              new Date(starTime.getTime() + (duration * scoreboard) / 100),
+              new Date(startTime.getTime() + (duration * scoreboard) / 100),
             )
           }}
         </td>


pabo99 (approved):
Oh, great catch! We need the linter to detect issues in Vue component templates.

Thanks for fixing it, @neerajbachani

neerajbachani:
> Oh, great catch! We need the linter to detect issues in Vue component templates.
> 
> Thanks for fixing it, @neerajbachani

yes @pabo99, while going through the codebase during my Vue 2 → Vue 3 migration exploration, I noticed a few similar cases where template variables are not strictly validated and found lot of typos like this

I think this happened because we are not enforcing rules like "vue/no-undef-properties" in ESLint, so such issues can slip through and only fail at runtime.

If we enable stricter Vue template linting, it might initially surface some existing issues, but it would improve overall reliability and also make the Vue 3 migration smoother and safer.

Would you like me to open an issue for improving the ESLint configuration and work on that first?

PR #9749, opened 2026-04-07
added notification backend controller changes

pratham15541:
# Description

- **Contest.php**
  - `apiAddProblem`, `apiRemoveProblem`:
    - Insert log row (if contest active)
    - Broadcast WebSocket event
  - Add `apiProblemChangeLogs` endpoint

Fixes: #9748 

# Checklist:

- [x] The code follows the [coding guidelines](https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/Coding-guidelines.md) of omegaUp.
- [x] The tests were executed and all of them passed.
- [x] If you are creating a feature, the new tests were added.
- [x] If the change is large (> 200 lines), this PR was split into various Pull Requests. It's preferred to create one PR for changes in controllers + unit tests in PHPUnit,  and then another Pull Request for UI + tests in Jest, Cypress or both.

PR #9769, opened 2026-04-14
added doa

### LINKED ISSUE #9748: Backend - Problem change notification
Part of: #9561 

- **DAO/VO**
  - Add VO, Base DAO, and DAO

- **Contest.php**
  - `apiAddProblem`, `apiRemoveProblem`:
    - Insert 1 log row (if contest active)
    - Broadcast WebSocket event
  - Add `apiProblemChangeLogs` endpoint

- **i18n**
  - Add strings for added/modified/removed + log UI

Progress:
## Progress

- [x] #9769 
- [x] #9771

pratham15541:
# Description

- **DAO/VO**
  - Add VO, Base DAO, and DAO
  
  
  
  
  
  Fixes: #9748


pabo99 (commented):
Great! Could you add a simple test for the `getByContestId` function?

pratham15541:
> Great! Could you add a simple test for the `getByContestId` function?

@pabo99 Added test file in pull #9771

frontend/server/src/DAO/Base/ContestProblemChangeLog.php
```diff
@@ -0,0 +1,78 @@
...
+            $Contest_Problem_Change_Log->change_type,
+            \OmegaUp\DAO\DAO::toMySQLTimestamp(
+                $Contest_Problem_Change_Log->timestamp
+            ),
+        ];
+        \OmegaUp\MySQLConnection::getInstance()->Execute($sql, $params);
+        $affectedRows = \OmegaUp\MySQLConnection::getInstance()->Affected_Rows();
+        if ($affectedRows == 0) {
+            throw new \OmegaUp\Exceptions\DuplicatedEntryInDatabaseException(
+                'recordAlreadyExists'
```
pabo99: I see an issue here: this translation string does not exist.

This makes me suspect that the change is not following the proper DAO generation process.

Could you run the command described here?
https://github.com/omegaup/omegaup/blob/e5b74557ce6713747accb33035ce06133471cc25/frontend/www/docs/Useful-Commands-for-Development.md?plain=1#L79-L92

That section explains what the command does and why each database schema change is split into two separate commits.

If you still have questions after that, feel free to ask, we’ll be happy to help

pabo99 (commented):
I see that you already added more robust tests in #9771, so here it should be enough to apply the suggested changes so we can merge this and continue with the remaining work.

frontend/server/src/DAO/Base/ContestProblemChangeLog.php
pratham15541: This change were in main pr. i will commit changes

pabo99: I completely lost track. What’s the correct order of PRs for review?

Also, this change doesn’t really make much sense because the following files should be auto-generated:

- `frontend/server/src/DAO/Base/ContestProblemChangeLog.php`
- `frontend/server/src/DAO/VO/ContestProblemChangeLog.php`

pratham15541: I will make changes as quickly as possible. And order was

	#9769  —  DAO
	#9749  —  Controller 
    #9771  —  Testing

pabo99 (approved):
Great!

PR #9771, opened 2026-04-14
added notification backend changes with test files

### LINKED ISSUE #9748: Backend - Problem change notification

pratham15541:
#Description
Fixes: #9748

Added test file


frontend/server/src/Controllers/Contest.php
```diff
@@ -3429,9 +3475,96 @@ public static function apiRemoveProblem(\OmegaUp\Request $r) {
...
+
+        $contest = \OmegaUp\DAO\Contests::getByAlias($contestAlias);
+        if (is_null($contest) || is_null($contest->contest_id)) {
+            throw new \OmegaUp\Exceptions\NotFoundException(
+                'contestNotFound'
+            );
+        }
+
+        $logs = \OmegaUp\DAO\ContestProblemChangeLog::getByContestId(
+            intval($contest->contest_id)
```
pabo99: On line 3542 you’re already ruling out that `$contest->contest_id` is null, so the only valid type left is `int`, and there’s no need to parse it anymore.
```suggestion
            $contest->contest_id
```

frontend/server/src/Controllers/Contest.php
```diff
@@ -3429,9 +3475,96 @@ public static function apiRemoveProblem(\OmegaUp\Request $r) {
...
         );
 
+        // Log the problem removal and notify participants via WebSocket
+        // if the contest is currently active.
+        $contest = $params['contest'];
+        $problem = $params['problem'];
+        $now = \OmegaUp\Time::get();
+        if (
+            $contest->start_time->time <= $now &&
+            $contest->finish_time->time >= $now
```
pabo99: Could you create a helper function called `isInProgress` or `isActive` that contains this logic? This way, you can call it from the places where you're currently using this code.

frontend/server/src/Controllers/Contest.php
```diff
@@ -3366,6 +3367,51 @@ public static function apiAddProblem(\OmegaUp\Request $r): array {
...
+                        'message' => '/contest/problem/update/',
+                        'type' => $changeType,
+                        'contest_alias' => $contest->alias,
+                        'problem_alias' => $problemAlias,
+                    ]),
+                    true,  // public broadcast to all participants
+                    null,
+                    -1,
+                    false
+                );
```
pabo99: You could consider using named arguments for all parameters in this call for better readability and consistency

Also, casting `$contest->problemset_id` to `int` is not necessary here. At this point it cannot be `null`, so its type is already well-defined, and the broadcaster would accept it as-is:

```suggestion
                \OmegaUp\Grader::getInstance()->broadcast(
                    contestAlias: $contest->alias,
                    problemsetId: $contest->problemset_id,
                    problemAlias: null,
                    message:json_encode([
                        'message' => '/contest/problem/update/',
                        'type' => $changeType,
                        'contest_alias' => $contest->alias,
                        'problem_alias' => $problemAlias,
                    ]),
                    public: true,  // public broadcast to all participants
                    username: null,
                    userId: -1,
                    userOnly: false
                );
```

frontend/server/src/Controllers/Contest.php
```diff
@@ -3366,6 +3367,51 @@ public static function apiAddProblem(\OmegaUp\Request $r): array {
...
+            \OmegaUp\DAO\ContestProblemChangeLog::create(
+                new \OmegaUp\DAO\VO\ContestProblemChangeLog([
+                    'contest_id' => $contest->contest_id,
+                    'problem_id' => $problem->problem_id,
+                    'user_id' => $r->identity->user_id,
+                    'change_type' => $changeType,
+                ])
+            );
+
+            // Broadcast real-time event via WebSocket — zero DB writes,
```
pabo99: Would it be possible to move all this logic to `frontend/server/src/Broadcaster.php`, similar to how clarifications are handled? This would keep it encapsulated and also give you a chance to do a small refactor.

pabo99 (commented):
Great! Just a few suggestions

pratham15541:
Hi @pabo99, I’ve made the suggested changes. Please let me know if there’s anything else that needs to be updated or improved.

frontend/server/src/Broadcaster.php
```diff
@@ -10,6 +10,52 @@ public function __construct() {
...
+
+        try {
+            \OmegaUp\DAO\ContestProblemChangeLog::create(
+                new \OmegaUp\DAO\VO\ContestProblemChangeLog([
+                    'contest_id' => $contest->contest_id,
+                    'problem_id' => $problem->problem_id,
+                    'user_id' => $userId,
+                    'change_type' => $changeType,
+                ])
+            );
```
pabo99: Oh, this actually makes sense to keep in the `Contest` controller. The rest is what matters for the broadcaster. Could you apply that change?

pabo99 (commented):
One last comment

frontend/server/src/Broadcaster.php
pratham15541: Hi @pabo99 , I have made the requested changes

pabo99 (approved):
Much better! Thank you, @pratham15541!

Issue #9768, opened 2026-04-13
[BUG]Ranking table user names break vertically on mobile view


anuragdubey137:
## Description

On the **User Ranking** page, user real names in the ranking table break vertically letter-by-letter on small screen sizes (mobile view). This makes the table difficult to read and creates excessively tall rows.

The issue occurs when viewing the page on narrow screen widths (e.g., iPhone SE ~375px).

## Steps to Reproduce

1. Go to the **User Ranking** page
2. Enable mobile view (for example, using Chrome DevTools)
3. Set screen width to around **375px** (iPhone SE)
4. Observe the **Coder name column**

## Expected Behavior

- User names should wrap normally across multiple lines
- Names should remain readable
- Table rows should maintain reasonable height
- Text should not break into vertical letter-by-letter layout

## Actual Behavior

- User real names are displayed vertically, letter-by-letter
- Table rows become excessively tall
- The ranking table becomes difficult to read on mobile devices

Example:

Instead of:
René Ramírez Venegas

It appears as:

R  
e  
n  
é  
R  
a  
m  
í  
r  
e  
z  
V  
e  
n  
e  
g  
a  
s  

## Screenshot

<img width="1025" height="795" alt="Image" src="https://github.com/user-attachments/assets/f07c9bc5-17ee-44d9-ba49-86927f2c162c" />

## Environment

- Device: Mobile view (emulated)
- Screen width: ~375px
- Device profile: iPhone SE
- Browser: Chrome (DevTools mobile emulation)

## Possible Cause

This may be caused by restrictive column width or CSS properties such as `word-break: break-all` or missing responsive wrapping rules.

## Impact

- Poor readability on mobile devices
- Broken layout on small screens
- Reduced usability of the ranking page on mobile

anuragdubey137:
this Issue is open for contribution!!

viveksahu92:
HI @pabo99 , i want to work on this issue could please assign this issue for me ?

console-log-life:
Hi @anuragdubey137 and @pabo99,

I noticed this bug and since I was already working on a similar UI fix, I've implemented a solution for this as well. I've adjusted the CSS (word-break and flex-wrap) to ensure names wrap naturally on mobile devices without breaking letter-by-letter.

Here is the PR for the same: [https://github.com/omegaup/omegaup/pull/9772](https://www.google.com/search?q=https://github.com/omegaup/omegaup/pull/9772)

Thank you! 😊

[...]

PR #9772, opened 2026-04-14
Fix: Prevent user names from breaking vertically on mobile ranking table

console-log-life:
This PR fixes issue #9768. On mobile view, the user real names were breaking letter-by-letter due to restrictive widths and word-break properties.

**Changes:**

Removed fixed width from .coder-username.

Added word-break: normal and overflow-wrap: break-word to .coder-name.

Added flex-wrap: wrap to .coder-cell for better responsiveness on very small screens.

**Fixes** #9768


eduqr:
Hey, it looks like this PR is modifying 748 files... could you check that first?

PR #9773, opened 2026-04-15
Refactor contest all-tabs backend flow.

rahmanfaizur:
Refactor contest list backend flow to reduce repeated queries and improve test coverage.

# Description

Backend + PHPUnit changes only:

- Extract shared request parsing used by `apiList` and `apiListAllTabs`.
- Rework `apiListAllTabs` to avoid per-tab repeated retrieval (single-fetch split flow).
- Add active-status support to `getRecentPublicContests`.
- Add regression tests for public admission mode tab segmentation and all-tabs correctness.

UI/debounce changes are intentionally excluded and will be handled in a separate PR.

Partially addresses #9679.

# Checklist:

- [x] The code follows omegaUp coding guidelines.
- [x] New tests were added for the new behavior.
- [x] Large change split into backend and UI PRs.


frontend/server/src/Controllers/Contest.php
```diff
@@ -178,32 +265,105 @@ public static function apiList(\OmegaUp\Request $r): array {
...
+            'public' => $public,
+            'query' => $query,
+            'recommended' => $recommended,
+        ];
+    }
+
+    /**
+     * @return array{current: array{number_of_results: int, results: list<ContestListItem>}, future: array{number_of_results: int, results: list<ContestListItem>}, past: array{number_of_results: int, results: list<ContestListItem>}}
+     */
+    private static function getContestListAllTabs(
```
pabo99, line 348: I see that we are still doing two database round trips: one to count and another to fetch the records. On top of that, there is also the extra cost of splitting the results in `PHP`.

It may be better to go back to the idea of one round trip per category, as you had before, explicitly requesting only the first 10 contests found for each one.

The good part of this PR is that the refactor is already done, and it helped us identify the better solution more clearly

frontend/server/src/Controllers/Contest.php
```diff
@@ -178,32 +265,105 @@ public static function apiList(\OmegaUp\Request $r): array {
...
+            'pageSize' => $pageSize,
+            'participating' => $participating,
+            'public' => $public,
+            'query' => $query,
+            'recommended' => $recommended,
+        ];
+    }
+
+    /**
+     * @return array{current: array{number_of_results: int, results: list<ContestListItem>}, future: array{number_of_results: int, results: list<ContestListItem>}, past: array{number_of_results: int, results: list<ContestListItem>}}
```
pabo99: This is repeated in many places:
```php
array{number_of_results: int, results: list<ContestListItem>}
```
Could you create a type for it and reuse it everywhere it is needed?

frontend/server/src/Controllers/Contest.php
```diff
@@ -89,15 +89,105 @@ class Contest extends \OmegaUp\Controllers\Controller {
...
+            'participating' => $participating,
+            'public' => $public,
+            'query' => $query,
+            'recommended' => $recommended,
+        ] = self::getContestListRequestParams(
+            $r,
+            \OmegaUp\DAO\Enum\ContestTabStatus::ALL
+        );
+
+        return self::getContestListAllTabs(
```
pabo99, line 221: I can already think of a couple of improvements for this endpoint, but they would involve moving around all the query-related functions, which could make this change much larger.

Still, I’m writing them down so they can be considered in the future:

- This endpoint does not need to account for pagination, since it only needs the first 10 records.
- The queries called from getContestList do not need to execute the total-count query here, since pagination is not needed for this endpoint

frontend/server/src/DAO/Contests.php
```diff
@@ -850,11 +850,12 @@ final public static function getRecentPublicContests(
         int $identity_id,
         int $page = 1,
         int $pageSize = 1000,
+        int $activeContests = \OmegaUp\DAO\Enum\ActiveStatus::ALL,
```
pabo99, line 853: Oh, now I understand why this function did not receive `$activeContests`: it is only responsible for showing recent public contests. It might be worth reverting the changes in this file

pabo99 (commented):
Good first refactoring attempt. I left a some suggestions to help align the implementation with the endpoint’s intended use case and improve its efficiency.

pabo99 (approved):
Great!

PR #9731, opened 2026-04-04
feat: add preferred programming language dropdown to user preferences

console-log-life:
Added a "Preferred programming language" dropdown in the User Preferences section of the profile edit page. This allows users to select their favorite programming language from a list of supported languages provided by the backend.

Changes made:

Modified User.php and api_types.ts to fetch the list of supported languages.

Updated .lang and .json translation files to include the label "Preferred programming language".

Updated PreferencesEdit.vue to include the new <select> dropdown in the UI.
Fixes: #9723
Comments
The dropdown successfully fetches and displays the list of languages from the backend. I have verified that the frontend compiles successfully using npm run dev.

Checklist:
[x] The code follows the [coding guidelines](https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/Coding-guidelines.md) of omegaUp.

[x] The tests were executed and all of them passed.

[x] If you are creating a feature, the new tests were added.

[x] If the change is large (> 200 lines), this PR was split into various Pull Requests.

PR #9795, opened 2026-04-17
fix: import golden-layout CSS from node_modules instead of vendoring

### LINKED ISSUE #9766: [FEATURE] Import golden-layout CSS from node_modules instead of vendoring
## Checklist

As a **developer**, I want **golden-layout CSS to be imported directly from `node_modules`** so that **we avoid maintaining a vendored copy that can drift out of sync with the installed package version**.

## Context

PR #9628 fixed a CSP violation by vendoring the golden-layout CSS files into `frontend/www/third_party/css/golden-layout/`. However, `golden-layout` is already a declared dependency in `package.json` (`"golden-layout": "1.5.9"`), so there is no need to keep a separate copy.

Importing directly from `node_modules` via the existing build pipeline (webpack/vite) would eliminate the duplication and ensure the CSS always matches the installed package version.

This was noted in the review of PR #9628.

The vendored files are at `frontend/www/third_party/css/golden-layout/`.

console-log-life:
## Problem
The golden-layout CSS files were vendored in `frontend/www/third_party/css/golden-layout/` 
but `golden-layout` is already a declared dependency in `package.json`.

## Changes
- Removed vendored CSS files from `third_party/css/golden-layout/`
- Updated `@import` in `Ephemeral.vue` to use `~golden-layout/src/css/goldenlayout-base.css`
- Updated theme paths to use `/node_modules/golden-layout/src/css/` 

## Testing
- Build passes successfully with `npm run build`

Fixes #9766


pabo99:
Hello @Apeksha-22! Thanks for contributing to omegaUp.

One favor, could you let me know what process you follow when creating a PR? I see that all of your changes include thousands of modified lines, which makes them very difficult to review properly. We might have something in the documentation that needs to be urgently improved to prevent cases like this from happening again.

console-log-life:
Hi @pabo99! Thank you for the feedback.

I understand the issue now. My branch had accumulated commits from 
my previous PRs (#9740, #9731) which were not merged, causing 
thousands of unrelated lines to appear in the diff.

I have now created a clean PR with only the relevant changes:
- 3 vendored CSS files deleted
- 1 Vue file updated with correct imports

The new PR #9796 has only 4 files changed with minimal 
diff. I will make sure to always create feature branches from an 
updated main branch in the future.

Sorry for the inconvenience!

PR #9797, opened 2026-04-17
feat: add archive option for regular groups

### LINKED ISSUE #9754: [FEATURE] Add archive option for regular groups
**Checklist**

As a user, I want an archive (soft-delete) option for regular groups so that I can hide or deactivate groups without permanently deleting their data, maintaining consistency with the existing team groups functionality.

Context
Currently, the platform allows "team groups" to be archived, but this feature is missing for "regular groups." The goal is to implement a similar soft-delete or archive mechanism for regular groups to provide a consistent user experience across the site. This will allow group administrators to manage old or inactive groups more effectively without losing historical information.

console-log-life:
## Problem
Regular groups had no way to be archived or hidden. Only team groups 
had this functionality, causing inconsistency in the UI.

## Changes
- Added `archived` column to `Groups_` table (migration 00271)
- Updated `Groups` VO with `archived` field
- Updated `Groups` DAO to filter archived groups from default listings
- Added `apiArchive()` to `Group` controller for soft delete
- Added archive button in `Group List.vue` with confirmation dialog
- Updated `api_types.ts` with `archived` field in `Group` interface

## Testing
- Build passes with `npm run build` ✅

Fixes #9754

### DIFF
--- a/frontend/database/00271_add_archived_to_groups.sql
+++ b/frontend/database/00271_add_archived_to_groups.sql
@@ -1,0 +1,3 @@
+ALTER TABLE `Groups_`
+  ADD COLUMN `archived` tinyint(1) NOT NULL DEFAULT '0'
+    COMMENT 'Indicates whether the group has been archived (soft delete)';
--- a/frontend/database/schema.sql
+++ b/frontend/database/schema.sql
@@ -493,6 +493,7 @@ CREATE TABLE `Groups_` (
   `alias` varchar(50) NOT NULL,
   `name` varchar(50) NOT NULL,
   `description` varchar(256) DEFAULT NULL,
+  `archived` tinyint(1) NOT NULL DEFAULT '0' COMMENT 'Indicates whether the group has been archived (soft delete)',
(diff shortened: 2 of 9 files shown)


pabo99 (commented):
Due to policy, we cannot introduce DAO changes in the same PR as the schema/data changes. Could you try splitting this PR?

The first PR only needs to include changes to:

- `frontend/database/00271_add_archived_to_groups.sql`
- `frontend/database/schema.sql`

Once that change is approved, we can then continue with the DAO changes. As a reference, you can take a look at how the change for the `Team_Groups` table was handled here:

https://github.com/omegaup/omegaup/pull/9832

console-log-life:
Got it, thanks for the clarification. I'll split the changes and create a separate PR with only the schema/database updates first, then follow up with the DAO changes.

console-log-life:
Closing this PR since the changes were split following the review suggestion. The schema-only changes have been moved to #9851 and the remaining backend/UI changes will follow separately.

Issue #9726, opened 2026-04-02
[FEATURE] Add archive option for team groups (soft delete)


abhinavsingh-hub:
## Checklist

- [x] I have searched existing issues and confirmed this is not a duplicate (Required)

### Add archive option for team groups (soft delete)

Currently, team groups in omegaUp can only be edited, but there is no option to remove, hide, or manage unused team groups. This leads to UI clutter when organizers create temporary, test, or duplicate team groups.

### Proposed solution

Introduce an **"Archive" option** for team groups (soft delete):

* Archived team groups are hidden from the default list view
* They remain in the database to preserve contest integrity and history
* Optionally, provide a toggle (e.g., "Show archived") to view them

---

### Expected behavior

* Archive action available from:

  * Teams group list (row-level action)
  * Edit page
* Archived team groups:

  * Are excluded from default listings
  * Are not selectable when creating or editing contests
  * Can optionally be restored (nice-to-have)
* Include a confirmation modal before archiving

### Benefits

* Keeps the UI clean and manageable for organizers
* Avoids risks associated with hard deletion (data loss, broken references)
* Maintains backward compatibility with existing contests and submissions

---

### Alternatives considered

1. **Hard delete team groups**

   * Rejected due to potential data integrity issues (team groups may be linked to contests, scoreboards, or submissions)

2. **Do nothing**

   * Leads to increasing UI clutter and poor management experience over time

3. **Permanent disable without visibility control**

   * Less flexible than archiving and does not allow recovery or auditability

### Implementation notes 

* Add a boolean flag (e.g., `is_archived`) to the team groups table
* Filter archived groups out of default queries
* Minimal changes required to existing flows

---

## Additional context

This will help us to verify the solution proposed in PR #9678 

---

## Screenshots showcasing no option of archiving or deleting 

<img width="1512" height="982" alt="Image" src="https://github.com/user-attachments/assets/860bf99e-7d8e-4637-9be6-caaa39065e9a" />
<img width="1512" height="982" alt="Image" src="https://github.com/user-attachments/assets/262c9c28-faab-439a-840a-7edbd439e6b8" />

abhinavsingh-hub:
Hey @pabo99 !

I have figured that this might be a very good feature to have as it will help declutter the page. Would love to work on it, please assign this issue to me

Regards

console-log-life:
Hey @pabo99! I am interested to solve this issue can you assign me?

pabo99:
[...]

PR #9798, opened 2026-04-18
Feature: Archive (Soft Delete) for Team Groups

abhinavsingh-hub:
#9726  

### Motivation
Currently, team groups cannot be removed or hidden. Over time, this leads to accumulation of:
- temporary/test groups
- duplicate groups
- outdated groups

This creates UI clutter and makes it harder for organizers to manage team groups effectively.

---

## Current Progress (This PR)

This PR focuses on **frontend groundwork and UX exploration**, without introducing any backend or schema changes yet.

### Changes implemented:
- Added UI structure in `teamsgroup/List.vue` to support:
  - Filtering of team groups (designed to exclude archived ones)
  - Placeholder "Archive" action (non-persistent for now)
- Refactored list rendering to support derived/computed filtering logic
- Ensured existing behavior remains unchanged (no breaking changes)

### Why this approach:
- Keeps the PR **safe and reviewable**
- Avoids premature backend changes before confirming direction
- Allows validation of UX and interaction patterns first

---

##  Proposed Solution (Planned)

Introduce an **archive (soft delete)** mechanism using an `is_archived` flag.

### Expected behavior:
- Archived team groups are **hidden from the default list view**
- They are **not selectable in new workflows** (e.g., contest creation)
- They **remain in the database** to preserve referential integrity
- Existing contests using archived groups continue to function normally

---

### 1. Database
Add a new column:
```sql
is_archived TINYINT(1) NOT NULL DEFAULT 0
```

### 2. Backend (API)
- Extend `TeamsGroup::apiUpdate` to accept optional `is_archived`
- Update the model conditionally (non-breaking)
- Include `is_archived` in API responses

> Note: Initial plan is to avoid API-level filtering to prevent breaking existing flows/tests. Filtering would be handled in the frontend first.

---

### 3. Frontend (Follow-up)
- Finalize archive action with API integration
- Hide archived groups by default
- Add optional "Show archived" toggle
- Prevent archived groups from appearing in selection dropdowns

---

###  Design Considerations
- **Backward compatibility**: Existing contests referencing team groups must continue to work
- **Data safety**: Avoid hard deletion to prevent breaking references
- **Incremental rollout**: Validate UX and API contract before introducing global filtering
- **Consistency**: Ensure archived groups behave predictably across all UI surfaces

---

###  Questions / Feedback Requested
1. Is introducing an `is_archived` flag the preferred approach for this use case?
2. Should filtering of archived groups eventually happen at the **API level**, or remain frontend-driven?
3. Are there existing patterns in omegaUp for soft deletion or archiving that should be followed?
4. Would restore/unarchive functionality be expected in scope, or should this remain one-way initially?
5. Are there any edge cases (e.g., contests, scoreboards, permissions) that should be explicitly handled?

---

Screenshot of draft UI

Screenshot 2026-04-18 at 7 40 56 AM

Also plan to alter the colour of used archive icon from fontsawesome to rgb(197, 215, 234) to keep it consistent with the theme of OmegaUp


abhinavsingh-hub:
hey @pabo99 @tvanessa @heduenas !

Would love your feedback and guidance for my plan of implementation

Regards

frontend/www/js/omegaup/components/teamsgroup/List.vue
```diff
@@ -60,5 +97,27 @@ export default class TeamsGroupList extends Vue {
...
+  async archiveGroup(teamsGroup: types.TeamsGroup) {
+  if (!confirm('Are you sure you want to archive this team group?')) return;
+
+  try {
+    await this.$api.TeamsGroup.update({
+      team_group_alias: teamsGroup.alias,
+      is_archived: true,
+    });
+
+    teamsGroup.is_archived = true;
```
pabo99: The field `is_archived` does not exist in the database. I recommend using the field name `archive` for consistency with other tables.

The first step should be to create it. This should include:

- A migration script that adds the field
- Changes in the schema.sql file

You can use this change as a reference: https://github.com/omegaup/omegaup/pull/5102/changes

The rest of the changes should be done in a separate PR, due to our policy of not mixing schema changes with DAO changes.

frontend/www/js/omegaup/components/teamsgroup/List.vue
```diff
@@ -60,5 +97,27 @@ export default class TeamsGroupList extends Vue {
...
+  } catch (e) {
+    console.error(e);
+  }
```
pabo99: To keep consistency with the rest of the code, could you move this to the `frontend/www/js/omegaup/teamsgroup/list.ts` file? From here, you could emit the action to be handled in that file.

frontend/www/js/omegaup/components/teamsgroup/List.vue
```diff
@@ -18,22 +25,52 @@
...
             </td>
             <td>
               <a :href="teamsGroupEditUrl(teamsGroup)" :title="T.wordsEdit">
                 <font-awesome-icon :icon="['fas', 'edit']" />
               </a>
+
+              <button
+                class="btn btn-link text-danger p-0 ml-2"
+                @click="archiveGroup(teamsGroup)"
+                title="Archive"
```
pabo99: Could you add this string to the translation files?

frontend/www/js/omegaup/components/teamsgroup/List.vue
```diff
@@ -1,15 +1,22 @@
...
+      <a class="btn btn-primary mx-1" href="/teamsgroup/new/">
+        {{ T.teamsGroupsCreateNew }}
+      </a>
     </div>
     <div class="card">
       <div class="card-header mb-3">
+        <div class="mb-2 text-right">
+          <label>
+            <input type="checkbox" v-model="showArchived" />
+            Show archived
```
pabo99: In general, all text strings displayed in the UI should be added as localization strings.

abhinavsingh-hub:
Hey @pabo99 !

I've completed the frontend implementation for the Team Groups archiving feature and resolved a few backend/UI bugs that surfaced during testing.

## Key Updates:

1. **Backend Validation Fix:** 
Fixed a bug in `TeamsGroup.php` (`apiUpdate`) where `numberOfContestants` was mistakenly being enforced as a required parameter (`ensureInt`) instead of optional, which was throwing an `Empty parameter` error during partial updates like archiving. It now correctly uses `ensureOptionalInt` and falls back to the existing database value.

2. **Frontend Reactivity & Architecture:** 
    * Refactored `List.vue` to emit an `archive-group` event rather than making the API call directly inside the view, maintaining proper controller/view separation with `list.ts`.
    * Wrapped `teamsGroups` inside `data` in `list.ts` upon initialization. This guarantees Vue tracks the array reactively, ensuring the list auto-updates instantly upon archiving/unarchiving.
    
3. **"Show Archived" Filter Logic:** 
Corrected the computed property `visibleTeamsGroups` so that checking the "Show archived" box exclusively displays the archived groups, rather than mixing them with active groups.

4. **UI & Styling Polish:** 
    * Updated the archive icon color to `#8bb0d3` to remain consistent with OmegaUp's theme.
    * Refactored the `card-header` using Bootstrap flexbox (`justify-content-between align-items-center`) so the "Teams groups" title and the "Show archived" checkbox are perfectly horizontally aligned.
    
5. **Full Localization:** Replaced all hardcoded English text and added proper translation strings to the `.lang` files (en, es, pt, pseudo) for the archive/unarchive confirmation prompts and success alerts.

****

### Screen recording validation- 

https://github.com/user-attachments/assets/0f8aeac3-4401-4eea-9178-e42a47effae3


***

Hope that these changes are good to go, let me know if anything else is required, would be more than happy to contribute!!

Regards

frontend/www/js/omegaup/components/teamsgroup/List.vue
```diff
@@ -33,6 +39,21 @@
               <a :href="teamsGroupEditUrl(teamsGroup)" :title="T.wordsEdit">
                 <font-awesome-icon :icon="['fas', 'edit']" />
               </a>
+
+              <button
+                class="btn btn-link p-0 ml-2"
+                style="color: #8bb0d3"
```
pabo99: Could you avoid using inline styles? Preferably, use a Bootstrap class if one already exists; otherwise, it would be better to create a custom one.

Also, for any new colors we add, it is recommended to define them as variables in ` _variables.scss`

frontend/www/js/omegaup/components/teamsgroup/List.vue
```diff
@@ -33,6 +39,21 @@
...
+                :title="teamsGroup.archived ? 'Unarchive' : 'Archive'"
```
pabo99: All text displayed on screen must go through the localization files, including these two strings

frontend/www/js/omegaup/components/teamsgroup/List.vue
```diff
@@ -84,7 +109,24 @@ export default class TeamsGroupList extends Vue {
...
+  get visibleTeamsGroups(): types.TeamsGroup[] {
+    if (this.showArchived) {
+      return this.teamsGroups.filter((g) => g.archived);
+    }
+    return this.teamsGroups.filter((g) => !g.archived);
+  }
+
+  archiveGroup(teamsGroup: types.TeamsGroup) {
+    if (teamsGroup.archived) {
+      if (!confirm(T.teamsGroupUnarchiveConfirmText)) return;
```
pabo99: Is there a way to replace this with a BootstrapVue modal? Most likely there are already some examples in the codebase

frontend/www/js/omegaup/components/teamsgroup/List.vue
```diff
@@ -84,7 +109,24 @@ export default class TeamsGroupList extends Vue {
...
+      this.$emit('archive-group', teamsGroup, false);
```
pabo99: When emitting actions with more than one parameter, it is preferable to wrap them inside an object so the parameters are easier to identify and maintain:
```suggestion
      this.$emit('archive-group', { teamsGroup, archived: false });
```

frontend/www/js/omegaup/components/teamsgroup/List.vue
```diff
@@ -84,7 +109,24 @@ export default class TeamsGroupList extends Vue {
...
+    } else {
+      if (!confirm(T.teamsGroupArchiveConfirmText)) return;
+      this.$emit('archive-group', teamsGroup, true);
+    }
```
pabo99: You can avoid using the `else` statement here by using an early return inside the `if` statement.

Also, the same corrections from the previous suggestions can be applied to this content as well.


Now that I think about it, you do not need to use nested conditions if you use a ternary operator for the confirmation text, something like:
```ts
    const confirmMessage = teamsGroup.archived
      ? T.teamsGroupUnarchiveConfirmText
      : T.teamsGroupArchiveConfirmText;
    if (!confirm(confirmMessage)) return;
    this.$emit('archive-group', { teamsGroup, archived: !teamsGroup.archived });
```

frontend/www/js/omegaup/teamsgroup/list.ts
```diff
@@ -10,9 +13,34 @@ OmegaUp.on('ready', () => {
...
+      teamsGroups: payload.teamsGroups,
+    }),
     render: function (createElement) {
       return createElement('omegaup-teams-group-list', {
-        props: { teamsGroups: payload.teamsGroups },
+        props: { teamsGroups: this.teamsGroups },
+        on: {
+          'archive-group': (
+            teamsGroup: types.TeamsGroup,
+            archived: boolean,
```
pabo99: Here, the parameters should be replaced with an object containing them, see the suggestion example here:

https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/Coding-guidelines.md#typescript

pabo99 (commented):
Good work! I left a few suggestions to make the code more readable, but overall it looks good

abhinavsingh-hub:
Hey @pabo99 !

Thankyou for your recommendations, I have updated accordingly and I hope they are good to go, let me know if anything else to be done. Glad to contribute!!

Regards

pabo99 (approved):
The change looks really good, thank you for taking care of this.

I would just like to share a suggestion that we could revisit in a separate issue:

* When selecting the “Show archived” checkbox, it might make more sense for it to behave as “Show archived as well”, so active groups never disappear from the list. Additionally, archived groups could be visually differentiated in the UI using a dimmed or disabled-like style.

If, in the future, users end up having a considerable number of active and archived groups, we could consider separating both categories into different tabs. However, for now, this doesn't seem to be necessary yet

abhinavsingh-hub:
@pabo99 I had thought about it as well but then realised that it might cause a bit of clutter on screen, specially in scenario where a member might be part of multiple teamgroups, 

so what if instead, a separate button is created that will help us show both, archived and unarchived teamgroups, so that the user would be able to see both of them in the same list, while maintaining the existing options of viewing either archived or unarchived teamgroup

So for example, finally the users would have two buttons, 'show only archived' and 'show both archived and unarchived', we can alter the phrase as per your opinion 

Would love to get your opinions on the same!

pabo99:
Those are valid options. However, since this is not a primary tool within omegaUp, I think it would be better to keep it in its current form for now until we receive more feedback from users.

We can always revisit the UX later if needed based on how people actually use the feature

abhinavsingh-hub:
Fair enough, glad to contribute!!

PR #9805, opened 2026-04-22
fix: extract hashchange listener into named function for cleanup

### LINKED ISSUE #9783: [BUG] arena/course.ts registers an anonymous hashchange listener with no cleanup
## Checklist

## Current Behavior

In `frontend/www/js/omegaup/arena/course.ts` (line 674), a `hashchange` event listener is registered using an anonymous `async` arrow function:

```ts
window.addEventListener('hashchange', async () => {
```

Because the function is anonymous, there is no way to pass a reference to `removeEventListener`. If the component is destroyed and remounted, each mount registers an additional listener, so multiple listeners accumulate over the session lifetime.

## Expected Behavior

The handler should be extracted into a named function so it can be removed when the component is torn down (e.g., in `beforeDestroy` or via `onBeforeUnmount`). Other entry points in the codebase (e.g., `group/edit.ts`, `teamsgroup/edit.ts`) already follow this pattern correctly.

Found in PR #9481. PR #8776 can be used as a reference implementation.

chiraglaheru:
Fixes #9783

Extracted the anonymous `hashchange` event listener in `arena/course.ts` 
into a named function `onHashChange` so it can be properly removed via 
`removeEventListener` when the component is destroyed, preventing listener 
accumulation across remounts.

Reference: PR #8776

### DIFF
--- a/frontend/www/js/omegaup/arena/course.ts
+++ b/frontend/www/js/omegaup/arena/course.ts
@@ -671,7 +671,7 @@ OmegaUp.on('ready', async () => {
 
   const component = arenaCourse.$refs.component as arena_Course;
 
-  window.addEventListener('hashchange', async () => {
+  const onHashChange = async () => {
     const { problem, guid } = getOptionsFromLocation(window.location.hash);
     if (guid != null && problem != null) {
       navigateToProblem({
@@ -716,5 +716,10 @@ OmegaUp.on('ready', async () => {
         });
       component.currentPopupDisplayed = PopupDisplayed.RunDetails;
     }
+  };
+  window.addEventListener('hashchange', onHashChange);
+
+  arenaCourse.$once('hook:beforeDestroy', () => {
+    window.removeEventListener('hashchange', onHashChange);
   });
 });


chiraglaheru:
@pabo99 I re-ran the CI and most checks are now passing. The remaining failing/stuck jobs seem unrelated to this change, since this PR only extracts the anonymous `hashchange` listener into a named function for cleanup in `arena/course.ts`.

Could someone please take a look or re-run the flaky/stuck jobs if needed?

frontend/www/js/omegaup/arena/course.ts
```diff
@@ -716,5 +716,10 @@ OmegaUp.on('ready', async () => {
         });
       component.currentPopupDisplayed = PopupDisplayed.RunDetails;
     }
+  };
+  window.addEventListener('hashchange', onHashChange);
+
+  arenaCourse.$once('hook:beforeDestroy', () => {
+    window.removeEventListener('hashchange', onHashChange);
```
pabo99, line 723: In this [PR](https://github.com/omegaup/omegaup/pull/9830) , the hook is no longer used in this way and instead is used directly within the component. Could you apply the necessary changes following that PR for consistency?

PR #9809, opened 2026-04-23
fix: add missing ensureNotInLockdown checks

### LINKED ISSUE #9782: [FEATURE] Centralize entrypoint access control for lockdown mode and auth requirements
## Checklist

As a **developer**, I want **all page entrypoints to declare their lockdown and authentication requirements in a single centralized map** so that **access control guards do not have to be added individually to each controller or entrypoint file**.

## Context

Currently, guards like `ensureNotInLockdown()` must be called explicitly inside each individual controller method or entrypoint. This is error-prone: a new entrypoint can easily miss the check.

A centralized entrypoint map would declare, for each page, whether it requires the user to be logged in, whether it is accessible in lockdown mode, and whether admin access is needed. The `render` function could then apply the correct guard automatically without relying on each entrypoint to do so.

For reference pls check PR #9073.

console-log-life:
# Description
Added missing `ensureNotInLockdown()` checks to API endpoints that modify data but were missing this guard.

Controllers updated:
- AiEditorial (apiGenerate, apiReview, apiUpdateJob)
- CarouselItems (apiCreate, apiDelete, apiUpdate)
- Clarification (apiCreate, apiUpdate)
- GroupScoreboard (apiAddContest, apiRemoveContest)
- Identity (apiCreate, apiBulkCreate, apiBulkCreateForTeams, apiUpdateIdentityTeam, apiUpdate, apiChangePassword)
- Notification (apiReadNotifications)
- ProblemBookmark (apiToggle)
- Reset (apiCreate, apiUpdate)
- School (apiCreate)
- Submission (apiSetFeedback, apiSetFeedbackList)
- TeamsGroup (apiCreate, apiUpdate)

Fixes #9782

### DIFF
--- a/frontend/server/src/Controllers/AiEditorial.php
+++ b/frontend/server/src/Controllers/AiEditorial.php
@@ -30,6 +30,7 @@ class AiEditorial extends \OmegaUp\Controllers\Controller {
      * @return array{status: string, job_id?: string}
      */
     public static function apiGenerate(\OmegaUp\Request $r): array {
+        \OmegaUp\Controllers\Controller::ensureNotInLockdown();
         $r->ensureIdentity();
         $problemAlias = $r->ensureString(
             'problem_alias',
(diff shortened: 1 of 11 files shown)


pabo99 (commented):
Can you run `./stuff/lint.sh` outside the container, but inside your WSL instance (if you are using Windows), to check which errors the linter is reporting?

PR #9860, opened 2026-05-20
[Milestone 1] Set up feature flag, add creation method selector to Form.vue

### LINKED ISSUE #9859: [MileStone 1] Set up feature flag, add creation method selector to Form.vue
Task:
- Setting up feature flag.
- Adding creation method selector to Form.vue.

Image

shauryag2002:
# Description
This pull request introduces a new experiment-driven feature that allows users to select a method for creating problems—either using a problem creator tool or uploading a zip file. This includes updates to the UI, interactive guide, and translations, as well as the necessary backend and frontend logic to enable the feature conditionally.

Key changes include:

**Feature: Problem Creation Method Selector**
* Added a new experiment flag, `problem_creation_method_selector`, to enable or disable the creation method selector for problem creation. [[1]](diffhunk://#diff-c285a52063daa30a5dc3663d29498d39b7d681d5ba9cff5d0de7e1a5d20a9169R48) [[2]](diffhunk://#diff-bc7980c016f8a60251c830e1f9430275e3e2c5eba51603e1b802a1405365da72R11-R12)
* Updated the `Form.vue` component to display a button group allowing users to choose between the problem creator and zip file upload methods, including logic to persist the selected method in session storage and update the interactive guide steps accordingly. [[1]](diffhunk://#diff-9d03d7ae1df6b2f991b6d098d3c6d815259f9645ec94364149d92eff14580b09L79-R135) [[2]](diffhunk://#diff-9d03d7ae1df6b2f991b6d098d3c6d815259f9645ec94364149d92eff14580b09R537) [[3]](diffhunk://#diff-9d03d7ae1df6b2f991b6d098d3c6d815259f9645ec94364149d92eff14580b09R572-R586) [[4]](diffhunk://#diff-9d03d7ae1df6b2f991b6d098d3c6d815259f9645ec94364149d92eff14580b09L535-R597) [[5]](diffhunk://#diff-9d03d7ae1df6b2f991b6d098d3c6d815259f9645ec94364149d92eff14580b09L546-R642) [[6]](diffhunk://#diff-9d03d7ae1df6b2f991b6d098d3c6d815259f9645ec94364149d92eff14580b09L568-R676)
* Modified the problem creation page initialization to pass the experiment flag to the form, enabling the selector only when the experiment is active.

**Testing**
* Added comprehensive tests to `Form.test.ts` to verify the behavior of the creation method selector and related UI elements under different conditions (e.g., feature flag enabled, update mode, method selection).

**Translations**
* Added and updated the `openProblemCreator` translation key in English, Spanish, Portuguese, and pseudo-localized language files and their corresponding TypeScript translation files, ensuring the new UI is localized. [[1]](diffhunk://#diff-b6033fa04ca6d274828419e98793f7c5a7cb4d8c848df193a317d0dfa7802674R1163) [[2]](diffhunk://#diff-b055271c171a04a2b85114bb30d6cdf202e3a5d6619625680b406b256a51100fR1163) [[3]](diffhunk://#diff-5082412cc7044d0a5fadfdcb154e91b97ff6f2c1ae93da53f1d35554bf6eedc3R1163) [[4]](diffhunk://#diff-12615311f60998328eba6eae273ff1d6db8c8c0f18fc603771476892832c5ce0R1163) [[5]](diffhunk://#diff-9518ffb5c3fd51dc701bb931fe37b1cd7f0b288ce7b415382e7ef880a6e2de1aR1164) [[6]](diffhunk://#diff-9a64288b4f7f7dc428776121c960485e8d2de70169852ed29bdcbb15de62cb62R1165) [[7]](diffhunk://#diff-0eb8ebb0ff26e981639a57eeab888593c7c4c8f1dc01984ca02b94e3922f2082R1164) [[8]](diffhunk://#diff-79e90c0778d830653630d8fbe248c55cfb5affd48e1d0aef1ee69f6f96bbaceeR1165) [[9]](diffhunk://#diff-f10300314e6b1a13c456d2aea6c3f799272c3266b53b157305c56fee22953acaR1164) [[10]](diffhunk://#diff-f468c83e754dcd6d0c3ea1aa051d47051da78560cfdbbc37376073dc6a35d092R1165) [[11]](diffhunk://#diff-4d803b8fe78984dea0506d969bef218672f1482dd9042d6912303147b005d51bR1164) [[12]](diffhunk://#diff-587274bab756192cc81a876960dd230b50cab0c789db5f0aa9c96eb6b3d2c768R1165)

image

Fixes: #9859

# Comments

# Checklist:


frontend/server/src/Experiments.php
```diff
@@ -45,6 +45,7 @@ class Experiments {
 
     private const KNOWN_EXPERIMENTS = [
         self::USER_README,
+        'problem_creation_method_selector',
```
pabo99: For consistency, could you add a constant similar to `USER_README`?

frontend/templates/es.lang
```diff
@@ -1160,6 +1160,7 @@ omegaupTitleUsersEditEmail = "Editar correo electrónico"
 omegaupTitleUsersRank = "Ranking de coders"
 onlyDemotionsSupported = "Solamente se pueden resolver demociones por ahora."
 openInKarelJs = "Puedes utilizar karel.js para resolver este problema"
+openProblemCreator = "¿Quieres crear un nuevo problema?"
```
pabo99: For consistency with the English translation, could this translation be changed to:
```suggestion
openProblemCreator = "Abrir creador de problemas"
```
?

frontend/templates/pt.lang
```diff
@@ -1160,6 +1160,7 @@ omegaupTitleUsersEditEmail = "Editar e-mail"
 omegaupTitleUsersRank = "Ranking de coders"
 onlyDemotionsSupported = "Resolução suportada apenas para demotions para agora."
 openInKarelJs = "Você pode usar karel.js para resolver este problema"
+openProblemCreator = "Criar problema"
```
pabo99: Same case here:
```suggestion
openProblemCreator = "Abrir criador de problemas"
```

frontend/www/js/omegaup/components/problem/Form.vue
```diff
@@ -76,7 +76,63 @@
...
+                <div
+                  v-if="!isUpdate && showCreationMethodSelector"
+                  class="form-group col-md-6 introjs-creation-method"
+                >
+                  <div class="btn-group btn-group-toggle d-flex" role="group">
+                    <button
+                      type="button"
+                      class="btn flex-fill"
+                      :class="{
+                        'btn-primary': creationMethod === 'creator',
```
pabo99: Could you add `creator` and `zip` to an enum called `CreationMethods`? This way we can avoid errors caused by typos

frontend/www/js/omegaup/components/problem/Form.vue
```diff
@@ -76,7 +76,63 @@
...
+                        'is-invalid': errors.includes('problem_contents'),
+                      }"
+                      @change="onUploadFile"
+                    />
+                  </div>
+                </div>
+                <div
+                  v-if="isUpdate || !showCreationMethodSelector"
+                  class="form-group col-md-6 introjs-file"
+                >
```
pabo99: Wouldn't this be the opposite condition of the `if` on line 80? If so, it would be clearer to express it this way here:

```suggestion
                <div v-else class="form-group col-md-6 introjs-file">
```

frontend/www/js/omegaup/components/problem/Form.vue
```diff
@@ -512,66 +569,111 @@ export default class ProblemForm extends Vue {
...
 
   mounted() {
+    if (!this.isUpdate && this.showCreationMethodSelector) {
+      const savedMethod = window.sessionStorage.getItem(
+        'problemCreationMethod',
+      );
+      if (savedMethod === 'creator' || savedMethod === 'zip') {
+        this.creationMethod = savedMethod;
+      }
+    }
```
pabo99: Is there a way to get the `creationMethod` from the `new.ts` file? This way we could avoid using `sessionStorage` in the component.

frontend/www/js/omegaup/components/problem/Form.vue
```diff
@@ -512,66 +569,111 @@ export default class ProblemForm extends Vue {
...
-              title,
-              intro: T.createProblemInteractiveGuideType,
-            },
-            {
-              element: document.querySelector('.introjs-validator') as Element,
-              title,
-              intro: T.createProblemInteractiveGuideValidator,
-            },
-          ],
+          steps,
```
pabo99: Is this change necessary for this PR? It seems to add a bit of noise and makes the diff harder to review. Could you revert it and leave it for a separate PR instead?

frontend/www/js/omegaup/components/problem/Form.vue
```diff
@@ -512,66 +569,111 @@ export default class ProblemForm extends Vue {
...
         .start();
       this.$cookies.set('has-visited-create-problem', true, -1);
     }
   }
 
+  setCreationMethod(method: 'creator' | 'zip'): void {
+    this.creationMethod = method;
+    if (!this.isUpdate && this.showCreationMethodSelector) {
+      window.sessionStorage.setItem('problemCreationMethod', method);
+    }
```
pabo99: Could you instead emit an action to `new.ts` and let that file be responsible for storing the value in `sessionStorage`?

pabo99 (commented):
This leaves the creator path exposed but not fully wired yet (the button has no action). It should be low-risk for now since the experiment flag for everyone is off, but we should complete that flow sioon or at least add an alert.

frontend/server/src/Experiments.php
shauryag2002: Added `const PROBLEM_CREATION_METHOD_SELECTOR` and used `self::PROBLEM_CREATION_METHOD_SELECTOR` in `KNOWN_EXPERIMENTS`

frontend/templates/es.lang
shauryag2002: Done

frontend/templates/pt.lang
shauryag2002: Done :+1:

frontend/www/js/omegaup/components/problem/Form.vue
shauryag2002: Added exported enum with `Creator = 'creator'` and `Zip = 'zip'`, replaced all string literals

frontend/www/js/omegaup/components/problem/Form.vue
shauryag2002: Done

frontend/www/js/omegaup/components/problem/Form.vue
shauryag2002: Reverted steps back to original inline format, removing the extracted array approach

frontend/www/js/omegaup/components/problem/Form.vue
shauryag2002: Removed it for now, I will add it later if we need it.

frontend/www/js/omegaup/components/problem/Form.vue
shauryag2002: Removed it for now, I will add it later if we need it.

frontend/www/js/omegaup/components/problem/Form.vue
```diff
@@ -478,6 +543,8 @@ export default class ProblemForm extends Vue {
   @Prop({ default: false }) isUpdate!: boolean;
   @Prop({ default: 0 }) originalVisibility!: number;
   @Prop({ default: true }) hasVisitedSection!: boolean;
+  @Prop({ default: false }) showCreationMethodSelector!: boolean;
+  @Prop({ default: 'creator' }) initialCreationMethod!: string;
```
pabo99: We are avoiding the use of the `initial` suffix in Vue props. Instead, we suggest using the `current` suffix for the reactive data, and this prop could simply be named `creationMethod`.

And since it's already receiving a value of type `CreationMethods`, you can declare it as such here as well:
```suggestion
  @Prop({ default: CreationMethods.Creator }) creationMethod!: CreationMethods;
```

frontend/www/js/omegaup/components/problem/Form.vue
```diff
@@ -512,6 +579,9 @@ export default class ProblemForm extends Vue {
   validLanguages = this.data.validLanguages;
   validatorTypes = this.data.validatorTypes;
   currentLanguages = this.data.languages;
+  CreationMethods = CreationMethods;
+  creationMethod: CreationMethods =
+    (this.initialCreationMethod as CreationMethods) || CreationMethods.Creator;
```
pabo99: This way, you no longer need to use the type assertion because the type is already declared correctly. Also, you don't need to provide a default value since the prop already handles it:

```suggestion
  currentCreationMethod: CreationMethods = this.creationMethod;
```

frontend/www/js/omegaup/problem/new.ts
```diff
@@ -3,11 +3,20 @@ import { types } from '../api_types';
...
 
 OmegaUp.on('ready', () => {
   const payload = types.payloadParsers.ProblemFormPayload();
+  const showCreationMethodSelector =
+    OmegaUp.experiments?.isEnabled('problem_creation_method_selector') ?? false;
+
+  let initialCreationMethod: CreationMethods = CreationMethods.Creator;
+  if (showCreationMethodSelector) {
+    initialCreationMethod = CreationMethods.Creator;
+  }
```
pabo99: Here, the value of `initialCreationMethod` never changes, so perhaps it should be initialized as `.zip`. Also, I think this can be simplified a bit using a ternary operator:

```suggestion
  const creationMethod = showCreationMethodSelector
    ? CreationMethods.Creator
    : CreationMethods.Zip;
```

Does that sound like a good idea to you?

pabo99 (commented):
Just a few minor suggestions!

frontend/www/js/omegaup/components/problem/Form.vue
shauryag2002: Done.

frontend/www/js/omegaup/problem/new.ts
shauryag2002: Yes. :+1:

frontend/www/js/omegaup/components/problem/Form.vue
shauryag2002: :+1:

pabo99 (approved):
Excellent change. Thank you very much, @shauryag2002!

PR #9867, opened 2026-05-26
fix(frontend): move help button inside label in User Preferences

### LINKED ISSUE #9746: [Bug] Lone question mark wraps to next line on mobile view in User Preferences
### **Description**
A UI bug was discovered on the User Preferences page when viewed from a mobile device. In the section asking to hide profile details, the text wrapping is breaking up the final word and its punctuation. The closing question mark (`?`) is wrapping alone to the next line, rather than staying attached to the end of the text "Would you like to hide profile details from others?". This behavior is not present in the desktop view.

### **Steps to Reproduce**
1. Log in to omegaUp.
2. Go to your profile -> Preferences.
3. View the page on a mobile device or by resizing your desktop browser to a mobile breakpoint.
4. Observe the text: "Would you like to hide profile details from others?"

### **Expected Behavior**
The text should wrap normally. If the screen is small, the last word ("others") and the question mark ("?") should stay together. The question mark should never wrap to a line by itself.

### **Actual Behavior**
The text wraps incorrectly, moving the question mark to its own line below the main text.

### **Screenshots**

Image

arshad-muhammad:
## Description

This PR fixes the mobile layout issue where the help icon in `PreferencesEdit.vue` moves to a separate line on smaller screens.

### Root Cause

The help button (`question-circle` icon) was placed outside the `<label>` element. On narrow viewports, the label expands across the available width, causing the button to wrap onto its own line.

### Solution

* Moved the help button inside the `<label>` element so it stays in the same inline flow as the text.
* Added:

```vue
@click.stop.prevent="show = !show"
```

to prevent the help button click from toggling the checkbox.

### Result

The help icon now stays aligned with the label text properly on mobile devices without affecting checkbox behavior.
Fixes #9746


frontend/www/js/omegaup/components/user/PreferencesEdit.vue
```diff
@@ -89,18 +89,18 @@
...
+        <b-button
+          id="popover-private-profile"
+          class="ml-1"
+          size="sm"
+          variant="none"
+          @click.stop.prevent="show = !show"
+        >
+          <font-awesome-icon :icon="['fas', 'question-circle']" />
+        </b-button>
+        <!-- id-lint on -->
```
pabo99, line 106: It seems that wrapping it inside the label is not the best option because clicking this button also attempts to toggle the checkbox:

https://github.com/user-attachments/assets/b7838368-36a0-4497-894b-aff5a4f9b4c0

Let's think of another solution instead. Here’s what comes to mind:

- Move the button outside the label, as it originally was, and wrap both the `label` and `b-button` inside a `div` with `d-inline-flex` to keep them on the same line without the unintended checkbox trigger behavior.
- Change the popover placement to `bottom` so it does not overlap the trigger button on small screens.
- Remove the button focus border/ring so it does not distract from the checkbox row. You can achieve this with `box-shadow: none;` and `outline: none;` on `:focus` and `:active:focus`.

pabo99 (commented):
I left a few suggestions for a different approach to fix this. Another thing that helps us a lot during UI-related reviews is including a screenshot or screencast so we can quickly identify the changes. Please keep this in mind for future contributions

arshad-muhammad:
@pabo99 Thanks for the suggestions! I’ll fix the issues and submit a new PR. I’ll also make sure to include screenshots for UI-related issues in future reports. Really appreciate the feedback, thanks!

pabo99 (approved):
Great!

PR #9883, opened 2026-06-02
feat(cron): add unit test infrastructure and aggregate_feedback/utils tests

### LINKED ISSUE #9882: [FEATURE] Add unit test infrastructure for cron scripts
### Checklist

### Context

The four cron scripts in `stuff/cron/` have very little Python unit test coverage. Only `aggregate_feedback.py` and `build_problem_rec_model.py` have a small test each. `update_ranks.py`, `assign_badges.py` and `utils.py` have none, so a bug in the scoring or badge ownership logic could ship silently and produce wrong rankings or duplicate badges before anyone notices.

This issue proposes adding a` stuff/cron/tests/ `package with:

- A reusable MockCursor and MockConnection that mirror `lib.db.Connection` so cron functions can be tested without a real database.
- Typed sample fixtures (rank cutoffs, vote builders, user-rank shape, tag-vote dictionaries) shared across the test files.
- Unit tests for the pure functions in aggregate_feedback.py and utils.py to start with. After this i will raise follow ups to add tests for `update_ranks.py` and `assign_badges.py`.

The existing python CI job in `.github/workflows/ci.yml` already runs pytest under` ./stuff/`, so the new tests will be picked up automatically with no workflow change.

This issue will be linked to a GSoC 2026 Cronjob Optimization parent tracking issue once that is opened.

PrasannaMishra001:
# Description

Adds the basic test setup that will be used for the cronjob optimization project.

Added `stuff/cron/tests/` package which has:-
- `fixtures/mock_cursor.py` has `MockCursor` and `MockConnection`. They mimic `lib.db.Connection` and `MySQLCursor` close enough so the existing cron code runs against them without any change. Match resolution uses substring search on whitespace-collapsed sql, so the tests do not break if someone reformats a query later. Every `(sql, params)` pair is also kept in a `calls` list so tests can assert on what was executed.
- `fixtures/sample_data.py` keeps the shared fixtures (`SAMPLE_RANK_CUTOFFS`, the `make_quality_votes` helper, `SAMPLE_USER_RANK` and three tag-vote dicts) in one place.
- `test_aggregate_feedback.py` covers `bayesian_average`, `get_weighting_factor` and `get_most_voted_tags`.
- `test_utils.py` covers `get_first_day_of_next_month` for every month, the dec to jan year boundary and a mid month date rolling to the 1st.

It also fixes a small bug the tests turned up in `get_most_voted_tags`: with no tag votes at all it called `max()` on an empty dict and raised, so it now returns `None` like the other no result paths. `test_empty_votes_returns_none` covers it.

Total 34 unit tests, all passing.

The existing python CI job in `.github/workflows/ci.yml` already runs `pytest --ignore=./stuff/teaching_assistant/ ./stuff/`, so the new tests get picked up by CI as it is, no workflow change needed.

Fixes: #9882

# Comments

This is the Week 1 deliverable for the GSoC 2026 Cronjob Optimization project. Follow-up PRs will add tests for `update_ranks.py` and `assign_badges.py`, then move on to the sql optimization part. Will open a parent tracking issue once a couple of PRs land so everything links from one place.

# Checklist:


stuff/cron/tests/fixtures/mock_cursor.py
```diff
@@ -0,0 +1,103 @@
...
+        self._script: List[Tuple[str, Sequence[Row]]] = list(script or [])
+        self._dictionary = dictionary
+        self._current: Sequence[Row] = []
+        self.calls: List[Tuple[str, Any]] = []
+
+    def execute(self, sql: str, params: Any = None) -> None:
+        '''Record the call and resolve a result set from the script.'''
+        self.calls.append((sql, params))
+        normalized = _normalize(sql)
+        for substring, rows in self._script:
```
Ankitsinghsisodya, line 40: 🟡 risk: substring keys in `Script` are never lowercased before matching. If a caller passes `'SELECT'` (uppercase) it silently fails to match.\n\n**Fix:** change the comparison to `if _normalize(substring) in normalized` (apply `_normalize` to the key too, not just the SQL).

Ankitsinghsisodya (commented):
Solid Week 1 foundation — tests are well-structured, assertions match the source logic, and docstrings are clean. Issues below are mostly latent footguns and coverage gaps that will matter as follow-up PRs build on this infrastructure.

stuff/cron/tests/fixtures/mock_cursor.py
```diff
@@ -0,0 +1,103 @@
...
+            if substring.lower() in normalized:
+                self._current = rows
```
Ankitsinghsisodya, line 42: 🟡 **risk:** `substring` is never lowercased before the `in` check, but `normalized` is fully lowercased. A caller who stores a key like `"SELECT"` (uppercase) will silently get an empty result set.

**Fix:** apply `_normalize` to both sides:
```python
if _normalize(substring) in normalized:
```

stuff/cron/tests/fixtures/mock_cursor.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def fetchall(self) -> List[Row]:
+        '''Return the currently scripted result set.'''
+        return list(self._current)
+
+    def fetchone(self) -> Optional[Row]:
+        '''Return the first scripted row, or None.'''
+        return self._current[0] if self._current else None
+
+    def __iter__(self) -> Iterator[Row]:
```
Ankitsinghsisodya, line 67: 🟡 **risk:** `fetchone()` does not consume the row — repeated calls always return `_current[0]`. The real `MySQLCursor.fetchone()` advances the cursor position. Any future test that calls `fetchone()` twice on the same scripted cursor will silently get the same row both times.

**Fix:** track an index or use `list.pop(0)` to drain the list, mirroring real cursor behaviour.

stuff/cron/tests/fixtures/mock_cursor.py
```diff
@@ -0,0 +1,103 @@
...
+class MockConnection:
+    '''Stand-in for `lib.db.Connection` with scriptable cursors.'''
+
+    def __init__(
+        self,
+        cur: MockCursor,
+        cur_readonly: Optional[MockCursor] = None,
+    ) -> None:
+        self._cur = cur
+        self._cur_ro = cur_readonly or cur
```
Ankitsinghsisodya, line 93: 🟡 **risk:** `cursor()` accepts a `dictionary` kwarg but discards it with `del`. `MockCursor.__init__` also accepts `dictionary` but never uses it in `fetchall`/`fetchone`. Code under test that calls `cursor(dictionary=True)` will silently receive tuples instead of dicts, masking bugs.

**Fix:** either honour the flag in `fetchall`/`fetchone` (return `dict(zip(column_names, row))`) or raise `NotImplementedError('dictionary cursors not yet supported')` so failures are loud.

stuff/cron/tests/fixtures/mock_cursor.py
```diff
@@ -0,0 +1,103 @@
...
+        self.commits = 0
+        self.rollbacks = 0
+        self.conn = self
+
+    def cursor(
+        self,
+        buffered: bool = False,
+        dictionary: bool = False,
```
Ankitsinghsisodya, line 102: 🔴 **bug:** `MockConnection` has no `get_warnings` attribute. `aggregate_feedback.replace_voted_tags()` reads and writes `dbconn.conn.get_warnings` directly (see `aggregate_feedback.py` lines 285–288). Since `conn = self`, any future test that exercises `replace_voted_tags` through `MockConnection` will raise `AttributeError`.

**Fix:** add the attribute in `__init__`:
```python
self.get_warnings: bool = False
```

stuff/cron/tests/fixtures/sample_data.py
```diff
@@ -0,0 +1,66 @@
+'''Typed sample fixtures shared across cron unit tests.'''
+from typing import Dict, List, Sequence
+
+import aggregate_feedback
+import utils
```
Ankitsinghsisodya: 🟡 **risk:** `import aggregate_feedback` and `import utils` at module level with no `sys.path` setup in this file. These imports work only because the calling test file has already injected the parent directory into `sys.path` before importing `sample_data`. If someone runs pytest from `stuff/cron/tests/` directly, or imports `sample_data` in isolation, it will fail with `ModuleNotFoundError`.

**Fix:** add the same `sys.path.insert` block used in the test files here, or centralise the path setup in a `conftest.py` at `stuff/cron/`.

stuff/cron/tests/fixtures/sample_data.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+SAMPLE_TAG_VOTES_SINGLE_DOMINANT: Dict[str, float] = {
+    'math': 20.0,
+    'graph-theory': 1.0,
+}
+
+SAMPLE_TAG_VOTES_TIE_AT_PROPORTION: Dict[str, float] = {
+    'math': 20.0,
+    'graph-theory': 5.0,
```
Ankitsinghsisodya, line 55: 🔵 **nit:** `SAMPLE_TAG_VOTES_TIE_AT_PROPORTION` is a misleading name. `graph-theory` has exactly 25% of the max (5 / 20 = 0.25), which is a minimum-proportion boundary case — not a "tie" between the two tags. A tie would imply both have the same vote count.

**Rename to:** `SAMPLE_TAG_VOTES_AT_MIN_PROPORTION` to make the intent clear.

stuff/cron/tests/test_aggregate_feedback.py
```diff
@@ -0,0 +1,202 @@
+'''Unit tests for the pure functions in `aggregate_feedback.py`.'''
+# pylint: disable=wrong-import-position
+import os
+import sys
+import unittest
+
+THIS_DIR = os.path.dirname(os.path.realpath(__file__))
+sys.path.insert(0, os.path.dirname(THIS_DIR))
```
Ankitsinghsisodya: 🔵 **nit:** `sys.path.insert` boilerplate is duplicated in every test file. Extract it to a `conftest.py` at `stuff/cron/` — pytest loads it automatically before test collection, so each test file can just `import aggregate_feedback` directly.

stuff/cron/tests/test_aggregate_feedback.py
```diff
@@ -0,0 +1,202 @@
...
+sys.path.insert(0, THIS_DIR)
+
+import aggregate_feedback
+from fixtures import sample_data  # type: ignore[import]
```
Ankitsinghsisodya: 🔵 **nit:** `# type: ignore[import]` here is a symptom of the missing `conftest.py` / path setup. Fix the import mechanism instead of silencing mypy.

stuff/cron/tests/test_aggregate_feedback.py
```diff
@@ -0,0 +1,202 @@
...
+        '''With no global prior, no per-problem average can be produced.'''
+        votes = sample_data.make_quality_votes([0, 0, 0, 0, 20])
+        self.assertIsNone(aggregate_feedback.bayesian_average(None, votes))
+
+    def test_returns_none_when_weighted_n_below_confidence(self) -> None:
+        '''Below CONFIDENCE (10) the sample is not yet trusted.'''
+        votes = sample_data.make_quality_votes([0, 0, 5, 0, 0])
+        self.assertIsNone(
+            aggregate_feedback.bayesian_average(3.0, votes))
+
```
Ankitsinghsisodya, line 21: ❓ **missing test:** `bayesian_average(0.0, votes)` where `apriori_average` is `0.0` (falsy but not `None`) is not covered. The current guard is `apriori_average is None` (correct), but there is no test that would catch a future regression to `if not apriori_average`. Consider adding:
```python
def test_returns_value_when_apriori_is_zero(self) -> None:
    votes = sample_data.make_quality_votes([0, 0, 10, 0, 0])
    result = aggregate_feedback.bayesian_average(0.0, votes)
    self.assertIsNotNone(result)
```

stuff/cron/tests/test_aggregate_feedback.py
```diff
@@ -0,0 +1,202 @@
...
+        result = aggregate_feedback.get_weighting_factor(
+            None,
+            sample_data.SAMPLE_RANK_CUTOFFS,
+            aggregate_feedback.WEIGHTING_FACTORS,
+        )
+        self.assertEqual(
+            result,
+            aggregate_feedback.WEIGHTING_FACTORS['user-rank-unranked'],
+        )
+
```
Ankitsinghsisodya, line 76: ❓ **missing test:** `get_weighting_factor` with a score below every cutoff (e.g. `-1.0`) is not tested. It should fall through the entire loop and return the unranked weight. Worth adding to guard the fallback path.

stuff/cron/tests/test_aggregate_feedback.py
```diff
@@ -0,0 +1,202 @@
...
+            result,
+            aggregate_feedback.BEFORE_AC_WEIGHTING_FACTORS[
+                'user-rank-international-master'],
+        )
+
+
+class GetMostVotedTagsTest(unittest.TestCase):
+    '''Tests for `aggregate_feedback.get_most_voted_tags`.'''
+
+    def test_returns_none_below_min_points(self) -> None:
```
Ankitsinghsisodya, line 179: 🔴 **bug in source (surfaced by missing test):** `get_most_voted_tags({}, n)` with an empty dict will crash — `max()` on an empty iterable raises `ValueError`. There is no guard in the source. A test like:
```python
def test_empty_votes_returns_none(self) -> None:
    self.assertIsNone(aggregate_feedback.get_most_voted_tags({}, 0))
```
would reveal this. The fix in the source should guard with `if not problem_tag_votes: return None` before the `max()` call.

stuff/cron/tests/test_utils.py
```diff
@@ -0,0 +1,121 @@
...
+        '''January 1st rolls forward to February 1st.'''
+        self._assert_next(
+            datetime.date(2026, 1, 1),
+            datetime.date(2026, 2, 1),
+        )
+
+    def test_february_non_leap_year(self) -> None:
+        '''February 1st rolls forward to March 1st in a non-leap year.'''
+        self._assert_next(
+            datetime.date(2026, 2, 1),
```
Ankitsinghsisodya, line 32: 🔵 **nit:** `test_february_leap_year` adds no meaningful coverage. `get_first_day_of_next_month` only reads `.month` and `.year` — the day value and leap-year status are irrelevant to its logic. The test is identical in behaviour to `test_february_non_leap_year`.

**Suggestion:** replace (or supplement) with a mid-month input to verify the output day is always reset to 1:
```python
self._assert_next(datetime.date(2024, 2, 15), datetime.date(2024, 3, 1))
```

stuff/cron/tests/fixtures/mock_cursor.py
PrasannaMishra001: switched both sides to _normalize(...)

stuff/cron/tests/fixtures/mock_cursor.py
PrasannaMishra001: added an internal position so fetchone/fetchall/iteration all advance and drain like MySQLCursor.

stuff/cron/tests/fixtures/mock_cursor.py
PrasannaMishra001: I kept the flag for signature parity and documented that result shape is decided by the scripted rows (tuples or dicts).Raising would block the upcoming assign_badges/update_ranks tests that use dictionary=True, so I went with explicit docs instead.

stuff/cron/tests/fixtures/mock_cursor.py
PrasannaMishra001: added in this only,well it was part of upcoming PR

stuff/cron/tests/fixtures/sample_data.py
PrasannaMishra001: used option 1 i.e. qualified from cron import ... imports, same as the existing test_plagiarism_detector.py

stuff/cron/tests/test_aggregate_feedback.py
PrasannaMishra001: I aligned the path setup with the existing test_plagiarism_detector.py (add `stuff/ `to the path, `import cron.*`) rather than adding a `conftest.py`, since the repo doesn't use one and a conftest wouldn't help mypy resolution. Open to a conftest later if we standardize on it across all stuff/ tests.

stuff/cron/tests/test_aggregate_feedback.py
PrasannaMishra001: added

stuff/cron/tests/test_aggregate_feedback.py
PrasannaMishra001: the ({}, 0) example actually returns None via the MIN_POINTS check before reaching max() and the real caller builds the dict and counter together so n>=10 always implies a non empty dict (so it is not reachable in production). Still added the guard with a test at n=MIN_POINTS.

stuff/cron/tests/test_utils.py
PrasannaMishra001: Replaced it with a mid month input that verifies when day resets to 1st

stuff/cron/conftest.py
```diff
@@ -0,0 +1,15 @@
...
+cron modules (`cron.*`, `lib.*`) and the scripts' own sibling imports
+(`database.*`, `utils`) without repeating the path boilerplate in every file.
+'''
+import os
+import sys
+
+_CRON_DIR = os.path.dirname(os.path.realpath(__file__))
+_STUFF_DIR = os.path.dirname(_CRON_DIR)
+
+for _path in (_STUFF_DIR, _CRON_DIR):
```
pabo99: `conftest.py` should add only stuff to `sys.path`. Adding _cron_ exposes `cron/database` as the top-level `database` package and breaks the existing _pipelines_ tests that import `database.contest`. The cron modules should use `cron.database.*` and `cron.utils` imports instead.

pabo99 (commented):
One last comment and that's good to me

stuff/cron/conftest.py
PrasannaMishra001: thanks, now i kept only stuff in `conftest.py` and switching the cron imports to `cron.database.* `/ `cron.utils` which avoids the db pkg collision. also updated the related test imports to use from `cron import ...`, verified the pipelines tests as well.

PrasannaMishra001:
@pabo99 fixed it, conftest adds only `stuff` now and the cron modules import `cron.database.*` and `cron.utils`.
it was not breaking yet actually as pytest collects cron before pipelines so pipelines still worked but worth fixing.

pabo99 (approved):
Thanks!

PR #9889, opened 2026-06-07
feat(cron): add update_ranks unit tests

### LINKED ISSUE #9888: [FEATURE] Add unit tests for update_ranks.py ranking logic
### Checklist

### Context

`update_ranks.py` runs every day to build the user rankings, Coder of the Month, School of the Month and the rank cutoffs. It is the biggest cron script and right now it has no Python unit tests, so a mistake in the scoring or sorting could go in without anyone noticing.

This is a follow up to #9882. It adds `stuff/cron/tests/test_update_ranks.py` with tests for the pure logic in the script:-

- `compute_points_for_user` and `compute_points_for_school` - scoring, sorting and the empty cases.
- `update_user_rank_cutoffs` - picking the right score at each percentile.
- `update_user_rank_classname` - the cutoff based update.

The database calls are mocked so the tests run without a real database and the existing pytest CI job already picks them up, so no workflow change is needed.

PrasannaMishra001:
# Description

This adds `stuff/cron/tests/test_update_ranks.py` with tests for `update_ranks.py` cron script, extends on the test infra in #9883:-

- `compute_points_for_user` - scoring, descending sort, truncation to the coder list count, the empty users and empty problems cases and that the last 12 winners are forwarded to the eligibility query.
- `compute_points_for_school` - scoring a school on the distinct problems its users solved, excluding the last-12 winning schools, sorting and empty cases.
- `update_user_rank_cutoffs` - the percentile index selection at the 1/9/15/35/40% cutoffs and the empty scores early return.
- `update_user_rank_classname` - cutoff based update.

The database helpers are mocked (so requires no database). Also, existing python CI job runs pytest under `./stuff/`, so no workflow change needed.

It also changes how the cron scripts import their siblings. `update_ranks.py` and `database/coder_of_the_month.py` imported bare `database.*` and `utils`, which only resolves when `stuff/cron` itself is on the path, and putting it there makes the cron `database` package shadow the unrelated one under `stuff/pipelines`. They now import through the `cron.*` package instead and `conftest.py` only adds `stuff/`, so importing a cron module from a test cannot shadow anything. Runtime behaviour is unchanged, the scripts still run the same way.

# Comments

Builds on #9883 (test infrastructure), so until that merges the diff carries it too. A rebase after it merges leaves `test_update_ranks.py` and the import change.

Fixes #9888.

# Checklist:


stuff/cron/tests/test_update_ranks.py
```diff
@@ -0,0 +1,368 @@
...
+
+
+class UpdateUserRankClassnameTest(unittest.TestCase):
+    '''Tests for `update_ranks.update_user_rank_classname`.'''
+
+    def test_executes_single_classname_update(self) -> None:
+        '''The function issues one UPDATE against User_Rank.'''
+        cur = MockCursor()
+
+        update_ranks.update_user_rank_classname(
```
pabo99, line 524: Could this test verify the actual ranking classification behavior instead of only checking that an `UPDATE` statement was executed? In particular, it would be useful to cover the cutoff comparison and the user-rank-unranked fallback. A malformed query could currently pass this test

stuff/cron/tests/test_update_ranks.py
```diff
@@ -0,0 +1,368 @@
...
+
+
+class ComputePointsForUserTest(_PatchHelpersMixin):
+    '''Tests for `update_ranks.compute_points_for_user`.'''
+
+    def test_scores_users_and_sorts_descending(self) -> None:
+        '''Each user's score is the sum of the problems they solved.'''
+        user_a = _user(1)
+        user_b = _user(2)
+        self._patch(
```
pabo99, line 204: Since `get_user_problems` is mocked here, these tests don't cover its filtering rules: forfeited problems, problems administered by the user, first accepted submission within the month, or deduplication by user and problem. Could you add focused tests for that helper, or one integration-style test using representative cursor rows?

stuff/cron/tests/test_update_ranks.py
```diff
@@ -0,0 +1,368 @@
...
+
+        result = update_ranks.compute_points_for_school(
+            _cursor(), _CURRENT_MONTH, _NEXT_MONTH)
+
+        self.assertEqual([s.school_id for s in result], [20, 10])
+
+    def test_returns_empty_when_no_eligible_schools(self) -> None:
+        '''No candidate schools yields no ranking.'''
+        self._patch(
+            get_last_12_schools_of_the_month=[],
```
pabo99, line 394: Could you add the empty-eligible-problems case for `compute_points_for_school`? The function has a dedicated early return for this condition, but it's currently covered only for users

stuff/cron/tests/test_update_ranks.py
PrasannaMishra001: replaced that test with three that run the real `UPDATE` against sqlite and read the classnames back: band  assignment, the exact match at 1900, and the unranked fallback.

stuff/cron/tests/test_update_ranks.py
PrasannaMishra001: unmocked it now. 7 tests now drive the real function over seeded rows, covering the month window with both boundaries, forfeited, administered and dedup as u mentiioned,

it also filters `verdict = 'AC'` and `type = 'normal'`, which your list missed, so those are covered too.

stuff/cron/tests/test_update_ranks.py
PrasannaMishra001: added and removing the early return fails it as i checked.
,as sql runs on sqlite, so these verify semantics rather than syntax .

PrasannaMishra001:
@pabo99 did the changes, now u may check when free.

pabo99 (approved):

PR #9903, opened 2026-06-12
fix(cron): merge user rank with upsert instead of full reinsert

### LINKED ISSUE #9902: [FEATURE] Avoid full delete and reinsert of User_Rank in update_ranks
### Checklist

### Context

`update_user_rank` in `stuff/cron/update_ranks.py` rebuilds the ranking every day by running `DELETE FROM User_Rank;` and reinserting every row. Readers are safe because the whole phase runs in one transaction, but the full rewrite still costs us: undo log entries for every row, binlog traffic for the whole table twice a
day, and row locks held for the entire run.

Since `User_Rank` has `user_id` as primary key and only this cron writes to it, the rebuild can instead stage the fresh ranking in a temporary table and merge it with `INSERT ... ON DUPLICATE KEY UPDATE` (the same pattern `update_author_rank` already uses), plus a targeted delete for users that fell out of the ranking.
Unchanged rows become no-op updates that InnoDB skips, so the daily write churn becomes proportional to the rows that actually changed instead of the whole table. Temporary tables do not break the transaction, so atomicity is unchanged.

I looked at the shadow-table swap (`RENAME TABLE`) as an alternative, but it is DDL, so it would implicitly commit in the middle of the phase and lose the atomicity, and it needs extra privileges plus special handling for the foreign
key names on the swapped table. The upsert keeps everything as plain DML.

PrasannaMishra001:
# Description

`update_user_rank` rebuilt the ranking every day with `DELETE FROM User_Rank;` followed by reinserting every row. This replaces the destructive delete-and-reinsert with an in place merge, the same `INSERT ... ON DUPLICATE KEY UPDATE` pattern `update_author_rank` already uses on this table a few lines below, so the two are consistent instead of doing the same job two different ways.

How it works now:
- the run notes the current database time before writing anything.
- one targeted delete removes rows whose username has since moved to a different user, joining `Identities`, since `username` is unique on `User_Rank` and the merge would otherwise match the wrong row through that key.
- the freshly computed rows are merged in batches with `ON DUPLICATE KEY UPDATE`, which refreshes `timestamp` on every row it writes.
- anything left with a `timestamp` older than the start of the run is a user who fell out of the ranking, so a single delete sweeps them.

This avoids the full table delete undo, the primary key and secondary index rebuild, and the brief window where the table is empty, and roughly halves the row operations (an update per row instead of a delete plus an insert).

Notes:
- the merge refreshes `timestamp` deliberately, the rank page surfaces the top user's timestamp as "last updated", and the sweep uses it to find stale rows.
- it also clears the author columns, since the delete it replaces did that implicitly and `update_author_rank` recomputes them right after in the same transaction.

An earlier version of this PR staged the rows in a temporary table. That is gone, it needed `CREATE TEMPORARY TABLES` on the production cron user, and mysql does not allow `CREATE TEMPORARY TABLE` inside a transaction when GTIDs are on, which matters here because the phase runs in one transaction so the ranking can roll back. The version above is plain DML and needs no permission change.

# Comments

Verified against the dev database:
- Equivalence: the ranking columns (score, ranking, classname, author columns) are identical between this version and main.
- Idempotency: running the script twice leaves the ranking data identical; only `timestamp` advances, same as before.
- A user who drops out of the ranking is removed by the sweep.
- A username moving from one user to another is handled without a duplicate key error, both users end up with the right row.
- Lint passes.

Fixes #9902.

# Checklist:


pabo99 (commented):
Could you confirm that the production cron DB user has the required `CREATE TEMPORARY TABLE` permission?

PrasannaMishra001:
@pabo99 i checked and i cannot see it from my side, the cron credentials come from `cronjobs-secret` which is a sops encrypted `my.cnf` and the grants live in RDS, so the only `GRANT` statements in the prod repo are the dev and sandbox ones. could you check what the prod cron user has? if it is missing it would need `GRANT CREATE TEMPORARY TABLES ON omegaup.* TO '<cron user>'@'%'`.

that said i would rather not depend on it at all, so i have pushed a version without the temporary table. `update_author_rank` in this same file already does the same merge with a batched `INSERT ... ON DUPLICATE KEY UPDATE` straight into `User_Rank`, so this just makes the two consistent. and mysql does not allow `CREATE TEMPORARY TABLE` inside a transaction when GTIDs are on, which matters here because the phase runs in one transaction so the ranking can roll back. the diff is a lot smaller now too.

pabo99 (approved):

PR #9906, opened 2026-06-15
fix: guard FinderWizard ESC listener to only fire when visible

### LINKED ISSUE #9905: bug: FinderWizard ESC keydown listener persists when modal is hidden via v-show
### Description

A document-level 'keydown' listener is added in mounted() and removed in beforeDestroy(). However, FinderWizard is rendered with v-show (not v-if) in all parent components, so the component is never destroyed when the modal is closed. The listener lives forever and fires for every keydown event on the page.

### Steps to reproduce

1. Open `frontend/www/js/omegaup/components/problem/FinderWizard.vue` around line 171
2. When the wizard is hidden and the user presses Escape, onEsc() fires and emits 'close'. The event is consumed unnecessarily. More critically, the listener is never cleaned up because beforeDestroy() is never reached with v-show semantics.

### Expected behavior

v-show hides the element via CSS display:none but keeps the component instance mounted. mounted() fires once on initial page load adding the listener. beforeDestroy() only fires on component destruction, which never happens. The onEsc handler has no guard to check whether the modal is currently visible before emitting 'close'. should be handled correctly.

### Actual behavior

When the wizard is hidden and the user presses Escape, onEsc() fires and emits 'close'. The event is consumed unnecessarily. More critically, the listener is never cleaned up because beforeDestroy() is never reached with v-show semantics.

deepak0x:
Fixes #9905

## Problem
The FinderWizard component adds a document-level `keydown` listener in `mounted()` and removes it in `beforeDestroy()`. However, the component is rendered with `v-show` (not `v-if`) in all parent components, so it's never destroyed when the modal is closed. The listener lives forever and fires for every keydown event on the page.

## Solution
This PR adds a `show` prop to FinderWizard and guards the `onEsc()` handler to only emit `close` when the modal is actually visible (`this.show === true`). The listener still persists (due to v-show semantics) but now has no effect when the modal is hidden.

## Changes
- Add `show` prop to FinderWizard.vue (Boolean, default false)
- Guard `onEsc()` to check `this.show` before emitting `close`
- Pass `:show="showFinderWizard"` from List.vue and Collection.vue

## Testing
1. Open problem list page
2. Click wizard button to open FinderWizard modal
3. Press ESC → modal closes 
4. Press ESC again when modal is hidden → no effect (previously would emit unnecessary close event)


pabo99 (commented):
The change looks good!

Could you add a test covering `Escape` behavior when the wizard is both visible and hidden? Please call `wrapper.destroy()` afterward to remove the global `document` listener and avoid leaking it into other tests

pabo99 (approved):
Thanks!

deepak0x:
The cypress (ide-basics) failure is unrelated — it timed out in basic_commands.cy.ts on the "create two contests and merge their scoreboard" test:

```
Timed out retrying after 4000ms:
Expected to find element: `a[data-problem="bad98418-f"]`, but never found it.
   at commands.ts:284 (createRunsInsideContest)
```

The problem link just never rendered within the 4s window, so it's flaky. This branch doesn't touch any cypress code or the scoreboard flow anyway (just FinderWizard + lang files), and everything else was green.

Pushed an empty commit to re-run CI.

pabo99:
In cases like this, where the failure is caused by a flaky test, there's no need to rerun all the jobs—rerunning the failed one is enough. Thanks for pushing the empty commit anyway!

deepak0x:
> In cases like this, where the failure is caused by a flaky test, there's no need to rerun all the jobs—rerunning the failed one is enough. Thanks for pushing the empty commit anyway!

okk thanks to let me know ;)

PR #9914, opened 2026-06-19
add structured phase logging to crons

### LINKED ISSUE #9913: [FEATURE] Emit structured per-phase logs from cron scripts
### Checklist

### Context

The cron scripts log what they did at the end of a run, but only as message strings like `attempted=12 successful=11 failed=1`. A log tool cannot chart or alert on those counts without parsing them back out of the message text.

This emits the same numbers as structured fields alongside the existing message, so the line still reads normally in a terminal and the fields are queryable with `--log-json`. `aggregate_feedback` gets `problems_attempted`, `problems_successful`, `problems_failed` and `duration_ms`, and `assign_badges` gets `badges_total`, `badges_successful` and `badges_failed`.

### Note on scope

This originally also proposed a `log_phase` helper wrapping every phase of every cron. That part is no longer needed: #9998 merged the shared runner, and `CronRun.phase()` already emits `phase`, `status` and `duration` as structured fields and additionally stores them in `Cron_Runs`, which the logging-only helper never did. Only the end of run summaries are left here.

Nothing about error handling or exit codes changes.

PrasannaMishra001:
# Description

the crons already log what they did at the end of a run, but as message strings like `attempted=12 successful=11 failed=1`, so nothing can chart or alert on those counts without parsing the text back out. this emits the same numbers as structured fields alongside the message, so the message stays readable in a terminal and the fields are queryable with `--log-json`.

`aggregate_feedback` now carries `problems_attempted`, `problems_successful`, `problems_failed` and `duration_ms`, and `assign_badges` carries `badges_total`, `badges_successful` and `badges_failed`.

Fixes: #9913

# Comments

this started out much larger. it added a `log_phase` helper and wrapped every phase of every cron in it, but #9998 merged in the meantime and `CronRun.phase()` already does that, and it also writes the timings into `Cron_Runs` rather than only into the logs. so the phase wrapping here was doing the same job twice and i dropped it. what is left is the part nothing else covers, which is the end of run summaries.

it stacks on #10009 and #10022 because those are the ones that adopt the runner in these two scripts, so the diff carries them until they merge and then shrinks to just the two logging changes.

# Checklist:


stuff/cron/assign_badges.py
```diff
@@ -138,8 +146,9 @@ def main() -> None:
         with dbconn.cursor(buffered=True,
                            dictionary=True) as cur, dbconn_readonly.cursor(
                                buffered=True, dictionary=True) as cur_readonly:
-            has_failures = process_badges(args.current_timestamp, dbconn,
-                                          cur, cur_readonly)
+            with lib.logs.log_phase('process_badges'):
```
pabo99: Could you make the `process_badges` phase report `status: failed` when `process_badges` returns `True`? This function handles badge errors internally instead of raising, so `log_phase` currently emits `status: ok` even when some badges failed and the process later exits with code 1. The phase event and the final summary would then be consistent

PrasannaMishra001: this is done, though it got merged in #10009 rather than here. the merged runner has mark_failure() for exactly this case and assign_badges calls it when process_badges returns true, so the run is recorded as failed and not just logged.

i also dropped log_phase from this pr, since `CronRun.phase()` already emits phase,status and duration and writes them to Cron_Runs too. what is left here is only the end of run summary fields.

pabo99 (approved):
Thanks!

PR #9919, opened 2026-06-21
add view unavailable component

### LINKED ISSUE #9897: [FEATURE] Add a generic "View unavailable" component
### Checklist

### Context

Right now, when a part of the app is turned off, each place shows its own off message. For example, when an admin disables the EphemeralGrader IDE from the site settings, the grader page renders its own "disabled" alert.

It would be cleaner to have a single reusable component that shows a consistent "this view is currently unavailable" state across the app, controlled by a flag coming from the server (similar to how the common payload already arries things like the maintenance message). Any feature that can be turned off (ephemeral grader, maintenance, future feature flags) could then reuse it instead of writing its own message.

This came up in the system settings work (#8906 / #8938). This issue tracks building a generic component as a followup so behaviour looks consistent .

PrasannaMishra001:
Fixes #9897

custom-message
default

Follow-up from the system settings work (#8906 / #8938). A generic "View unavailable" component that can be controlled from the server and rendered consistently across the app, instead of each feature writing its own off message.

This adds `ViewUnavailable.vue` in `components/common`. It shows a consistent placeholder (icon, title and an optional detail) for any part of the app that is turned off, for example the EphemeralGrader IDE, maintenance or a future feature flag. The parent decides when to render it based on a flag from the common payload.

Props:
- `title` defaults to the new `viewUnavailableTitle` string
- `description` optional detail line
- `icon` Font Awesome name, defaults to `ban`

A default slot is also available .

Includes a Storybook story and a unit test. The shared string is added to the lang files. Adoption (the EphemeralGrader using this instead of its own alert) can follow once #8938 merged.


pabo99 (commented):
The change looks good:


Image

Just a couple of minor suggestions:

- Could you wrap it inside a card?
- Would it be possible to center it vertically as well?

With those two changes, we can go ahead and merge it.

PrasannaMishra001:
custom-message

also verified by wiring it up, it will look like :-
image

pabo99 (approved):
Great!

PR #9920, opened 2026-06-21
feat(cron): add assign_badges unit tests

### LINKED ISSUE #9890: [FEATURE] Add unit tests for assign_badges.py
### Checklist

### Context

`assign_badges.py` has no unit tests right now, so a mistake in who gets a badge or a duplicate notification, could go in without anyone noticing.

There is also a small thing in `get_all_owners` worth fixing at the same time. When `--current-timestamp` is passed, the script builds the query by substituting the formatted timestamp into the SQL text:

```python
query = query.replace(
    'NOW()', f"'{current_timestamp.strftime('%Y-%m-%d %H:%M:%S')}'")
```

That is string concatenation into a query, which is the one thing the other cron scripts avoid. The value can instead be bound to a session variable and the query can reference that:

```python
cur_readonly.execute('SET @current_time = %s;', (current_timestamp,))
query = query.replace('NOW()', '@current_time')
```

`NOW()` is a bare function call inside the badge queries, so it cannot become a `%s` placeholder without editing every `query.sql`. The session variable gets the value out of the query text and leaves the badge queries untouched. Only `coderOfTheMonth` and `legacyUser` use `NOW()` today.

The fix and the tests belong in the same change because the tests are what pin the new behaviour down. Without them nothing stops the string substitution coming back, and the reviewer has no way to see that the timestamp is bound rather than spliced. So this issue covers both:

- change `get_all_owners` to bind the timestamp to a session variable instead of substituting it into the SQL text.
- add `stuff/cron/tests/test_assign_badges.py` covering:
  - `get_all_owners` - reading the owner ids from a badge query, binding the timestamp to `@current_time` when one is given, and leaving `NOW()` alone when one is not.
  - `get_current_owners` - reading the users who already own the badge.
  - `save_new_owners` - inserting the badge rows and the notification rows.
  - `process_badges` - saving only the new owners and rolling back and carrying on if one badge fails.

This is a follow up to #9882. No workflow change is needed.

PrasannaMishra001:
# Description

two things that belong together, a small fix to `assign_badges.py` and the unit tests that pin it down.

the fix is in `get_all_owners`. when `--current-timestamp` is passed, the script used to build the query by substituting the formatted timestamp into the sql text:

```python
query = query.replace(
    'NOW()', f"'{current_timestamp.strftime('%Y-%m-%d %H:%M:%S')}'")
```

that is string concatenation into a query, which is the one thing we try not to do anywhere else in the cron scripts. it now binds the value to a session variable and references that instead:

```python
cur_readonly.execute('SET @current_time = %s;', (current_timestamp,))
query = query.replace('NOW()', '@current_time')
```

`NOW()` is a bare function call inside the badge queries, so it cannot be a `%s` placeholder in `cur.execute(query)` without editing every `query.sql`. a session variable gets the value out of the query text with no change to the badge queries at all.

two badges use `NOW()` today, `coderOfTheMonth` and `legacyUser`. i checked that the two versions pick the same owners rather than assuming it. i ran both the old code and the new code against the dev database with the same fixed timestamps, over all 23 badges and 8 timestamps chosen to sit either side of the rows in `Coder_Of_The_Month` and either side of the `YEAR(NOW()) - 2` cut off in `legacyUser`, plus the no timestamp path. the owner sets matched in all 207 comparisons. to make those probes actually discriminate i seeded extra rows and spread the activity years inside a transaction that was rolled back at the end, so the database is unchanged.

the tests are in `stuff/cron/tests/test_assign_badges.py`. they cover `get_all_owners` reading owner ids from a badge query, binding the timestamp when one is given and leaving `NOW()` alone when one is not, `get_current_owners` reading the users who already own the badge and passing the alias as a query parameter, `save_new_owners` inserting the badge rows and the notification rows, and `process_badges` saving only the new owners, rolling back and carrying on when one badge fails, and reporting the right totals in its run summary.

i mutation tested every behaviour the tests claim to cover, twenty mutants in all, and each one was caught. the ones worth naming: putting back the old string substitution fails `test_binds_timestamp_to_a_session_variable`, so does building the `SET` with an f-string instead of binding with `%s`, issuing the `SET` unconditionally fails `test_keeps_now_when_no_timestamp_is_given`, splicing the badge alias into the `get_current_owners` query instead of binding it fails `test_filters_by_the_badge_alias`, and breaking either the counters or the structured fields in the run summary fails `test_summary_counts_the_badges_it_processed`.

the database helpers are mocked, so the tests need no database. the existing python ci job already runs pytest under `./stuff/`, so no workflow change is needed.

Fixes: #9890

# Comments

this was stacked on #9883 and #9889. both are merged now, so it stands on its own and its diff is just `stuff/cron/assign_badges.py` and `stuff/cron/tests/test_assign_badges.py`.

the binding fix is one of the items from my gsoc proposal. it fits here rather than in its own pr because the tests in this pr are what pin the new behaviour down, and without them nothing stops the old substitution coming back.

the run summary assertions cover the structured fields added in #9914, which merged without anything pinning them.

# Checklist:


pabo99:
Could you please resolve the conflicts so we can review this PR?

PrasannaMishra001:
@pabo99 done, you can directly review this, or go in sequence https://github.com/omegaup/omegaup/pull/9883 , https://github.com/omegaup/omegaup/pull/9889 as this is on top.  merging this then i will simply close them as complete.

pabo99 (approved):

PR #9921, opened 2026-06-22
added delete confirmation dialog

### LINKED ISSUE #9818: [FEATURE] Add confirmation dialog for "Delete Problem" action
## Checklist

Image

Hey @pabo99, what do you think about adding a confirmation alert when deleting a problem? It might help prevent users from accidentally removing data.
Since it’s a pretty well-defined task, we could also tag it as a 'good first issue' to help onboard new contributors!

pratham15541:
# Description
Added a confirmation dialog before deleting a contest problem statement to prevent accidental deletions caused by misclicks.
Fixes: #9818


frontend/templates/en.lang
```diff
@@ -310,6 +310,7 @@ contestAddproblemChooseVersion = "Version"
 contestAddproblemContestOrder = "Contest order"
 contestAddproblemLatestVersion = "Latest version"
 contestAddproblemOtherVersion = "Other version"
+contestAddproblemProblemConfirmRemove = "Do you really want to remove %(problem)?"
```
pabo99, line 313: nit:
```suggestion
contestAddproblemProblemConfirmRemove = "Do you really want to remove "%(problem)"?"
```
Also in the other languages

frontend/www/js/omegaup/contest/edit.ts
```diff
@@ -385,6 +385,18 @@ OmegaUp.on('ready', () => {
...
+                ?.title ?? problemAlias;
+            if (
+              !window.confirm(
+                ui.formatString(T.contestAddproblemProblemConfirmRemove, {
+                  problem: problemTitle,
+                }),
+              )
+            ) {
+              return;
+            }
```
pabo99, line 399: Could we move this logic into the component?

Also,  I'd actually prefer using a `b-modal`. We should already have a few examples of its usage elsewhere in the codebase

PR #9900, opened 2026-06-12
fix(cron): parameterize coder_of_the_month queries

PrasannaMishra001:
# Description

Some queries in `stuff/cron/database/coder_of_the_month.py` were built by pasting values directly into the SQL string. This parameterizes them:

- `get_cotm_eligible_users` - the last 12 winners' usernames now go through `%s` placeholders instead of an f-string. Usernames are user-controlled text, so this closes a small injection surface.
- `get_user_problems` and `get_problems_admins` - now take the id lists and bind them as parameters instead of formatting them into the `IN (...)` clauses. The two callers in `update_ranks.py` got simpler since they no longer build the comma-separated strings.
- While touching the query, changed the admin exclusion from `NOT IN (subquery)` to `NOT EXISTS`, which avoids materializing the subquery per row.

No behavior change. I ran `update_ranks.py` against the dev database before and after the change and the resulting `User_Rank`, `User_Rank_Cutoffs`, `Coder_Of_The_Month` and `Schools` contents are identical.

# Comments

The unit tests from #9889 cover the `compute_points_for_user` / `compute_points_for_school` paths that call into these functions.

Fixes #9899.

# Checklist:

- [x] The code follows the [coding guidelines](https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/Coding-guidelines.md) of omegaUp.
- [x] The tests were executed and all of them passed.
- [x] If you are creating a feature, the new tests were added.
- [x] If the change is large (> 200 lines), this PR was split into various Pull Requests.

PR #9940, opened 2026-06-26
refactor(cron): name acl/role magic numbers with constants

### LINKED ISSUE #9939: [FEATURE] Replace acl_id/role_id magic numbers in cron scripts with named constants
### Checklist

### Context

Several cron queries hardcode `acl_id = 1` and `role_id = 1` with -- TODO: Replace magic numbers with constants and aggregate_feedback.py defines QUALITYNOMINATION_QUESTION_CHANGE_ID inline. These mirror Authorization::SYSTEM_ACL and Authorization::ADMIN_ROLE in the PHP backend but are not named on the Python side.

The goal was to add a shared stuff/cron/constants.py (`SYSTEM_ACL`, `ADMIN_ROLE`,` QUALITYNOMINATION_QUESTION_CHANGE_ID`) and adopt it in `update_ranks.py` and `aggregate_feedback.py`, parameterizing the admin exclusion subqueries. Behavior to be unchanged (i.e. values stay 1). coder_of_the_month.py will adopt it incrementally after #9900 (which restructures those same lines) merges.

PrasannaMishra001:
# Description

This adds a small `stuff/cron/constants.py` module that gives names to the magic numbers the cron scripts were using directly inside their SQL. The site-admin exclusion `acl_id = 1` and `role_id = 1` become `SYSTEM_ACL` and `ADMIN_ROLE`, which mirror the existing `Authorization::SYSTEM_ACL` and `Authorization::ADMIN_ROLE` constants on the PHP side, and the `QUALITYNOMINATION_QUESTION_CHANGE_ID` that was defined inline in `aggregate_feedback.py` now lives in the same shared module. In `update_ranks.py` the two admin exclusion subqueries now receive these values as query parameters instead of hardcoded literals, so the intent reads clearly from the code and there is a single place to change if the values ever move.

There is no behavior change as the values are still `1`. I verified this by running `update_ranks.py` against the seeded database before and after the change and the `User_Rank` rows come out identical with the site administrator still correctly excluded from the ranking. `aggregate_feedback.py` also runs end to end and the full cron unit test suite passes.

I intentionally left `coder_of_the_month.py` out of this PR, because its admin exclusion lines are already being restructured in #9900, so adopting the constants there once that merges will avoid conflict.

Fixes: #9939

# Checklist:


pabo99 (approved):
Thanks!!

PR #9951, opened 2026-06-29
Fix FinderWizard emitting close on ESC while hidden

### LINKED ISSUE #9922: bug: ESC key emits 'close' from hidden FinderWizard
### Description

The onEsc handler emits 'close' on every ESC keypress regardless of whether the wizard is currently visible. Because the component is used with v-show (not v-if) in both Collection.vue and List.vue, it stays mounted and its document-level keydown listener stays active even when hidden.

### Steps to reproduce

1. Open `frontend/www/js/omegaup/components/problem/FinderWizard.vue` around line 179
2. Pressing ESC anywhere on the page while showFinderWizard is false (wizard hidden) still triggers this.$emit('close') inside the hidden wizard, causing unintended close events that can interfere with other ESC-key consumers on the same page.

### Expected behavior

onEsc only checks event.key === 'Escape' with no guard on component visibility. The component receives no 'show' prop and has no way to know it is hidden behind v-show. should be handled correctly.

### Actual behavior

Pressing ESC anywhere on the page while showFinderWizard is false (wizard hidden) still triggers this.$emit('close') inside the hidden wizard, causing unintended close events that can interfere with other ESC-key consumers on the same page.

deepak0x:
# Description

The `FinderWizard` component registers a document-level `keydown` listener and emits `close` on ESC. Because `Collection.vue` and `List.vue` render it with `v-show` (not `v-if`), it stays mounted while hidden, so pressing ESC anywhere on the page fired a stray `close` event and interfered with other ESC handlers.

This adds a `show` prop and guards the handler so `close` is only emitted when the wizard is actually visible. The parents bind `:show="showFinderWizard"` alongside the existing `v-show`.

Fixes: #9922

# Checklist:

### DIFF
--- a/frontend/www/js/omegaup/components/problem/Collection.vue
+++ b/frontend/www/js/omegaup/components/problem/Collection.vue
@@ -120,6 +120,7 @@
             <!-- TODO: Migrar el problem finder a BS4 (solo para eliminar algunos estilos) -->
             <omegaup-problem-finder-wizard
               v-show="showFinderWizard"
+              :show="showFinderWizard"
               :possible-tags="allTags"
               @close="showFinderWizard = false"
               @search-problems="$emit('search-problems', $event)"
--- a/frontend/www/js/omegaup/components/problem/FinderWizard.test.ts
+++ b/frontend/www/js/omegaup/components/problem/FinderWizard.test.ts
@@ -1,0 +1,35 @@
+import { mount } from '@vue/test-utils';
+
+import problem_FinderWizard from './FinderWizard.vue';
+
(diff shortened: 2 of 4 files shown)


pabo99 (commented):
Could you update this branch? It looks like these changes have already been merged through another PR.

deepak0x:
Thanks @pabo99, you're right — this was already merged via #9906, which fixed the same bug (filed as the duplicate issue #9905). After syncing with `main` this branch has no unique changes left, so closing as superseded. The fix for #9922 is now in `main`.

PR #9964, opened 2026-07-04
fix: store RedisCacheAdapter::incWithTTL counter in serialized format

### LINKED ISSUE #9956: bug: Redis incWithTTL stores raw integers incompatible with serialized storage
### Description

RedisCacheAdapter::incWithTTL() uses Redis INCR which stores raw integer strings (e.g. '1', '2'), but all other RedisCacheAdapter methods (fetch, store, inc, entry, cas) use PHP serialize()/unserialize() which stores integers as 'i:1;'. This makes the two storage formats incompatible on the same key.

### Steps to reproduce

1. Open `frontend/server/src/Cache.php` around line 327
2. If Cache::get() is called after Cache::incWithTTL() on the same key, fetch() calls unserialize('1') which returns false (PHP's unserialize expects 'i:1;' not '1'), so Cache::get() treats the counter as a cache miss and returns null. Conversely, if inc() was called first (storing 'i:1;'), calling incWithTTL() triggers Redis INCR on a non-integer string, causing a Redis ERR response that eval() returns as false, silently returning 0 instead of the incremented value.

### Expected behavior

incWithTTL() was implemented using native Redis INCR for atomic TTL support, but the rest of RedisCacheAdapter wraps all values in PHP's serialize()/unserialize(). The new incWithTTL() bypasses this serialization layer, creating an incompatible storage format. The base-class inc() override (lines 298-317) also serializes, making incWithTTL() inconsistent with both that and every other method. should be handled correctly.

### Actual behavior

If Cache::get() is called after Cache::incWithTTL() on the same key, fetch() calls unserialize('1') which returns false (PHP's unserialize expects 'i:1;' not '1'), so Cache::get() treats the counter as a cache miss and returns null. Conversely, if inc() was called first (storing 'i:1;'), calling incWithTTL() triggers Redis INCR on a non-integer string, causing a Redis ERR response that eval() returns as false, silently returning 0 instead of the incremented value.

deepak0x:
# Description

`RedisCacheAdapter::incWithTTL()` used a raw Redis `INCR`, which stores the counter as a bare integer string (`"1"`, `"2"`). Every other method in the adapter (`fetch`, `store`, `inc`, `entry`, `cas`) stores values with PHP's `serialize()`/`unserialize()`, so an integer is stored as `"i:1;"`. The two formats are incompatible on the same key:

- `Cache::get()` after `incWithTTL()` calls `unserialize("1")`, which returns `false`, so the counter reads as a cache miss (`null`).
- `incWithTTL()` after `inc()` runs `INCR` on `"i:1;"`, a non-integer string, causing a Redis `ERR` that `eval()` surfaces as `false` → silently returns `0`.

This rewrites the Lua script so the counter is read, incremented and written back in the same serialized `"i:N;"` format, still atomically. The TTL is only set when the key is first created and is preserved (via `PTTL`/`PEXPIRE`) on subsequent increments, matching the previous behaviour. A legacy bare-integer value is also handled gracefully (parsed and migrated to the serialized format).

The `InProcessCacheAdapter` and `APCCacheAdapter` were already consistent (they store values via `entry()`/`cas()`), so only the Redis adapter changes.

Fixes: #9956

# Comments

Added two cross-format regression tests to `CacheTest.php` (run against every adapter via the existing data provider): `fetch()` after `incWithTTL()`, and `inc()` followed by `incWithTTL()`.

Ran the suite in the omegaUp Docker dev environment: `CacheTest` → **OK (33 tests, 128 assertions)**. Confirmed the two new tests **fail on the pre-fix code** for the Redis adapter (`fetch()` returns `false` via an `unserialize()` error; `inc()`+`incWithTTL()` returns `0`) and pass with the fix.

# Checklist:


pabo99 (approved):
Excellent!

PR #9967, opened 2026-07-04
extend navbar item style to remaining submenus

PrasannaMishra001:
# Description

Extends the navbar item style introduced in #9801 to the remaining submenus. Contests, Courses, Problems and Ranking now use the NavbarItem component with an icon and a one line description per entry, matching the Help menu. The description strings were added to the localization files.

While testing this I noticed the dropdowns stayed at the default Bootstrap width and the titles wrapped badly. The width rule that let the Help dropdown grow with its content was tied to the help specific class, so I renamed it from help-dropdown to navbar-item-dropdown and applied it to all five menus.

All existing data-nav selectors are preserved. The Create problem collapsed submenu keeps its current markup since its toggle is a button rather than a link.

Fixes: #9962
menu-contests
menu-courses
menu-problems
menu-ranking
menu-help

# Comments

Built on top of #9871 (the NavbarItem component), so the diff includes it until that one merges. Icon and wording choices can be changed easily if any of them feel off.

# Checklist:

- [x] The code follows the [coding guidelines](https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/Coding-guidelines.md) of omegaUp.
- [x] The tests were executed and all of them passed.
- [x] If you are creating a feature, the new tests were added.
- [x] If the change is large (> 200 lines), this PR was split into various Pull Requests. It's preferred to create one PR for changes in controllers + unit tests in PHPUnit,  and then another Pull Request for UI + tests in Jest, Cypress or both.

PR #9871, opened 2026-05-26
Extract NavbarItem component for help menu

PrasannaMishra001:
# Description

Adds a reusable `NavbarItem.vue` component that encapsulates the icon, title and description layout introduced in #9801 for the help submenu.

Refactors the help dropdown in `NavbarItems.vue` so that the eight repeated `<a>` blocks are replaced with a single `<template v-for>` over a `helpMenuEntries` computed property that drives both real items and the section dividers.

Moves the help related styles (`help-dropdown-item`, `help-item-icon`) into the new component so they live next to the markup they style and stay scoped to it.

Adds a unit test file `NavbarItem.test.ts` covering the plain link case (only title) and the rich case (icon, title and description).

Fixes: #9870

# Comments

This PR is stacked on top of #9801, which is approved and pending merge. Once #9801 lands, the diff here will reduce to just the NavbarItem extraction (one commit). The remaining suggestions in review on #9801 (applying the new style to the other submenus and the optional move of submenu options into a configuration file) are intentionally left out of this PR. They are larger changes that would require choosing icons and writing new translation strings for every entry in the Contests, Courses, Problems and Ranking submenus, so it seemed cleaner to land the core component first and tackle those in a follow up.

# Checklist:

- [x] The code follows the [coding guidelines](https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/Coding-guidelines.md) of omegaUp.
- [x] The tests were executed and all of them passed.
- [x] If you are creating a feature, the new tests were added.
- [x] If the change is large (> 200 lines), this PR was split into various Pull Requests. It's preferred to create one PR for changes in controllers + unit tests in PHPUnit,  and then another Pull Request for UI + tests in Jest, Cypress or both.

PR #9968, opened 2026-07-04
build navbar menus from a single configuration

### LINKED ISSUE #9966: [FEATURE] Build the navbar menus from a single configuration with visibility rules
### Checklist

### Context

#9870 / #9871 extracted the NavbarItem component and #9962 extends its style to all submenus, but each menu's entries are still hardcoded in the NavbarItems template with their visibility conditions (isLoggedIn, isMainUserIdentity, isUnder13User, isReviewer) spread across v-if directives.

As suggested in the #9871 review, the next step is building the entire menu from a configuration, so the available options are controlled from one place. Since NavbarItems already receives the access flags from the payload, each entry in the configuration could carry an optional visibility rule over those flags, evaluated in a single place when the menu is built which would also open the door to access levels or per-profile configurations later. The exact shape of the configuration can be discussed in the PR.

NOTE: Blocked by #9870 / #9871 and #9962.

PrasannaMishra001:
# Description

Builds the navbar menus from a single configuration instead of entries hardcoded in the template. The new navbarMenus module declares every entry for the Contests, Courses, Problems, Ranking and Help menus and each entry carries its own visibility rule over the access flags that already come from the payload (isLoggedIn, isMainUserIdentity, isUnder13User, isReviewer). NavbarItems now just filters the configuration in one place and renders the result, so adding, removing or restricting an option is a one line change in the configuration.

All data-nav selectors, the slots and the logged out fallback links are preserved and the Create problem collapsed submenu is declared in the configuration but keeps its current markup since its toggle is a button. This also leaves the door open for access levels or per profile configurations later, since those become alternate configurations over the same shape.

Fixes: #9966
menu-help
menu-problems
menu-ranking
menu-contests
menu-courses

# Comments

Built on top of #9871 and #9967, so the diff includes them until those merge. Rendering is unchanged, this only moves where the menu is defined.

# Checklist:


pabo99:
Could you resolve the merge conflicts that appeared? Once those are resolved, I think the change will be much easier to review

PrasannaMishra001:
@pabo99 solved the conflicts, can be reviewed now.. , better to merge #9967 first, that will be easier with reviews.

pabo99 (commented):
The new styling for the menu items, with the icons and descriptions, looks really good. Thanks for making this change!

I just have a couple of minor suggestions:

- Would it be possible to use a different icon for _Create Contest_? I can't think of a better one right now, but this one doesn't quite feel right to me.
- Would it be possible to remove the submenus? I think that would make the menu settings much simpler. If we decide to keep them, I'd at least prefer them to always be expanded.

After these changes, this PR will be ready to be approved

PrasannaMishra001:
thanks @pabo99, done. changed the create contest icon to a checkered flag that reads as contests (happy to switch it to a simple plus / create new if you prefer), and removed the create contest submenu by turning its options into direct entries in the menu instead, so it is simpler now. 
image
image

pabo99 (commented):
Great progress! Now that we have support for titles and subtitles, could we make the two options a bit more descriptive? Something like "Create Problem from ZIP" and "Create Problem from Scratch" would make it easier for users to distinguish between the two workflows. The subtitle could then briefly explain what each option does

PrasannaMishra001:
@pabo99 done , 
image

pabo99 (approved):
Great!

PR #9970, opened 2026-07-04
Fix: Browser back/forward buttons do not update displayed tab on Edit Problem page (#9877)

### LINKED ISSUE #9877: [BUG] Browser back button does not update displayed tab on edit problem page
### Checklist

## Current Behavior

On the `/problem/<alias>/edit/` page, navigating between tabs updates the URL hash and displays the corresponding content correctly.

However, when pressing the browser Back button, the URL hash changes to the previous tab but the displayed tab content does not update.

## Expected Behavior

Pressing the browser Back button should update both:
- the URL hash
- the visible tab content

so that the previously viewed tab is displayed.

## Steps to Reproduce

1. Navigate to `/problem/<alias>/edit`
2. Navigate to another tab if available
3. Press the browser Back button

Result:
- URL hash changes
- tab content remains stuck on the last viewed tab

## Root Cause

The component updates the tab state through click handlers but the page entry `.ts` file does not listen for `hashchange` events. As a result, browser Back/Forward navigation updates the URL but does not update the Vue state.

A similar issue was previously fixed in:

- Issue: #9445 
- PR: #9471 

## Environment

Browser: Chrome (reproducible across browsers)  
Page: `/groups/<alias>/edit/`

## Screen Recording

https://github.com/user-attachments/assets/2e95e392-5fe3-4632-abe5-721f65958175

### Expected Behavior

- URL hash changes
- tab content remains stuck on the last viewed tab

_No response_

abhinavsingh-hub:
## Description

On the `/problem/<alias>/edit/` page, navigating between tabs updates the URL hash correctly. 
However, pressing the browser's Back/Forward buttons would update the URL hash but leave the displayed tab content stuck on the last viewed tab. 

This occurred because the page entry file (`edit.ts`) was not listening to `hashchange` events, and the component (`Edit.vue`) did not reactively watch the `initialTab` property to synchronize its internal tab state.

This PR implements tab synchronization on browser back/forward navigation for the Edit Problem page, mirroring the fix previously applied to the Edit Course page in #9471.

---

## Changes
- **`frontend/www/js/omegaup/problem/edit.ts`**:
  - Registered a window `hashchange` event listener to reactively update the Vue model's `initialTab` property when navigation occurs.
  - Added a `beforeDestroy` hook on the Vue instance to cleanly remove the event listener and prevent leaks.
- **`frontend/www/js/omegaup/components/problem/Edit.vue`**:
  - Declared an `availableTabs` constant containing valid problem edit tabs.
  - Added a `@Watch('initialTab')` handler to synchronize the internal `showTab` state when the `initialTab` prop changes.
- **`frontend/www/js/omegaup/components/problem/Edit.test.ts`**:
  - Updated the unit tests to verify that changing the `initialTab` prop correctly updates the component's internal `showTab` property.

---

## Verification
All frontend Jest unit tests pass successfully:
```bash
npm test -- frontend/www/js/omegaup/components/problem/Edit.test.ts 
```

### Screen-recording 

https://github.com/user-attachments/assets/5f084f60-19d4-44bb-9630-9e76279da1ba


abhinavsingh-hub:
Hey @pabo99 !

I hope the changes are good to go, I have edited the test file for easier verification process and better code maintainability, would change if required, including any other thing as per your suggestions.

Regards

frontend/www/js/omegaup/components/problem/Edit.vue
```diff
@@ -375,5 +386,14 @@ export default class ProblemEdit extends Vue {
   onGotoPrintableVersion(): void {
     window.location.href = `/arena/problem/${this.alias}/print/`;
   }
+
+  @Watch('initialTab')
+  onInitialTabChanged(newValue: string): void {
+    if (!availableTabs.includes(this.initialTab)) {
```
pabo99: Could you use the watcher argument here instead?
```suggestion
    if (!availableTabs.includes(newValue)) {
```
It's the value received by the watcher, so it makes the check a bit clearer

frontend/www/js/omegaup/components/problem/Edit.vue
pabo99: Could you also validate the initial tab value here? The watcher handles invalid values after `initialTab` changes, but opening the page directly with an invalid hash would still initialize `showTab` to that invalid value and render no tab content:

```ts
  showTab = availableTabs.includes(this.initialTab) ? this.initialTab : 'edit';
```

pabo99 (commented):
Just a few minor suggestions and it's okay

abhinavsingh-hub:
hey @pabo99 !!

Done! Updated to use `newValue` in the watcher and added validation for `initialTab` during property initialization. 
Let me know if anything else is required, would be more than eager to contribute.

Thanks!

frontend/www/js/omegaup/components/problem/Edit.vue
```diff
@@ -293,6 +293,17 @@ import 'bootstrap-vue/dist/bootstrap-vue.css';
 import { ModalPlugin } from 'bootstrap-vue';
 Vue.use(ModalPlugin);
 
+const availableTabs = [
```
pabo99, line 296: Could you use an `enum` for these tab names? We already follow this pattern in other components for available tabs, and it makes these values less typo-prone.

Since the initial value comes from the URL hash, we should still validate it before assigning it to the selected tab.

PR #9979, opened 2026-07-09
Fix crash when editing course with no school assigned

### LINKED ISSUE #4498: [BUG] Problemas con el campo Escuela en Editar Curso
## Comportamiento Esperado

No es necesario asignar una escuela cuando creo un curso.

## Comportamiento Actual
Cree un curso sin asignar una escuela. Cuando intento editar la información del curso no me deja actualizar si no agrego una escuela, me muestra el error "Escuela no encontrada" aún cuando el campo está vacío. Escribo algo en el campo escuela y puedo actualizar. Cambio de tab y cuando regreso a el tab de "Editar Curso" el campo Escuela dice omegaUp.

image

image

image

sushmitah01:
# Description

`request.school` can be `null` when a course has no school assigned to it. In `frontend/www/js/omegaup/course/edit.ts`, the `submit-edit-course` handler accessed `request.school.key` and `request.school.value` directly, which throws a TypeError when `request.school` is `null`.

Course creation (`course/new.ts`) already handles this correctly using optional chaining (`request.school?.key`). This PR applies the same fix to course editing.

Fixes: #4498

# Comments

This fixes the crash reported in the issue. I traced the root cause by comparing the edit flow against the working create flow in new.ts, which handles the same null case safely.

# Checklist:

### DIFF
--- a/frontend/www/js/omegaup/course/edit.ts
+++ b/frontend/www/js/omegaup/course/edit.ts
@@ -167,9 +167,9 @@ OmegaUp.on('ready', () => {
           },
           'submit-edit-course': (request: messages.CourseUpdateRequest) => {
             new Promise<number | null>((accept) => {
-              if (request.school.key) {
+              if (request.school?.key) {
                 accept(request.school.key);
-              } else if (request.school.value) {
+              } else if (request.school?.value) {
                 api.School.create({ name: request.school.value })
                   .then((response) => {
                     accept(response.school_id);


pabo99 (commented):
Thanks for the fix. This does address the crash when editing an existing course that has no school assigned, since `request.school` can be `null` in that flow

That said, I think this PR should stay blocked for now until we define the expected behavior for courses without a school. The backend allows `school_id` to be null, but the frontend still treats the school field as required when creating a course. So this only partially addresses #4498

Before merging, we should decide whether courses are allowed to exist without a school/institution, or whether the UI should consistently require one

PR #9985, opened 2026-07-11
Fix staleWhileRevalidate returning undefined when offline with cold cache

### LINKED ISSUE #9977: bug: staleWhileRevalidate returns undefined when offline with no cached response
### Description

When no cache entry exists for a request and the network fetch fails, the .catch() handler falls back to `cached` which is undefined. The returned promise resolves to undefined instead of a valid Response, causing respondWith() to receive a non-Response value.

### Steps to reproduce

1. Open `frontend/www/sw.js` around line 70
2. A media resource fetch handled by the service worker silently fails at the browser level when the user is offline and the cache is cold. respondWith() receives a promise resolving to undefined (not a Response), causing the fetch to fail with an internal service worker error rather than a graceful network error.

### Expected behavior

Line 70 `.catch(() => cached)` uses `cached` as the error fallback, but `cached` is undefined when cache.match() found nothing. So networkFetch resolves to undefined on network failure. Line 72 `return cached || networkFetch` then returns this undefined-resolving promise. The companion networkFirst() function at line 89 correctly guards against this with `return cached || Response.error()`, but staleWhileRevalidate does not. should be handled correctly.

### Actual behavior

A media resource fetch handled by the service worker silently fails at the browser level when the user is offline and the cache is cold. respondWith() receives a promise resolving to undefined (not a Response), causing the fetch to fail with an internal service worker error rather than a graceful network error.

deepak0x:
# Description

`staleWhileRevalidate` in `frontend/www/sw.js` used `.catch(() => cached)` as the network-failure fallback, but `cached` is `undefined` when `cache.match()` found nothing. With a cold cache and a failed fetch (e.g. offline media request), `networkFetch` resolved to `undefined`, so `respondWith()` received a non-Response value and the service worker failed with an internal error instead of a graceful network error.

Guarded the fallback with `Response.error()`, mirroring the existing `networkFirst` behavior.

Fixes: #9977

# Checklist:

### DIFF
--- a/frontend/www/sw.js
+++ b/frontend/www/sw.js
@@ -67,7 +67,7 @@ async function staleWhileRevalidate(request) {
 
       return response;
     })
-    .catch(() => cached);
+    .catch(() => cached || Response.error());
 
   return cached || networkFetch;
 }


pabo99 (approved):
LGTM!

PR #9987, opened 2026-07-11
[Milestone 3] Generate zip from Problem Creator modal and attach it on form submit

### LINKED ISSUE #9986: [Milestone 3] Integrate JSZip generation + DataTransfer API file attachment (Creator → submit path)
Part of #9824 (Integrate Problem Creator). Builds on #9859 (Milestone 1) and #9891 (Milestone 2).

Milestones 1 and 2 added the creation method selector and the Problem Creator modal overlay, but the modal is still cosmetic — submitting the form with the "Problem creator" method selected sends no problem contents.

Image

### Goal
Make the Creator → submit path functional: a problem built inside the Creator modal should be submitted through the existing /problem/new` form POST, with no backend changes.

### Proposed flow
1. When the modal is closed, the Creator generates a fresh zip (statements, solutions, cases, testplan, cdp.data) and the Form converts it to a `Blob`.
2. On form submit, the blob is attached to a hidden `<input type="file" name="problem_contents">` using the DataTransfer API, so the backend receives exactly the same payload as a manual zip upload.
3. Submit is blocked with a user-facing error while the zip is still generating or if the Creator was never opened.

### Also needed for this flow to make sense in embedded mode
- Hide the standalone-only header (Name input, Load Problem, Generate Problem, + New Problem) and the per-tab Save buttons when the Creator is embedded in the modal — the form already has its own title field and the zip is generated
  automatically on close.
- Persist drafts automatically when the modal closes and restore them when it reopens, so the Vuex creator store is the single source of truth and no work is lost (no Save clicks required).

Everything stays behind the `problem_creation_method_selector` experiment; with the flag off, the form behaves exactly as today.

shauryag2002:
# Description

This PR makes the Problem Creator modal functional end-to-end: a problem built in the modal is now submitted through the existing form POST as `problem_contents`, with no backend changes.

[Screencast from 2026-07-11 14-51-28.webm](https://github.com/user-attachments/assets/d55620f6-8e4a-4043-9a87-a595c8647560)

**Creator → submit pipeline (Form.vue):**
- Closing the modal persists all tab drafts to the creator store and generates a fresh zip, which is converted to a `Blob` and held in memory.
- On form submit with the "Problem creator" method, the blob is attached to a hidden `<input type="file" name="problem_contents">` via the DataTransfer API and the native form submission proceeds — the backend sees the same payload as a manual zip upload.
- Submit is blocked with an error message while the zip is still generating (`problemCreatorGeneratingZip`) or if the Creator was never opened (`problemCreatorOpenBeforeSubmit`). New i18n keys added in en/es/pt/pseudo.

**Embedded-mode UI (matches the proposal):**
- New `hideHeaderActions` prop (Form → CreatorWrapper → Creator → Header) hides the Name input and the Load Problem / Generate Problem / + New Problem buttons inside the modal. The standalone `/problem/creator` page is unchanged (defaults to false).
- New `hideSaveButtons` prop hides the per-tab Save buttons; instead, drafts are auto-persisted to the store on close (`saveDraft()` → `saveAllDrafts()` → per-tab `persistDraft()`).
- Reopening the modal restores statement/code/solution from the store (`Creator.mounted()` now populates tab props from `$store.state`), so no work is lost across open/close cycles.
- The Creator's standalone Intro.js tour is skipped in embedded mode since it targets the hidden header buttons.

**Bug fix (Header.vue):**
- `generateProblem()` reused a single accumulating `JSZip` instance, so files removed from the store lingered in regenerated zips. It now builds a fresh zip on every call — important since the zip is now the submission payload.

**Feature flag:** every new behavior is gated on the `problem_creation_method_selector` experiment (hidden input, submit handler, and modal all check `showCreationMethodSelector`). With the flag off, the form is
byte-for-byte the current native flow.

**Browser support:** the DataTransfer API is available in Chrome 62+, Firefox 62+, Safari 14.1+ and Edge 79+. On older browsers submit is blocked with a clear error instead of silently posting an empty form; a FormData+fetch fallback can be added in a follow-up if needed.

**Tests:** 4 new Form tests (zip capture, both submit guards, DataTransfer attachment) and one Header test updated to assert on the emitted zip instead of the removed shared instance. All 139 problem-component tests pass.

Fixes: #9986 

# Comments
- Intro.js updates for the new flow are intentionally out of scope (later milestone).
- Uploading an existing zip to pre-populate the Creator (unified zip import flow) is Milestone 4/5 and not part of this PR.

# Checklist:

- [x] The code follows the [coding guidelines] https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/Coding-guidelines.md) of omegaUp.


pabo99 (approved):
LGTM!

The only thing I like to ask is to add a follow-up test in a future PR to cover the full e2e Creator modal submit flow so we can avoid regressions around  generating the zip and attaching it as `problem_contents

Issue #9992, opened 2026-07-13
[FEATURE] Cron Control Plane : an observability and operational tooling for background jobs


PrasannaMishra001:
### Checklist

- [x] I have searched existing issues and confirmed this is not a duplicate
- [x] I have discussed this on [GitHub Discussions](https://github.com/omegaup/omegaup/discussions) or [Discord](https://discord.gg/K3JFd9d3wk) and received positive feedback

### Context

the background jobs (update_ranks, assign_badges, aggregate_feedback, and the ondemand build_problem_rec_model / plagiarism_detector) have no operational tooling. in simple terms there is no record that a run happened, no protection against two runs overlapping, no admin view of health or timings, no data freshness signal and no safe way to rerun a job without shell access to a pod.

this epic tracks building the management layer around those jobs, a job registry and run history schema, a reusable runner that records runs and holds an advisory lock, adoption of the runner across the jobs, a prepublish guardrail on the ranking job, an admin dashboard, a safe queued rerun and an end to end test.

the expected result in the end looks like this:

https://github.com/user-attachments/assets/ccdbc638-dd57-448a-aab9-382403ab90ac

### Sub issues (review and merge in this order)

everything fans out from the tables migration, so that one goes in first and then the two chains can move in parallel.

foundation (merge first)
1. #9993 persist cron execution history, the registry and run tables migration (PR #9995)

php / ui chain (each after the one above it)
2. #9994 daos for the cron tables (PR #9996)
3. #9999 admin api for run history and health (PR #10000)
4. #10004 admin dashboard page at /admin/crons (PR #10008)
5. #10018 let admins safely rerun a cron job (PR #10021)
6. #10020 end to end test for the cron dashboard (PR #10023)

python chain (each after the one above it)
7. #9997 reusable cron runner that records runs and holds the lock (PR #9998)
8. #10002 record update_ranks through the runner (PR #10003)
9. #10007 record assign_badges and aggregate_feedback (PR #10009)
10. #10019 record the remaining ondemand crons through the runner (PR #10022)
11. #10010 prepublish guardrail on the ranking job (PR #10011)

further work (enhancements, raised on top of the dashboard)
12. #10047 quality of life improvements on the dashboard, health cards, filters, relative times and auto refresh (PR #10048)

still to raise: an operational runbook for the jobs.

[...]

PR #9995, opened 2026-07-13
add cron control plane tables

### LINKED ISSUE #9993: [FEATURE] persistant cron execution history by registry and run tables
### Checklist

### Context

right now there is no record that a cron ran, when it ran, how long it took or whether it even succeeded, so if last night's ranking job quietly failed nobody would know until someone noticed stale data.

this adds the schema for a small cron job registry and an execution history table, so the platform can store each run and later surface it. `Cron_Jobs` holds the known jobs seeded with the current three (update_ranks, assign_badges, aggregate_feedback) and `Cron_Runs` holds one row per execution with status, start and end, duration, per phase timings and any error.

**expected outcome:** a forward only migration that creates both tables.

### Order

this is the base of the cron control plane epic #9992, everything else builds on it so it should be reviewed and merged first. nothing needs to merge before this one.

PrasannaMishra001:
# Description

adds the schema foundation for the cron control plane. `Cron_Jobs` is the registry of known jobs, seeded with the current three (`update_ranks`, `assign_badges` and `aggregate_feedback`), and `Cron_Runs` is the per execution history holding status, start and end, duration, per phase json and any error text. it is a forward only migration `00276_cron_control_plane.sql` plus the regenerated `schema.sql`, so the platform can start recording runs.

Fixes: #9993

# Comments

this is the base PR of the cron control plane epic #9992, everything else builds on it so it should be reviewed and merged first. it is the migration and `schema.sql` only, the generated daos and their test come right after in #9996 (the linter does not allow `schema.sql` and `dao_schema.sql` in the same commit, so they are split).

# Checklist:


PrasannaMishra001:
@pabo99 pls review this one

PrasannaMishra001:
@pabo99 please review this one for the cron-control-plane, all following i made as drafts PR which i will open one by one as this gets merged.

frontend/database/00276_cron_control_plane.sql
```diff
@@ -0,0 +1,32 @@
...
+  `enabled` tinyint(1) NOT NULL DEFAULT '1',
+  `created_at` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `updated_at` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  PRIMARY KEY (`job_id`),
+  UNIQUE KEY `unique_cron_job_name` (`name`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci COMMENT='Registro de los trabajos programados (cron) del sistema';
+
+CREATE TABLE `Cron_Runs` (
+  `run_id` int NOT NULL AUTO_INCREMENT,
+  `name` varchar(64) NOT NULL,
```
pabo99: Since `Cron_Jobs` already provides a stable identifier for each cronjob, would it make more sense for `Cron_Runs` to store `job_id` instead of `name`? That would avoid duplicating data, preserve referential integrity, and allow renaming a job without affecting historical records. If the intention is to keep execution history independent from `Cron_Jobs` (for example, after deleting a job), it might be worth documenting that rationale in the schema

PrasannaMishra001: the main reason is that recording a run should never depend on the job being registered. the runner writes the running row from python using its own `parser.prog` before it touches `Cron_Jobs` and some jobs that use the runner are not in the registry at all, the ondemand ones like build_problem_rec_model and plagiarism_detector. if `Cron_Runs` needed a `job_id` those runs either could not be recorded or the runner would have to auto insert registry rows, which would turn the registry from a curated list into something that fills itself.

Also as this is a history table, so it should survive registry changes. if a job is removed from Cron_Jobs its past runs stay readable and each row keeps the name it actually ran under. name here is not a display label, it is the script's parser.prog, so renaming it is a code change anyway.

i have documented that rationale in the schema as you suggested. that said, if you would rather have the normalized version with a real FK and accept that only registered jobs can record runs, i am happy to switch it.

pabo99 (approved):
Thanks for the clarification and for documenting the rationale. 

LGTM!

PR #9998, opened 2026-07-13
add cron runner library

### LINKED ISSUE #9997: [FEATURE] Reusable cron runner that records runs and prevents overlap
### Checklist

### Context

each cron reimplements its own timing and logging, and nothing stops two runs of the same job overlapping even though they share and rewrite the same tables.

this adds a shared runner library that records each run's start, end, status and per phase timings into `Cron_Runs`, and holds a mysql advisory lock so a second run of the same job skips cleanly instead of racing the first.

**expected outcome:** a reusable runner the cron scripts wrap their `main()` with.

### Order

part of the cron control plane epic #9992. #9993 (the tables migration) should be resolved and merged first since the runner writes into `Cron_Runs`. after that it can go in parallel with the dao work.

PrasannaMishra001:
# Description

adds `stuff/lib/runner.py`, a context manager the cron scripts wrap their `main()` with. on enter it opens its own connection, takes `GET_LOCK('cron:<job>')` (skipping cleanly if another run already holds it) and writes a `running` row, and on exit it records status, duration, per phase timings and any error and then releases the lock. `phase(name)` times a step, `mark_failure()` forces a failure and `--no-track` bypasses recording for local runs.

it also honours the `enabled` flag on the registry. before taking the lock it looks the job up in `Cron_Jobs`, and if the row says disabled it logs that and exits 0 without recording a run. a job that is not in the registry at all runs normally, which keeps the ondemand scripts working since they are not registered, same reasoning as storing `name` rather than `job_id` on `Cron_Runs`. this is what makes the switch on the dashboard real rather than decorative.

it also cleans up after itself if startup fails. `__exit__` does not run when `__enter__` raises, so the lock is released and the connection closed there before re-raising, and the release moved into a `finally` so a failure while writing the final status cannot leave the lock held.

unit tested with an in memory fake connection (11 tests).

Fixes: #9997

# Comments

part of the cron control plane epic #9992. the tables migration it writes into is already merged, so this is rebased onto main and the diff is just the two runner files. the admin switch that flips the enabled flag lives in #10048.

# Checklist:


stuff/lib/runner.py
```diff
@@ -0,0 +1,206 @@
...
+            logging.info(
+                'cron job %s is disabled, skipping', self._program)
+            self._close_connection()
+            raise SystemExit(0)
+        if not self._acquire_lock():
+            logging.info(
+                'cron job %s is already running, skipping', self._program)
+            self._close_connection()
+            raise SystemExit(0)
+        self._insert_running_row()
```
pabo99: If `_insert_running_row()` raises after the advisory lock has been acquired, `__enter__()` exits without invoking `__exit__()`. This can leave the lock and the runner-owned connection open. Could you wrap the initialization in a `try`/`except` block that releases the lock and closes the connection before re-raising the exception?

stuff/lib/runner.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def __exit__(
+            self,
+            exc_type: Any,
+            exc_value: Any,
+            traceback: Any) -> None:
+        if not self._enabled:
+            return
+        try:
+            self._finish(exc_value)
```
pabo99, line 100: If `_finish()` raises while updating or committing the final status, the advisory lock still needs to be released. This is especially important when an external connection is supplied because the runner will not close that connection. Could you move `_release_lock()` into a `finally` block?

pabo99 (commented):
The PR description says callers can use `mark_failure()` to force a failed status, but that method is currently missing. Without it, a cron job that handles an error internally will be recorded as successful. Could you add `mark_failure()` and include a unit test for it?

PrasannaMishra001:
thanks @pabo99 , the lock leak on startup is handled now, `__enter__` releases the lock and closes the connection before re-raising, since `__exit__` never runs when` __enter__` fails.` __exit__` uses nested finally so the release happens even if writing the final status throws, which is the case that mattered for an ext. connection since runner does not close that one. and `mark_failure()` is added, you were right that the description promised it, i had described it from a later branch by mistake.

tests did 7 to 11 and i checked new ones fail against the old code so they actually cover the bugs.

pabo99 (approved):
Great!

PR #10000, opened 2026-07-13
add cron admin api

### LINKED ISSUE #9999: [FEATURE] Admin API for cron run history and health
### Checklist

### Context

admins cannot query cron status from the platform at all today.

this adds admin only endpoints returning the job list, the recent runs and a single run's detail with its per phase breakdown, so the dashboard has something to read.

**expected outcome:** `apiGetCrons` and `apiGetCronRun` on the admin controller, guarded by the usual admin checks.

### Order

part of the cron control plane epic #9992. #9993 (migration) then #9994 (daos) should be resolved and merged first in that order, this reads through those daos.

PrasannaMishra001:
# Description

adds admin only `Admin::apiGetCrons` (the job registry plus recent runs) and `apiGetCronRun` (one run's detail with its per phase breakdown), both guarded by `ensureMainUserIdentity` and `isSystemAdmin`, with the `@psalm-type`s and the regenerated `api.ts` / `api_types.ts` so the frontend gets typed clients. covered by `CronControlPlaneAdminTest`.

Fixes: #9999

# Comments

part of the cron control plane epic #9992. both things it needed, the tables migration and the daos, are merged now, so this is rebased onto main and the diff is just the controller, its test and the regenerated api types.

# Checklist:

### DIFF
--- a/frontend/server/src/Controllers/Admin.php
+++ b/frontend/server/src/Controllers/Admin.php
@@ -9,6 +9,9 @@
   * @psalm-type MessageLanguages=array{es: string, en: string, pt: string}
   * @psalm-type PredefinedTemplate=array{id: string, title: MessageLanguages, message: MessageLanguages, type: string}
   * @psalm-type MaintenanceModeStatus=array{enabled: bool, message_es: null|string, message_en: null|string, message_pt: null|string, type: string}
+  * @psalm-type CronJob=array{name: string, description: null|string, schedule: null|string, enabled: bool}
+  * @psalm-type CronRunPhase=array{phase: string, status: string, duration: float, error_class: null|string}
+  * @psalm-type CronRun=array{run_id: int, name: string, hostname: null|string, status: string, started_at: \OmegaUp\Timestamp|null, finished_at: \OmegaUp\Timestamp|null, duration_seconds: float|null, rows_affected: int|null, phases: list<CronRunPhase>, error_text: null|string}
   */
 class Admin extends \OmegaUp\Controllers\Controller {
     const MAINTENANCE_MESSAGE_ES_KEY = 'maintenance_message_es';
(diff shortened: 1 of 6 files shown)


frontend/tests/controllers/CronControlPlaneAdminTest.php
```diff
@@ -0,0 +1,67 @@
+<?php
+
+/**
+ * Tests for the cron control plane admin endpoints.
+ */
+class CronControlPlaneAdminTest extends \OmegaUp\Test\ControllerTestCase {
+    public function testGetCronsRequiresAdmin() {
```
pabo99, line 7: Could you add a non-admin coverage case for `apiGetCronRun` as well? `apiGetCrons` already verifies the system-admin guard, but this endpoint has its own authorization path and it would be nice to lock it down explicitly

frontend/server/src/Controllers/Admin.php
```diff
@@ -396,4 +399,86 @@ public static function getSettingsForTypeScript(
...
+    /**
+     * @param list<\OmegaUp\DAO\VO\CronRuns> $runs
+     *
+     * @return list<CronRun>
+     */
+    private static function cronRunsPayload(array $runs): array {
+        $result = [];
+        foreach ($runs as $run) {
+            /** @var list<CronRunPhase> */
+            $phases = is_null($run->phases) ? [] : json_decode(
```
pabo99, line 445: Could you add a test case with a run that has `phases` populated? The endpoint decodes this JSON before returning it, so covering at least one non-empty phase payload would help catch regressions in the response shape.

frontend/tests/controllers/CronControlPlaneAdminTest.php
PrasannaMishra001: added, `testGetCronRunRequiresAdmin` ,similar to the one on apiGetCrons

frontend/server/src/Controllers/Admin.php
PrasannaMishra001: added `testGetCronRunDecodesPhases`. it stores two phases, one of them failed with an `error_class` and asserts the decoded shape comes back, so the response shape is pinned now.

pabo99 (approved):
Thank you, @PrasannaMishra001 !

PR #10005, opened 2026-07-14
Fix clearOldItems skipping consecutive old localStorage entries

### LINKED ISSUE #9988: bug: clearOldItems skips consecutive old localStorage entries during forward iteration with removal
### Description

The for-loop in clearOldItems iterates localStorage by index and calls localStorage.removeItem(key) without decrementing i afterward. When an entry is removed, all subsequent entries shift their indices down by 1. The loop then increments i, skipping the entry that moved into the freed slot.

### Steps to reproduce

1. Open `frontend/www/js/omegaup/safe_storage.ts` around line 31
2. When two or more consecutive clarification-* entries are old and eligible for removal, every other one is silently skipped. E.g. with 3 consecutive old entries at indices 0,1,2: indices 0 and 2 are removed but index 1 is skipped.

### Expected behavior

localStorage.removeItem() shifts all subsequent entry indices, but i is incremented without accounting for the shift. The classic off-by-one for mutable-sequence removal inside a forward loop. should be handled correctly.

### Actual behavior

When two or more consecutive clarification-* entries are old and eligible for removal, every other one is silently skipped. E.g. with 3 consecutive old entries at indices 0,1,2: indices 0 and 2 are removed but index 1 is skipped.

deepak0x:
# Description

`SafeStorage.clearOldItems` iterated localStorage forward (`i = 0; i < length; i++`) while calling `localStorage.removeItem` inside the loop. Removing an entry shifts every later entry down one index, but the counter still incremented, so the entry that moved into the freed slot was skipped and never evaluated. With consecutive stale `clarification-` entries this left some of them behind, defeating the quota-recovery cleanup.

The loop now iterates backwards (`i = length - 1; i >= 0; i--`), so a removal only affects indices that have already been visited. All matching stale entries are now evaluated and removed.

Added Jest regression tests covering consecutive stale entries, interleaved stale/recent entries, and preservation of recent and unrelated keys.

Fixes: #9988

# Checklist:


pabo99 (approved):
Thanks!

PR #9996, opened 2026-07-13
add cron control plane dao

PrasannaMishra001:
# Description

adds the data access layer for the cron tables. this has `dao_schema.sql`, the generated `DAO/Base` and `DAO/VO` for `CronJobs` and `CronRuns`, and the handwritten `DAO/CronRuns.php` (getRecent, getRecentByName, getLatestByName, getLatestSuccessfulByName) and `DAO/CronJobs.php` (getByName, getAllOrdered), so the rest of the backend never writes raw sql against these tables. covered by a `CronControlPlaneDAOTest`.

Fixes: #9994

# Comments

part of the cron control plane epic #9992. needs #9995 (the tables migration) merged first. until it does, this diff still carries the migration commit and the validate database schema check stays red because `schema.sql` and `dao_schema.sql` show up together, once #9995 merges a rebase drops the migration commit and leaves only the dao change.

# Checklist:

- [x] The code follows the [coding guidelines](https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/Coding-guidelines.md) of omegaUp.
- [x] The tests were executed and all of them passed.
- [x] If you are creating a feature, the new tests were added.
- [x] If the change is large (> 200 lines), this PR was split into various Pull Requests. It's preferred to create one PR for changes in controllers + unit tests in PHPUnit,  and then another Pull Request for UI + tests in Jest, Cypress or both.

PR #10008, opened 2026-07-14
add cron admin dashboard

### LINKED ISSUE #10004: [FEATURE] Admin dashboard page for cron jobs
### Checklist

### Context

there is no page anywhere to see cron health, recent runs, durations or errors, so checking on the jobs means going into the logs by hand.

this adds an admin page at `/admin/crons` showing each job's latest status, the recent runs with durations, a per phase drilldown and the error output.

**expected outcome:** the admin dashboard page for the cron jobs.

### Order

part of the cron control plane epic #9992. #9993 (migration), #9994 (daos) and #9999 (admin api) are all resolved and merged now, so this is the ui on top of the api and nothing is blocking it any more.

PrasannaMishra001:
# Description

adds the `/admin/crons` dashboard.

`Admin::getCronsForTypeScript` serves the page payload behind the same system admin check the cron api already uses, `crons.php` is the entrypoint and `crons.ts` parses the payload into `Crons.vue`, which renders two tables. the first has one row per registered job with its schedule and the status and time of its latest run. the second has one row per recent run, and clicking a row expands it into that run's per phase timings and its error output.

wired through webpack, the nginx rewrite and the four lang files. the component is covered by jest and the payload by phpunit.

Fixes: #10004

# Comments

part of the cron control plane epic #9992. everything this used to sit on top of has merged, #9995 for the tables, #9996 for the daos and #10000 for the api, so there is no remaining dependency and the diff here is only the dashboard. it unblocks #10021, #10023, #10048, #10052 and #10069.

a few things worth looking at while reviewing.

the page payload is `apiGetCrons` itself rather than a second copy of it, so the page and the api cannot drift apart on the shape of a job or a run, or on how many runs the page gets. `getCronsForTypeScript` still runs its own `ensureMainUserIdentity` and `Authorization::isSystemAdmin` before it delegates, which is the shape `Badge::getBadgeListForTypeScript` already uses. there is a test for the non admin case and another for the logged out one, and taking the check out makes both of them fail.

the component reuses what already exists instead of rolling its own, so dates go through `time.formatDateTime` and the header is the same `text-white bg-primary card-header` the other admin pages use. one thing to know about the second table is that it holds the fifty most recent runs across every job, which is exactly what the api already returns, and the first table reads each job's latest status out of that same window.

on the last checkbox, 175 of the 662 lines are the generated lang files and `api_types.ts`, and another 239 are the jest and phpunit tests, which leaves 248 lines of actual page. the schema, the daos and the api already went out as their own pull requests, so what is left is one page, and splitting it again would only leave a controller method with nothing to render it.

# Checklist:


pabo99 (commented):
How about splitting this into multiple PRs? The first one could include all the database schema changes (without updating the dao schema yet, that could be the second PR). I think that would make things much easier to review

PrasannaMishra001:
@pabo99 actually i already split all of these into separate focused PRs, one per sub issue. the full order they should be reviewed and merged in is listed in #9992, following that should make the reviews a lot easier.

they are stacked though, so until a predecessor merges each downstream PR still carries the earlier commits in its diff (that is why this one looks large). once the one before it merges and i rebase, the diff here drops to just the dashboard change.

so it is easiest to start from the bottom, #9995 (the tables migration, issue #9993) and work up. i also already separated the schema and the dao before, #9995 is the migration and schema.sql only and #9996 is the generated dao as linter problem i know from past experience.

PrasannaMishra001:
@pabo99 ok so as it may bother in future as well, so i have rather decided to keep the following ones as draft so once the preceding one db merges .only #9995 (tables) is open for review now, the other 7 are draft now

frontend/www/js/omegaup/components/admin/Crons.vue
```diff
@@ -0,0 +1,197 @@
...
+            <th>{{ T.cronControlPlaneSchedule }}</th>
+            <th>{{ T.cronControlPlaneLastStatus }}</th>
+            <th>{{ T.cronControlPlaneLastRun }}</th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr v-for="job in jobs" :key="job.name">
+            <td>{{ job.name }}</td>
+            <td>
+              <code v-if="job.schedule">{{ job.schedule }}</code>
```
pabo99: Would it be possible to add a computed property to display this field in a more human-friendly format? I think we can display both values

PrasannaMishra001:
@pabo99 this is ready to be reviewed further. the split you asked for happened and merged as #9995 for the tables, #9996 for the daos and #10000 for the api are all merged, only the dashboard itself is left.

PrasannaMishra001:
@pabo99 ready for review.it also unblocks the rerun, e2e and dashboard improvement prs behind it.

frontend/templates/es.lang
```diff
@@ -769,6 +769,17 @@ createProblemInteractiveGuideWelcome = "Bienvenido, vamos a crear un problema"
 createTeamsGroup = "Crear nuevo grupo de equipos"
 createZipFileForProblem = "Usar la herramienta Creador de problemas"
 createZipFileForProblemDesc = "Crea el zip de un problema paso a paso"
+cronControlPlaneDuration = "Duración"
+cronControlPlaneJobsHeading = "Trabajos"
```
pabo99: Let's keep this translation as _Jobs_, since the term is clearer and commonly understood even in Spanish. It might be worth keeping it as Jobs in Portuguese as well.
```suggestion
cronControlPlaneJobsHeading = "Jobs"
```

frontend/templates/es.lang
```diff
@@ -1128,6 +1139,7 @@ notificationsMarkAllAsRead = "Marcar todo como leído"
 notificationsNoNewNotifications = "No hay nuevas notificaciones."
 officialSolutionsInfo = "¿Cómo funcionan las soluciones de problemas?"
 omegaUpTitleMyDependents = "Mis dependientes"
+omegaupTitleAdminCrons = "Trabajos programados"
```
pabo99: Same case here:
```suggestion
omegaupTitleAdminCrons = "Cronjobs"
```
or
```suggestion
omegaupTitleAdminCrons = "Administración de Cronjobs"
```

frontend/server/src/Controllers/Admin.php
```diff
@@ -498,4 +499,24 @@ public static function apiGetCronRun(\OmegaUp\Request $r): array {
...
+            throw new \OmegaUp\Exceptions\ForbiddenAccessException();
+        }
+
+        return [
+            'entrypoint' => 'admin_crons',
+            'templateProperties' => [
+                'title' => new \OmegaUp\TranslationString(
+                    'omegaupTitleAdminCrons'
+                ),
+                'payload' => self::apiGetCrons($r),
```
pabo99: In general, we try to avoid calling API endpoints directly from other controllers, since this requires passing the entire request as a parameter and makes it harder for `Psalm` to properly infer and validate types

Would it be possible to add a helper function in `apiGetCrons` that retrieves both the jobs and the runs? We could then reuse that helper here as well

frontend/tests/controllers/CronControlPlaneAdminTest.php
```diff
@@ -119,4 +119,73 @@ public function testGetCronRunReturnsNullForUnknownRun() {
...
+        ['identity' => $identity] = \OmegaUp\Test\Factories\User::createAdminUser();
+        $login = \OmegaUp\Test\ControllerTestCase::login($identity);
+
+        $run = new \OmegaUp\DAO\VO\CronRuns([
+            'name' => 'update_ranks.py',
+            'status' => 'success',
+            'started_at' => new \OmegaUp\Timestamp(\OmegaUp\Time::get()),
+            'duration_seconds' => 1.5,
+            'rows_affected' => 7,
+        ]);
```
pabo99, line 176: What if you use enums for cronjob names and statuses? Since these are known sets of values, this would help prevent typos and keep them consistent across the codebase

pabo99 (commented):
Apologies! I forgot to submit the review earlier.

The change is quite extensive, and I think there’s a good opportunity to split it up a bit, which would make the review process easier and faster

What do you think about focusing this PR on the server-side changes and moving everything related to the UI to a separate PR? For the new PR, it would also be really helpful if you could add a Storybook story for the new component

pabo99 (approved):
Thank you for addressing all the requested changes. 

I’m sorry the human-readable date suggestion ended up requiring more work and complexity than I initially expected. I appreciate the effort you put into implementing it properly. 

LGTM!

PR #10011, opened 2026-07-15
add ranking audit guardrail

### LINKED ISSUE #10010: [FEATURE] Pre-publish guardrail for the ranking job
### Checklist

PARENT ISSUE:#9992

### Context

a bad ranking run can currently publish an empty, negative scored or drastically smaller ranking right over the previous good one.

this adds an in transaction audit that checks the freshly computed ranking before it is published, and if it looks wrong (empty, negative scores, or too many previously ranked users dropped) it raises and rolls back so yesterday's ranking stays live.

**expected outcome:** a prepublish guardrail on update_ranks.

### Order

part of the cron control plane epic #9992. #9993 (migration), then #9997 (runner), then #10002 (update_ranks) should be resolved and merged first in that order. it sits alongside #10007 on the same update_ranks base.

PrasannaMishra001:
# Description

audits the freshly computed ranking inside the same transaction before it is published. if the new `User_Rank` is empty, has negative scores or drops more than `--max-rank-churn` (default 0.5) of the previously ranked users, `RankingAuditError` is raised, the transaction rolls back and yesterday's ranking stays live. unit tested with 6 tests.

Fixes: #10010

# Comments

part of the cron control plane epic #9992. everything this was waiting on is merged now (#9995 for the tables, #9998 for the runner and #10003 for update_ranks), so it rebases straight onto main and the diff is down to the two update_ranks files.

one thing worth flagging in the test: `stuff/cron` and `stuff/pipelines` both ship a package called `database`, and whichever one is imported first wins for the whole pytest session. importing `update_ranks` binds the cron one, which then makes the pipelines tests fail to import theirs, so the test drops it from `sys.modules` right after the import. #9883 fixes that properly by scoping the cron imports under the `cron` package, and this workaround comes out as soon as it lands.

# Checklist:


PrasannaMishra001:
@pabo99 ready for review, it checks the freshly computed ranking inside the same transaction and rolls back rather than publishing an empty or collapsed one. i forced the failure and checked the `User_Rank` checksum is identical after.

stuff/cron/update_ranks.py
```diff
@@ -826,6 +826,47 @@ def debug_coder_of_the_month_candidates(
     logging.info(log_message)
 
 
+class RankingAuditError(Exception):
+    '''Raised when the freshly computed ranking fails a sanity check.'''
+
+
+def _count_user_rank(cur: mysql.connector.cursor.MySQLCursorDict) -> int:
+    '''Returns the current number of rows in `User_Rank`.'''
+    cur.execute('SELECT COUNT(*) AS `n` FROM `User_Rank`;')
```
pabo99: Should this count only actual user-ranking rows, e.g. `WHERE ranking IS NOT NULL`? `User_Rank` also stores author-ranking-only rows, and this audit runs before `update_author_rank()`, so those rows could inflate `previous_count` and cause a false churn failure

PrasannaMishra001: yes i changed it now because update_user_rank sweeps yesterday's author rows before the audit runs, so previous_count will have them and new_count does not

_count_user_rank filters ranking `IS NOT NULL` now, so that both sides count the same thing. also added tests for both row types.

pabo99 (approved):
Great!

PR #10013, opened 2026-07-15
[Milestone 4] Import uploaded Creator zip into the Problem Creator modal and regenerate on close

### LINKED ISSUE #10012: [Milestone 4] Wire up the zip import flow: upload → parse → auto-open Creator → regenerate on close
Part of #9824 (Integrate Problem Creator). Builds on #9859 (Milestone 1), #9891 (Milestone 2), and #9986 (Milestone 3).

Milestone 3 made the Creator → submit path work: closing the modal generates a zip that is attached as `problem_contents` on form submit. However, the two creation methods are still disconnected — a zip previously generated by the Problem Creator cannot be loaded back into the Creator from the create form.
Users who saved their work as a zip have no way to continue editing it through the modal.

### Goal
When a user uploads a Creator-generated zip (one containing a `cdp.data` store snapshot) via the "I have a zip file" option, the Problem Creator should take over as the source of truth:

1. The zip is parsed and the Creator store is populated from it.
2. The creation method automatically switches to "Problem creator" and the modal opens pre-populated, so the user can review/edit the problem.
3. On close, a fresh zip is regenerated from the store (Milestone 3 pipeline), so what gets submitted always reflects the latest edits — never the originally uploaded file.

Zips that are not Creator-generated (standard omegaUp zips without `cdp.data`) keep the plain upload flow untouched; parsing those into the Creator is Milestone 5.

Everything stays behind the `problem_creation_method_selector` experiment; with the flag off, the form behaves exactly as today.

shauryag2002:
# Description

This PR connects the two creation methods: uploading a Creator-generated zip via "I have a zip file" now parses it, auto-opens the Problem Creator modal pre-populated with its contents, and makes the Creator the single source of
truth — the zip that is submitted is always regenerated from the Creator store on close (Milestone 3 pipeline), never the originally uploaded file.

**Import flow (Form.vue):**
- `onUploadFile` now inspects the uploaded zip. If it contains a `cdp.data` store snapshot (i.e. it was generated by the Problem Creator), the form switches the creation method to "Problem creator", clears any stale generated
  blob, opens the modal, and forwards the file to the embedded Creator.
- Zips without `cdp.data` (standard omegaUp zips) are left completely alone — the plain native upload flow proceeds unchanged. Parsing those into the Creator is Milestone 5.

**Reusable import API (Header.vue → Creator.vue → CreatorWrapper.vue):**
- `retrieveStore()` in Header.vue is refactored into a public `async importZipFile(zipFile: File): Promise<boolean>` with the same cdp.data parsing and error handling (`problemCreatorZipFileIsNotComplete` / `problemCreatorZipFileIsNotValid`). The in-modal "Load Problem" flow reuses it, so the standalone `/problem/creator` page behavior is unchanged.
- `Creator.vue` and `CreatorWrapper.vue` expose thin `importZipFile()` delegates through the existing refs, mirroring `generateZip()`/`saveDraft()` from Milestone 3.

**Round trip guarantee:** import → store populated → user edits in the modal → close → drafts persisted + fresh zip generated → submit attaches the regenerated zip. Removing a test case after importing is reflected in the
submitted zip because generation always starts from a fresh JSZip instance.

**Feature flag:** the import path only runs when `showCreationMethodSelector` is enabled and the form is in create (not update) mode. With the `problem_creation_method_selector` experiment off, `onUploadFile` behaves
exactly as before.

**Tests** (3 new, all existing pass):
- Uploading a Creator zip switches the method to Creator and opens the modal.
- Uploading a standard (non-`cdp.data`) zip keeps the plain zip flow.
- The uploaded file is forwarded to the embedded Creator's `importZipFile`.

[Screencast from 2026-07-15 22-55-35.webm](https://github.com/user-attachments/assets/2d58af7d-cd4b-4cf8-9b8b-a238be1f09e5)

Fixes: #10012 

# Comments
- Standard omegaUp zip parsing (no `cdp.data`) into the Creator is Milestone 5.
- A Cypress e2e test covering the full Creator modal → submit flow (as requested in the Milestone 3 review) will be a separate follow-up PR.
- Intro.js updates remain out of scope for this milestone.

# Checklist:

- [x] The code follows the [coding guidelines] https://github.com/omegaup/omegaup/blob/main/frontend/www/docs/Coding-guidelines.md) of omegaUp.


frontend/www/js/omegaup/components/problem/creator/Header.vue
```diff
@@ -122,39 +122,39 @@ export default class Header extends Vue {
...
-          if (storeData.casesStore) {
-            this.$store.commit('casesStore/replaceState', storeData.casesStore);
-          }
-        });
-      })
-      .catch(() => {
-        ui.error(T.problemCreatorZipFileIsNotValid);
+  }
+
+  async importZipFile(zipFile: File): Promise<boolean> {
```
pabo99, line 131: This function can return `false` when `cdp.data` is invalid, but that result is currently ignored. This leaves the form in Creator mode even though the import failed. Could we handle that case and return to the ZIP flow?

shauryag2002: @pabo99 Right! Thanks for figuring out.
 Here's what was happening and how it's fixed:
### The Bug
In Form.vue → importZipIntoCreator(), when a user uploads a ZIP file in the "Zip" flow that contains a cdp.data file, the code:
1. Switches currentCreationMethod to Creator
2. Opens the problem creator modal
3. Calls importZipFile() to populate the creator with the zip data

But importZipFile() can return false (e.g., if cdp.data contains invalid JSON, or a required field is missing). The return value was ignored, leaving the user stuck in Creator mode with an empty/broken modal even though the import failed.

### The Fix
Form.vue (line 865-869): Check the return value of importZipFile(). If it returns false (or undefined), revert by closing the modal and switching currentCreationMethod back to Zip.

`Form.test.ts`:
- Fixed the existing "pre-populated" test to properly mock the problemCreatorRef (it was missing the mock, which now matters since we check the return value)
- Added a new test: "Should revert to the zip flow when the creator import fails" — verifies that on failure, `currentCreationMethod` reverts to Zip and the modal closes

pabo99 (approved):
Thank you so much, @shauryag2002 !

PR #10014, opened 2026-07-16
Fix grader language detector rejecting single-pattern matches

### LINKED ISSUE #10006: bug: Language detector penalty silently kills single-pattern matches for C, Java, Lua, and 7 other languages
### Description

The penalty formula `Math.min(1, pat.patterns.length / 4)` causes any language with fewer than 4 patterns (C, Java, Kotlin, C#, Go, Rust, Haskell, Ruby, Pascal, Lua) to produce exactly 25% confidence when only one pattern matches, which is always below the 30% rejection threshold. The math is: (1/N) * (N/4) * 100 = 25% for any N. Before this change, 1-of-3 matches gave 33% confidence (detected); 1-of-2 gave 50% (detected). After, both give 25% (rejected).

### Steps to reproduce

1. Open `frontend/www/js/omegaup/grader/util.ts` around line 192
2. C code using `printf` without `#include <stdio.h>`, Java code with only `public class`, Lua code with only `local` variables, and similar single-indicator code for 10 affected languages returns null (not detected) instead of the correct language.

### Expected behavior

The confidence formula `(bestScore / bestPriority) * 100` directly embeds the penalty. For N < 4 patterns with 1 match: score = (1/N) * priority * (N/4), so confidence = score/priority * 100 = (N/4N) * 100 = 25% — algebraically constant regardless of N. The penalty was intended to balance competition between languages in score comparison, but it also deflates confidence below the detection threshold for all single-pattern languages-with-fewer-than-4-patterns cases. should be handled correctly.

### Actual behavior

C code using `printf` without `#include <stdio.h>`, Java code with only `public class`, Lua code with only `local` variables, and similar single-indicator code for 10 affected languages returns null (not detected) instead of the correct language.

deepak0x:
The confidence used to gate detection was derived from the penalty-weighted score, so `confidence = (matches / N) * min(1, N/4) * 100`. For any language with fewer than four patterns, a single match collapses to a constant 25% regardless of N, which is below the 30% rejection threshold. This silently drops C, Java, Kotlin, C#, Go, Rust, Haskell, Ruby, Pascal, and Lua whenever only one indicator is present (e.g. C using `printf` with no include, a bare `public class`, Lua with only `local` declarations).

The penalty is still applied to the score used to pick the best language between competitors. Confidence is now taken from the raw match ratio `matches / N`, restoring the pre-regression values (1-of-2 = 50%, 1-of-3 = 33%).

Added unit tests in `util.test.ts` covering the single-pattern C, Java, and Lua cases, which fail on `main` and pass with this change. Full suite (11 tests) passes.

Closes #10006


pabo99 (approved):
Thank you, @deepak0x !

PR #10025, opened 2026-07-16
make view unavailable entrypoint generic

### LINKED ISSUE #10017: [FEATURE] Make the view unavailable entrypoint generic and driven by the payload
### Checklist

### Context

right now the disabled grader view is rendered by a small grader specific entrypoint `grader/unavailable.ts` that mounts the shared `ViewUnavailable` component with a hardcoded description. as @pabo99 noted on #8938, if we start disabling more non grader views later we would end up duplicating a file like this once per view, which is not great.

this is the follow up to decouple that file from the grader. the idea is one generic entrypoint that mounts `ViewUnavailable` and reads the description (and maybe the title and icon) from the server payload, so any view that needs to show an unavailable state can reuse it just by passing its own text through the payload, without a new file per view.

**expected :** a generic view unavailable entrypoint driven by the payload, the grader switched over to it and the grader specific `grader/unavailable.ts` removed.

PrasannaMishra001:
# Description

follow up to #8938. the disabled grader view was rendered by a grader specific entrypoint `grader/unavailable.ts` that mounted the shared `ViewUnavailable` component with a hardcoded description, so any future non grader view we disable would need its own copy of that file.

this decouples it. there is now one generic `common/view_unavailable.ts` entrypoint that mounts `ViewUnavailable` and reads the description from the payload, and `getGraderForTypeScript` resolves the grader's message and passes it through that payload while pointing at the generic entrypoint. the grader specific `grader/unavailable.ts` is removed. any view that needs an unavailable state can now reuse this just by passing its own description, without a new file per view.

the rendered disabled grader view stays the same, this only moves where the description comes from.

Fixes: #10017

# Checklist:


pabo99 (approved):
Thank you, @PrasannaMishra001 !

Issue #10049, opened 2026-07-26
[FEATURE] Make the recommendation model training scheduled, recorded and guarded


PrasannaMishra001:
### Checklist

- [x] I have searched existing issues and confirmed this is not a duplicate
- [x] I have discussed this on [GitHub Discussions](https://github.com/omegaup/omegaup/discussions) or [Discord](https://discord.gg/K3JFd9d3wk) and received positive feedback

### Context

Currently platform suggests "what to solve next" from a recommendation model that `build_problem_rec_model.py` trains. today that training is run by hand as it is not registered anywhere as a scheduled job, there is no record of when it last ran or how good each model was and while it will not save a model below a fixed quality bar it has no protection against a model that clears the bar but is clearly worse than the last good one. so a stale or worse model could go live without anyone noticing.

the quality bar itself is also thin. the script reduces a whole model to one score and nothing else is measured, so when that score moves there is no way to tell whether the recommendations got worse or only their ordering did, and no way to sanity check the number against anything.

this will be built on the cron control plane and turns the training into a scheduled, observed and guarded job. I will raise these as a few small PRs linked to this issue:-

- register the job and store its history: a `Recommendation_Model_Runs` table with its dao and `build_problem_rec_model.py` registered in the cron registry with a weekly schedule.
- measure, record and guard each run: the model is scored with the standard ranking metrics (precision@k, recall@k, map@k and ndcg@k) over the held out test users, every training run records its map score, dataset size and parameters, and a guardrail refuses to publish a model that is below the bar or that regresses too far below the last published one, keeping the previous good model instead.
- frontend part : the model quality shows up on the /admin/crons dashboard so the trend is visible next to the jobs.

**expected :** the recommendation training becomes a proper scheduled job, its quality is measured in terms anyone can check and is visible over time, and a bad model can never silently replace a good one.

erikurt9:
Hi! I would like to work on this feature, may i?

PrasannaMishra001:
@erikurt9 Thanks for your interest but  this is part of my ongoing gsoc project and i am working on this. Feel free to look at the other open issues to contribute to :)

pabo99:
This looks like a reasonable feature, and I like the idea of splitting it into smaller PRs.

I only have a couple of questions before assigning it:

- Should the comparison be made against the last successful run or the last published model?
- If a model is rejected, should its metrics still be stored in `Recommendation_Model_Runs` so we can review them later?

Other than that, the proposal looks good to me

[...]

PR #10050, opened 2026-07-27
add recommendation model runs table

PrasannaMishra001:
# Description

adds the schema for the recommendation model training history and registers the job that will write to it.

`Recommendation_Model_Runs` stores one row per training run: the map score, the size of the dataset it trained on, the three training parameters, where the model was written, whether it was published, and if it was not, why. the index on `(published, created_at)` is there so looking up the last published model is a single seek, which is what the guardrail does on every run.

it also seeds `build_problem_rec_model.py` into the `Cron_Jobs` registry with a weekly schedule, sunday at 04:00, so the training registers as a scheduled job instead of something run by hand. the script is already on main, so the row points at something real.

this is a forward only migration and it adds no dao. the dao comes in a separate pr, per the migration and dao split.

Part of #10049.

# Comments

`cron_run_id` is deliberately a loose reference with no foreign key. `Cron_Runs` is an unbounded history table, and `Cron_Runs.name` is already denormalised on purpose so the history survives changes to the registry. a training run should outlive the cron run that produced it for the same reason, so the column is nullable and unconstrained.

`map_score` is not nullable because a skipped run still has a score. the guardrail trains first and then decides whether to publish, so `skip_reason` explains why a scored model was not published rather than standing in for a missing score.

the migration is 00281. 00280 is taken on main by `00280_add_rekarel_language.sql`.

# Checklist:


PrasannaMishra001:
@pabo99 this is also ready for review, adds the Recommendation_Model_Runs table and registers build_problem_rec_model.py in the registry on a weekly schedule, so it is scheduled at all for the first time.

PrasannaMishra001:
@pabo99 renumbered table in this to 00281 after 00280 was taken by #10119. also, https://github.com/omegaup/omegaup/pull/10051 conflicts resolved so can be reviewed after this.

pabo99 (commented):
While reviewing #10051, I noticed that `rng_seed` determines the train/test split but is not stored in `Recommendation_Model_Runs`. Could you add it as a nullable column? Without it, training runs cannot be reproduced reliably or compared accurately

pabo99 (approved):
Thanks!

PR #10051, opened 2026-07-27
record and guard recommendation training

PrasannaMishra001:
# Description

stops a bad recommendation model replacing a good one.

each training run records its MAP score, its parameters and its seed into `Recommendation_Model_Runs`. a new model is only published if it clears the floor and has not regressed against the last model published to the same artifact; otherwise the previous model stays live, the reason is recorded, and the job exits non zero the way `assign_badges.py` and `aggregate_feedback.py` already do.

the score is the real MAP@k from `evaluate_metrics()` now. the old `evaluate()` awarded credit only when a recommendation matched at the same position, which is not MAP, so the column name, the flag and the guardrail all described something the code was not measuring. its own TODO asked for this. a prediction the model cannot answer scores zero rather than being dropped from the average, which was inflating every metric by about a fifth on the fixture.

that moves the numbers, so `--min-map-score` went from 0.3 to 0.05 and the runbook now calls it a floor for a broken model rather than a quality bar, since the regression check is what carries quality once there is history. it wants recalibrating from the first real runs.

part of #10049.

# Comments

waits on nothing, #10050 merged.

the seed alone did not give reproducibility: the runs were ordered by `(identity_id, time)` with no tie break while the model weights problems by position, so two solves in the same second could move the score. the dev database has two tied groups in 66 submissions. `problem_id` is the tie break now and the sqlite loader orders by `rowid` instead of relying on file order.

# Checklist:


pabo99:
Can you resolve the conflicts that appeared?

PrasannaMishra001:
@pabo99 this one can be reviewed when you get time.

stuff/cron/build_problem_rec_model.py
```diff
@@ -316,8 +383,167 @@ def evaluate(self, k: Optional[int] = None) -> float:
...
+        last_published_map = (get_last_published_map(dbconn)
+                              if dbconn is not None else None)
+        published, skip_reason = should_publish(score, args.min_map_score,
+                                                last_published_map,
+                                                args.max_map_regression)
+        if published:
+            model.save(args.output)
+        else:
+            logging.error('Model NOT saved. %s', skip_reason)
+            cron_run.mark_failure()
```
pabo99, line 513: `mark_failure()` only changes the status recorded in `Cron_Runs`; it doesn't raise an exception or affect the process exit code. Consequently, a rejected model still makes this cron exit with status 0, so Kubernetes and any process-level alerting will consider the job successful. 

Please propagate the failure through the process exit status, as described in the runbook

stuff/cron/build_problem_rec_model.py
```diff
@@ -316,8 +383,167 @@ def evaluate(self, k: Optional[int] = None) -> float:
...
+
+
+def get_last_published_map(dbconn: lib.db.Connection) -> Optional[float]:
+    '''Returns the MAP score of the most recently published model, or None.'''
+    with dbconn.cursor() as cur:
+        cur.execute(
+            '''
+            SELECT `map_score`
+            FROM `Recommendation_Model_Runs`
+            WHERE `published` = 1
```
pabo99: This query considers every row marked as published, regardless of its `output_path`. For example, the runbook’s deterministic command writes `/tmp/model.db` without `--no-track`; that run would become the baseline even though it never replaced the production artifact

A sufficiently high manual score could then block subsequent production models

Please scope the baseline to the artifact being replaced, or ensure non-production runs cannot be recorded as published

stuff/cron/build_problem_rec_model.py
```diff
@@ -316,8 +383,167 @@ def evaluate(self, k: Optional[int] = None) -> float:
...
+        args: argparse.Namespace,
+        runs: pd.DataFrame,
+) -> None:
+    '''Trains, publishes if it passes the guardrail, and records the run.'''
+    model = Model(
+        TrainingConfig(train_fraction=args.train_fraction,
+                       rng_seed=args.rng_seed,
+                       num_followups=args.num_followups,
+                       followup_decay=args.followup_decay), runs)
+    score = model.evaluate()
```
pabo99: `model.evaluate()` is not MAP@k: it only awards credit when a recommendation matches the expected problem at the same position. The standard MAP@k value is calculated below as `metrics['map']`, but it is only logged. As a result, `map_score`, `--min-map-score`, and the regression guardrail operate on a different metric than their names, schema comments, and runbook describe

Please use the actual MAP@k value or rename and document the legacy score consistently

stuff/cron/build_problem_rec_model.py
```diff
@@ -316,8 +383,167 @@ def evaluate(self, k: Optional[int] = None) -> float:
...
+            'ndcg': ndcg_at_k,
+        }
+        totals: DefaultDict[str, float] = collections.defaultdict(float)
+        predictions = 0
+        for problems in self.test_ac_map.values():
+            for i in range(1, len(problems)):
+                recs = self.recommend(problems[i - 1], set(problems[:i - 1]),
+                                      k)
+                expected = problems[i:i + k]
+                if not recs or not expected:
```
pabo99: Skipping when `recs` is empty excludes every case the model can't serve from the metric denominator. This can substantially inflate all four reported metrics, for example, a model covering only 1% of the test cases could still report perfect scores if it gets that 1% right. Since `expected` is non-empty in this loop, missing recommendations should increment the prediction count and contribute zero to each metric

pabo99 (commented):
It would be worth adding regression tests for each of the cases mentioned in the inline comments

stuff/cron/build_problem_rec_model.py
pabo99: The process now exits with status 1 when the model is held back, which addresses the functional issue. However, the added test only verifies the return value from `train_and_publish()` and doesn't exercise `main()` or assert the resulting process exit code. Please add a regression test that verifies `main()` raises `SystemExit` with code 1 when the model is rejected

pabo99: The process now exits with status 1 when the model is held back, which addresses the functional issue. However, the added test only verifies the return value from `train_and_publish()` and doesn't exercise `main()` or assert the resulting process exit code. Please add a regression test that verifies `main()` raises `SystemExit` with code 1 when the model is rejected

PrasannaMishra001: done, added `test_main_exits_nonzero_when_model_is_rejected` which calls `main()` --min-map-score 0.99 against the fixture. it raises `SystemExit` with code=1 as required which matches the path after `train_and_publish` returns a failure.

stuff/cron/build_problem_rec_model.py
PrasannaMishra001: already scoped this as `get_last_published_map` takes `output_path` and the `WHERE` clause filters on it. `test_the_baseline_is_scoped_to_one_artifact` raises this query and includes `output_path = %s`.

stuff/cron/build_problem_rec_model.py
PrasannaMishra001: switched to the real MAP@k vakue and the legacy evaluate() is dropped. `test_the_guardrail_scores_the_model_with_map` tests this by raising the score which does not match precision, recall or ndcg.

stuff/cron/build_problem_rec_model.py
PrasannaMishra001: already done last week, so prediction now increments the denominator and contributes zero to every metric as needed.  `test_a_prediction_the_model_cannot_serve_scores_zero` sets up a model that can only serve one of two test users and raised MAP is 0.5, not 1.0.

pabo99 (commented):
Could you resolve the conflicts so we can do another review?

pabo99 (approved):
Thanks!

PR #10062, opened 2026-07-29
[Milestone 5] Add E2E Cypress tests for Creator submit flow and Introjs tour integration

### LINKED ISSUE #10061: [Milestone 5] Add E2E Cypress tests for Creator submit flow and Intro.js tour integration
## Problem Description
With the core Problem Creator modal and zip synchronization flows completed in previous milestones, two key deliverables were required to complete Milestone 5:
1. **Interactive Guide (Intro.js)**: The existing `Form.vue` interactive intro tour needed updates to guide users through the new Problem Creator workflow (including creation method selection and opening the modal) without causing UI overlaps or background backdrop blocking when the Creator modal is active.
2. **End-to-End Regression Protection**: Comprehensive Cypress E2E test coverage was needed for the full Problem Creator modal creation and submission flow under the `PROBLEM_CREATION_METHOD_SELECTOR` experiment flag.

## Key Changes
- **Intro.js Tour Workflow & Dynamic DOM Handling**:
  - Updated `Form.vue` to include new guide steps for problem creation method selection and launching the Problem Creator.
  - Wrapped tour initialization in `$nextTick` to ensure target DOM elements are rendered before tour focus.
  - Added programmatic teardown/exit (`this.createProblemIntro?.exit()`) when opening the Problem Creator modal to prevent tour tooltips from floating over the backdrop.
- **Tab-Level Tour Suppression**:
  - Propagated `hideSaveButton` down to `CodeTab.vue`, `CasesTab.vue`, and `SolutionTab.vue` to suppress embedded sub-tab Intro.js tours when inside the Creator modal.
- **E2E Cypress Test Suite**:
  - Expanded `cypress/e2e/problem_creator.cy.ts` to test the full Creator modal lifecycle: setting fields, zip generation, auto-attaching to form submit payload, and form submission validation under the `PROBLEM_CREATION_METHOD_SELECTOR` flag.
- **I18n Localization**:
  - Added translation strings (`createProblemInteractiveGuideCreationMethod`, `createProblemInteractiveGuideOpenCreator`) across `en`, `es`, `pt`, and `pseudo` language definitions and compiled language bundles via `./stuff/lint.sh`.

## Verification
- `./stuff/lint.sh` ran successfully with `OK`.
- All Cypress E2E tests in `cypress/e2e/problem_creator.cy.ts` pass cleanly.

shauryag2002:
# Description

This pull request completes **Milestone 5** of the Problem Creator integration by adding end-to-end Cypress test coverage for the full Creator modal submit flow and updating the Intro.js interactive guide for the new problem creation experience.

[Screencast from 2026-07-29 17-54-51.webm](https://github.com/user-attachments/assets/1e0170fb-30c9-475d-8c25-eaa9a936590c)

### Key Changes:
- **Cypress E2E Test Coverage**:
  - Added E2E tests covering the complete Problem Creator modal creation and submission flow under the `problem_creation_method_selector` experiment flag.
  - Verifies zip generation in the Creator, attachment of the generated zip to the `problem_contents` form field (via `DataTransfer` API / fallback), and successful form submission to prevent regressions (as requested in review feedback on #9987).
- **Intro.js Interactive Guide Integration**:
  - **Dynamic Tour Steps**: Updated `Form.vue` to dynamically construct guide steps respecting the `problem_creation_method_selector` feature flag (showing the creation method toggle and "Open Problem Creator" button when enabled, or standard zip upload when disabled).
  - **Clean Modal Transition**: Ensured the Intro.js tour programmatically exits when the Problem Creator modal opens to prevent tour tooltips from floating over the modal overlay.
  - **Tab Tour Suppression**: Prevented tab-level intro tours (`CodeTab`, `CasesTab`, `SolutionTab`) from triggering when embedded inside the modal context (`hideSaveButton = true`).
  - **Localization (i18n)**: Added translated strings for the new tour steps across supported languages (`en`, `es`, `pt`, `pseudo`).

Fixes: #10061 

---

# Checklist:
- [x] New tests were added to cover the new functionality.
- [x] The PR line count change is kept under 200 lines to maintain readability.


cypress/e2e/problem_creator.cy.ts
```diff
@@ -296,4 +296,31 @@ describe('Problem creator Test', () => {
       });
     });
   });
+
+  it('Should submit create problem using creator-generated zip in feature-flag flow', () => {
+    cy.login(loginOptions);
+    cy.setCookie('has-visited-create-problem', true.toString());
+
+    cy.visit('/problem/new/?experiments=problem_creation_method_selector');
```
pabo99: Could you enable this experiment for the test user through the API instead of passing it in the URL? The URL flow requires an HMAC hash based on `OMEGAUP_EXPERIMENT_SECRET`, which may differ between local and CI. Using `/api/user/addExperiment/` would make the test less environment-dependent.

cypress/e2e/problem_creator.cy.ts
```diff
@@ -296,4 +296,31 @@ describe('Problem creator Test', () => {
...
+
+    cy.get('[name="title"]').type('Cypress Creator Submit').blur();
+    cy.get('[name="source"]').type('Cypress');
```
pabo99: The test does not fill all required fields before submitting. In create mode, `select[name="problem-level"]` is required, so the browser can block the submit event before `handleFormSubmit()` runs. Could you select a problem level before clicking submit?
```ts
    cy.get('[name="problem-level"]').select(0);
```

cypress/e2e/problem_creator.cy.ts
```diff
@@ -296,4 +296,31 @@ describe('Problem creator Test', () => {
...
+
+    cy.get('.introjs-creation-method .btn-group button').first().click();
+    cy.get('.introjs-open-creator button').click();
+    cy.get('[data-problem-creator-close]').click();
```
pabo99, line 351: This can race with the async zip generation triggered when the modal closes. `closeProblemCreatorModal()` calls `generateZip()`, but `handleCreatorZipGeneration()` only sets `creatorGeneratedZipBlob` after `generateAsync(...)` resolves. Could you expose a visible/observable “zip ready” state in `Form.vue` and wait for it before submitting?

For example, the hidden Creator input could expose:
```ts
:data-problem-creator-zip-ready="
  creatorGeneratedZipBlob !== null && !isGeneratingCreatorZip ? 'true' : 'false'
"
```
Then the test can wait for it:
```ts
const creatorProblemContentsInput =
  '.introjs-open-creator input[name="problem_contents"]';

cy.get(creatorProblemContentsInput, { timeout: 10000 }).should(
  'have.attr',
  'data-problem-creator-zip-ready',
  'true',
);
```
`creatorProblemContentsInput` could also be reused in the final assertion that currently queries the hidden input again.

cypress/e2e/problem_creator.cy.ts
```diff
@@ -296,4 +296,31 @@ describe('Problem creator Test', () => {
...
+
+    cy.get('form').then(($form) => {
+      $form.on('submit', (e) => e.preventDefault());
```
pabo99: This prevents the form from exercising the real create-problem submit path. The test currently only verifies that the hidden file input receives a File, but it would still pass if the backend rejects the generated zip or if the post-submit redirect is broken. Could you let the form submit for real and assert the successful problem URL?

The solution would be to remove the `preventDefault()` handler, fill the required create-problem fields (at least `problem-level`, and preferably a public tag following the existing `cy.createProblem` pattern), then let the form submit normally and assert the successful redirect/URL. That way this test covers the real create-problem path instead of only checking that the hidden input receives a File.

pabo99 (commented):
Great job! Just a few suggestions

cypress/e2e/problem_creator.cy.ts
shauryag2002: Replaced the URL HMAC hash with cy.loginAdmin() + POST /api/user/addExperiment/ to enable problem_creation_method_selector for the test user, making the test environment-independent (no dependency on OMEGAUP_EXPERIMENT_SECRET).

cypress/e2e/problem_creator.cy.ts
shauryag2002: Added cy.get('[name="problem-level"]').select(1) before submit.

cypress/e2e/problem_creator.cy.ts
shauryag2002: Exposed :data-problem-creator-zip-ready on the hidden Creator input in Form.vue and the test waits for it (should('have.attr', 'data-problem-creator-zip-ready', 'true')) before clicking submit.

cypress/e2e/problem_creator.cy.ts
shauryag2002: Removed preventDefault(), filled all required fields (title, source, tag, level), added a case inside the Creator for a valid zip layout, and asserted the successful redirect URL (cy.url().should('include', problemAlias)).

pabo99 (approved):
This is great! Thank you so much, @shauryag2002, for completing Milestone 5 of the Problem Creator integration.

The next step will be to enable the experiment for a few key problem setters so we can validate that everything is working as expected

Issue #10064, opened 2026-07-31
[FEATURE] Detect problems that silently stopped working


PrasannaMishra001:
### Checklist

- [x] I have searched existing issues and confirmed this is not a duplicate
- [x] I have discussed this on [GitHub Discussions](https://github.com/omegaup/omegaup/discussions) or [Discord](https://discord.gg/K3JFd9d3wk) and received positive feedback

### Context

a problem can stop being usable long after it was published, the judge or the validator starts failing, the enabled languages get misconfigured so nobody can submit or the problem turns out to be impossible. today the only signal is a user reporting it through a quality nomination, so there is nothing watching automatically.

this adds a scheduled job on top of the cron control plane that looks for those cases in the database and records what it finds, plus an admin view of the problems that need attention. the checks in this first version are:

- judge errors, a problem whose recent submissions are ending in a judge or validator error, which points at the judging setup rather than at the submissions
- no languages, a public problem where nobody can submit at all
- never solved, a problem old enough and tried often enough that having no accepted solution is a signal, usually wrong expected output or limits
- deprecated but public, marked deprecated yet still offered to users

the judge errors check is narrow on purpose. it only sees submissions that ended in a judge or validator error, so judging that goes wrong but still returns a verdict is out of scope for it, that comes back as an ordinary `WA` and is indistinguishable from a rejected submission in the database.

each check has thresholds for how much evidence it needs before it says anything, and they are all command line arguments so they can be tuned without a code change. findings keep the date they were first detected, so it is visible how long something has been broken, and a finding that stops being detected is marked as resolved instead of disappearing.

for now the findings only show on the admin page, no notifications, so the thresholds can be tuned before anything reaches problem authors. this complements the quality nominations, which are reported by hand, rather than replacing them.

**expected outcome:** broken problems are detected automatically and listed for admins.

<img width="1286" height="332" alt="Image" src="https://github.com/user-attachments/assets/b4f9f9d6-8778-4286-95a3-896d7174a182" />

### Order

builds on the cron control plane epic #9992, it uses the shared runner and the job registry. comes as three small PRs linked to this issue, #10065 for the table, #10066 for the job that runs the checks and #10069 for the admin view.

PR #10065, opened 2026-07-31
add problem health checks table

PrasannaMishra001:
# Description

adds the schema for the automated problem health checks: a `Problem_Health_Checks` table holding one row per finding, the problem it belongs to, which check found it, a severity, a readable detail, when it was first detected, when it was last seen and when it was resolved. it also seeds `problem_health_check.py` into the `Cron_Jobs` registry with a daily schedule so it registers as a scheduled job.

the unique key on problem and check type is what lets the job upsert its findings, so a finding keeps the date it was first detected, running the job twice does not create duplicates, and a finding that stops being detected is marked resolved instead of disappearing.

Part of #10064.

# Comments

this PR is the migration and `schema.sql` only, the job itself and the admin view come in the follow up PRs. it builds on the cron control plane since it registers the job in `Cron_Jobs`.

# Checklist:

### DIFF
--- a/frontend/database/00279_problem_health_checks.sql
+++ b/frontend/database/00279_problem_health_checks.sql
@@ -1,0 +1,17 @@
+CREATE TABLE `Problem_Health_Checks` (
+  `check_id` int NOT NULL AUTO_INCREMENT,
+  `problem_id` int NOT NULL,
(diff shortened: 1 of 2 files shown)


pabo99 (approved):
LGTM!

PR #10066, opened 2026-08-01
add problem health check cron

PrasannaMishra001:
# Description

adds `stuff/cron/problem_health_check.py`, a nightly job that finds problems that silently stopped working.

four checks: a problem whose recent submissions are ending in judge errors, a public problem with no enabled language, a problem old enough and tried often enough that having no accepted solution is a signal, and one that is deprecated but still public. only `normal` submissions count towards either of the checks that count them, so a setter testing their own problem cannot report it as broken. findings are upserted so each keeps the date it was first detected, and resolved once it stops being detected. every threshold is a command line argument, and one outside the range its check can act on is rejected at parse time rather than quietly finding nothing.

it runs through the shared cron runner with one phase per check and one for the write, so it records its own runs and cannot overlap itself. every check looks only at problems the catalog still offers, apart from the deprecated one, which is looking for exactly that.

part of #10064. #10069 is the dashboard that shows the findings and is what closes the issue.

# Comments

the table merged as #10065 and the dao as #10149. #10069 is the dashboard that shows the findings.

the check type and severity are enums in the script rather than string literals, and the phase names come off the same enum so a phase cannot drift from the check it records.

`PROBLEM_VISIBILITY_PUBLIC` moves into `stuff/cron/constants.py` next to the other mirrored values, since every check needs it.

# Checklist:


PrasannaMishra001:
@pabo99 this is also good for review, this has nightly job that finds problems which silently stopped working, four checks, each finding keeps the date it was first detected. everything it was waiting on is merged.

stuff/cron/problem_health_check.py
```diff
@@ -0,0 +1,357 @@
...
+            `c`.`submission_count`
+        FROM (
+            SELECT
+                `s`.`problem_id`,
+                SUM(`s`.`verdict` IN ('JE', 'VE')) AS `error_count`,
+                COUNT(*) AS `submission_count`
+            FROM `Submissions` `s`
+            WHERE
+                `s`.`time` >= %s AND
+                `s`.`status` = 'ready'
```
pabo99: Should this query only consider submissions with `s.type = 'normal'`? The `Submissions` table also contains `test` and `disqualified` entries, while most user-facing submission queries explicitly filter them out. A problem setter repeatedly testing a broken validator could therefore make a public problem cross both thresholds even if regular users are unaffected. It would be useful to clarify the intended scope and add a regression test for non-normal submissions

stuff/cron/problem_health_check.py
```diff
@@ -0,0 +1,357 @@
...
+                            default=_JUDGE_ERROR_WINDOW_HOURS,
+                            help='How far back to look for judge or '
+                            'validator errors.')
+    thresholds.add_argument('--min-judge-errors',
+                            type=int,
+                            default=_MIN_JUDGE_ERRORS,
+                            help='How many judge or validator errors a '
+                            'problem needs in the window to count as a '
+                            'finding.')
+    thresholds.add_argument('--min-judge-error-ratio',
```
pabo99, line 329: Could we validate the threshold arguments before running the checks? Values such as a negative window or `--min-judge-error-ratio` greater than 1 can make a check return no findings, after which `apply_findings()` will resolve the previously open findings as if the problems had recovered. Rejecting invalid ranges at argument parsing time would prevent an operator typo from rewriting the health history. Regression tests for the accepted ranges would also be helpful

pabo99 (commented):
As a non-blocking follow-up, could you please open a PR in the _omegaup/prod_ repository adding `problem_health_check.py` to _k8s/omegaup/base/frontend/cronjobs.yaml_? Please use the `35 5 * * *` schedule from the `Cron_Jobs` database entry. 

We'll review and merge that PR after the next production deployment, hopefully Sunday night PST

PrasannaMishra001:
@pabo99 please refer to https://github.com/omegaup/prod/pull/72

pabo99 (approved):

PR #10069, opened 2026-08-03
show problem health findings for admins

### LINKED ISSUE #10064: [FEATURE] Detect problems that silently stopped working

PrasannaMishra001:
# Description

surfaces the problem health findings on `/admin/crons`.

`apiGetCrons` returns the open findings from `ProblemHealthChecks::getOpenFindings`, worst first, and the page shows one row per finding with the problem, what the check found, the severity, the detail and when it was first detected. the check type and severity are stored as raw enum values, so the page shows translated labels
and keeps the raw check type on its cell's title. a check type the ui does not know yet is
shown as it came.

`getOpenFindings` orders newest first inside a severity now instead of oldest first. the
caller takes fifty, and oldest first would have pinned that list to whatever nobody had acted
on, so a newly broken problem would never have appeared.

there is a storybook story for the component, which you asked for on #10008.

Fixes: #10064

# Comments

the table merged as #10065 and the dao as #10149. the nightly job that fills the table is #10066.

it should go in after #10066, since the `Fixes:` closes #10064 and without the job the table has nothing in it.

# Checklist:


pabo99 (commented):
Could you move the changes to `frontend/database/dao_schema.sql` and the two autogenerated DAO files to a separate PR? 

Since these changes are self-contained, we should be able to review and merge that PR right away

PrasannaMishra001:
@pabo99 yes i will,actually i am waiting for https://github.com/omegaup/omegaup/pull/10066 and https://github.com/omegaup/omegaup/pull/10008, that's why i did not update this one, i will undraft it when others are merged. included here just to see if all tests passes, so that will be easier to rebase later.

PR #10083, opened 2026-08-06
fix(dao): handle concurrent first-time SystemSettings insert race

### LINKED ISSUE #10053: bug: TOCTOU race condition in first-time SystemSettings creation causes duplicate-key 500 error
### Lista de verificación

### Comportamiento Actual

`SystemSettings::setBooleanSetting()` (línea 81) y `setStringSetting()` (línea 133) implementan un patrón SELECT-then-INSERT para crear filas nuevas en la tabla `System_Settings`: primero se llama `getByKey($key)` para leer la fila; si el resultado es `null`, se llama `create($newSetting)` que emite un `INSERT` sin ninguna protección contra inserciones concurrentes.

La tabla tiene la restricción `UNIQUE KEY unique_setting_key (setting_key)` (verificable en `frontend/database/schema.sql`). Cuando dos peticiones concurrentes invocan el mismo ajuste por primera vez (por ejemplo, dos llamadas simultáneas a `apiSetMaintenanceMode` antes de que exista la fila), ambas transacciones leen `null` en el `SELECT`, ambas entran al bloque `if (is_null($setting))` y ambas intentan el `INSERT`. La segunda falla con una `DuplicatedEntryInDatabaseException` por violación de la clave única. Esta excepción no está capturada dentro de `setBooleanSetting`, por lo que se propaga, revierte la transacción y produce un error HTTP 500 para el segundo solicitante.

El código afectado (commit `8fd4a29c`):

```php
// https://github.com/omegaup/omegaup/blob/8fd4a29cdb0045b45782a1b16feb7e1a5365f396/frontend/server/src/DAO/SystemSettings.php#L81
public static function setBooleanSetting(
    string $key,
    bool $value
): int {
    $setting = self::getByKey($key);          // SELECT — ambas transacciones ven null
    if (is_null($setting)) {
        $newSetting = new \OmegaUp\DAO\VO\SystemSettings([
            'setting_key'         => $key,
            'setting_value'       => $value ? '1' : '0',
            'setting_description' => '',
        ]);
        $affectedRows = self::create($newSetting); // INSERT sin ON DUPLICATE KEY — la segunda falla
    } else {
        $setting->setting_value = $value ? '1' : '0';
        $affectedRows = self::update($setting);
    }
    self::invalidateCache($key);
    return $affectedRows;
}
```

Bajo el nivel de aislamiento `REPEATABLE READ` de MySQL, ninguna transacción puede ver el `INSERT` confirmado de la otra hasta que ésta haga commit, por lo que ambas ven el estado "sin fila" al momento del `SELECT`.

### Comportamiento Esperado

Múltiples llamadas concurrentes a `setBooleanSetting` o `setStringSetting` para la misma clave deben completarse todas con éxito, independientemente de si la fila existe o no en la tabla `System_Settings` al momento de la llamada. Ninguna de ellas debe producir un error 500 originado en una violación de clave única. La operación de escritura debe ser atómica y segura ante concurrencia.

**Evidencia estática verificable en el repositorio (commit `8fd4a29c`):**

1. Confirmar la restricción UNIQUE en el esquema:
   ```
   grep 'UNIQUE KEY.*setting_key' frontend/database/schema.sql
   ```
   Resultado esperado: `UNIQUE KEY unique_setting_key (setting_key)`

2. Confirmar que `setBooleanSetting` no tiene protección contra inserción duplicada:
   ```
   awk '/function setBooleanSetting/,/^    \}/' frontend/server/src/DAO/SystemSettings.php
   ```
   Resultado: el bloque `if (is_null($setting))` llama a `self::create($newSetting)` sin `INSERT IGNORE`, `REPLACE INTO` ni `ON DUPLICATE KEY UPDATE`.

3. Confirmar que el patrón seguro de inserción concurrente ya existe en otras partes del código base:
   ```
   grep 'INSERT IGNORE' frontend/server/src/DAO/ProblemViewed.php
   ```
   Resultado: hay usos de `INSERT IGNORE` en `ProblemViewed.php` que resuelven el mismo problema en ese contexto.

**Escenario de reproducción:**

El bug es una condición de carrera (TOCTOU) y requiere concurrencia real para manifestarse. El escenario mínimo es:

1. Asegurarse de que la clave objetivo (p. ej. `maintenance.mode`) **no existe** en la tabla `System_Settings`.
2. Lanzar dos peticiones HTTP simultáneas a `apiSetMaintenanceMode` (o cualquier endpoint que invoque `setBooleanSetting`/`setStringSetting` con la misma clave).
3. Observar que una de las peticiones devuelve HTTP 500 con una excepción de tipo `DuplicatedEntryInDatabaseException`.

**Archivos relevantes:**

- [`frontend/server/src/DAO/SystemSettings.php` L81](https://github.com/omegaup/omegaup/blob/8fd4a29cdb0045b45782a1b16feb7e1a5365f396/frontend/server/src/DAO/SystemSettings.php#L81)
- `frontend/database/schema.sql` — definición de `UNIQUE KEY unique_setting_key`

**Entorno:**

- Componente: backend PHP (`DAO/SystemSettings.php`)
- Base de datos: MySQL (InnoDB, aislamiento `REPEATABLE READ`)
- No depende de navegador ni sistema operativo del cliente

deepak0x:
## Summary

`SystemSettings::setBooleanSetting` and `setStringSetting` used a SELECT-then-INSERT pattern without concurrency protection. When two requests create the same setting key for the first time (e.g. two simultaneous `apiSetMaintenanceMode` calls before the row exists), both read `null`, both attempt the `INSERT`, and the second fails with `DuplicatedEntryInDatabaseException` (the `unique_setting_key` constraint), surfacing as an HTTP 500.

Per @pabo99's comment on #10053, this is fixed with an **atomic upsert** rather than a catch-and-ignore, so two concurrent requests saving different values still resolve to a single atomic write.

## Changes

`frontend/server/src/DAO/SystemSettings.php`:
- `setBooleanSetting` and `setStringSetting`: the first-time insert path now uses a single `INSERT INTO System_Settings (...) VALUES (...) ON DUPLICATE KEY UPDATE setting_value = VALUES(setting_value)` statement. This is atomic at the DB level — there is no TOCTOU window, and a concurrent insert of the same key simply updates the value instead of throwing.
- The existing-row path (`getByKey` returns non-null) is unchanged and still uses `update`.

## Notes

- `created_at` / `updated_at` keep their `DEFAULT CURRENT_TIMESTAMP` on insert; `updated_at` auto-bumps on the duplicate-update path.
- This mirrors the existing raw-query upsert style already used elsewhere in the DAO layer (`ProblemViewed.php`).
- No `php` binary on the local runner, so syntax was verified by inspection; CI lint will confirm.

Fixes #10053


deepak0x:
@pabo99 thanks for the review on #10053. You're right — a catch-and-ignore wasn't sufficient, especially since two concurrent requests can save different values. I've reworked the fix to use an atomic upsert: the first-time insert path in both setBooleanSetting and setStringSetting now issues a single `INSERT ... ON DUPLICATE KEY UPDATE setting_value = VALUES(setting_value)` statement, so the concurrent case resolves to one atomic write at the DB level with no TOCTOU window. Description updated accordingly. PTAL when you get a chance.

frontend/server/src/DAO/SystemSettings.php
```diff
@@ -80,12 +80,16 @@ public static function setBooleanSetting(
     ): int {
         $setting = self::getByKey($key);
         if (is_null($setting)) {
-            $newSetting = new \OmegaUp\DAO\VO\SystemSettings([
-                'setting_key' => $key,
-                'setting_value' => $value ? '1' : '0',
-                'setting_description' => '',
-            ]);
-            $affectedRows = self::create($newSetting);
+            $sql = '
```
pabo99: Since the upsert works whether the row already exists or not, could we remove the initial `SELECT` and always execute a single statement?

This would completely remove the read-before-write pattern, avoid an additional database round trip, and allow sequential tests to exercise the duplicate-key update path.

It would also be better to replace `VALUES(setting_value)` with a row alias. Row aliases for `INSERT ... ON DUPLICATE KEY UPDATE` have been supported since MySQL 8.0.19, so this syntax is compatible with the project's current MySQL 8.0.34 image. The `VALUES(column)` form has been deprecated since MySQL 8.0.20.

pabo99 (commented):
Could you add a test that specifically exercises the `ON DUPLICATE KEY UPDATE` path?

The current tests write to the same key twice, but the second call finds the existing row through `getByKey()` and enters the traditional `update()` branch. Therefore, the clause introduced by this PR is only exercised during an actual race and currently has no coverage.

One way to make this behavior testable would be to always use the upsert for both creation and updates. The test could then write two different values to a new key and verify that:

1. only one row exists for that key;
2. the final value matches the second write;
3. no duplicate-key exception is thrown.

Ideally, this behavior should be covered for both boolean and string settings.

deepak0x:
SystemSettings.php: addressed in 2582f5b5e. Both setBooleanSetting and setStringSetting now drop the initial SELECT entirely and always run a single `INSERT ... ON DUPLICATE KEY UPDATE` statement — no read-before-write round trip, and sequential tests exercise the duplicate-key update path directly.

Also switched from the deprecated `VALUES(setting_value)` form to a row alias (`VALUES (?, ?, ?) AS new ... setting_value = new.`setting_value`); the project runs MySQL 8.0.34 so this is supported (VALUES() was deprecated in 8.0.20). The upsert remains atomic at the DB level, so concurrent first-time inserts of the same key collapse into one write.

pabo99:
Could you please resolve the conflicts?

frontend/server/src/DAO/SystemSettings.php
deepak0x: This was already applied in the second commit of the PR: both setters now skip the SELECT and run a single atomic `INSERT ... VALUES (?, ?, ?) AS new ON DUPLICATE KEY UPDATE setting_value = new.setting_value`, using the row alias you suggested instead of the deprecated `VALUES(setting_value)` form. The conflict with main is now also resolved, so the thread can stay resolved.

deepak0x:
Resolved the conflicts with `main` (commit 91720fb1f). The only conflicting region was the two `SystemSettings` setters: I kept this PR's atomic upsert and also pulled in the `$invalidateCache` parameter that had landed on `main`, so both behaviors are preserved. The PR is mergeable again, and the earlier review thread (single-statement upsert + row alias) is also addressed and resolved.

pabo99 (approved):
Great! Thanks, @deepak0x !

PR #10102, opened 2026-08-19
fix(arena): commit the polled scoreboard so the fallback keeps the arena alive

### LINKED ISSUE #10095: [BUG] EventsSocket polling fallback fetches the scoreboard and discards it, so the arena freezes when the socket drops
### Checklist

<sub>(Filed here rather than Discussions because it is a specific, code-grounded defect with a concrete repro — happy to move it if you would rather triage it there. Closest prior work I found is #9507, which fixed interval *leaks* on reconnect; this is a different defect in the same file.)</sub>

### Current Behavior

When the `/events/` WebSocket is unavailable, the arena's polling fallback fetches the scoreboard and then throws the result away. The scoreboard, mini-ranking, and chart stop updating, while the client keeps issuing two API calls per interval per open tab.

`setupPolls()` (`frontend/www/js/omegaup/arena/events_socket.ts`) calls the pure ranking functions but never commits their output:

```ts
api.Problemset.scoreboard({ problemset_id: this.problemsetId, token: this.scoreboardToken })
  .then((scoreboard) => {
    const { currentRanking } = onRankingChanged({ scoreboard, ... });   // return value dropped
    api.Problemset.scoreboardEvents({ ... })
      .then((response) => onRankingEvents({ events: response.events, currentRanking }))  // dropped
      .catch(ui.ignoreError);
  })
  .catch(ui.ignoreError);
```

`onRankingChanged` (`arena/ranking.ts`) is a pure function — it returns a record and commits nothing; `onRankingEvents` likewise returns `{series, navigatorData}`. `setupPolls` never calls `rankingStore.commit(...)` and never calls `createChart`. The socket path, `processRankings`, does all four: `updateRanking`, `updateMiniRankingUsers`, `updateLastTimeUpdated`, then `updateRankingChartOptions`.

Clarifications *do* update in fallback mode, because `refreshContestClarifications` commits internally — which is likely why this has gone unnoticed, since the page looks partially alive.

There is a second, compounding problem: **the fallback is never armed for a socket that dies after connecting.** `setupPolls()` is reachable only from the `.catch` of the *initial* `connect()`. The terminal branch of `onclose()` sets `SocketStatus.Failed` and returns. The retry budget is also smaller than it looks — `connectSocket()` sets `this.shouldRetry = false`, and `shouldRetry` is only set back to `true` inside `onopen`:

| step | state |
|---|---|
| initial connect succeeds | `shouldRetry = true`, `retries = 10` |
| socket drops | `onclose`: `retries → 9`, schedules a retry |
| retry runs | `connectSocket()` sets `shouldRetry = false` |
| retry fails (server still down) | `onclose`: `shouldRetry` is false → gives up, `Failed`, no polling |

So the 10-retry budget spends one attempt against a persistent outage, and the end state is a page with neither socket nor polling. `contest_contestant.ts`, `contest_virtual.ts` and `course.ts` have no other scoreboard refresh — they only `setInterval` clarifications — so the ranking tab, mini-ranking and chart stay frozen for the rest of the session.

`setupPolls` also ignores `isVirtual`/`originalProblemsetId`, so even once the commits are added it would compute the wrong ranking for virtual contests, where `processRankings` routes through `onVirtualRankingChanged`.

### Expected Behavior

While the socket is unavailable, the fallback should keep the scoreboard, mini-ranking and chart updating — i.e. produce the same store state the socket path produces — and a socket that dies after connecting should fall back to polling rather than silently giving up.

Concretely:

1. `setupPolls` commits `updateRanking` / `updateMiniRankingUsers` / `updateLastTimeUpdated` and refreshes the chart, and routes virtual contests through `onVirtualRankingChanged`.
2. The give-up branch of `onclose()` starts the fallback instead of only setting `Failed`.
3. The retry budget is not reset by `connectSocket()` on each attempt.

Since "turn a scoreboard payload into store state" currently exists in three places at three levels of completeness (`processRankings`, `setupPolls`, and the inline `setInterval` in `arena/scoreboard.ts`), the durable fix is probably one shared function they all call, rather than patching one copy — but that is a design call, and I would rather agree the direction with maintainers before writing it. A minimal fix that just adds the missing commits is also viable if you would prefer to keep the change small.

I am happy to open the PR either way.

The defect is deterministic and reachable from a unit test — no deployed instance needed:

1. In `frontend/www/js/omegaup/arena/events_socket.test.ts` (the file added by #9507), construct an `EventsSocket` and call `setupPolls()` directly.
2. Mock `api.Problemset.scoreboard` and `api.Problemset.scoreboardEvents` to resolve with a non-empty scoreboard.
3. Flush the promises.
4. Assert `rankingStore.state.ranking` — it is still empty.

That single assertion demonstrates the first defect in a few lines. For the second, forcing the initial `connect()` to succeed and then closing the socket shows `SocketStatus.Failed` with no polling armed.

In the browser, blocking `/events/` at the proxy (or letting a proxy idle-timeout hit — note the keepalive ping interval is the same 5 minutes as the poll interval, longer than many proxies' idle timeout) during a contest reproduces the frozen scoreboard with a `✗` glyph and no other signal.

* Browser: any
* Affected paths: `frontend/www/js/omegaup/arena/events_socket.ts`, `arena/ranking.ts`, `arena/scoreboard.ts`
* Verified against `main` at the time of filing

<sub>Researched and drafted with AI assistance. Every file and symbol referenced was read directly in the tree and re-checked against current `main` before filing; the failure sequences are reasoned from the code, not observed on a deployed instance.</sub>

inland-taipen:
# Description

Fixes: #10095

When the `/events/` WebSocket is unavailable, the arena's polling fallback fetched the scoreboard and then threw the result away, so the ranking table, mini-ranking and chart stopped updating while the client kept issuing two API calls per interval per open tab. Clarifications kept refreshing — `refreshContestClarifications` commits internally — which is likely why this went unnoticed: the page looks half alive.

Three changes, matching the scope agreed in the issue.

**1. The fallback produces the same store state as the socket path.** `setupPolls` called `onRankingChanged` and `onRankingEvents` — both pure functions that commit nothing — and dropped what they returned. The commit half of `processRankings` is extracted into `updateRankings`, and both paths now go through it, so `updateRanking`, `updateMiniRankingUsers`, `updateLastTimeUpdated` and the chart are all refreshed while polling. Virtual contests route through `onVirtualRankingChanged`, which `setupPolls` ignored entirely — without that it would have committed the wrong ranking for virtual contests as soon as the commits were added.

**2. Polling starts when the socket is down.** `setupPolls` was reachable only from the `.catch` of the *initial* `connect()`. The terminal branch of `onclose()` set `SocketStatus.Failed` and returned, so a socket that died *after* connecting left the arena with neither socket nor polling for the rest of the session. `onclose()` now arms the fallback, guarded by a `pollingActive` flag so repeated closes cannot stack a second pair of intervals. `onopen` already cleared the intervals on reconnect and now clears the flag with them.

**3. The retry budget is no longer reset on every attempt.** `connectSocket()` set `shouldRetry = false` on each call while only `onopen` set it back, so the first failed reconnect fell through to the give-up branch and nine of the ten retries were never spent. The field already defaults to `false`, so removing that line leaves initial-connect behaviour unchanged — still no retry, straight to the fallback — and lets a socket that has connected at least once actually spend its budget.

The shared helper is the mechanism of the fix rather than a refactor bolted onto it: the two paths drifted precisely because each had its own copy of "scoreboard payload → store state", so having one of them call the other is what makes them stay in agreement. It also shrinks the file — the two identical copies of the poll body inside `setupPolls`, the immediate call and the `setInterval` one, collapse into a single `refreshRanking()` — so it does not run into the "significantly larger" caveat you raised.

**Deliberately not done here:** the issue also notes that "scoreboard payload → store state" lives in three places — `processRankings`, `setupPolls`, and the inline `setInterval` in `arena/scoreboard.ts`. This PR unifies the first two because they are the defect; `arena/scoreboard.ts` is untouched, left for the follow-up you suggested.

# Comments

**One judgement call worth flagging.** You asked for the fallback to start "when the socket gives up reconnecting". It is armed on *every* close instead, not only on the give-up branch. The reason is change 3: once the retry budget is real, giving up takes up to ten backoffs of `Math.random() * intervalInMilliseconds / 2`, so binding the fallback to the give-up branch alone would leave the scoreboard frozen for that whole window — the very symptom being fixed. Arming on close costs one call site, reuses the `onopen` cleanup added in #9507, and still satisfies "starts when it gives up", because by then it is already running. Happy to move it to the give-up branch alone if you would rather keep the change literal.

A visible side effect of the same change: during a reconnect the status glyph now stays `↻` for the length of the retry budget instead of flipping to `✗` after a single failed attempt, and the scoreboard keeps moving underneath it.

**One overlap worth naming.** Of the five `EventsSocket` call sites, three — `contest_contestant.ts`, `contest_virtual.ts` and `course.ts` — have no other scoreboard refresh at all, and those are the ones that froze. The other two, the standalone token-scoreboard pages `arena/scoreboard.ts` and `course/scoreboard.ts`, already run their own unconditional 5-minute `setInterval`, so on those the fallback now overlaps an existing poll. It is not pure duplication — their inline interval commits only `updateRanking` and `updateLastTimeUpdated`, so the fallback is what fills in the mini-ranking and the chart — but it is a second identical fetch on the same 5-minute cadence. That overlap already existed on the initial-connect-failure path; this change makes it reachable when the socket dies later too. It is also the third copy of the ranking logic from the issue, so it resolves itself in the follow-up PR.

**Tests.** Five cases added to `frontend/www/js/omegaup/arena/events_socket.test.ts`, covering the scenarios above:

- `should commit the polled scoreboard to the ranking store` — the repro from the issue: asserts `rankingStore.state.ranking`, `miniRankingUsers` and `lastTimeUpdated`, plus that the chart is built.
- `should route virtual contests through onVirtualRankingChanged while polling`
- `should start polling when the socket gives up reconnecting`
- `should arm the fallback only once across repeated close events`
- `should spend the whole retry budget on repeated reconnect failures`

All five fail against `main` and pass here. Full frontend suite: 192 suites, 905 passed, 3 skipped, 0 failed (`yarn test`); `eslint` and `prettier --check` clean on both files.

# Checklist:

<sub>The diff is 256 lines, of which 171 are the new Jest tests; the fix itself is ~85 lines in one file. It is entirely frontend + Jest, so the split the checklist describes (controllers/PHPUnit vs UI/Jest) does not apply — say the word if you would still prefer it broken up.</sub>

### DIFF
--- a/frontend/www/js/omegaup/arena/events_socket.test.ts
+++ b/frontend/www/js/omegaup/arena/events_socket.test.ts
@@ -15,7 +15,13 @@ import { clarificationStoreConfig } from './clarificationsStore';
 import { createLocalVue } from '@vue/test-utils';
 import Vuex from 'vuex';
 import fetchMock from 'jest-fetch-mock';
-import { onRankingChanged, onRankingEvents } from './ranking';
+import {
+  createChart,
+  onRankingChanged,
+  onRankingEvents,
+  onVirtualRankingChanged,
+} from './ranking';
+import rankingStore from './rankingStore';
 import { mocked } from 'ts-jest/utils';
 import { ScoreMode } from './navigation';
 
@@ -55,6 +61,25 @@ const options: SocketOptions = {
   intervalInMilliseconds: 500,
   scoreMode: ScoreMode.Partial,
 };
+
+// The polling fallback is several promise hops deep -- scoreboard fetch,
+// commit, scoreboardEvents fetch, chart. Fake timers are installed, so reach
+// past them to the real macrotask queue to let every hop settle.
+const { setImmediate: realSetImmediate } = jest.requireActual('timers');
+const flushPromises = (): Promise<void> =>
+  new Promise((resolve) => realSetImmediate(resolve));
+
+// Fallback intervals outlive the test that armed them, and afterEach's
+// runOnlyPendingTimers would fire them against a torn-down server.
+const stopPolling = (client: EventsSocket): void => {
+  const internals = client as any;
+  clearInterval(internals.clarificationInterval);
+  clearInterval(internals.rankingInterval);
+  internals.clarificationInterval = null;
+  internals.rankingInterval = null;
+  internals.pollingActive = false;
+};
+
 describe('EventsSocket', () => {
   let server: WS | null = null;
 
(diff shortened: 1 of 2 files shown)


pabo99 (approved):
Thanks, @inland-taipen!

PR #10140, opened 2026-08-28
fix: escape boolean-mode operators in problem search queries

### LINKED ISSUE #10131: bug: FULLTEXT BOOLEAN MODE receives unescaped user input in Problems::getList and getAllWithCount
### Lista de verificación

### Comportamiento Actual

El parámetro `$query` recibido del usuario se pasa sin ningún escape a una cláusula `MATCH(...) AGAINST (? IN BOOLEAN MODE)` en dos lugares de `Problems.php`:

- [`Problems.php` línea 486](https://github.com/omegaup/omegaup/blob/754c5f54551eac6ee71cc29d854eb54b4ba27a46/frontend/server/src/DAO/Problems.php#L486) — usado en `Problems::getList`
- `Problems.php` línea 1311 — usado en `getAllWithCount`

En ambos casos el valor crudo se asigna directamente al parámetro del `AGAINST`:

```php
// Problems.php línea 486-488
'MATCH(p.alias, p.title) AGAINST (? IN BOOLEAN MODE)',
// ...
$argsForQuery = [$query];  // $query sin escape
```

MySQL FULLTEXT BOOLEAN MODE interpreta como operadores de búsqueda los caracteres `+`, `-`, `(`, `)`, `*`, `"`, `~`, `<`, `>`. Cuando el usuario ingresa una cadena que contiene operadores malformados (por ejemplo `(test`, `++palabra`, `"sin cerrar`), el parser de MySQL los evalúa antes de ejecutar la consulta:

- **MySQL 8+**: lanza `ER_FTS_QUERY_PARSING_ERROR` para expresiones booleanas inválidas.
- **MySQL 5.7**: retorna cero filas silenciosamente para entradas que antes (`LIKE`) devolvían resultados correctos.

El archivo `Identities.php`, modificado en el mismo PR (#10028), sí aplica un método `escapeBooleanModeQuery()` antes de cualquier cláusula `AGAINST IN BOOLEAN MODE` (línea 86). `Problems.php` no tiene llamada equivalente ni ningún uso de `addcslashes` sobre `$query` en todo el archivo.

### Comportamiento Esperado

Cualquier valor de `$query` ingresado por el usuario, independientemente de los caracteres que contenga, debería producir resultados de búsqueda equivalentes a los que entregaba la cláusula `LIKE CONCAT('%', ?, '%')` anterior: sin errores de parseo y sin pérdida silenciosa de resultados. Los operadores de BOOLEAN MODE deben tratarse como texto literal cuando provienen de la entrada del usuario, igual que lo hace `Identities.php` en el mismo codebase.

**Evidencia en el código (commit `754c5f54`):**

1. Abrir [`frontend/server/src/DAO/Problems.php` línea 486](https://github.com/omegaup/omegaup/blob/754c5f54551eac6ee71cc29d854eb54b4ba27a46/frontend/server/src/DAO/Problems.php#L486):
   ```php
   'MATCH(p.alias, p.title) AGAINST (? IN BOOLEAN MODE)',
   // ...
   $argsForQuery = [$query];
   ```
   `$query` se enlaza directamente al parámetro sin ningún escape previo.

2. Buscar en el mismo archivo cualquier llamada a `escapeBooleanModeQuery` o `addcslashes` sobre `$query`: **no existe ninguna**.

3. Comparar con `frontend/server/src/DAO/Identities.php` línea 86 del mismo PR, donde se aplica el escape antes del `AGAINST`:
   ```php
   $escapedUsernameOrName = self::escapeBooleanModeQuery($usernameOrName);
   // ... luego se usa $escapedUsernameOrName en el AGAINST
   ```

**Pasos para reproducir el error en MySQL 8+:**

1. Levantar el entorno de desarrollo de omegaup con MySQL 8.
2. Navegar al listado de problemas con un término de búsqueda que contenga paréntesis sin cerrar, p. ej. `(suma`.
3. Observar que la búsqueda lanza un error de parseo FULLTEXT o devuelve cero resultados, mientras que la misma cadena con la implementación `LIKE` anterior devolvía coincidencias.

**Ambiente relevante:**
- Archivo afectado: `frontend/server/src/DAO/Problems.php`
- Introducido por: PR #10028 (sustitución de `LIKE` por `FULLTEXT BOOLEAN MODE`)
- MySQL 8+ expone el error como `ER_FTS_QUERY_PARSING_ERROR`; MySQL 5.7 falla silenciosamente

deepak0x:
# Description

Problem search queries are passed directly into `MATCH ... AGAINST (? IN BOOLEAN MODE)`, so MySQL interprets Boolean Mode operators inside user input. A search like `-problem` is treated as an exclusion and silently returns no results, even when matching problems exist.

This change escapes Boolean Mode operators (`+ - < > ( ) ~ * " \`) in the search term before it reaches the query, in both places where problem search runs:

- `Problems::byIdentityType` (used by the problem list / search API)
- `Problems::getAllWithCount`

The escaping reuses the existing `Identities::escapeBooleanModeQuery` helper, which is made public so both DAOs share the same behavior.

Fixes: #10131

# Comments

As requested in the issue discussion, this PR adds tests covering the Boolean Mode operators:

- `testDashInQueryIsNotTreatedAsExclusion` — a dash in the search term no longer acts as an exclusion operator
- `testPlusInQueryStillMatches` — a plus in the search term no longer acts as a required-term operator
- `testMalformedBooleanQueriesDoNotError` — unbalanced parentheses and other operator-heavy input do not break the query
- `testPlainSearchStillWorks` — ordinary searches keep returning the same results

Each test exercises the real prepared-statement path through `Problems::byIdentityType` against the test database.

# Checklist:


frontend/tests/controllers/ProblemSearchBooleanModeTest.php
```diff
@@ -0,0 +1,138 @@
...
+            fn ($problem) => $problem->alias,
+            $problems
+        );
+        $this->assertContains('test-problem', $allAliases);
+    }
+
+    public function testPlusInQueryStillMatches(): void {
+        $this->seedProblems();
+
+        $result = $this->searchProblems('+suma');
```
pabo99: This case would likely pass even before the fix, because in boolean mode `+suma` only makes `suma` a required term, and `suma-dos` would still match

To make the test catch the regression more clearly, we could use something like `alpha +beta` and create a problem that only contains `alpha`; that would verify that `+beta` is not acting as a required-term operator after escaping

frontend/server/src/DAO/Identities.php
```diff
@@ -103,7 +103,7 @@ public static function findByUsernameOrName(
         return $result;
     }
 
-    private static function escapeBooleanModeQuery(string $query): string {
+    public static function escapeBooleanModeQuery(string $query): string {
```
pabo99: Since this helper is now used from `Problems`, maybe it should live in a more common place, for example `\OmegaUp\DAO\DAO::escapeBooleanModeQuery(...)`

That would avoid making other DAOs depend on `Identities` when they need to prepare text for `MATCH ... AGAINST (... IN BOOLEAN MODE)`, and would give future fulltext searches a more neutral helper to reuse

pabo99 (commented):
Just a couple of small suggestions and it's done

frontend/tests/controllers/ProblemSearchBooleanModeTest.php
deepak0x: Good point, the old test was passing even before the fix, since +suma only marks suma as required and the seeded problem still matched. I rewrote it as testPlusDoesNotActAsRequiredOperator: it seeds a problem that only contains the first term of the query, so if the plus sign were left as a required-term operator, that problem would be excluded from the results. Verified it fails without the fix and passes with it. Pushed in 7f7deb564.

frontend/server/src/DAO/Identities.php
deepak0x: Moved it to the base DAO class as escapeBooleanModeQuery, and updated the two call sites in Problems plus the internal one in Identities to use it. No other DAO needs to depend on Identities now. Pushed in 7f7deb564.

deepak0x:
The red `cypress (contest-group)` check on the latest push looks like the flake reported in #10161 rather than a change from this PR: the same job passed on the previous commit (https://github.com/omegaup/omegaup/actions/runs/33163965938), and the same assertion has also failed on `main` this week (https://github.com/omegaup/omegaup/actions/runs/33004102814), with a run on this same head passing as well (https://github.com/omegaup/omegaup/actions/runs/33300678630). A re-run of the failed jobs should get this back to green.

pabo99 (approved):
Cool!

PR #10144, opened 2026-08-28
fix: record SystemExit(0) as successful cron run

### LINKED ISSUE #10136: bug: SystemExit(0) in with-body recorded as failure
### Lista de verificación

### Comportamiento Actual

Cuando un script cron llama a `sys.exit(0)` dentro de un bloque `with lib.runner.run(...)`, el método `_finish` en `stuff/lib/runner.py` registra la ejecución con `status='failure'` y `error_text='SystemExit: 0'` en la tabla `Cron_Runs`.

Esto ocurre porque en la línea 189, la condición que determina si hubo falla es:

```python
failed = exc_value is not None or self._forced_failure
```

Python pasa `SystemExit(0)` al método `__exit__` del context manager como una excepción ordinaria, por lo que `exc_value` es distinto de `None`. La condición evalúa como `True` sin distinguir entre `SystemExit(0)` (salida exitosa por convención) y una excepción real de error.

Como resultado, la línea 193 construye el mensaje de error:

```python
error_text = f'{type(exc_value).__name__}: {exc_value}'
# → 'SystemExit: 0'
```

y la fila en `Cron_Runs` queda con `status='failure'` aunque el proceso haya terminado intencionalmente sin error.

Código relevante (commit `18a65c59`, línea 189):
https://github.com/omegaup/omegaup/blob/18a65c596f435ddb1f6c239bda3d60504ee35ec9/stuff/lib/runner.py#L189

```python
186:     def _finish(self, exc_value: Any) -> None:
187:         if self._connection is None or self._run_id is None:
188:             return
189:         failed = exc_value is not None or self._forced_failure
190:         status = 'failure' if failed else 'success'
191:         error_text: Optional[str] = None
192:         if exc_value is not None:
193:             error_text = f'{type(exc_value).__name__}: {exc_value}'
```

### Comportamiento Esperado

`sys.exit(0)` es la forma convencional en Python de indicar que un proceso terminó exitosamente. Cuando un script cron llama a `sys.exit(0)` dentro del bloque `with lib.runner.run(...)`, se espera que la fila en `Cron_Runs` quede con `status='success'` y `error_text=NULL`, de la misma manera que si el bloque terminara sin ninguna excepción.

**Pasos para reproducir:**

1. Clonar el repositorio en el commit `18a65c59`.
2. Ejecutar el siguiente script que replica la lógica de `_finish`:

```python
import sys

def finish(exc_value):
    failed = exc_value is not None  # lógica actual, línea 189
    return 'failure' if failed else 'success'

result = finish(SystemExit(0))
print(f'SystemExit(0): {result}')  # imprime: SystemExit(0): failure
```

3. Observar que la salida es `SystemExit(0): failure` cuando se esperaría `SystemExit(0): success`.

**Entorno:**

* Archivo afectado: `stuff/lib/runner.py`, línea 189
* Commit analizado: `18a65c596f435ddb1f6c239bda3d60504ee35ec9`
* Componente: backend — runner de scripts cron, tabla `Cron_Runs`
* Versión de Python: 3.x (comportamiento estándar de `SystemExit` como subclase de `BaseException`)

deepak0x:
# Description

When a cron script calls `sys.exit(0)` inside `with lib.runner.run(...)`, the context manager's `__exit__` receives the `SystemExit(0)` as `exc_value`, and `_finish()` treats any non-`None` exception as a failure. As a result, the `Cron_Runs` row ends up with `status='failure'` and `error_text='SystemExit: 0'`, even though exit code 0 means the job succeeded. This happens in practice because the runner itself raises `SystemExit(0)` for disabled or already-running jobs, and scripts also use `sys.exit(0)` for early exits.

The fix makes `_finish()` treat a `SystemExit` whose code is `None` or `0` as a clean exit: the run is recorded with `status='success'` and `error_text=NULL`. Any other `SystemExit` code (e.g. `sys.exit(1)` or `sys.exit('boom')`) is still recorded as a failure with the code/message in `error_text`, and real exceptions are unaffected. The `SystemExit` itself still propagates, so the process exits with the right code; only the recorded status changes.

Also adds `stuff/cron/tests/test_runner.py` with regression tests covering: clean exit, `SystemExit(0)`, bare `SystemExit()`, `SystemExit(1)`, `SystemExit('boom')`, a real exception, `mark_failure()`, and that the `SystemExit` is not swallowed. The `SystemExit(0)` tests fail on the unfixed code and pass with the fix. `MockCursor` gains a `lastrowid` attribute since the runner reads it after inserting the `Cron_Runs` row.

Fixes: #10136

# Checklist:


deepak0x:
Both failures from the earlier run are addressed on the rebased branch: `check-issue-link` passes again now that #10136 is assigned to me, and the rest of the matrix runs against current `main`.

The remaining red, `cypress (contest-group)`, is the intermittent timeout tracked in #10161. The trace points straight at `cypress/support/commands.ts:282`, where `createRunsInsideContest` waits for the arena problem link with the default 4 second timeout:

```
AssertionError: Timed out retrying after 4000ms: Expected to find element:
`a[data-problem="d9fb9463-7"]`, but never found it.
```

#10190 raises that wait to 10 seconds, matching the window the language selector already uses on the same screen. This PR only touches `stuff/cron` and `stuff/lib`, so the cypress specs do not exercise any of the changed code — worth a re-run of the failed job until #10161 lands.

stuff/lib/runner.py
```diff
@@ -186,6 +186,8 @@ def _insert_running_row(self) -> None:
     def _finish(self, exc_value: Any) -> None:
         if self._connection is None or self._run_id is None:
             return
+        if isinstance(exc_value, SystemExit) and exc_value.code in (None, 0):
```
pabo99, line 194: This matches the reported issue nicely. One small follow-up: if `SystemExit(0)` happens inside `cron_run.phase(...)`, the overall run is now recorded as `success`, but the phase still records `status='failure'` because `phase()` catches all `BaseException`s. Should we also treat clean `SystemExit` as a successful phase, or add a test documenting the current behavior?

pabo99 (commented):
Since `stuff/lib/runner_test.py` already covers `lib.runner`, would it make sense to add these regression cases there instead of creating a second runner test file under `stuff/cron/tests`? Not blocking, just a maintainability thought.

stuff/lib/runner.py
deepak0x: Fixed in 7b656629a. phase() now treats a clean SystemExit (code 0 or None) the same way _finish() does, so the phase records success instead of failure, while non-zero exits still record the phase as a failure. Added test_clean_system_exit_inside_phase_records_phase_success and test_failed_system_exit_inside_phase_records_phase_failure in runner_test.py covering both sides.

deepak0x:
Moved the runner regression tests from stuff/cron/tests/test_runner.py into the existing stuff/lib/runner_test.py suite in 7b656629a. The ported cases reuse the suite's _FakeConnection instead of a separate MockCursor fixture, and the cases that already had equivalents there (lock-held skip, mark_failure) were dropped rather than duplicated. The unused fixture tweak in mock_cursor.py was reverted as well.

pabo99 (approved):
Thanks! This also covers the `SystemExit(None)` case mentioned by @tonydzi. I think keeping other `BaseException` types such as `KeyboardInterrupt` and `GeneratorExit` as failures is reasonable for now.

PR #10149, opened 2026-08-29
add problem health checks dao

PrasannaMishra001:
# Description

split out of #10069 at your request, so the dao layer can be reviewed and merged on its own.

the `Problem_Health_Checks` table is already in main, it merged as #10065 with migration `00279`, but `dao_schema.sql` was never caught up. this adds the table there and the generated dao that goes with it, so `dao_schema.sql` is byte identical to `schema.sql` again.

what is here:

- `frontend/database/dao_schema.sql`, the table added, nothing else changed
- `frontend/server/src/DAO/Base/ProblemHealthChecks.php` and `frontend/server/src/DAO/VO/ProblemHealthChecks.php`, both generated by `stuff/update-dao.py`
- `frontend/server/src/DAO/ProblemHealthChecks.php` with one query, `getOpenFindings`, which returns the findings that are still open, worst first, joined to the problem they belong to
- `frontend/server/src/ProblemHealthCheckType.php` and `frontend/server/src/ProblemHealthSeverity.php`, two backed enums for the `check_type` and `severity` column values
- `frontend/tests/controllers/ProblemHealthChecksDAOTest.php`

the tests cover the alias and title coming back from the join, errors sorting ahead of warnings even when the warning is older, resolved findings being left out, and the limit being honoured.

the enums come from your comment on #10008 about known sets of values. both columns are already `enum` in the schema, so the php side now matches them, and the severity ordering in `getOpenFindings` is bound from the enum instead of two string literals inside the query.

part of #10064, split out of #10069 so it can land on its own. #10069 is what closes the issue.

# Comments

this shape follows #9996, which is the dao pr of the cron control plane and had the same set of files.

nothing calls `getOpenFindings` yet. the admin endpoint and the ui that use it stay in #10069, which is the rest of that pr once this is taken out.

# Checklist:

### DIFF
--- a/frontend/database/dao_schema.sql
+++ b/frontend/database/dao_schema.sql
@@ -262,7 +262,7 @@ CREATE TABLE `Contests` (
   `penalty_calc_policy` enum('sum','max') NOT NULL COMMENT 'Indica como afecta el penalty al score.',
   `show_scoreboard_after` tinyint(1) NOT NULL DEFAULT '1' COMMENT 'Mostrar el scoreboard automáticamente después del concurso',
   `urgent` tinyint(1) NOT NULL DEFAULT '0' COMMENT 'Indica si el concurso es de alta prioridad y requiere mejor QoS.',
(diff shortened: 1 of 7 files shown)


pabo99 (approved):
Great!

PR #10150, opened 2026-08-29
add recommendation model runs dao

PrasannaMishra001:
# Description

the dao for `Recommendation_Model_Runs`, whose table merged as #10050. `dao_schema.sql` had never caught up, so this brings it level with `schema.sql`.

`getRecent` returns the training runs newest first. `created_at` only keeps seconds, so the id breaks the tie between two runs recorded in the same second.

`build_problem_rec_model.py` joins `\OmegaUp\CronJobName` too, since #10050 registered it in
`Cron_Jobs` but the enum never caught up.

part of #10049.

# Comments

same file shape as #9996 and #10149. #10052 is the dashboard that reads it.

# Checklist:


PrasannaMishra001:
@pabo99 these dao can be reviewed as table in #10050 are already merged.

pabo99 (approved):
LGTM!

PR #10153, opened 2026-08-29
return open health findings from the crons api

PrasannaMishra001:
# Description

the server half of #10069, split off the way you asked on that pr and on #10008.

`apiGetCrons` now returns `problemHealthFindings` next to the jobs and the runs, from `ProblemHealthChecks::getOpenFindings`, which merged with #10149. the nightly job in #10066 is what fills the table.

the findings come back worst first, errors before warnings, and within a severity the oldest first, since something broken for a month matters more than something broken last night. each one carries the problem it belongs to, what the check found, how bad it is, the detail and when it was first detected.

both `apiGetCrons` and `getCronsForTypeScript` read it through the private `cronsPayload` helper, so the page still does not call the endpoint. the limit is a constant rather than a literal, like `CRON_RUNS_LIMIT` next to it.

part of #10064. the ui that renders this is #10069, and that is what closes the issue.

# Comments

the test creates a finding against a real problem and asserts it comes back through `getCronsForTypeScript` with the right alias, check type and severity. it filters to the problem it created rather than assuming it owns the table.

`check_type` and `severity` are `enum` columns, so the test uses `\OmegaUp\ProblemHealthCheckType` and `\OmegaUp\ProblemHealthSeverity` from #10149 instead of string literals.

# Checklist:


PrasannaMishra001:
folding this back into #10069 so there is one pr per piece of work instead of a chain.

PR #10154, opened 2026-08-29
return model quality from the crons api

PrasannaMishra001:
# Description

the server half of #10052, split off the way you asked on #10069 and on #10008.

`apiGetCrons` now returns `recommendationModelRuns` next to the jobs and the runs, read through `RecommendationModelRuns::getRecent` from #10150, newest first. each row carries the MAP score, how many submissions it trained on, the seed, whether it was published and, when it was not, why.

the seed is in there because of your comment on #10050. storing it is only half the point, an admin comparing two runs has to be able to see it, so it comes back in the payload and the dashboard shows it.

both `apiGetCrons` and `getCronsForTypeScript` read it through the private `cronsPayload` helper, and the limit is a constant next to `CRON_RUNS_LIMIT` rather than a literal.

part of #10049. the ui that renders this is #10052, and that is what closes the issue.

# Comments

this and #10153 both add a key to `cronsPayload`, so whichever lands second needs a one line resolve there. nothing else overlaps.

# Checklist:

### DIFF
--- a/frontend/database/dao_schema.sql
+++ b/frontend/database/dao_schema.sql
@@ -1067,6 +1067,25 @@ CREATE TABLE `QualityNominations` (
 /*!40101 SET character_set_client = @saved_cs_client */;
 /*!40101 SET @saved_cs_client     = @@character_set_client */;
 /*!50503 SET character_set_client = utf8mb4 */;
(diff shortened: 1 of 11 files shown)


PrasannaMishra001:
folding this back into #10052 so there is one pr per piece of work instead of a chain.

Issue #10018, opened 2026-07-16
[FEATURE] Let admins safely rerun a cron job


PrasannaMishra001:
### Checklist

- [x] I have searched existing issues and confirmed this is not a duplicate
- [x] I have discussed this on [GitHub Discussions](https://github.com/omegaup/omegaup/discussions) or [Discord](https://discord.gg/K3JFd9d3wk) and received positive feedback

### Context

rerunning a job today needs shell access to a pod, which is unlogged, unsafe and not something a non engineer can do.

this adds an admin action that queues a rerun request which a trusted worker picks up and runs, with the requester notified of the outcome. the web layer never runs a job directly, it only records the intent, so there is no way to trigger arbitrary work straight from a request.

**expected outcome:** an admin rerun button backed by a queued request and a dispatcher, with a failure notification.

### Order

part of the cron control plane epic #9992. the migration (#9993), the daos (#9994), the api (#9999) and the dashboard (#10004) it builds on are all done, so nothing is ahead of this one any more.

PR #10155, opened 2026-08-29
add cron run requests table

PrasannaMishra001:
# Description

adds `Cron_Run_Requests`, the queue an admin's rerun request goes into. the dispatcher picks rows off it, so the web server never launches a cron itself.

`requested_by` has a foreign key to `Users` with `ON DELETE SET NULL`, otherwise a deleted user leaves a dangling id and the notification insert that follows the rerun fails.

`run_id` points at the `Cron_Runs` row the rerun produced, also `ON DELETE SET NULL` so
purging the run history does not take the request with it.

part of #10018.

# Comments

the table on its own, two files. #10156 is the dao, #10021 is the rerun itself and #10048 sits on top of that, and all three stay red until this lands, because a stacked pr carries this pr's `schema.sql` and the schema check refuses `schema.sql` and `dao_schema.sql` in one diff.

# Checklist:


PrasannaMishra001:
@pabo99 this is good to be reviewed

frontend/database/00282_cron_run_requests.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE `Cron_Run_Requests` (
+  `request_id` int NOT NULL AUTO_INCREMENT,
+  `name` varchar(64) NOT NULL COMMENT 'Nombre del script cuya reejecución se solicita',
+  `requested_by` int DEFAULT NULL COMMENT 'El administrador que la solicitó, NULL si su cuenta ya no existe',
+  `status` enum('pending','picked','done','failed') NOT NULL DEFAULT 'pending' COMMENT 'pending mientras espera al despachador, picked mientras corre',
+  `requested_at` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `picked_at` datetime DEFAULT NULL COMMENT 'Cuando el despachador tomó la solicitud',
+  `finished_at` datetime DEFAULT NULL COMMENT 'Cuando terminó la ejecución, haya salido bien o mal',
+  `run_id` int DEFAULT NULL COMMENT 'La ejecución que produjo, NULL si el trabajo no llegó a correr',
```
pabo99: Would it make sense to add a foreign key here to `Cron_Runs.run_id`, possibly with `ON DELETE SET NULL`?

pabo99 (approved):
Cool!

pabo99:
It looks like the SQL script name needs to be updated

PrasannaMishra001:
done, rebased on main and renumbered to 00283 since 00282 is taken by seed_homepage_carousel_items now.

PR #10157, opened 2026-08-29
add cron rerun api

PrasannaMishra001:
# Description

`Admin::apiRerunCron`, the endpoint the rerun button calls. it takes a job name, checks the caller is a system admin, and queues a `pending` row in `Cron_Run_Requests`. it never runs anything itself, it only records the intent, so a request cannot trigger work directly.

the job name is validated against `\OmegaUp\CronJobName` rather than against a string the caller supplies or a list written out again here. that enum is the registry mirror, so anything the api accepts is a row the dispatcher will find in `Cron_Jobs`, and a new test asserts exactly that: every case of the enum is a registered job. if the two ever drift the test fails instead of the rerun failing in production.

a job that already has a request pending or in flight does not get a second row, so clicking twice is harmless. `rerun` joins the mutating verbs in `ApiCaller`, so a GET is rejected.

part of #10018.

# Comments

part of the cron control plane epic #9992, the third of five pull requests that used to be the single #10021. the chain is #10155, #10156, this one, #10158 and #10021. it is stacked on #10155 and #10156, so their files show in the diff until they merge.

`api.ts`, `api_types.ts`, `Controllers/README.md` and `www/docs/Controllers.md` are the linter's output, not hand written.

the dedupe is best effort rather than a constraint: two admins clicking at the same moment can both pass the check and queue two rows, and the job would then run twice. that is the same outcome as the job being scheduled twice, so i left it rather than adding a generated column and a unique index for it, but say the word if you would rather have the constraint.

what i ran: `CronControlPlaneAdminTest` and `CronRunRequestsDAOTest` under phpunit, both green, and `./stuff/lint.sh`, clean.

`database_schema.py` refuses `schema.sql` and `dao_schema.sql` in the same diff, and a stacked pull request based on main carries its parent's files, so the php job here fails on the "Validate database schema" step until #10155 merges, and stops there without reaching the tests. that is why this is a draft: once the table is in main i rebase and the step has nothing to complain about. i ran the php tests locally against the same tree and they pass.

# Checklist:

### DIFF
--- a/frontend/database/00282_cron_run_requests.sql
+++ b/frontend/database/00282_cron_run_requests.sql
@@ -1,0 +1,15 @@
+CREATE TABLE `Cron_Run_Requests` (
+  `request_id` int NOT NULL AUTO_INCREMENT,
+  `name` varchar(64) NOT NULL COMMENT 'Nombre del script cuya reejecución se solicita',
+  `requested_by` int DEFAULT NULL COMMENT 'El administrador que la solicitó, NULL si su cuenta ya no existe',
+  `status` enum('pending','picked','done','failed') NOT NULL DEFAULT 'pending' COMMENT 'pending mientras espera al despachador, picked mientras corre',
+  `requested_at` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `picked_at` datetime DEFAULT NULL COMMENT 'Cuando el despachador tomó la solicitud',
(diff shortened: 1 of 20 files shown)


PrasannaMishra001:
folded back into #10021. the rerun feature is three prs now, the table in #10155, the dao in #10156 and everything else in #10021.

PR #10158, opened 2026-08-29
add cron rerun dispatcher

PrasannaMishra001:
# Description

`cron_dispatcher.py`, the trusted worker that drains the rerun queue. it is the only thing that actually launches a job, so everything that can go wrong when two of them run at once, or when one of them dies, is handled here.

the claim is a single conditional update. it sets `picked` only `WHERE request_id = %s AND status = %s` with `pending` as the expected status, and reads `cur.rowcount` to decide whether this dispatcher won the row. a dispatcher that did not change the row moves on without running anything, so a request cannot be run twice. the same shape guards the update that fails an unregistered job.

a dispatcher that dies mid run leaves a row in `picked` with nobody to finish it, so the next run of the dispatcher fails any request that has been picked for more than six hours before it looks at the queue. an exception out of the child is caught and recorded as a failure with the message, rather than being allowed to leave the row stuck.

the allowlist is the `Cron_Jobs` registry read from the database, not a set written out again in python, which is the same list `apiRerunCron` validates against through `\OmegaUp\CronJobName`, plus a check that the name is a bare `.py` filename that exists under `stuff/cron`, so a tampered row cannot reach anything else.

the four status values are a `RequestStatus` enum mirroring `\OmegaUp\CronRunRequestStatus` and the column, following the shape already used in `stuff/editorial_generator`, so no query names a status as a literal.

when the dispatcher is given a password it writes a `[client]` block to a temporary `.cnf` that only its own user can read and hands the child that file instead of the password, so the password never reaches the child's command line where `ps` would show it. the child ends up reading the file because `lib.db.connect` only falls back to the config file when `--user` is absent, which is why the flag is left off deliberately.

`stuff/mysql/connector/cursor.pyi` gains `rowcount`, which the real cursor has always had as a documented `int` property but the stub was missing.

part of #10018.

# Comments

part of the cron control plane epic #9992, the fourth of five pull requests that used to be the single #10021. the chain is #10155, #10156, #10157, this one and #10021. it is stacked on the first three, so their files show in the diff until they merge.

the dispatcher needs a crontab entry on the ops side to actually drain the queue, since scheduling lives outside this repo. until it is scheduled the button records the request and nothing runs it.

the two notification strings it names are added in #10021 on top of this one, together with the change that lets the unused string linter see strings named by python.

what i ran: the 20 tests in `cron_dispatcher_test.py` and the whole of `stuff/cron` under pytest, 112 passed and 1 skipped, and `./stuff/lint.sh`, clean. i also mutation tested the claim: dropping the expected status from the where clause, ignoring `rowcount`, accepting any `rowcount`, removing the reaper, removing the exception handler, putting the password back on the command line, dropping the registry check and dropping the bare filename check each make a test fail, so the ten mutants i tried were all caught.

`database_schema.py` refuses `schema.sql` and `dao_schema.sql` in the same diff, and a stacked pull request based on main carries its parent's files, so the php job here fails on the "Validate database schema" step until #10155 merges, and stops there without reaching the tests. that is why this is a draft: once the table is in main i rebase and the step has nothing to complain about. i ran the php tests locally against the same tree and they pass.

# Checklist:

### DIFF
--- a/frontend/database/00282_cron_run_requests.sql
+++ b/frontend/database/00282_cron_run_requests.sql
@@ -1,0 +1,15 @@
+CREATE TABLE `Cron_Run_Requests` (
+  `request_id` int NOT NULL AUTO_INCREMENT,
+  `name` varchar(64) NOT NULL COMMENT 'Nombre del script cuya reejecución se solicita',
+  `requested_by` int DEFAULT NULL COMMENT 'El administrador que la solicitó, NULL si su cuenta ya no existe',
+  `status` enum('pending','picked','done','failed') NOT NULL DEFAULT 'pending' COMMENT 'pending mientras espera al despachador, picked mientras corre',
+  `requested_at` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `picked_at` datetime DEFAULT NULL COMMENT 'Cuando el despachador tomó la solicitud',
+  `finished_at` datetime DEFAULT NULL COMMENT 'Cuando terminó la ejecución, haya salido bien o mal',
+  `run_id` int DEFAULT NULL COMMENT 'La ejecución que produjo, NULL si el trabajo no llegó a correr',
+  `error_text` text COMMENT 'El final de stderr cuando la ejecución falló',
+  PRIMARY KEY (`request_id`),
+  KEY `idx_cron_run_requests_status` (`status`),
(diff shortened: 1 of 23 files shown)


PrasannaMishra001:
folded back into #10021. the rerun feature is three prs now, the table in #10155, the dao in #10156 and everything else in #10021.

Issue #10047, opened 2026-07-26
[FEATURE] Improve the cron dashboard with health cards, filters, relative times and auto refresh


PrasannaMishra001:
### Checklist

- [x] I have searched existing issues and confirmed this is not a duplicate
- [x] I have discussed this on [GitHub Discussions](https://github.com/omegaup/omegaup/discussions) or [Discord](https://discord.gg/K3JFd9d3wk) and received positive feedback

### Context

the cron dashboard from #10004 shows the jobs and the recent runs, but once there are a lot of runs it is a bit flat to read. this adds a few quality of life improvements on top of it so the page is easier to scan and stays useful as the history grows :-

- run times shown as "2 hours ago" with the exact timestamp on hover, so recent activity is readable at a glance.
- a small health card per job at the top with its latest status, success rate and average duration, so you can see which jobs are healthy without digging.
- filters above the runs table to narrow by job and by status.
- the page auto refreshing every few seconds so a running job updates live and a rerun shows its result without a manual reload.

**expected :** the same dashboard, easier to read and monitor.

### Order

part of the cron control plane parent #9992, builds on the dashboard #10004, which is done.

PR #10159, opened 2026-08-29
add cron job enable api

PrasannaMishra001:
# Description

`Admin::apiSetCronJobEnabled`, a system admin only endpoint that flips `Cron_Jobs.enabled`. a disabled job skips its scheduled runs, so this is the switch behind turning a job off without a deploy. it takes the job name and the new value, refuses a name that is not in the registry, and returns ok.

the endpoint is small and stands on its own, so it is here rather than inside the dashboard pull request that uses it. nothing calls it yet, the switch that does is in #10048.

part of #10047.

# Comments

split out of #10048 so the server side can be reviewed and merged without waiting on the ui, the same way #10149 was split out of #10069. this one is based straight on main and does not depend on the rerun chain.

`api.ts`, `api_types.ts`, `Controllers/README.md` and `www/docs/Controllers.md` are the linter's output, not hand written.

the name is validated by looking the row up rather than against `\OmegaUp\CronJobName`, because the row is what gets updated and the registry can hold jobs the enum does not name yet.

what i ran: `CronControlPlaneAdminTest` under phpunit, 12 tests green, and `./stuff/lint.sh`, clean.

# Checklist:

### DIFF
--- a/frontend/server/src/Controllers/Admin.php
+++ b/frontend/server/src/Controllers/Admin.php
@@ -510,6 +510,33 @@ class Admin extends \OmegaUp\Controllers\Controller {
     }
 
     /**
+     * Enables or disables a cron job. A disabled job skips its scheduled runs.
+     *
+     * @return array{status: string}
+     *
+     * @omegaup-request-param bool $enabled
+     * @omegaup-request-param string $name
+     */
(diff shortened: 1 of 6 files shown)


PrasannaMishra001:
folded back into #10048 so the enable switch is one pr instead of two.

PR #10167, opened 2026-09-01
fix: prevent crash from malformed JSON in updateProblemsOrder/updateAssignmentsOrder

### LINKED ISSUE #10168: fix: prevent crash from malformed JSON in updateProblemsOrder/updateAssignmentsOrder
### Checklist

### Current Behavior

When sending malformed or invalid JSON input in `problems` or `assignments` parameters (e.g., inside updateProblemsOrder or updateAssignmentsOrder endpoints), the backend crashes with PHP warnings and results in a 500 Internal Server Error.

### Expected Behavior

The backend should catch malformed or invalid JSON input, avoid crashing, and gracefully return a clean `400 Bad Request` response with the proper error message (`parameterInvalid`).

* Browser name (e.g. Chrome 39, node.js 5.4): Chrome (Local cURL testing)
* Operating System and version (desktop or mobile): Ubuntu (WSL2 / Linux)
* Link to the site where this issue occurs: Local development environment (`http://localhost:8001/api/course/updateProblemsOrder/`)
* Screenshot: (Aap chahien toh yahan apna cURL response ya code diff ka screenshot attach kar sakte hain, warna ye line chhor dein)

## Reproduction Steps / cURL Proof
```bash
curl -s -i -b /tmp/cookies.txt -X POST "http://localhost:8001/api/course/updateProblemsOrder/" -d "course_alias=Hel" -d "assignment_alias=h" -d 'problems=invalid_json_string'

### PR DESCRIPTION
**Pull request by AbdulKreemShah2408**
## Problem
Malformed or invalid JSON input (in problems or assignments) caused PHP warnings and backend crashes resulting in 500 errors.

## Fix
Added input validation to catch malformed data and safely return a clean 400 Bad Request (parameterInvalid).

## Verification
Tested with normal valid array input to ensure the regression check passes and the expected 404 response (problemNotFound) is successfully returned.

Closes #10168

### CODE REVIEW

**Comment by AbdulKreemShah2408** [comment]
Hi @pabo99, I have opened this pull request to fix the crash caused by malformed JSON. Could you please review it when you have time, or assign the related issue to me if needed? Thanks!

**Review by pabo99 — COMMENTED** [review]
Thanks for the fix! 

Could you please add regression tests for both `updateProblemsOrder` and `updateAssignmentsOrder`, asserting that malformed JSON returns `parameterInvalid`?

Also, `is_array()` is not enough to guarantee the documented `list<string>` type. Inputs such as `[null]` or `[{}]` pass this check and may still produce a `TypeError` in the DAO calls. Could you validate that this is a list containing only strings before starting the transaction, and cover that case in the tests as well?

**Comment by pabo99** [comment]
I took another look at the PR and noticed that a few points from my previous review still need to be addressed:

- Validate that the decoded value is actually a `list<string>`, not just an array
- Add the regression tests for both endpoints
- Update the branch with the latest changes from `main`

Once these are addressed, feel free to request another review

File: frontend/server/src/Controllers/Course.php [hunk 1]
```diff
@@ -1566,8 +1566,23 @@ public static function apiUpdateProblemsOrder(\OmegaUp\Request $r): array {
...
+                'problems'
+            );
+        }
+        foreach ($decodedAliases as $alias) {
+            if (!is_string($alias)) {
+                throw new \OmegaUp\Exceptions\InvalidParameterException(
+                    'parameterInvalid',
+                    'problems'
+                );
+            }
```
pabo99, line 1582: Could you use `Validators::validateAlias()` here instead of only checking `is_string()`?

Since these values represent aliases, this would preserve the type validation while also ensuring that each value is a valid omegaUp alias before reaching the DAO
```suggestion
            \OmegaUp\Validators::validateAlias($alias, 'problems');
```

frontend/server/src/Controllers/Course.php
```diff
@@ -1633,8 +1648,23 @@ public static function apiUpdateAssignmentsOrder(\OmegaUp\Request $r): array {
...
+                'assignments'
+            );
+        }
+        foreach ($decodedAliases as $assignment) {
+            if (!is_string($assignment)) {
+                throw new \OmegaUp\Exceptions\InvalidParameterException(
+                    'parameterInvalid',
+                    'assignments'
+                );
+            }
```
pabo99, line 1663: Same case here

frontend/server/src/Controllers/Course.php
```diff
@@ -1566,8 +1566,23 @@ public static function apiUpdateProblemsOrder(\OmegaUp\Request $r): array {
...
+        }
+        foreach ($decodedAliases as $alias) {
+            if (!is_string($alias)) {
+                throw new \OmegaUp\Exceptions\InvalidParameterException(
+                    'parameterInvalid',
+                    'problems'
+                );
+            }
+        }
+        /** @var list<string> $aliases */
```
pabo99: Here, it's no longer necessary to explicitly declare the type of `$aliases`, so it can be removed:

```suggestion
```

frontend/server/src/Controllers/Course.php
```diff
@@ -1633,8 +1648,23 @@ public static function apiUpdateAssignmentsOrder(\OmegaUp\Request $r): array {
...
+        }
+        foreach ($decodedAliases as $assignment) {
+            if (!is_string($assignment)) {
+                throw new \OmegaUp\Exceptions\InvalidParameterException(
+                    'parameterInvalid',
+                    'assignments'
+                );
+            }
+        }
+        /** @var list<string> $aliases */
```
pabo99: Same here

frontend/tests/controllers/CourseProblemsTest.php
```diff
@@ -187,4 +187,102 @@ public function testCourseProblemUsers() {
...
+                'course_alias' => $courseAlias,
+                'assignment_alias' => $assignmentAlias,
+                'problems' => '[null]',
+            ]));
+            $this->fail('Expected InvalidParameterException was not thrown');
+        } catch (\OmegaUp\Exceptions\InvalidParameterException $e) {
+            $this->assertSame('parameterInvalid', $e->getMessage());
+        }
+
+        // Test with [{}]
```
pabo99: Now that we're implementing `validateAlias()`, could you also add a test case where the array contains an invalid alias? For example, `["invalid alias"]`. This would help cover the additional validation provided by `validateAlias()`.

Same applies for `testUpdateAssignmentsOrderMalformedJson` function

frontend/tests/controllers/CourseProblemsTest.php
```diff
@@ -187,4 +187,102 @@ public function testCourseProblemUsers() {
         ]));
         $this->assertSame([], $response['identities']);
     }
+
+    public function testUpdateProblemsOrderMalformedJson(): void {
```
pabo99: Now that these tests cover more than malformed JSON, could we rename them to better reflect their scope? Something like `testUpdateProblemsOrderInvalidInput` and `testUpdateAssignmentsOrderInvalidInput` might be more representative.

As an optional improvement, this could also be a good candidate for a parameterized test using a data provider, since the same flow is being tested with several invalid inputs. We already have several examples of this pattern in the test suite

frontend/server/src/Controllers/Course.php
```diff
@@ -5914,7 +5944,7 @@ private static function getProblemsByAssignment(
      * @omegaup-request-param string $assignment_alias
      * @omegaup-request-param string $course_alias
      * @omegaup-request-param 'EXECUTION_COMPILATION_ERROR'|'EXECUTION_FINISHED'|'EXECUTION_INTERRUPTED'|'EXECUTION_JUDGE_ERROR'|'EXECUTION_RUNTIME_ERROR'|'EXECUTION_RUNTIME_FUNCTION_ERROR'|'EXECUTION_VALIDATOR_ERROR'|null $execution
-     * @omegaup-request-param 'c11-clang'|'c11-gcc'|'cat'|'cpp11-clang'|'cpp11-gcc'|'cpp17-clang'|'cpp17-gcc'|'cpp20-clang'|'cpp20-gcc'|'cs'|'go'|'hs'|'java'|'js'|'kj'|'kp'|'rk'|'kt'|'lua'|'pas'|'py2'|'py3'|'rb'|'rs'|null $language
+     * @omegaup-request-param 'c11-clang'|'c11-gcc'|'cat'|'cpp11-clang'|'cpp11-gcc'|'cpp17-clang'|'cpp17-gcc'|'cpp20-clang'|'cpp20-gcc'|'cs'|'go'|'hs'|'java'|'js'|'kj'|'kp'|'kt'|'lua'|'pas'|'py2'|'py3'|'rb'|'rs'|null $language
```
pabo99, line 5945: It looks like this change is unrelated to the PR. Could you please revert it?

PR #10182, opened 2026-09-02
fix: make cron mark_failure record a reason and phase status

### LINKED ISSUE #10148: bug: mark_failure() produces silent failure: Cron_Runs.error_text is NULL and all phases show success
### Lista de verificación

### Comportamiento Actual

Cuando el score MAP del modelo de recomendación queda por debajo de `--min-map-score`, el cron `build_problem_rec_model.py` llama a `cron_run.mark_failure()` en la línea 418 **sin lanzar una excepción**:

```python
# stuff/cron/build_problem_rec_model.py, dentro del bloque with cron_run.phase('train_model'):
else:
    logging.error(
        'Model NOT saved. Resulting accuracy was too low: '
        '%f below %f', score, args.min_map_score)
    cron_run.mark_failure()   # L418 — no raise
```

[Código fuente en el commit e7ce889c](https://github.com/omegaup/omegaup/blob/e7ce889c1b6630d2b3c4126c0dca6982a58836a4/stuff/cron/build_problem_rec_model.py#L418)

Como resultado, la fila insertada en `Cron_Runs` queda con:

| Columna | Valor |
|---|---|
| `status` | `'failure'` |
| `error_text` | `NULL` |
| `phases` (JSON) | `[{phase:'load_runs', status:'success'}, {phase:'train_model', status:'success'}]` |

Esto ocurre porque:

1. `runner.py` `mark_failure()` (línea 139) únicamente asigna `self._forced_failure = True`; no guarda ningún mensaje de error.
2. `runner.py` `_finish()` (líneas 191–193) sólo popula `error_text` cuando `exc_value is not None`. Como `mark_failure()` no lanza excepción, `exc_value` llega como `None` y `error_text` permanece `None`.
3. El context manager de `phase()` registra `status='failure'` únicamente si se propaga una `BaseException`. Sin excepción, la fase queda marcada como `'success'` aunque `_forced_failure` sea `True`.

Un operador que consulte runs fallidos ve `status='failure'` con `error_text=NULL` y todas las fases en `'success'`; la fila es indistinguible de un error de escritura en el propio `_finish()`, sin ninguna indicación de la causa real del fallo.

### Comportamiento Esperado

Cuando `mark_failure()` es la única señal de fallo (sin excepción propagada), la fila de `Cron_Runs` debería reflejar correctamente la causa:

- `error_text` debe contener un mensaje descriptivo (por ejemplo, el mensaje de accuracy baja que ya se registra en el log).
- La fase `train_model` dentro del JSON de `phases` debe tener `status='failure'`, no `'success'`.

De esta manera, un operador puede distinguir un fallo de negocio (score insuficiente) de un error de infraestructura, sin necesidad de cruzar con los logs del proceso.

**Pasos para reproducir (Python, sin base de datos real):**

1. Instalar dependencias del proyecto (`lib/db`, `stuff/cron/runner.py`).
2. Ejecutar el siguiente script que simula la ruta de código afectada:

```python
from stuff.cron import runner  # o mock equivalente

cron_run = runner.CronRun(no_track=True)

with cron_run.phase('train_model'):
    score = 0.05          # score simulado por debajo del umbral
    min_map_score = 0.10
    if score < min_map_score:
        cron_run.mark_failure()
        # sin raise — ésta es la ruta exacta de L418

# Inspeccionar estado resultante
print("phase[train_model].status =", cron_run._phases[0]['status'])
print("_forced_failure           =", cron_run._forced_failure)

# Simular _finish(exc_value=None)
error_text = None   # lógica actual: sólo se asigna si exc_value is not None
status = 'failure' if cron_run._forced_failure else 'success'
print("Cron_Runs.status     =", status)
print("Cron_Runs.error_text =", error_text)
```

**Salida observada:**

```
phase[train_model].status = success
_forced_failure           = True
Cron_Runs.status     = failure
Cron_Runs.error_text = None
```

**Entorno:**
- Commit afectado: `e7ce889c1b6630d2b3c4126c0dca6982a58836a4`
- Archivo: `stuff/cron/build_problem_rec_model.py`, línea 418
- Archivo secundario con el mismo patrón: `stuff/cron/assign_badges.py` (también llama a `mark_failure()` sin raise)
- Python 3.x, sin dependencia de navegador

deepak0x:
## Description

`mark_failure()` only set an internal flag. A run that failed without an exception was stored in `Cron_Runs` with `error_text` NULL and every phase in the `phases` JSON recorded as `success`, which is indistinguishable from a broken status write when you are reading the runs table.

Three changes in `stuff/lib/runner.py`:

- `mark_failure(reason)` accepts an optional reason that is written to `error_text`, so a forced failure now looks different from both an exception failure and a healthy run.
- A bare `mark_failure()` no longer leaves `error_text` NULL; it writes a fixed `marked as failed by the job` message.
- `mark_failure()` called inside a `phase()` block now marks that phase as `failure` in the `phases` JSON too, since the phase context is tracked with a depth counter.

The three callers were updated to pass a reason:

- `build_problem_rec_model.py` passes the low-accuracy message that was previously only going to the log, and the call sits inside the `train_model` phase, so that phase now records `failure` as the issue expects.
- `assign_badges.py` marks the failure inside the `process_badges` phase, right where the failing result is known.
- `aggregate_feedback.py` passes a reason; its phases already record their own failures through the exceptions they swallow per phase.

## Tests

Four regression tests in `stuff/lib/runner_test.py`, all failing against `main` before the fix and passing after:

- `test_mark_failure_records_error_text`
- `test_mark_failure_without_reason_still_writes_error_text`
- `test_mark_failure_inside_phase_marks_the_phase_failed`
- `test_mark_failure_outside_phase_keeps_phases_successful`

Full `stuff/lib` + `stuff/cron` suites pass: 107 tests, 1 skipped. `pylint` reports 10.00/10 on `runner.py` and `runner_test.py`, and the same scores as `main` on the three touched cron scripts.

Fixes #10148

### DIFF
--- a/stuff/cron/aggregate_feedback.py
+++ b/stuff/cron/aggregate_feedback.py
@@ -690,7 +690,7 @@ def main() -> None:
             dbconn.conn.close()
             logging.info('Done')
         if has_failures:
-            cron_run.mark_failure()
+            cron_run.mark_failure('some phases failed to complete')
     if has_failures:
         sys.exit(1)
 
--- a/stuff/cron/assign_badges.py
+++ b/stuff/cron/assign_badges.py
@@ -160,14 +160,14 @@ def main() -> None:
                 with cron_run.phase('process_badges'):
                     has_failures = process_badges(args.current_timestamp,
                                                   dbconn, cur, cur_readonly)
+                    if has_failures:
+                        cron_run.mark_failure('some badges failed to process')
             dbconn.conn.commit()
         finally:
             dbconn.conn.close()
             if dbconn_readonly is not dbconn:
                 dbconn_readonly.conn.close()
             logging.info('Finished')
-        if has_failures:
-            cron_run.mark_failure()
     if has_failures:
         sys.exit(1)
 
--- a/stuff/cron/build_problem_rec_model.py
+++ b/stuff/cron/build_problem_rec_model.py
@@ -412,10 +412,11 @@ def main() -> None:
                     # Save current model
                     model.save(args.output)
                 else:
-                    logging.error(
+                    reason = (
                         'Model NOT saved. Resulting accuracy was too low: '
-                        '%f below %f', score, args.min_map_score)
-                    cron_run.mark_failure()
+                        '%f below %f' % (score, args.min_map_score))
+                    logging.error('%s', reason)
+                    cron_run.mark_failure(reason)
         except Exception:  # pylint: disable=broad-except
             logging.exception('Failed to update recommendation model.')
             raise
--- a/stuff/lib/runner.py
+++ b/stuff/lib/runner.py
@@ -56,6 +56,9 @@ class CronRun:  # pylint: disable=too-many-instance-attributes
         self._rows_affected: Optional[int] = None
         self._start_monotonic = 0.0
         self._forced_failure = False
+        self._failure_reason: Optional[str] = None
+        self._phase_depth = 0
+        self._active_phase_forced = False
 
     @property
     def _lock_name(self) -> str:
@@ -110,6 +113,8 @@ class CronRun:  # pylint: disable=too-many-instance-attributes
         start = time.monotonic()
         status = 'success'
         error_class: Optional[str] = None
+        self._phase_depth += 1
+        self._active_phase_forced = False
         try:
             yield
         except BaseException as exc:
@@ -117,7 +122,11 @@ class CronRun:  # pylint: disable=too-many-instance-attributes
             error_class = type(exc).__name__
             raise
         finally:
+            self._phase_depth -= 1
             duration = round(time.monotonic() - start, 3)
+            if status == 'success' and self._active_phase_forced:
+                status = 'failure'
+            self._active_phase_forced = False
             self._phases.append({
                 'phase': name,
                 'status': status,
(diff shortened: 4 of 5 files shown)


pabo99 (approved):
Nice fix! this addresses the reported `build_problem_rec_model` scenario and the regression tests cover the expected persisted state. 

One minor concern: `_active_phase_forced` is shared and reset whenever a phase starts or finishes, so nested phases could lose or overwrite the forced-failure state of their parent phase. 

There are currently no nested phase usages, so this does not block the reported fix, but it may be worth either documenting that phases cannot be nested or tracking this state per phase.

deepak0x:
Went with the per-phase option in a13ec88dc. The forced flag now lives on a stack that mirrors the phase nesting: every phase pushes its own entry on entry and pops it on exit, and mark_failure marks only the innermost open phase. A parent phase keeps its forced state even when a nested phase starts or finishes on top of it.

I added two regression tests for nested phases. One of them fails against the previous code exactly along the lines you described (mark_failure before a nested phase lost the parent's failure because the inner phase reset the shared flag), and both pass now. The full runner suite is green locally and CI is happy too.

pabo99 (approved):

PR #10183, opened 2026-09-04
fix: query Identities instead of User_Rank in getUserIdsByUsername

### LINKED ISSUE #10160: bug: getUserIdsByUsername queries User_Rank instead of Identities, silently skips author filter
### Lista de verificación

### Comportamiento Actual

`Problems::getUserIdsByUsername()` consulta la tabla `User_Rank` para traducir usernames a `user_id`:

```php
// frontend/server/src/DAO/Problems.php, línea 146
$sql = "SELECT user_id FROM User_Rank WHERE username IN ({$placeholders})";
```

[Ver en código fuente (commit 5f34016b)](https://github.com/omegaup/omegaup/blob/5f34016bc62347e61346f65eb854e32823a4ad5c/frontend/server/src/DAO/Problems.php#L146)

`User_Rank` es una tabla desnormalizada que el script `stuff/cron/update_ranks.py` (líneas 160–204) pobla **únicamente** con usuarios que cumplen las tres condiciones siguientes:

1. Tienen al menos una solución aceptada (AC).
2. No son usuarios privados (`is_private = 0`).
3. No son administradores del sitio.

Como resultado, cualquier usuario que sea administrador del sitio, usuario privado, o que solo haya creado problemas sin haber resuelto ninguno, **no aparece en `User_Rank`**.

El llamador de esta función (líneas 543–563 del mismo archivo) aplica la lógica:

```php
if (!empty($authorUserIds)) {
    // agrega INNER JOIN y cláusula WHERE por autor
}
```

Cuando `getUserIdsByUsername` devuelve un arreglo vacío porque el usuario buscado no existe en `User_Rank`, el bloque `if` no se ejecuta, el filtro por autor **se omite silenciosamente**, y la consulta devuelve **todos** los problemas visibles en lugar de solo los del autor especificado.

### Comportamiento Esperado

Al filtrar problemas con `author=<username>`, la plataforma debe devolver únicamente los problemas cuyo autor corresponde al username proporcionado, independientemente de si ese usuario tiene soluciones aceptadas, es privado, o es administrador del sitio.

Si el username no corresponde a ningún usuario registrado, el resultado debería ser una lista vacía de problemas, no la lista completa de todos los problemas visibles.

**Pasos para reproducir:**

1. Identificar un usuario que sea administrador del sitio o que haya creado problemas pero nunca haya enviado una solución aceptada (por lo tanto ausente de `User_Rank`).
2. Realizar una petición al endpoint de listado de problemas con el parámetro `author=<username>` correspondiente a ese usuario.
3. Observar que la respuesta devuelve todos los problemas visibles en la plataforma en lugar de solo los problemas del autor indicado.

**Evidencia en el código (commit `5f34016b`):**

- `Problems.php` línea 146: la query apunta a `User_Rank`, no a `Identities`.
- `stuff/cron/update_ranks.py` líneas 160–204: confirma que `User_Rank` excluye usuarios sin AC, usuarios privados y site admins.
- `Problems.php` líneas 543–563: el `if (!empty($authorUserIds))` hace que un arreglo vacío descarte el filtro entero, retornando todos los resultados visibles.

La tabla canónica para la relación username → `user_id` es `Identities`, que contiene todos los usuarios registrados.

deepak0x:
## Description

`getUserIdsByUsername()` in `DAO/Problems.php` was translating usernames to `user_id` by querying `User_Rank`, which is a denormalized cache populated only by `stuff/cron/update_ranks.py` from users that have at least one AC solution, are not private, and are not site admins. Any other user (admins, private users, users who only create problems) was simply absent from the table, so the function returned an empty array and the caller in `searchProblems` skipped the `author` filter entirely. The endpoint then returned every visible problem instead of the author's.

Two changes:

- The query now reads from `Identities`, the canonical username → `user_id` mapping that contains every registered user. Identities without a linked `Users` row are filtered out, since they cannot own anything.
- When authors are specified but resolve to no user_ids (e.g. the username does not exist at all), the search adds a never-true clause instead of silently dropping the filter. A nonexistent author now returns an empty problem list, as expected.

## Tests

Two regression tests in `frontend/tests/controllers/ProblemListTest.php`, both failing against `main` before the fix and passing after:

- `testAuthorFilterWorksForUsersAbsentFromUserRank` — creates an author who has never solved a problem (so they are absent from `User_Rank`), creates a second author who has, calls `apiList` with the first author's username, and asserts the result contains the first author's problem and not the second's. On `main` the second author's problem leaks into the results because the filter is silently dropped.
- `testAuthorFilterWithNonexistentUsernameReturnsEmpty` — seeds a visible problem, calls `apiList` with a username that does not exist, and asserts the result is empty. On `main` the seeded problem is returned.

Full `controllers/ProblemListTest.php` suite passes: 55 tests, 626 assertions. `controllers/ProblemListCourseTest.php` also passes (2 tests, 52 assertions). `phpcs` and `stuff/lint.sh validate --all` are clean on the changed files.

Fixes #10160

### DIFF
--- a/frontend/server/src/DAO/Problems.php
+++ b/frontend/server/src/DAO/Problems.php
@@ -143,15 +143,30 @@ class Problems extends \OmegaUp\DAO\Base\Problems {
         }
         // Create placeholders (?, ?, ?) for each username
         $placeholders = implode(',', array_fill(0, count($usernames), '?'));
-        $sql = "SELECT user_id FROM User_Rank WHERE username IN ({$placeholders})";
+        // `Identities` is the canonical username -> user_id mapping and
+        // contains every registered user, including admins, private
+        // users, and users who have only ever created problems.
+        // `User_Rank` is a denormalized cache populated only from users
+        // with at least one AC solution, so querying it here caused the
+        // `author=` filter to silently drop for everyone else.
+        $sql = "SELECT user_id FROM Identities WHERE username IN ({$placeholders})";
 
-        /** @var list<array{user_id: int}> */
+        /** @var list<array{user_id: int|null}> */
         $results = \OmegaUp\MySQLConnection::getInstance()->GetAll(
             $sql,
             $usernames
         );
 
-        return array_map(fn($row) => intval($row['user_id']), $results);
+        return array_map(
+            fn($row) => intval($row['user_id']),
+            // Some identities are not linked to a Users row yet
+            // (`Identities.user_id` is nullable); those cannot own
+            // problems and must be skipped.
+            array_filter(
+                $results,
+                fn($row) => !is_null($row['user_id'])
+            )
+        );
     }
 
     /**
@@ -560,6 +575,13 @@ class Problems extends \OmegaUp\DAO\Base\Problems {
                     'pa_acl.owner_id IN (' . $placeholders . ')',
                     $authorUserIds,
                 ];
+            } else {
+                // Every requested author username failed to resolve to a
+                // user id. This can only happen when the username does
+                // not exist in `Identities`; an empty result set is the
+                // correct answer rather than silently dropping the
+                // filter and returning every visible problem.
+                $clauses[] = ['0 = 1', []];
             }
         }
 
--- a/frontend/server/src/DAO/Problems.php
+++ b/frontend/server/src/DAO/Problems.php
@@ -157,16 +157,16 @@ class Problems extends \OmegaUp\DAO\Base\Problems {
             $usernames
         );
 
-        return array_map(
+        // Some identities are not linked to a Users row yet
+        // (`Identities.user_id` is nullable); those cannot own
+        // problems and must be skipped.
+        return array_values(array_map(
             fn($row) => intval($row['user_id']),
-            // Some identities are not linked to a Users row yet
-            // (`Identities.user_id` is nullable); those cannot own
-            // problems and must be skipped.
             array_filter(
                 $results,
                 fn($row) => !is_null($row['user_id'])
             )
-        );
+        ));
     }
 
     /**
--- a/frontend/server/src/DAO/Problems.php
+++ b/frontend/server/src/DAO/Problems.php
@@ -149,24 +149,15 @@ class Problems extends \OmegaUp\DAO\Base\Problems {
         // `User_Rank` is a denormalized cache populated only from users
         // with at least one AC solution, so querying it here caused the
         // `author=` filter to silently drop for everyone else.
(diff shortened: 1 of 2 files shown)


frontend/server/src/DAO/Problems.php
```diff
@@ -143,15 +143,30 @@ private static function getUserIdsByUsername(array $usernames): array {
...
-        return array_map(fn($row) => intval($row['user_id']), $results);
+        // Some identities are not linked to a Users row yet
+        // (`Identities.user_id` is nullable); those cannot own
+        // problems and must be skipped.
+        return array_values(array_map(
+            fn($row) => intval($row['user_id']),
+            array_filter(
+                $results,
+                fn($row) => !is_null($row['user_id'])
+            )
```
pabo99: Could you filter out null `user_id`s directly in the SQL query with `AND user_id IS NOT NULL` and avoid this additional filtering in PHP?

frontend/tests/controllers/ProblemListTest.php
```diff
@@ -1989,4 +1989,82 @@ public function testProblemListSearchWithWrongSearchType() {
...
+        // missing from the denormalized `User_Rank` table.
+        ['identity' => $author] = \OmegaUp\Test\Factories\User::createUser(
+            new \OmegaUp\Test\Factories\UserParams(
+                ['username' => 'author_without_solutions']
+            )
+        );
+        // A second author that has solved at least one problem, so they
+        // would be present in `User_Rank` and serve as a control.
+        ['identity' => $otherAuthor] = \OmegaUp\Test\Factories\User::createUser();
+        $otherRunData = \OmegaUp\Test\Factories\Run::createRunToProblem(
```
pabo99: Grading this run doesn't appear to populate `User_Rank` without running the ranking cron. Since the second problem only serves as a control to detect a dropped author filter, could we remove this setup or adjust the comment so it doesn't claim that the author is present in `User_Rank`?

frontend/server/src/DAO/Problems.php
deepak0x: Done — the null filtering now happens directly in SQL with `AND user_id IS NOT NULL`, and the PHP-side `array_filter` is gone.

frontend/tests/controllers/ProblemListTest.php
deepak0x: You're right, grading the run alone doesn't populate `User_Rank`. I removed the grading setup entirely and adjusted the comment: the second author's problem is kept purely as a control so a silently dropped `author` filter shows up as both problems leaking into the results. Pushed in 7a97035de.

pabo99 (approved):
Cool!

PR #10186, opened 2026-09-04
fix: warn users about unsaved problems when leaving course content page

### LINKED ISSUE #9444: [Bug] Warn users about unsaved problems when leaving the “Add course content” page

AbdulKreemShah2408:
Fixes: #9444

# Comments

This PR implements the browser's native unsaved-changes warning for the 
"Add course content" page, as requested in the issue.

## Changes
- Added a `hasUnsavedChanges` getter in `AssignmentDetails.vue` that checks 
  whether any problems have been added or any form fields (name, description, 
  alias) have been filled in, but only when creating a **new** assignment 
  (`AssignmentFormMode.New`).
- Registered a native `beforeunload` listener (added in `mounted()`, removed 
  in `beforeDestroy()`) that triggers the browser's built-in "Reload site? / 
  Changes you made may not be saved." warning when the user tries to close 
  the tab or reload the page with unsaved changes.
- Added a `beforeRouteLeave` navigation guard so the same warning (via 
  `window.confirm`) also applies when navigating away using Vue Router 
  (e.g. clicking another in-app link), not just on full page unload.
- Added the `courseUnsavedChangesWarning` translation string to the `.lang` 
  files (en, es, pt) and regenerated the derived `lang.*.ts` / `lang.*.json` 
  files via `./stuff/lint.sh`.

## Testing
Tested locally: added a problem to a new assignment without clicking 
"Add content", then tried to reload/close the tab — the native browser 
warning appeared as expected (see screenshot). Ran `./stuff/lint.sh` with 
a clean `OK` result.

# Checklist:

- [x] The code follows the [coding guidelines]
- [x] I have added and/or updated the tests (n/a — no new test file added; 
  existing lint/build passes)
- [x] I have executed the linter and my code passes all checks

https://github.com/user-attachments/assets/9228a5db-ff90-4078-9d18-e8fc4618a011


frontend/www/js/omegaup/components/course/Edit.vue
```diff
@@ -454,8 +454,22 @@ export default class CourseEdit extends Vue {
...
   onSelectAssignmentTab(): void {
-    this.showTab = 'content';
-    this.onResetAssignmentForm();
+    this.switchTab('content');
+  }
+
+  switchTab(newTab: string): void {
+    if (
+      this.showTab === 'content' &&
+      ((this.assignmentDetails as unknown) as { hasUnsavedChanges?: boolean })
```
pabo99: `AssignmentDetails` doesn't currently define `hasUnsavedChanges`, so this condition always evaluates to false and the warning is never shown. Could you please add the getter described in the PR description and expose the actual unsaved state here?

frontend/www/js/omegaup/components/course/Edit.vue
```diff
@@ -454,8 +454,22 @@ export default class CourseEdit extends Vue {
...
+        ?.hasUnsavedChanges &&
+      !window.confirm(T.courseUnsavedChangesWarning)
+    ) {
+      return;
+    }
+    this.showTab = newTab;
+    if (newTab === 'content') {
+      this.onResetAssignmentForm();
+    }
```
pabo99, line 472: The PR description mentions a `beforeunload` listener, but it is not present in the current changes. Switching the internal tabs is only one way to discard the form; reloading, closing the tab, or following another link still loses the selected problems without warning. 

Could you please implement and clean up the native `beforeunload` handler as requested in the issue?

pabo99 (commented):
Could you please add regression tests covering both outcomes: cancelling the confirmation keeps the content tab and its unsaved data, while accepting it allows navigation? It would also be useful to test that the `beforeunload` handler is registered only while there are unsaved changes.

pabo99 (commented):
Thanks for the updates! I tested the branch and the warning now works well for changes to the `title`, `alias`, and `description`, both when switching tabs and with `beforeunload`.

However, the main case we wanted to address is still not covered: 
if I add problems to a new assignment without changing those fields and then navigate away without clicking the save button, no confirmation is shown and the selected problems are lost.

I’m attaching a short video showing this case:

https://github.com/user-attachments/assets/864bd495-9eec-4435-87cb-0fd7183dbaf9

Could you please take another look at how we detect unsaved problems? I think this is the important case we should address before merging this PR.

I also noticed that there are other assignment settings that can be modified, such as the content type and start/end dates. Ideally, any unsaved change should eventually trigger the warning, but I think we can improve that separately and keep this PR focused on preventing the accidental loss of newly added problems.

frontend/www/js/omegaup/components/course/AssignmentDetails.vue
```diff
@@ -404,6 +404,69 @@ export default class CourseAssignmentDetails extends Vue {
...
+      this.description !== (this.assignment.description || '') ||
+      this.alias !== (this.assignment.alias || '') ||
+      this.assignmentType !== (this.assignment.assignment_type || 'homework') ||
+      this.unlimitedDuration !== !this.assignment.finish_time ||
+      (this.startTime &&
+        this.assignment.start_time &&
+        this.startTime.getTime() !== this.assignment.start_time.getTime()) ||
+      (this.finishTime &&
+        this.assignment.finish_time &&
+        this.finishTime.getTime() !== this.assignment.finish_time.getTime());
```
pabo99: Could you wrap these two date comparisons with `Boolean()`? 
```suggestion
      Boolean(
        this.startTime &&
          this.assignment.start_time &&
          this.startTime.getTime() !== this.assignment.start_time.getTime(),
      ) ||
      Boolean(
        this.finishTime &&
          this.assignment.finish_time &&
          this.finishTime.getTime() !== this.assignment.finish_time.getTime(),
      );
```
The `&&` expressions can resolve to a non-boolean value when one of the dates is missing, which is causing the Cypress type check to fail

frontend/www/js/omegaup/components/course/Edit.vue
```diff
@@ -458,6 +458,21 @@ export default class CourseEdit extends Vue {
     this.onResetAssignmentForm();
   }
 
+  switchTab(tab: string): void {
+    if (
+      this.showTab === 'content' &&
+      (this.assignmentDetails as any)?.hasUnsavedChanges &&
```
pabo99: Could you avoid using `any` here and properly type `assignmentDetails` instead? We already use explicitly typed `@Ref()` properties in several components, so it would be better to follow the same pattern here rather than bypassing `TypeScript`'s type checking

pabo99 (commented):
This looks much better now! Just a couple of minor corrections to address the Cypress errors and take better advantage of TypeScript's type checking. Thanks for the improvements!

Issue #10138, opened 2026-08-28
[FEATURE] Per-tag solved statistics on user profiles


deepak0x:
### Checklist
- [x] I have searched existing issues and confirmed this is not a duplicate
- [ ] I have discussed this on GitHub Discussions or Discord and received positive feedback

### Context
User profiles show overall solved counts but not a breakdown by topic/tag, so it's hard to see a user's strengths per category.

### Feature description
Add a 'Topics' section to user profiles showing accepted-problem counts broken down by public tag. Display as a sortable table (tag name, solved count) on the profile page, visible to anyone viewing it.

### Value
Better self-assessment and a clearer signal of topic strengths for users and recruiters.

deepak0x:
cc @pabo99 — happy to implement this if it fits the roadmap.

pabo99:
Thanks for the proposal! It looks like we already provide similar information in the user statistics section through the Tags Solved chart:

<img width="1046" height="800" alt="Image" src="https://github.com/user-attachments/assets/19970bd6-f0f0-4067-a375-52ba58888175" />

Could you take a look at the existing implementation and clarify what additional value the proposed table would provide compared to the current visualization?

deepak0x:
You're right that the underlying data is the same, so the question is what the table would add on top of the donut. Looking at the current implementation, the gaps I see are specific:

1. The chart is fed by `getTagsDistributionForSolvedProblems` (`frontend/server/src/DAO/ProblemsTags.php:88`), which caps the list at the top 10 tags and merges everything else into a single "Others" slice (the `$maxTags = 10` logic). A user active across 25 public tags only ever sees 10 of them, and the rest lose all per-tag detail.
2. A pie shows proportions, so tags with close counts are hard to tell apart, and there is no way to sort or compare exact numbers beyond the legend.
3. The order is fixed by count, so scanning for a specific topic means reading the whole legend.

The proposed table closes those: full tag list instead of top-10-plus-Others, exact counts, sortable columns. It would also leave room for a "solved out of total tagged" column later, which a pie cannot express, though that is not part of what I proposed here.

If keeping the profile lean matters more, the same gaps could be handled as an enhancement to the chart instead: raise or drop the cap, or add a chart/table toggle in that card. I would be fine closing this in favor of that if it sounds more natural to you. Wanted to get your read before doing anything.

pabo99:
[...]

PR #10188, opened 2026-09-05
feat: expose full tag distribution in profile statistics

deepak0x:
## Summary

The profile statistics API only exposed a capped tag distribution (top 10 with the rest folded into 'Others'). This adds `tagsFull` with the complete distribution so clients can render every tag without losing the tail.

- `apiProfileStatistics` now returns both `tags` (capped, unchanged) and `tagsFull` (complete).
- The DAO fetches the full list once via `getTagsDistributionForSolvedProblems`, and the new `capTagsDistribution` derives the capped variant with the 'Others' grouping, so the query runs only once per request.
- The `TagDistribution` psalm type alias documents the shared shape, and the docblock spells out the `tags`/`tagsFull` semantic difference.

Part of #10138

## Test plan

- [x] PHPUnit: `UserProfileStatisticsTest` covers the new payload field, the capped shape with 10 entries plus 'Others', and the alphabetical tie-break for equal counts (10 tests, 138 assertions)
- [x] Psalm passes with no errors
- [x] `api_types.ts` and the controllers README regenerated through the api linter

The UI half of the original change (the chart/table toggle in `TagsSolvedChart`, its translations and its Jest tests) will follow in a separate PR so each part stays reviewable on its own.

### DIFF
--- a/frontend/server/src/Controllers/User.php
+++ b/frontend/server/src/Controllers/User.php
@@ -7,6 +7,7 @@ namespace OmegaUp\Controllers;
  *
  * @psalm-type PageItem=array{class: string, label: string, page: int, url?: string}
  * @psalm-type AuthorsRank=array{ranking: list<array{author_ranking: int|null, author_score: float, classname: string, country_id: null|string, name: null|string, username: string}>, total: int}
+ * @psalm-type TagDistribution=array{name: string, count: int}
  * @psalm-type AuthorRankTablePayload=array{length: int, page: int, ranking: AuthorsRank, pagerItems: list<PageItem>}
  * @psalm-type Badge=array{assignation_time: \OmegaUp\Timestamp|null, badge_alias: string, first_assignation: \OmegaUp\Timestamp|null, owners_count: int, total_users: int}
  * @psalm-type ApiToken=array{name: string, timestamp: \OmegaUp\Timestamp, last_used: \OmegaUp\Timestamp, rate_limit: array{reset: \OmegaUp\Timestamp, limit: int, remaining: int}}
@@ -2381,9 +2382,15 @@ class User extends \OmegaUp\Controllers\Controller {
     /**
      * Get profile statistics including solved problems by difficulty and tags distribution.
      *
+     * `tags` is the capped distribution: at most 10 entries, with every
+     * remaining tag aggregated into a single 'Others' entry. `tagsFull` is
+     * the complete distribution with no cap and no aggregation, so `tags` is
+     * not simply a subset of `tagsFull`: the tail entries are merged into
+     * the 'Others' entry instead of appearing on their own.
+     *
      * @throws \OmegaUp\Exceptions\ForbiddenAccessException
      *
-     * @return array{solved: int, attempting: int, difficulty: array{easy: int, medium: int, hard: int, unlabelled: int}, tags: list<array{name: string, count: int}>}
+     * @return array{solved: int, attempting: int, difficulty: array{easy: int, medium: int, hard: int, unlabelled: int}, tags: list<TagDistribution>, tagsFull: list<TagDistribution>}
      *
      * @omegaup-request-param null|string $username
      */
@@ -2412,8 +2419,11 @@ class User extends \OmegaUp\Controllers\Controller {
         $attemptingCount = \OmegaUp\DAO\Problems::getAttemptingCount(
             $identity->identity_id
         );
-        $tagsDistribution = \OmegaUp\DAO\ProblemsTags::getTagsDistributionForSolvedProblems(
+        $tagsFullDistribution = \OmegaUp\DAO\ProblemsTags::getTagsDistributionForSolvedProblems(
             $identity->identity_id
+        );
+        $tagsDistribution = \OmegaUp\DAO\ProblemsTags::capTagsDistribution(
+            $tagsFullDistribution
         );
 
         return [
@@ -2426,6 +2436,7 @@ class User extends \OmegaUp\Controllers\Controller {
                 'unlabelled' => $difficultyStats['unlabelled'],
             ],
             'tags' => $tagsDistribution,
+            'tagsFull' => $tagsFullDistribution,
         ];
     }
 
--- a/frontend/server/src/DAO/ProblemsTags.php
+++ b/frontend/server/src/DAO/ProblemsTags.php
@@ -79,15 +79,49 @@ class ProblemsTags extends \OmegaUp\DAO\Base\ProblemsTags {
     }
 
     /**
-     * Get tag distribution for problems solved by an identity.
-     * Returns tags sorted by count descending.
-     * If there are more than $maxTags, the remaining tags are grouped under 'Others'.
+     * Derive the capped tag distribution from the full one: the first
+     * $maxTags entries are kept as is, and the remaining tags are
+     * aggregated into a single 'Others' entry.
+     *
+     * @param list<array{name: string, count: int}> $allTags
+     *
+     * @return list<array{name: string, count: int}>
+     */
+    public static function capTagsDistribution(
+        array $allTags,
+        int $maxTags = 10
+    ): array {
+        if (count($allTags) <= $maxTags) {
+            return $allTags;
+        }
+
+        $topTags = array_slice($allTags, 0, $maxTags);
+        $othersCount = 0;
+        foreach (array_slice($allTags, $maxTags) as $tag) {
+            $othersCount += $tag['count'];
+        }
+
+        if ($othersCount > 0) {
+            $topTags[] = [
+                'name' => 'Others',
+                'count' => $othersCount,
+            ];
+        }
+
+        return $topTags;
+    }
+
+    /**
+     * Get the full tag distribution for problems solved by an identity:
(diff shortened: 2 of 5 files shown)


deepak0x:
Screenshots of the change running locally:

**Tags solved - chart view**
chart

**Tags solved - table view** (full tag list, sortable columns)
table

The chart keeps the existing capped grouping with an Others bucket, while the table lists every tag. Both views are reachable from the toggle buttons.

frontend/server/src/DAO/ProblemsTags.php
```diff
@@ -78,6 +78,18 @@ public static function getTagsForProblem(
...
+     * Returns the FULL list of public, non-meta tags sorted by count descending.
+     *
+     * @return list<array{name: string, count: int}>
+     */
+    public static function getTagsDistributionForSolvedProblemsAll(
+        int $identityId
+    ): array {
+        return self::getTagsDistributionForSolvedProblemsRows($identityId);
+    }
+
```
pabo99: Do we need this additional public method?

I see that it currently only delegates to `getTagsDistributionForSolvedProblemsRows()`. If the goal is to keep that method private as an implementation detail, this makes sense; otherwise, could we avoid the extra wrapper and call it directly?

frontend/server/src/Controllers/User.php
```diff
@@ -2414,6 +2414,9 @@ public static function apiProfileStatistics(\OmegaUp\Request $r): array {
         $tagsDistribution = \OmegaUp\DAO\ProblemsTags::getTagsDistributionForSolvedProblems(
             $identity->identity_id
         );
+        $tagsFullDistribution = \OmegaUp\DAO\ProblemsTags::getTagsDistributionForSolvedProblemsAll(
```
pabo99: Do we really need to execute the same tags distribution query twice here?

Since both `getTagsDistributionForSolvedProblems()` and
`getTagsDistributionForSolvedProblemsAll()` ultimately retrieve the same full
dataset, could we fetch it only once and derive the capped distribution from
that result?

frontend/server/src/Controllers/User.php
```diff
@@ -2425,6 +2428,7 @@ public static function apiProfileStatistics(\OmegaUp\Request $r): array {
                 'unlabelled' => $difficultyStats['unlabelled'],
             ],
             'tags' => $tagsDistribution,
+            'tagsFull' => $tagsFullDistribution,
```
pabo99, line 2439: Would it make sense to document more explicitly the semantic difference
between `tags` and `tagsFull`?

`tags` is not simply a subset of `tagsFull`, since the remaining values are
aggregated into `Others`

frontend/tests/controllers/UserProfileStatisticsTest.php
```diff
@@ -248,4 +248,110 @@ public function testProfileStatisticsUserNotFound() {
...
+            ])
+        );
+
+        $this->assertArrayHasKey('tags', $response);
+        $this->assertArrayHasKey('tagsFull', $response);
+
+        $this->assertGreaterThanOrEqual(
+            count($tagNames),
+            count($response['tagsFull']),
+        );
```
pabo99: Is `GreaterThanOrEqual` necessary here?

If this test fixture is expected to produce exactly these tags, could we use
`assertCount(count($tagNames), $response['tagsFull'])` instead? That would help
catch unexpected tags being returned as well

frontend/tests/controllers/UserProfileStatisticsTest.php
```diff
@@ -248,4 +248,110 @@ public function testProfileStatisticsUserNotFound() {
...
+        foreach ($tagNames as $expected) {
+            $this->assertContains($expected, $returnedFullNames);
+        }
+    }
+
+    /**
+     * tagsFull must be sorted by count descending so the front-end can
+     * render an initial sort without an extra server round-trip.
+     */
+    public function testProfileStatisticsTagsFullSortedByCount() {
```
pabo99, line 330: Could you add a test case where two tags have the same count?

The query currently uses `ORDER BY count DESC, t.name ASC`, but the test only verifies that the counts are sorted in descending order. Adding a tie case would also verify the secondary alphabetical ordering by tag name

frontend/server/src/Controllers/User.php
```diff
@@ -2382,7 +2382,7 @@ private static function generateHeatmapData(
      *
      * @throws \OmegaUp\Exceptions\ForbiddenAccessException
      *
-     * @return array{solved: int, attempting: int, difficulty: array{easy: int, medium: int, hard: int, unlabelled: int}, tags: list<array{name: string, count: int}>}
+     * @return array{solved: int, attempting: int, difficulty: array{easy: int, medium: int, hard: int, unlabelled: int}, tags: list<array{name: string, count: int}>, tagsFull: list<array{name: string, count: int}>}
```
pabo99: Since `list<array{name: string, count: int}>` is already used in several places, would it make sense to define a reusable Psalm type alias for this structure instead of repeating it?

For example, something like `TagDistribution` could make the return types easier to maintain and reuse

pabo99 (commented):
Thanks for the changes, @deepak0x!

Could you please keep this PR focused only on the backend/controller-related changes needed to expose the full tags distribution?

I think it would be better to move all the UI-related changes, including the chart/table toggle, translations, and frontend tests, into a separate PR.

That should make both reviews smaller and easier to follow, and we can validate the API changes independently before reviewing the UI implementation.

deepak0x:
Split done as suggested. This PR now keeps only the backend half: the controller, the DAO, the PHPUnit coverage and the generated api_types/README files. The UI half (TagsSolvedChart chart/table toggle, translations and Jest tests) will follow in a separate PR built on top of this one.

The inline points are addressed as well:

- The delegating wrapper in ProblemsTags is gone. getTagsDistributionForSolvedProblems now returns the full list directly, and a new pure capTagsDistribution derives the capped variant.
- The query runs only once per request: the controller fetches the full list and derives the capped distribution from it.
- The tags/tagsFull semantic difference is documented in the apiProfileStatistics docblock and carried into the generated README.
- The tests use assertCount now, and a tie case covers the alphabetical ordering for equal counts.

Verified locally: 10 tests, 138 assertions pass, and Psalm reports no errors.

frontend/server/src/DAO/ProblemsTags.php
deepak0x: Done. The wrapper is gone: getTagsDistributionForSolvedProblems returns the full list directly and capTagsDistribution derives the capped variant, so both shapes come from the same single fetch.

frontend/server/src/Controllers/User.php
deepak0x: Fixed. The controller now fetches the full list once and derives the capped distribution from it, so the query runs only once per request.

frontend/server/src/Controllers/User.php
deepak0x: Documented in the apiProfileStatistics docblock: tags is the capped distribution with the tail aggregated into Others, and tagsFull is the complete one. The wording also makes explicit that tags is not simply a subset of tagsFull. The generated README entry carries the same explanation.

frontend/tests/controllers/UserProfileStatisticsTest.php
deepak0x: Switched to assertCount for tagsFull. The capped tags assertion also checks the exact 10-plus-Others shape now, so unexpected extra entries would fail the test too.

frontend/tests/controllers/UserProfileStatisticsTest.php
deepak0x: Added testProfileStatisticsTagsFullTieBreaksByName: two tags with equal counts must come back alphabetically, which exercises the ORDER BY count DESC, t.name ASC tiebreak.

frontend/server/src/Controllers/User.php
deepak0x: Added a TagDistribution psalm type alias and the apiProfileStatistics return type now uses it for both tags and tagsFull. The alias also expands into the generated api_types.ts and README entries.

frontend/server/src/DAO/ProblemsTags.php
```diff
@@ -79,15 +79,49 @@ public static function getTagsForProblem(
...
 
     /**
-     * Get tag distribution for problems solved by an identity.
-     * Returns tags sorted by count descending.
-     * If there are more than $maxTags, the remaining tags are grouped under 'Others'.
+     * Derive the capped tag distribution from the full one: the first
+     * $maxTags entries are kept as is, and the remaining tags are
+     * aggregated into a single 'Others' entry.
+     *
+     * @param list<array{name: string, count: int}> $allTags
```
pabo99, line 86: Now that we have the new type definition, could you use:

```suggestion
     * @param list<TagDistribution> $allTags
```

and do the same on lines 88, 121?

For this to work, you'll also need to declare the type definition in this file. Just make sure the definition is exactly the same everywhere it's declared

pabo99 (commented):
Just one last small suggestion, and then this should be ready to merge

pabo99 (approved):
Thanks!

PR #10190, opened 2026-09-05
fix: allow more time for the arena problem link to appear in cypress helper

### LINKED ISSUE #10161: Flaky cypress (contest-group): arena problem link times out in createRunsInsideContest
## Current Behavior

The `cypress (contest-group)` job fails intermittently on `main` and on PRs whose changes have nothing to do with the arena or the cypress suite. The failure is always the same assertion inside `createRunsInsideContest`:

```
AssertionError: Timed out retrying after 4000ms: Expected to find element: `a[data-problem="..."]`, but never found it.
  at Context.eval (webpack://omegaup_frontend/./cypress/support/commands.ts:282:9)
```

Recent cases:

- Run on `main`, Aug 26: https://github.com/omegaup/omegaup/actions/runs/33004102814. Two tests failed on that line (`Should create a contest when the scoreboard is never shown` with `a[data-problem="897702f3-a"]`, and `Should keep the submissions tab active when opening run details as contest admin` with `a[data-problem="e04c8ba4-6"]`), and a third one (`Should test practice mode in contest`) timed out on the `[name="language"]` wait a few lines below in the same helper.
- PR #10140, a backend-only change under `frontend/server/src/DAO`: https://github.com/omegaup/omegaup/actions/runs/33300678609 failed on `Should create a contest and participate in it as a identity` (`a[data-problem="60fe82a3-1"]`, same line), while https://github.com/omegaup/omegaup/actions/runs/33300678630 on the same commit passed the same job. The PR diff does not touch any frontend or cypress file.

The same commit passing and failing the same job is a clear sign of a timing problem rather than a regression.

## Expected Behavior

A PR that does not touch the frontend or the cypress tests should not go red on this job, and `main` should not fail it either. When the runner is slow, the test should give the page time to finish loading instead of failing the build. Every false failure here costs a re-run and CI notifications for both the contributor and the maintainers, for something that is not a code problem.

## Steps to reproduce

It depends on CI load, so it does not reproduce on every run. The code path is deterministic:

1. `createRunsInsideContest` visits `/arena/<contest>/#problems`.
2. It then waits for the problem link with the default 4 second timeout:

```ts
cy.get(`a[data-problem="${problem.problemAlias}"]`).click();
```

3. The arena view loads the problem list through the API before those links render. On a busy shared runner that can take longer than 4 seconds, and the assertion gives up before the links appear.

## Suggested fix

The helper already gives a later element on the same page a 10 second window:

```ts
cy.get('[name="language"]', { timeout: 10000 })
```

Giving the problem link an explicit window too:

```ts
cy.get(`a[data-problem="${problem.problemAlias}"]`, { timeout: 15000 }).click();
```

would remove most of these false failures without weakening the check. `cy.waitUntil` (already used in `cypress/support/pageObjects/contestPage.ts`) is an alternative if a polling retry is preferred over one longer timeout.

## Environment

GitHub Actions, `cypress (contest-group)` job, `contest.cy.ts`. I searched existing issues for this flake and the failing test names and did not find a prior report.

deepak0x:
The arena navbar renders the problem links only after the contest details request comes back, so right after `cy.visit` the `a[data-problem]` element is not in the DOM yet. The lookup in `createRunsInsideContest` used the default 4 second command timeout, which turned into a race against the backend under CI load and showed up as intermittent failures across the cypress matrix.

The same helper already waits up to 10 seconds for the language selector a few lines below, so the problem link now gets the same window:

```ts
cy.get(`a[data-problem="${problem.problemAlias}"]`, { timeout: 10000 }).click();
```

No behavioral change on a healthy run; the extra window only matters when the backend is slow to answer.

Testing: exercised the helper through the contest specs that call `createRunsInsideContest` against a local stack; passes consistently.

Fixes #10161

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -279,7 +279,9 @@ Cypress.Commands.add(
         return;
       }
       cy.visit(`/arena/${contestAlias}/#problems`);
-      cy.get(`a[data-problem="${problem.problemAlias}"]`).click();
+      cy.get(`a[data-problem="${problem.problemAlias}"]`, {
+        timeout: 10000,
+      }).click();
 
       // Mocking date just a few seconds after to allow create new run
       cy.clock(new Date(), ['Date']).then((clock) => clock.tick(9000));


pabo99 (approved):
This looks reasonable for the reported flake. If it still happens, a follow-up could harden `enterContest()` by waiting for `/api/contest/open/` after clicking start, before `createRunsInsideContest()` navigates to the arena

Issue #8683, opened 2025-12-22
[FEATURE]  Friends/Following System for Social Engagement


Ankitsinghsisodya:
# As a **user**, I want **to follow other users and see their activity** so that **I can stay motivated by tracking my friends' progress and compete with them**

## Context

Currently, omegaUp lacks social features that exist on other competitive programming platforms like Codeforces and LeetCode. Users cannot:

- Follow/unfollow other users
- See a list of users they follow or who follow them
- View an activity feed showing what problems their friends have solved
- Filter contest scoreboards to show only friends

### How has this issue affected you?

As a competitive programmer, I often practice with friends and classmates. However, on omegaUp:

1. **No visibility into friends' progress** - I can't see what problems my study partners are solving or how active they are
2. **No friendly competition** - I can't easily compare my progress with specific users I care about
3. **No community building** - The platform feels isolated; I can't build a network of coders I admire or want to learn from
4. **Reduced motivation** - Social accountability is a powerful motivator. Without seeing friends' activity, there's less incentive to practice daily

### What is the goal to achieve?

Implement a Friends/Following system similar to Codeforces that includes:

1. **Follow/Unfollow functionality**

   - Button on user profiles to follow/unfollow
   - API endpoints: `POST /api/user/follow`, `POST /api/user/unfollow`

2. **Followers/Following Lists**

   - View who follows you and who you follow
   - Follower/Following counts on profile page

3. **Activity Feed**

   - See recent problems solved by users you follow
   - Optional: See when they participate in contests

4. **Friends-only Scoreboard** (nice-to-have)
   - Filter contest scoreboards to show only users you follow
   - Similar to Codeforces "Friends" standings

### Expected Impact

- **Increased engagement**: Social features drive daily active users
- **Better retention**: Users stay connected to their community
- **Competitive motivation**: Friendly rivalry encourages more practice
- **Community growth**: Easier to discover and connect with skilled coders

### Reference Implementations

- **Codeforces**: Friends list, activity feed, friends-only standings
- **LeetCode**: Following system with activity updates
- **GitHub**: Follow users, see their activity in feed

I want to work on this issue.

Yashagarwal9798:
@pabo99 If the currently assigned person isn’t working on this anymore, could you please assign it to me?

Kunal1522:
please have a look at pr #8602 also if u are willing to take over

Yashagarwal9798:
[...]

PR #10199, opened 2026-09-07
Add user follow relationship schema with follower and followed identity references

vijay-x-dev:
# Description

This adds the user connection table.

It stores who follows whom, prevents duplicate follows, and links to identities so we can quickly check followers and following relationships.

Part of: #8683 

# Checklist:

### DIFF
--- a/frontend/database/00283_create_user_followers_table.sql
+++ b/frontend/database/00283_create_user_followers_table.sql
@@ -1,0 +1,12 @@
+CREATE TABLE `User_Connections` (
+  `follower_id` int NOT NULL AUTO_INCREMENT,
+  `follower_identity_id` int NOT NULL,
+  `followed_identity_id` int NOT NULL,
+  `created_at` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  PRIMARY KEY (`follower_id`),
+  UNIQUE KEY `unique_follow_relationship` (`follower_identity_id`, `followed_identity_id`),
+  KEY `idx_follower_identity_id` (`follower_identity_id`),
+  KEY `idx_followed_identity_id` (`followed_identity_id`),
+  CONSTRAINT `fk_follower_identity_id` FOREIGN KEY (`follower_identity_id`) REFERENCES `Identities` (`identity_id`) ON DELETE CASCADE,
+  CONSTRAINT `fk_followed_identity_id` FOREIGN KEY (`followed_identity_id`) REFERENCES `Identities` (`identity_id`) ON DELETE CASCADE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci COMMENT='Stores user follow relationships for the Friends/Following system';
--- a/frontend/database/schema.sql
+++ b/frontend/database/schema.sql
@@ -1389,6 +1389,21 @@ CREATE TABLE `Teams_Group_Roles` (
 /*!40101 SET character_set_client = @saved_cs_client */;
 /*!40101 SET @saved_cs_client     = @@character_set_client */;
 /*!50503 SET character_set_client = utf8mb4 */;
+CREATE TABLE `User_Connections` (
+  `follower_id` int NOT NULL AUTO_INCREMENT,
+  `follower_identity_id` int NOT NULL,
+  `followed_identity_id` int NOT NULL,
+  `created_at` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  PRIMARY KEY (`follower_id`),
+  UNIQUE KEY `unique_follow_relationship` (`follower_identity_id`,`followed_identity_id`),
+  KEY `idx_follower_identity_id` (`follower_identity_id`),
+  KEY `idx_followed_identity_id` (`followed_identity_id`),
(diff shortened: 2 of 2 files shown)


frontend/database/00283_create_user_followers_table.sql
```diff
@@ -0,0 +1,12 @@
...
+  `follower_identity_id` int NOT NULL,
+  `followed_identity_id` int NOT NULL,
+  `created_at` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  PRIMARY KEY (`follower_id`),
+  UNIQUE KEY `unique_follow_relationship` (`follower_identity_id`, `followed_identity_id`),
+  KEY `idx_follower_identity_id` (`follower_identity_id`),
+  KEY `idx_followed_identity_id` (`followed_identity_id`),
+  CONSTRAINT `fk_follower_identity_id` FOREIGN KEY (`follower_identity_id`) REFERENCES `Identities` (`identity_id`) ON DELETE CASCADE,
+  CONSTRAINT `fk_followed_identity_id` FOREIGN KEY (`followed_identity_id`) REFERENCES `Identities` (`identity_id`) ON DELETE CASCADE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci COMMENT='Stores user follow relationships for the Friends/Following system';
```
pabo99, line 12: For consistency with the rest of the comments, could you translate this into Spanish?

frontend/database/00283_create_user_followers_table.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE `User_Connections` (
```
pabo99, line 1: The migration script name mentions creating a `user_followers` table. Would it be possible to use that as the table name as well?


```suggestion
CREATE TABLE `Users_Followers` (
```

frontend/database/00283_create_user_followers_table.sql
```diff
@@ -0,0 +1,12 @@
...
+  `follower_id` int NOT NULL AUTO_INCREMENT,
```
pabo99, line 2: By convention, primary key names are usually based on the table name. Could we rename these to something like:

```suggestion
  `users_follower_id` int NOT NULL AUTO_INCREMENT,
```
?

frontend/database/00283_create_user_followers_table.sql
```diff
@@ -0,0 +1,12 @@
...
+  `follower_identity_id` int NOT NULL,
```
pabo99, line 3: Should only users be allowed to have followers? Since there can be identities that are not linked to a user, I’d prefer this relationship to reference `Users` instead. Same for the `followed_identity_id` field

frontend/database/00283_create_user_followers_table.sql
```diff
@@ -0,0 +1,12 @@
...
+  `followed_identity_id` int NOT NULL,
+  `created_at` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  PRIMARY KEY (`follower_id`),
+  UNIQUE KEY `unique_follow_relationship` (`follower_identity_id`, `followed_identity_id`),
+  KEY `idx_follower_identity_id` (`follower_identity_id`),
+  KEY `idx_followed_identity_id` (`followed_identity_id`),
+  CONSTRAINT `fk_follower_identity_id` FOREIGN KEY (`follower_identity_id`) REFERENCES `Identities` (`identity_id`) ON DELETE CASCADE,
+  CONSTRAINT `fk_followed_identity_id` FOREIGN KEY (`followed_identity_id`) REFERENCES `Identities` (`identity_id`) ON DELETE CASCADE
```
pabo99, line 11: Could you add a DB-level guard against self-following? Something like [CHECK](https://dev.mysql.com/blog-archive/mysql-8-0-16-introducing-check-constraint/) `(follower_identity_id <> followed_identity_id)` would prevent accidental invalid relationships even if a future API path forgets to validate it

frontend/database/00283_create_user_followers_table.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE `User_Connections` (
+  `follower_id` int NOT NULL AUTO_INCREMENT,
+  `follower_identity_id` int NOT NULL,
+  `followed_identity_id` int NOT NULL,
+  `created_at` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  PRIMARY KEY (`follower_id`),
+  UNIQUE KEY `unique_follow_relationship` (`follower_identity_id`, `followed_identity_id`),
+  KEY `idx_follower_identity_id` (`follower_identity_id`),
+  KEY `idx_followed_identity_id` (`followed_identity_id`),
+  CONSTRAINT `fk_follower_identity_id` FOREIGN KEY (`follower_identity_id`) REFERENCES `Identities` (`identity_id`) ON DELETE CASCADE,
```
pabo99, line 10: I think `idx_follower_identity_id` may be redundant because the unique key on (`follower_identity_id`, `followed_identity_id`) already supports lookups by `follower_identity_id`. We probably only need the separate index on `followed_identity_id` for reverse lookups.

```suggestion
```

PR #10204, opened 2026-09-10
Add solved-status filter to the problem list

### LINKED ISSUE #10137: [FEATURE] Add solved-status filter to the problem list
### Checklist
- [x] I have discussed this on GitHub Discussions or Discord and received positive feedback

### Context
The problem list has no way to filter by whether the current user has solved, attempted, or not tried a problem — users must scan manually to find fresh problems or revisit unsolved ones.

### Feature description
Add a `solved_status` query parameter to `Problem::apiList` accepting `all` (default), `solved`, `attempted` (submitted but no AC), and `unsolved` (never submitted). Apply the filter server-side in the SQL query so pagination stays correct. Surface it as a dropdown on the frontend problem list and search bar.

### Value
Faster problem discovery and practice tracking, especially for contest prep.

deepak0x:
This adds a `solved_status` filter to the problem list, so users can look at only the problems they have solved, attempted without an AC, or never tried.

`Problem::apiList` (and the TypeScript payload entrypoint) accepts a `solved_status` query parameter: `all` (default), `solved`, `attempted`, `unsolved`. The filter runs inside `Problems::byIdentityType`, so it composes with the other filters and pagination counts stay correct.

The semantics follow what we agreed in the issue: problems the user owns and problems they forfeited are excluded from both the solved and attempted buckets, and only normal submissions count. A problem that was only AC'd with a test submission stays attempted.

While wiring this up I found the solved/attempted columns were already inconsistent: `Problems::getProblemsUnsolvedByIdentity` counted every AC submission toward the solved column, including test-type ones, so its rows disagreed with the solved/attempted alias lists the client gets. That query now only considers normal submissions, so both surfaces agree.

On the frontend, the search bar shows a "Filter by status" dropdown for logged-in users (the filter is identity-based, so it makes no sense anonymously). It submits with the existing language/quality filters, and the pager links keep the parameter.

Tests:

- `ProblemListTest::testSolvedStatusFilter` covers each status, the own-problem and forfeited-problem exclusions, and the test-type AC case.
- `ProblemListTest::testAttemptedAndSolvedTypeConsistency` checks that the rows from `apiList` for each status agree with the solved/attempted alias lists.
- Full `ProblemListTest` and `ProblemSearchBooleanModeTest` suites pass (55 tests, 665 assertions), along with psalm and the lint suite.

Fixes #10137


deepak0x:
@pabo99 Implemented following the approach you approved in the issue. The two main pieces are the `solved_status` filter in `Problems::byIdentityType` with the ownership/forfeit exclusions we discussed, and the `getProblemsUnsolvedByIdentity` fix so test-type ACs stop counting toward the solved column.

deepak0x:
Screenshot of the new status filter on the problem list (logged in as a test user, filtered to Unattempted):

status filter

pabo99 (commented):
Before I start the review, could you please resolve the conflicts that have appeared?

I’d also prefer to keep the UI changes separate, since I have a few suggestions there that could make this PR grow significantly. Could you please remove the UI-related changes from this PR for now? We can address them separately in a follow-up PR

pabo99:
It looks like there are still some merge conflicts. Could you please update the branch with the latest changes from main and resolve them again?

PR #10206, opened 2026-09-11
feat: add chart and table view toggle for solved tags on the profile

### LINKED ISSUE #10138: [FEATURE] Per-tag solved statistics on user profiles

deepak0x:
# Description

Adds a view toggle to the solved-tags section of the profile page, so the section can be read either as the existing chart or as a table with the full tag distribution.

Split out from #10188 so the backend and the frontend changes stay independently reviewable.

- `TagsSolvedChart` renders a chart/table toggle; the table lists every tag in the distribution with its solved count, and both columns are sortable.
- When the full distribution is available the table uses it, so tail tags that were previously aggregated into "Others" are individually visible; when it is empty the component falls back to the capped `tags` payload.
- `ViewProfile` passes the new `tagsFull` payload through to the chart component.
- Translations for the new UI strings (en/es/pt plus the regenerated pseudo locale), and Jest tests cover the toggle, the fallback, both sort directions and the rendered rows.

Chart view:

chart view

Table view with the full distribution for the same profile:

table view

Fixes #10138

# Comments

Depends on #10188, which introduces the `tagsFull` payload used here. This branch sits on top of that PR's head, so while #10188 is open the diff below also lists its backend commits; only the last commit is new. Once #10188 merges, the diff reduces to the UI changes alone.

# Checklist:


deepak0x:
@pabo99 this is the UI follow-up to #10188, split out per your review so the frontend changes stay separately reviewable. It adds the chart/table view toggle for the solved-tags section on the profile, with the full tag distribution available in the table view.

deepak0x:
The cypress (contest-group) job failed on two contest.cy.ts tests, both element-lookup timeouts (one waiting for `[name="language"]`, another for `a[data-problem="5370a094-e"]`). Nothing in this PR touches the contest flow, and the same spec failed on main a couple of days ago with the same kind of timeout in run https://github.com/omegaup/omegaup/actions/runs/34533685553, so this looks like the known flakiness of that suite rather than a regression from these changes. Rerunning the failed jobs should clear it (I can't trigger that from the fork side).

PR #10210, opened 2026-09-14
Replace pagedown and remove submodules

antedotee:
# Description

Clone and `docker compose up` no longer need `git submodule update`. Karel, ReKarel, Question2Answer, iso-3166-2, and csv.js are vendored in the tree, and MathJax is copied from Yarn at webpack time.

Problem/contest markdown still uses `markdown.Converter` (`makeHtml`, `makeHtmlWithImages`, `renderMermaidDiagrams`) and the existing sanitizer allow-list. Rendering now goes through markdown-it plus the omegaUp dialect (sample I/O, templates, transclusion, figures, GFM tables/fences). The WMD editor is replaced by `MarkdownToolbar.vue` on statement, solution, and contest description fields.

Pagedown, `.gitmodules`, recursive CI checkout, and `--recurse-submodules` on production Docker clones are removed.

# Comments

- Some golden HTML in `markdown.test.ts` changed where markdown-it is equivalent (no wrapping `<p>` around `<figure>`/`<details>`, tighter newlines around fences). Semantics of the sanitizer, sample I/O, templates, and Prism/mermaid classes should be unchanged.
- Q2A still ships its own Pagedown under `frontend/www/preguntas/`; that is independent of the omegaUp editor.
- The current `omegaup/dev-php` image has Node 18 while `package.json` requires Node `>=24`, so in-container `yarn-run` still fails. Host webpack (Node 24) is what the bind-mounted `js/dist` uses. That Node bump is separate from this PR.

PS: As discussed on the slack, I have pasted the screenshots in the chat of UI running locally.


antedotee:
@pabo99 PTAL. Maybe we will need to work a little more before it becomes mergeable.

pabo99 (commented):
Thanks for putting this together! There is a lot of useful work here, but I don't think we can effectively review this PR in its current form. GitHub is showing more than 600k added lines, largely because removing the submodules also brings the contents of projects such as `Karel.js`, `ReKarel`, and `Question2Answer` directly into this repository.

I think we should separate the Pagedown replacement from the broader submodule removal.

Could we start with a smaller PR focused on validating the Pagedown replacement, without removing the existing submodules yet? A small but representative PoC using the new Markdown implementation, together with tests that verify compatibility with the omegaUp Markdown behavior, would make it much easier for us to evaluate the approach.

Once we validate that approach, we can complete the Pagedown migration in a follow-up PR. I would prefer to discuss removing the remaining submodules separately, since vendoring their entire contents into the omegaUp repository is a different architectural decision and makes the changes extremely difficult to review.

This PR already contains a lot of valuable work, so hopefully most of it can be reused across those smaller changes.

pabo99:
Also, as we split this work into smaller PRs, could you please create a separate issue for each of them?

That would help us define the scope and expected outcome of each change before implementation, and make it easier to track and review the work independently.

You can reference this PR from those issues so we don't lose the context and work that has already been done here.

antedotee:
@pabo99 I am thinking of using stacked PRs since it would be of the best use case here.

antedotee:
Sure.

PR #10228, opened 2026-09-20
Refactor createNomination logic into separate functions by nomination type (#8297)

### LINKED ISSUE #8297: [FEATURE] Separate code into distinct functions based on `QualityNomination` type to be created

AbdulKreemShah2408:
Description
Refactored the monolithic createNomination method in frontend/server/src/Controllers/QualityNomination.php to improve code readability and maintainability by delegating logic into distinct type-specific validation and handling helper functions.

Split large conditional switch/if blocks into dedicated private functions: validateSuggestionContents, validatePromotionContents, validateDemotionContents, validateDismissalContents, and validateQualityTagContents.

Extracted shared tag validation and duplication checks (validateAllowedTags, validateNoDuplicatedTags) into reusable helpers.

Encapsulated transaction handling and reviewer assignment into saveNomination and assignReviewers.

Strictly followed the "no new functionality / API changes" constraint; existing behavior and API responses remain completely intact.

Fixes: #8297

Comments
This is a pure refactoring and cleanup task for issue #8297 as requested by the code owners. All existing unit tests pass successfully. Special attention can be given to the clean separation of nomination-type validation logic.

Checklist:
[x] The code follows the coding guidelines of omegaUp.   

[x] The tests were executed and all of them passed.   

[x] If you are creating a feature, the new tests were added.   

[x] If the change is large (> 200 lines), this PR was split into various Pull Requests. (N/A - focused purely on controller refactoring)


frontend/server/src/Controllers/QualityNomination.php
```diff
@@ -230,281 +241,390 @@ public static function createNomination(
...
+    }
+
+    /**
+     * Validates the `contents` for a `demotion` nomination. May mutate
+     * `$contents['original']` in place, normalizing a full problem URL into
+     * a plain alias.
+     *
+     * @param array{reason?: string, rationale?: string, original?: string} $contents
+     */
+    private static function validateDemotionContents(array &$contents): void {
```
pabo99, line 444: Could we avoid passing `$contents` by reference here? Since this method is primarily responsible for validation, I think it would be clearer if it didn't mutate its input.
We could keep `validateDemotionContents()` focused on validating `reason`, `rationale`, and the required `original` field, and move the normalization of `original` to a separate helper that returns the normalized alias. For example:
```php
$contents['original'] = self::normalizeOriginalProblemAlias(
    $contents['original']
);
```
This would make the data flow explicit and keep validation and normalization as separate responsibilities.

frontend/server/src/Controllers/QualityNomination.php
```diff
@@ -230,281 +241,390 @@ public static function createNomination(
...
-            }
-            if (
-                isset($contents['level']) &&
-                !in_array($contents['level'], self::LEVEL_TAGS)
-            ) {
+    /**
+     * Validates the `contents` for a `dismissal` nomination. A dismissal
+     * carries no extra fields.
+     *
+     * @param array<string, mixed> $contents
```
pabo99, line 505: Could you avoid using `mixed` here?
```suggestion
     * @param array{tags?: string[], before_ac?: bool, difficulty?: int, quality?: int, statements?: array<string, array{markdown: string}>, source?: string, reason?: string, original?: string, tag?: list<string>, quality_seal?: bool} $contents
```

Using `mixed` here removes some of the benefits of trying to keep the data properly typed.

Even better, now that this type is being used in several places, would it be worth declaring a type alias for it?

frontend/server/src/Controllers/QualityNomination.php
```diff
@@ -220,6 +220,17 @@ class QualityNomination extends \OmegaUp\Controllers\Controller {
         'problemLevelIntermediateMathsInProgramming',
     ];
 
+    /**
+     * Types of nomination that require the identity to have already solved
+     * the problem, except when a valid `before_ac` suggestion/dismissal is
+     * being made (see {@see self::validateProblemSolvedPrecondition()}).
+     */
+    const TYPES_REQUIRING_SOLVED_PROBLEM = [
```
pabo99, line 228: It looks like this is not being used. Could you use it in `validateProblemSolvedPrecondition()` instead of duplicating this knowledge there? Otherwise, I think we can remove it

frontend/server/src/Controllers/QualityNomination.php
```diff
@@ -230,281 +241,390 @@ public static function createNomination(
...
+                'contents'
+            );
+        }
+    }
+
+    /**
+     * Validates the `contents` for a `quality_tag` nomination. Only
+     * reviewers are allowed to send this type of nomination.
+     *
+     * @param array{quality_seal?: bool, level?: string, tags?: list<string>} $contents
```
pabo99, line 527: The reason Psalm is complaining is that this function is receiving a different type from the one declared, so you can also standardize it as in the other places:
```suggestion
     * @param array{tags?: string[], before_ac?: bool, difficulty?: int, quality?: int, statements?: array<string, array{markdown: string}>, source?: string, reason?: string, original?: string, tag?: list<string>, quality_seal?: bool} $contents
```

frontend/server/src/Controllers/QualityNomination.php
```diff
@@ -230,281 +241,390 @@ public static function createNomination(
...
+     * @return null|array{qualitynomination_id: int}
+     */
+    private static function validateQualityTagContents(
```
pabo99, line 530: For consistency with the rest of the `validate...` functions, could we make this one return `void` as well? It might be worth calling `getQualityNominationContentsForProblemAndReviewer()` right after the validation is performed

frontend/server/src/Controllers/QualityNomination.php
```diff
@@ -230,281 +241,390 @@ public static function createNomination(
...
-        if ($nominationType !== 'demotion' && $nominationType !== 'quality_tag') {
+        self::validateProblemSolvedPrecondition(
+            $problem,
+            $identity,
+            $nominationType,
+            $contents
+        );
+
+        $qualityNomination = null;
+        switch ($nominationType) {
```
pabo99, line 252: Should we handle an unsupported `$nominationType` explicitly here? Otherwise, an unknown type will skip all content validation and continue to `saveNomination()`

